PR #10, opened 2016-11-09
Send content-length header on 404 not found response. Close #7

### LINKED ISSUE #7: Default 404 handler doesn't send content-length or transfer-encoding header.
Default implementation of HttpClient's 404 handler is following.

```java
						request.delegate()
						       .writeAndFlush(new DefaultHttpResponse(HttpVersion.HTTP_1_1,
								       HttpResponseStatus.NOT_FOUND));
```

This response doesn't include `Content-Length` header and `Transfer-Encoding` header. As a result, HTTP client can't determine end of the response. HTTP/1.1 server must return Content-Length or Transfer-Encoding(iirc).

Curl's verbose output is here:
```
$ curl -vvv http://localhost:8011/jkjk
*   Trying ::1...
* connect to ::1 port 8011 failed: Connection refused
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8011 (#0)
> GET /jkjk HTTP/1.1
> Host: localhost:8011
> User-Agent: curl/7.49.1
> Accept: */*
> 
< HTTP/1.1 404 Not Found
* no chunk, no close, no size. Assume close to signal end
< 
```

tokuhirom:


tokuhirom:
Note: Testing failure is not related to this patch.

smaldini:
Fixed in 0.6.0 already

PR #13, opened 2016-12-06
Fix overloaded method argument issue

rstoyanchev:
The changed methods are passing `uri()` where a `protocols` argument is expected.


smaldini:
Refactored since then

PR #29, opened 2017-01-16
Allow configuring a custom resolver for proxies in ClientOptions

tareqhs:
NoopAddressResolverGroup causes UnknownHostException in some scenarios where the client is configured to use http proxy

The fix is needed for CloudFoundry Java Client library to function behind proxies:
https://github.com/cloudfoundry/cf-java-client/issues/479

Signed-off-by: Tareq Sharafy <tareq.sha@gmail.com>


smaldini:
Hey there, do you mind signing https://cla.pivotal.io/sign/spring/icla for reactor we can add this for snapshot testing with CF. Thanks again, nice workaround !

tareqhs:
Hi @smaldini , done.

PR #32, opened 2017-01-31
fix #17 Add SNI support for all ClientContextHandlers

### LINKED ISSUE #17: Support client SNI
Currently, reactor uses the same code for creating a `SslHandler` regardless of the Netty channel being a client or server connection. (https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/ipc/netty/options/NettyOptions.java#L219)

Clients should use the `newHandler(ByteBufAllocator alloc, String peerHost, int peerPort)` variant so that the server name gets sent in the SSL handshake so that the server knows which certificate to send to the client. See https://github.com/netty/netty/issues/3801.

simonbasle:
cc @smaldini @mheath

### DIFF
--- a/src/main/java/reactor/ipc/netty/channel/ClientContextHandler.java
+++ b/src/main/java/reactor/ipc/netty/channel/ClientContextHandler.java
@@ -16,6 +16,8 @@
 
 package reactor.ipc.netty.channel;
 
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelPipeline;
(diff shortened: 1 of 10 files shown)


simonbasle:
got the following test failing (will restart them):
```
reactor.ipc.netty.tcp.TcpClientTests > readIdleDoesNotFireWhileDataIsBeingRead FAILED
    reactor.core.Exceptions$ReactiveException at TcpClientTests.java:309
        Caused by: io.netty.channel.AbstractChannel$AnnotatedConnectException

reactor.ipc.netty.tcp.TcpClientTests > consumerSpecAssignsEventHandlers FAILED
    reactor.core.Exceptions$ReactiveException at TcpClientTests.java:289
        Caused by: io.netty.channel.AbstractChannel$AnnotatedConnectException
```

smaldini:
unrelated issues unfortunately, clear to merge

smaldini:
Note that we need to review separately the SSL options to see if we can improve the API/builder

PR #44, opened 2017-02-16
Improve (or fix) several javadocs, use localhost:0 for TcpServer

simonbasle:
@smaldini I had doubts on the actual behavior for some methods, and
the relevance of some sentences, for which I put TODOs in place.


smaldini:
Should we consume this iteration before 0.6.1.RELEASE and resume onto 0.6.2.RELEASE ?

simonbasle:
@smaldini TODOs removed, also changed a bit of TcpServer, HttpServer to default to port 0 and 80 respectively. Also added info on how to chose random port and get the effective port.

Issue #46, opened 2017-02-16
Have `addEncoder`/`addDecoder` skip if handler already exist, like legacy `addHandler`


simonbasle:
`addHandler` used to check if there was an existing handler with the provided name, and skip the addition of the new handler if that was the case.
Since the move to `addEncoder`/`addDecoder`, this is no longer the case.

This should be fixed, for a minimum consistency with the previous API (even though the methods don't add at the same position as before).

Note that this is the inverse of #22 which would introduce a way of replacing instead of skipping.

Issue #22, opened 2016-12-28
Let the `NettyInbound.onReadIdle` and `NettyOutbound.onWriteIdle` replace existing idle handlers


rstoyanchev:
````
java.lang.IllegalArgumentException: Duplicate handler name: onChannelReadIdle
        at io.netty.channel.DefaultChannelPipeline.checkDuplicateName(DefaultChannelPipeline.java:1062)
        at io.netty.channel.DefaultChannelPipeline.filterName(DefaultChannelPipeline.java:293)
        at io.netty.channel.DefaultChannelPipeline.addBefore(DefaultChannelPipeline.java:250)
        at io.netty.channel.DefaultChannelPipeline.addBefore(DefaultChannelPipeline.java:240)
        at reactor.ipc.netty.channel.ChannelOperations.addHandler(ChannelOperations.java:141)
        at reactor.ipc.netty.channel.ChannelOperations.addHandler(ChannelOperations.java:64)
        at reactor.ipc.netty.NettyInbound.onReadIdle(NettyInbound.java:66)
        at org.springframework.messaging.tcp.reactor.ReactorNettyTcpConnection.onReadInactivity(ReactorNettyTcpConnection.java:67)
        at org.springframework.messaging.simp.stomp.StompBrokerRelayMessageHandler$StompConnectionHandler.initHeartbeats(StompBrokerRelayMessageHandler.java:706)
...
````

For context, on a TCP connection to a STOMP broker we use heartbeats to keep the connection from hanging. However heartbeats have to be negotiated first through an exchange of CONNECT and CONNECTED frames and so there is a possibility to hang before the heartbeats have started leading to a [resource leak](https://jira.spring.io/browse/SPR-14266). This is why we first register one task first check if we haven't received CONNECTED and then a heartbeat task after.

This should not cause `onChannelReadIdle` to fail. One option would be to replace the OnChannelReadIdle handler first (my current workaround) or the API could provide a clear choice whether the new task should be added or used as a replacement.

simonbasle:
72f4cbd doesn't completely fix this (as subsequent invocations of `onReadIdle` API went from failing to silently being discarded)...  It makes the workaround shorter, though, with the addition of `removeHandler`:

```java
public void onReadInactivity(Runnable runnable, long inactivityDuration) {

	// TODO: workaround for https://github.com/reactor/reactor-netty/issues/22
	this.inbound.removeHandler(NettyPipeline.OnChannelReadIdle);

[...]

PR #47, opened 2017-02-16
addEncoder/addDecoder skips existing, setEncoder/setDecoder replaces

simonbasle:
cc @smaldini @rstoyanchev @nebhale, this covers #22, #46 and #49


src/main/java/reactor/ipc/netty/channel/NettyContextSupport.java
```diff
@@ -129,29 +122,6 @@ public static final void addDecoderBeforeReactorEndHandlers(Channel channel, Str
 		ChannelHandler extractor = addExtractor.apply(handler);
 		boolean shouldAddExtractor = extractor != null;
 
-		if (exists) {
```
simonbasle: this is the fix for #22, why remove that?

src/main/java/reactor/ipc/netty/channel/NettyContextSupport.java
```diff
@@ -223,29 +188,6 @@ public static final void addEncoderAfterReactorCodecs(Channel channel, String na
...
```
simonbasle: same here, skipIfExist, extractorExists and this block are all part of the fix for #22

src/main/java/reactor/ipc/netty/NettyContext.java
```diff
@@ -16,88 +16,174 @@
...
 	 * @return this NettyContext
+
 	 */
-	default NettyContext addEncoder(ChannelHandler handler){
-		return addEncoder(handler.getClass().getSimpleName(), handler);
+	default NettyContext addHandler(ChannelHandler handler){
+		return addHandler(handler.getClass().getSimpleName(), handler);
 	}
-
+
```
simonbasle: extra space

src/main/java/reactor/ipc/netty/NettyContext.java
```diff
@@ -125,5 +246,33 @@ default NettyContext addDecoder(ChannelHandler handler){
...
+	 */
+	default NettyContext removeHandler(String name) {
+		ReactorNetty.removeHandler(channel(), name);
+		return this;
+	}
+
+	/**
+	 * Replace a named handler if present and return this context.
+	 * If handler wasn't present, an {@link RuntimeException} will be thrown
+	 *
```
simonbasle: Add a note "note that if your handler is of different type, there could be an extraneous 'extractor' still in place" so you should avoid this kind of replacement" or something like that?

src/main/java/reactor/ipc/netty/ReactorNetty.java
```diff
@@ -15,21 +15,212 @@
...
+
+	/**
+	 * A common implementation for the {@link NettyContext#addHandlerLast(String, ChannelHandler)}
+	 * method that can be reused by other implementors.
+	 * <p>
+	 * This implementation will look for reactor added handlers on the right hand side of
+	 * the pipeline, provided they are identified with the {@link NettyPipeline#RIGHT}
+	 * prefix, and add the handler just before the first of these.
+	 * <p>
+	 * It will also add a ByteBuf extractor for relevant encoders (and add/remove it as
```
simonbasle: this doesn't seem relevant any longer

src/main/java/reactor/ipc/netty/ReactorNetty.java
```diff
@@ -15,21 +15,212 @@
...
+
+	/**
+	 * A common implementation for the {@link NettyContext#addHandlerFirst(String, ChannelHandler)}
+	 * method that can be reused by other implementors.
+	 * <p>
+	 * This implementation will look for reactor added handlers on the left hand side of
+	 * the pipeline, provided they are identified with the {@link NettyPipeline#LEFT}
+	 * prefix, and add the handler just after the last of these.
+	 * <p>
+	 * It will also add a ByteBuf extractor for relevant encoders (and add/remove it as
```
simonbasle: this doesn't seem relevant any longer

src/main/java/reactor/ipc/netty/ReactorNetty.java
```diff
@@ -112,4 +303,27 @@ protected void channelIdle(ChannelHandlerContext ctx,
...
+
+		final BiConsumer<? super ChannelHandlerContext, Object> extractor;
+
+		ExtractorHandler(BiConsumer<? super ChannelHandlerContext, Object> extractor) {
+			this.extractor = Objects.requireNonNull(extractor, "extractor");
+		}
+
+		@Override
+		public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+			extractor.accept(ctx, msg);
```
simonbasle: I wouldn't have done it like that, but rather as a `Function<Object, Object>` that would translate inbound objects. Have it return null or the same instance to signal pass-through. Let the `ExtractorHandler` do the job of ctx.channelReadForward + the http specific version detect that a `LastHttpContent` has been converted (in which case we forward both the translation and `EMPTY_LAST_HTTP_CONTENT`.

This would lift the burden of not forgetting that last step from the user's shoulders...

src/main/java/reactor/ipc/netty/http/HttpOperations.java
```diff
@@ -187,4 +244,24 @@ protected final boolean markHeadersAsSent() {
 			AtomicIntegerFieldUpdater.newUpdater(HttpOperations.class,
 					"statusAndHeadersSent");
 
+	final static ChannelInboundHandler HTTP_EXTRACTOR = NettyPipeline.inboundHandler(
+			(ctx, msg) -> {
+				if (msg instanceof ByteBufHolder) {
+					ByteBuf bb = ((ByteBufHolder) msg).content();
```
simonbasle: (continuing) the 2 lines above would have been the user-provided lambda, with `else return msg`

src/main/java/reactor/ipc/netty/http/HttpOperations.java
```diff
@@ -187,4 +244,24 @@ protected final boolean markHeadersAsSent() {
...
+					if(msg instanceof FullHttpMessage){
```
simonbasle: all the lines below would have been responsibility of `ExtractorHandler`

src/main/java/reactor/ipc/netty/http/HttpOperations.java
```diff
@@ -187,4 +244,24 @@ protected final boolean markHeadersAsSent() {
...
+						// TODO convert into 2 messages if FullHttpMessage
+						ctx.fireChannelRead(msg);
+					}
+					else {
+						ctx.fireChannelRead(bb);
+						if (msg instanceof LastHttpContent) {
```
simonbasle: ... `HttpExtractorHandler` would only do that if for the lambda, the input was `LastHttpContent` and the result wasn't.

src/main/java/reactor/ipc/netty/ReactorNetty.java
smaldini: this is quite opinionated and already the http extractor has issues with the right type of message (not only last frame but start frame). Also the Consumer can condition the emission of some frame if required (empty for instance).

Issue #77, opened 2017-04-19
Support for http response body compression


nename0:
With spring-boot you can configure [Compression](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/api/org/springframework/boot/web/server/Compression.html) of http response body of configurable content-types. reactor-netty should have support for it. Its also not too hard to implement as netty has the [`io.netty.handler.codec.http.HttpContentCompressor`](https://netty.io/4.1/api/io/netty/handler/codec/http/HttpContentCompressor.html) class. You can to override it to filter out the Content-Type and User-Agent. About the minResponseSize I'm not really sure, that's not easy to implement in a reactive way.

nename0:
Fixed with #100

PR #78, opened 2017-04-22
Make Pipeline modification methods in ReactorNetty public

nename0:
I don't see why these methods shouldn't be public. If you want to modify the pipline in the `onChannelInit` or `afterChannelInit` options (e.g. adding a `HttpContentCompressor` (should be fixed with #77) or `HttpContentDecompressor`), these methods would be very useful.

### DIFF
--- a/src/main/java/reactor/ipc/netty/ReactorNetty.java
+++ b/src/main/java/reactor/ipc/netty/ReactorNetty.java
@@ -39,7 +39,7 @@ import reactor.util.Loggers;
  *
  * @author Stephane Maldini
  */
-final class ReactorNetty {
+public final class ReactorNetty {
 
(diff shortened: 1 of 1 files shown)


nename0:
I don't know whats the intended method to modifiy the pipeline but I now do this:
```
httpServer.newHandler((httpServerRequest, httpServerResponse) ->
{
      httpServerRequest.context().addHandler(new HttpContentDecompressor());
      httpServerRequest.context().addHandler(new HttpContentCompressor());
      return handler.apply(httpServerRequest, httpServerResponse);
})
```

nename0:
Nope, the code does not work as the handler is invoked asynchronous and therfore the handlers are too late

nename0:
Ok so making these methods public doesn't seem like a good solution, also because they require a `NettyContext`, which you don't have inside the `afterChannelInit` option :see_no_evil:
Moved to a issue #80

PR #98, opened 2017-05-19
conditional HTTP compression

mostroverkhov:
This PR adds support for HTTP compression, configured with `HttpServerOptions.Compression` `enabled` and `minResponseSize` parameters. Latter controls compression based on response body size, 0 value means all responses with non-empty body are compressed. Partially addresses #77

PR #100, opened 2017-05-19
conditional HTTP compression

mostroverkhov:
This PR adds support for HTTP compression. Server is configured with HttpServerOptions.Compression enabled and minResponseSize parameters. Latter controls compression based on response body size, 0 value means all responses with non-empty body are compressed. Client is configured with `HttpClientOptions.supportsCompression(boolean)`. Partially addresses #77. Relates to #98

### DIFF
--- a/src/main/java/reactor/ipc/netty/NettyPipeline.java
+++ b/src/main/java/reactor/ipc/netty/NettyPipeline.java
@@ -59,12 +59,15 @@ public interface NettyPipeline {
 	String ReactiveBridge     = RIGHT + "reactiveBridge";
 	String HttpEncoder        = LEFT + "httpEncoder";
 	String HttpDecoder        = LEFT + "httpDecoder";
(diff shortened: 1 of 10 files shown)


mostroverkhov:
Interesting fact - disabling `HttpContentCompressor` on client while compression on server is on still yields correct payload, which makes me wonder whether serverside compression is performed at all. Response `content-encoding` headers are there though. I ll need to investigate further

mostroverkhov:
This not ready to merge yet - looks like there are some missing pieces, and those need to be addressed

mostroverkhov:
Ok, issues were resolved, and content compression is working correctly. Still there are 2 concerns regarding this PR: 1. one allocation for each non compressed message 2. In case client requests compression, but does not support it, response body contains garbage - imo error notification is more appropriate here (maybe as configuration option).

smaldini:
Very f***** nice 👍

PR #108, opened 2017-06-20
fix #66 Expose duplicate of options for TCP/HTTP Client and Server

### LINKED ISSUE #66: Expose read-only version of HttpServerOptions in HttpServer
Currently, when the server startup fails, we don't have access to the port that was used to start the server. We would like to log the failure along with the port (in case of a BindException). 

/cc @bclozel

simonbasle:
This is useful when the handler cannot be created, to get the original
configuration of the Client/Server object.

The toString() methods for Client/Servers and Options have been aligned,
and each option type has 2 additional String representations (one
detailed as used in toString, the other simplified, as used by the
client/server's toString).


simonbasle:
**This depends on #107 being merged first** (currently this is rebased on the corresponding branch)

Issue #89, opened 2017-05-09
Mixed pooled connection operations


lucasponce:
Testing reactor-netty in migration PoC we are finding several hanged connections on the client.
Setting DEBUG mode the spot are in the following points:
```
11:38:57.245 [reactor-http-server-epoll-14] DEBUG r.i.n.c.ContextHandler - [id: 0x328b42c6, L:/127.0.0.1:8080 - R:/127.0.0.1:39188]Mixed pooled connection operations between GET:/hawkular/alerts/triggers/test-status-threshold - and a previous one PUT:/hawkular/alerts/triggers/enabled?triggerIds=test-status-threshold&enabled=true
11:41:09.804 [reactor-http-server-epoll-9] DEBUG r.i.n.c.ContextHandler - [id: 0xbc3df2e5, L:/127.0.0.1:8080 - R:/127.0.0.1:39284]Mixed pooled connection operations between GET:/hawkular/alerts/events/event/test-event-id - and a previous one POST:/hawkular/alerts/events/data
11:44:08.618 [reactor-http-server-epoll-15] DEBUG r.i.n.c.ContextHandler - [id: 0x29134eab, L:/127.0.0.1:8080 - R:/127.0.0.1:39354]Mixed pooled connection operations between GET:/hawkular/alerts/triggers/test-autoresolve-trigger - and a previous one PUT:/hawkular/alerts/triggers/enabled?triggerIds=test-autoresolve-trigger&enabled=true
11:46:23.075 [reactor-http-server-epoll-16] DEBUG r.i.n.c.ContextHandler - [id: 0x519a4a4c, L:/127.0.0.1:8080 - R:/127.0.0.1:39398]Mixed pooled connection operations between DELETE:/hawkular/alerts/triggers/test-manual-trigger - and a previous one DELETE:/hawkular/alerts/triggers/test-autoresolve-trigger
```
I wonder if this is a known issue, or if there is some advice to avoid these kind of errors.
In a first view it looks like a bug, as there is no clear evidence in the user's code, but I would like some help to confirm this.
Thanks.

smaldini:
This has been repoted on 0.6.3 is it ?

lucasponce:
It happens with react-netty 0.6.2 version and react 3.0.7. That's stack I'm using it.

lucasponce:
@smaldini any update about this issue ?
To confirm if it is a well known issue, or perhaps some workaround ?
Thanks

smaldini:
[...]

PR #111, opened 2017-06-21
issue #89: Update pendingResponses in the event for termination

violetagg:
Problem: pendingResponses variable is decreased when
`r.i.n.h.server.HttpServerHandler#write` is invoked with `LastHttpContent`,
but the channel attributes are cleared in the event for termination
`r.i.n.channel.ChannelOperations#onHandlerTerminate`.
New request can be received after invocation of the write operation and
before event for termination.
`r.i.n.h.server.HttpServerHandler#channelRead`
In this situation instead of pipelining the new request the implementation
will check the channel attributes and as they are still not cleared the
request will not be processed.


smaldini:
Would you have a reproducing test tot addd (e.g. one from the issue).

Issue #102, opened 2017-05-22
UnsupportedMessageTypeException: DefaultFileRegion when returning FileSystemResource from controller on TLS


oleg-alexeyev:
Stack: Spring WebFlux 5.0.0.M5 / Reactor Netty 0.6.2 / Reactor Core 3.0.7 / Netty 4.1.8.
Having a controller returning `Mono<FileSystemResource>`, everythinig works fine with plain HTTP endpoint, but fails on HTTPS one:
```
io.netty.handler.codec.UnsupportedMessageTypeException: io.netty.channel.DefaultFileRegion (expected: io.netty.buffer.ByteBuf)
	at io.netty.handler.ssl.SslHandler.write(SslHandler.java:594) ~[netty-handler-4.1.8.Final.jar:4.1.8.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:739) ~[netty-transport-4.1.8.Final.jar:4.1.8.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeWrite(AbstractChannelHandlerContext.java:731) ~[netty-transport-4.1.8.Final.jar:4.1.8.Final]
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:817) ~[netty-transport-4.1.8.Final.jar:4.1.8.Final]
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:724) ~[netty-transport-4.1.8.Final.jar:4.1.8.Final]
	at io.netty.handler.codec.MessageToMessageEncoder.write(MessageToMessageEncoder.java:111) ~[netty-codec-4.1.8.Final.jar:4.1.8.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:739) ~[netty-transport-4.1.8.Final.jar:4.1.8.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeWrite(AbstractChannelHandlerContext.java:731) ~[netty-transport-4.1.8.Final.jar:4.1.8.Final]
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:817) ~[netty-transport-4.1.8.Final.jar:4.1.8.Final]
[...]

PR #112, opened 2017-06-22
Make sendFile work on secure connections

simonbasle:
Issue #102

Work in progress: use a strategy pattern to fallback on the correct `ChunkedInput` variant whenever `SslHandler` is detected in the pipe while sending a file (`FileRegion` is
not compatible with Ssl).

**TODO**
 - [x] move the strategy classes out in their own files
 - [x] test (and adapt?) for TCP and UDP


src/main/java/reactor/ipc/netty/ByteBufFlux.java
```diff
@@ -234,6 +240,10 @@ public ByteBufFlux retain() {
 		return new ByteBufFlux(doOnNext(ByteBuf::retain), alloc);
 	}
 
+	public ByteBufFlux logByteBuf() {
```
smaldini: remove this for now

PR #114, opened 2017-06-27
fix #24 Allow shutdown of global TcpResources

### LINKED ISSUE #24: No way to shut down TcpClient resources
Currently TcpClient is created as follows:
````java
ClientOptions clientOptions = ClientOptions.create();
clientOptions.loopResources(TcpResources.get()).poolResources(TcpResources.get());
options.accept(clientOptions);
return new TcpClient(clientOptions.duplicate());
````
Even if I provide my own resource through `ClientOptions` the global resources are created anyway and after that there is no way to shut them down since `dispose()` is empty and `_dispose()` is protected.

Please support the use case to create a `TcpClient` and then shut down **all** associated resources. If I provide my own resources it shouldn't trigger creation of global ones. There should also be some way to shut down global resources through an explicit method designed for that.  Consider for example that when running in a server like Tomcat on shutdown you get ugly messages about active threads.

It would also be very useful for `ClientOptions` to expose a way to expose what an option is currently set to so that I can also accept external options and add my own only if others are not provided -- same thing that `TcpClient#create` would have to do.

### PR DESCRIPTION
**Pull request by simonbasle**
This commit allows to just shut down the currently set global
TcpResource without replacing it with a new one. It disposes underlying
resources, which helps in server shutdown scenarios.

### DIFF
--- a/src/main/java/reactor/ipc/netty/tcp/TcpResources.java
+++ b/src/main/java/reactor/ipc/netty/tcp/TcpResources.java
@@ -72,11 +72,15 @@ public class TcpResources implements PoolResources, LoopResources {
 	 * @return the global HTTP resources
 	 */
 	public static TcpResources reset() {
+		shutdown();
+		return getOrCreate(tcpResources, null, null, ON_TCP_NEW, "tcp");
+	}
+
+	public static void shutdown() {
 		TcpResources resources = tcpResources.getAndSet(null);
 		if (resources != null) {
 			resources._dispose();
 		}
-		return getOrCreate(tcpResources, null, null, ON_TCP_NEW, "tcp");
 	}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/java/reactor/ipc/netty/tcp/TcpResources.java [hunk 1]
```diff
@@ -72,11 +72,15 @@ public static TcpResources set(LoopResources loops) {
 	 * @return the global HTTP resources
 	 */
 	public static TcpResources reset() {
+		shutdown();
+		return getOrCreate(tcpResources, null, null, ON_TCP_NEW, "tcp");
+	}
+
+	public static void shutdown() {
```
smaldini: Add small javadoc

PR #115, opened 2017-06-27
fix #25 Allow listen until LoopResource/PoolResources disposal

### LINKED ISSUE #25: No way to wait for shutdown of TcpClient resources
Following the model of `TcpResources` I'm trying to create my own resources:
````java
this.channelGroup = new DefaultChannelGroup(ImmediateEventExecutor.INSTANCE);
this.loopResources = LoopResources.create("reactor-netty-tcp-client");
this.poolResources = PoolResources.fixed("reactor-netty-tcp-pool");

TcpClient.create(opts -> opts
		.channelGroup(this.channelGroup)
		.loopResources(this.loopResources)
		.poolResources(this.poolResources)
		.preferNative(false));
````
So I can then shut them down:
````java
ChannelGroupFuture future = this.channelGroup.close();
Mono<Void> completion = FutureMono.from(future)
		.doAfterTerminate((x, e) -> {
			this.loopResources.dispose();
			this.poolResources.dispose();
		});
````
Note however that unlike `ChannelGroup` there is no option to wait for the completion of the shutdown of resources whose dispose methods return void and don't wait.

### PR DESCRIPTION
**Pull request by simonbasle**
This commit adds a disposeAndListen method to both LoopResources and
PoolResources that allow to dispose these but also get a notification of
when the disposal is done, in a best effort fashion.

Subsequent calls however are not guaranteed to return a relevant Mono,
which could complete sooner than the actual disposal as triggered by the
first invocation.

### DIFF
--- a/src/main/java/reactor/ipc/netty/resources/DefaultLoopResources.java
+++ b/src/main/java/reactor/ipc/netty/resources/DefaultLoopResources.java
@@ -60,7 +60,7 @@ final class DefaultLoopResources extends AtomicLong implements LoopResources {
 			int selectCount,
 			int workerCount,
 			boolean daemon) {
-		this.running = new AtomicBoolean();
+		this.running = new AtomicBoolean(true);
 		this.daemon = daemon;
 		this.workerCount = workerCount;
 		this.prefix = prefix;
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/main/java/reactor/ipc/netty/resources/DefaultLoopResources.java [hunk 1]
```diff
@@ -85,24 +88,44 @@ static ThreadFactory threadFactory(DefaultLoopResources parent, String prefix) {
 
 	@Override
 	public void dispose() {
+		disposeAndListen();
+	}
+
+	public Mono<Void> disposeAndListen() {
```
smaldini: use deferDispose() and then Mono.defer inside to defer the actual dispose

src/main/java/reactor/ipc/netty/resources/DefaultPoolResources.java
```diff
@@ -188,19 +188,22 @@ public String toString() {
...
+			Pool pool;
+			for (SocketAddress key: channelPools.keySet()) {
+				pool = channelPools.remove(key);
+				if(pool != null){
+					pool.close();
+				}
 			}
-		}
-		return Mono.empty();
+			return Mono.empty();
```
smaldini: Could be Mono.fromRunnable then

src/main/java/reactor/ipc/netty/resources/DefaultLoopResources.java
```diff
@@ -87,45 +87,43 @@ static ThreadFactory threadFactory(DefaultLoopResources parent, String prefix) {
...
-			group = cacheNativeServerLoops.get();
-			if(group != null){
-				Future<?> cnsrvlFuture = group.shutdownGracefully();
-				cnsrvlMono = FutureMono.from((Future)cnsrvlFuture);
-			}
-
-			return Mono.when(clMono, slMono, cnclMono, cnslMono, cnsrvlMono).then();
-		}
-		return Mono.empty();
+			return Mono.empty();
```
smaldini: could be Mono.fromRunnable

smaldini:
Would deferDispose() works better than disposeDeferred ?

smaldini:
Use shutdownLater ?

simonbasle:
perfect 👌

PR #116, opened 2017-06-28
fix #61 Offer a simple blocking alternative API for client/server

### LINKED ISSUE #61: Provide a simpler embedded server API
Current Reactor Netty API is not very straightforward when it comes to just run it from a main, see for example [this code sample](https://github.com/mix-it/mixit/blob/ccc46a3077083beb7cb86b7c8ea8e89a96e34ba8/src/main/kotlin/mixit/support/Server.kt#L26) where `AtomicReference<NettyContext>` is needed.

Could we have something like `startAndAwait()` method and something to add handlers in a more friendly fashion?

simonbasle:


src/main/java/reactor/ipc/netty/tcp/BlockingNettyContext.java
```diff
@@ -0,0 +1,97 @@
...
+	 * @see InetSocketAddress#getHostString()
+	 */
+	public String getHost() {
+		return context.address().getHostString();
+	}
+
+	/**
+	 * Stop the {@link NettyContext} and wait for its termination, up to the {@link #setLifecycleTimeout(Duration) lifecycle timeout}.
+	 */
+	public void stop() {
```
smaldini: shutdown()

PR #124, opened 2017-07-06
fix #125 Allow configuration of hosts that bypass the proxy

### LINKED ISSUE #125: Allow proxy bypass
When using a proxy, allow to configure some addresses that won't go through the proxy, when using `HttpClient.get` (since this is no longer the base url and uses a copy of the configuration, we can remove the proxy part of the conf if the get url is absolute and is one of the bypassed hosts).

violetagg:
New configuration is added ClientProxyOptions.Builder.nonProxyHosts
that can be used to specify via regular expression
(using java.util.regex) the hosts that should bypass the proxy.

### DIFF
--- a/src/test/java/reactor/ipc/netty/http/client/HttpClientTest.java
+++ b/src/test/java/reactor/ipc/netty/http/client/HttpClientTest.java
@@ -22,8 +22,6 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.cert.CertificateException;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
@@ -33,7 +31,6 @@ import io.netty.buffer.Unpooled;
 import io.netty.handler.codec.http.DefaultFullHttpResponse;
 import io.netty.handler.codec.http.HttpContentDecompressor;
 import io.netty.handler.codec.http.HttpHeaderNames;
-import io.netty.handler.codec.http.HttpObjectAggregator;
 import io.netty.handler.codec.http.HttpResponseEncoder;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.HttpVersion;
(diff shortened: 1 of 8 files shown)


src/test/java/reactor/ipc/netty/http/client/HttpClientTest.java
```diff
@@ -245,6 +241,40 @@ public void proxy() throws Exception {
 		page.block(Duration.ofSeconds(30));
 	}
 
+	@Test
+	@Ignore
```
simonbasle: leftover test, or do you need to finish this one (no assertions)?

src/test/java/reactor/ipc/netty/options/ClientOptionsTest.java
```diff
@@ -76,4 +78,29 @@ public void toStringContainsAsDetailedString() {
...
+		assertThat(opsBuilder.build().useProxy(new InetSocketAddress("google.com", 123))).isTrue();
+
+		proxyBuilder.nonProxyHosts("localhost");
+		opsBuilder.proxyOptions(proxyBuilder.build());
+		assertThat(opsBuilder.build().useProxy((String) null)).isTrue();
+		assertThat(opsBuilder.build().useProxy((InetSocketAddress) null)).isTrue();
+		assertThat(opsBuilder.build().useProxy("hostName")).isTrue();
+		assertThat(opsBuilder.build().useProxy(new InetSocketAddress("google.com", 123))).isTrue();
+		assertThat(opsBuilder.build().useProxy("localhost")).isFalse();
+		assertThat(opsBuilder.build().useProxy(new InetSocketAddress("localhost", 8080))).isFalse();
```
simonbasle: does it work with e.g `127.0.0.1` as well?

violetagg: I added a test for `127.0.0.1`

src/test/java/reactor/ipc/netty/http/client/HttpClientTest.java
violetagg: I added assertions.
The test is marked as `@Ignore` because a proxy is needed.

simonbasle: ah makes sense... I wonder if we could simulate such a proxy easily? (ie spin a `HttpServer` that serves as the proxy, returning the requested url as a response rather than actually proxying)

simonbasle:
@violetagg I opened an issue for tracking this (to be referenced in commit message): #125

PR #131, opened 2017-07-10
Ensure accept gzip is applied when there is no handler provided

violetagg:
Ensure that accept gzip configuration is applied when the
HttpClient request is made without a handler


violetagg:
Closed #131 via de9dd7690f7dc54515d20e90f0d7ab4291b1fc92

PR #132, opened 2017-07-11
fix #80 Add NettyContext config callback, rework Channel config options

### LINKED ISSUE #80: Add a way to modify the pipeline on servers
Created from #78 

If you want to modify the pipeline on servers, you cannot use something like this, because the handler is invoked [asynchronous](https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/ipc/netty/channel/ContextHandler.java#L236):
```java
httpServer.newHandler((httpServerRequest, httpServerResponse) ->
{
      httpServerRequest.context().addHandler(new HttpContentDecompressor());
      httpServerRequest.context().addHandler(new HttpContentCompressor());
      return handler.apply(httpServerRequest, httpServerResponse);
})
```
It will be too late when the handler is invoked, as the data might already be gone through the pipeline.
So it seems the only way to do this is the `afterChannelInit` option but from there you cannot access the `NettyContext.**handler**` or `ReactorNetty.**handler**` methods.

simonbasle:
This commit adds a configuration callback to NettyOptions that deals
with newly initialized NettyContexts rather than just with the Channel.
This allows to get access to the handler management API for easier
insertion of handlers in the pipelines that have reactor-netty handlers.

Additionally, this commit does a bit of rework and cleanup around other
options that deal with pipeline configuration, notably afterChannelInit,
onChannelInit and channelGroup.

### DIFF
--- a/src/main/java/reactor/ipc/netty/channel/ContextHandler.java
+++ b/src/main/java/reactor/ipc/netty/channel/ContextHandler.java
@@ -29,7 +29,6 @@ import io.netty.channel.ChannelPipeline;
 import io.netty.channel.pool.ChannelPool;
 import io.netty.handler.logging.LoggingHandler;
 import io.netty.handler.ssl.SslHandler;
-import io.netty.util.AttributeKey;
 import io.netty.util.concurrent.Future;
 import org.reactivestreams.Publisher;
 import reactor.core.Disposable;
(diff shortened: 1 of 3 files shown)


violetagg:
Closed #132 via 00dd708f04c61c28c68b48469aa82b12589547ac

Issue #67, opened 2017-03-27
Add support for multiple connectors


mbhave:
/cc @bclozel

smaldini:
What issue was that @mbhave @bclozel ???

bclozel:
It all depends on the server concepts and infrastructure, but several servers support binding to several ports/addresses at the same time. Undertow can connect through multiple channels, Tomcat can be configured with multiple connectors.

Is there a similar construct in Reactor Netty? Or is the recommended approach to start multiple `HttpServer` instances?

smaldini:
I suppose yes, depends on what you expected to share. The handler callback cannot be shared in essence its fine to reuse it, the rest is pretty close to tcp level and should be kept isolated this way.

smaldini:
I think we can add that support directly once we revamp the high level lifecycle API/SPI for client/server which seems def. to be the most awaited improvement in our next major version.

andye2004:
I'm wondering when the next milestone release will be and if there will be any plans to pull this into it?

ctlove0523:
v1.0.0 version already released,from release notes i can't find any new features and improvements about reactor-netty support multiple connectors, the new version support this feature？

violetagg:
@ctlove0523 No this feature wasn't able to be part of the 1.0.0 GA and was left for post 1.0.0 GA
We will try to schedule this for some of the next SPs.

violetagg:
Having this recent issue https://github.com/netty/netty/issues/10781#issuecomment-723176224 and that by default Reactor Netty Servers/Clients share the `LoopResource`s and that Reactor Netty Server is 1:1 with `ServerTransport` (similar to `ServerBootstrap`), I tend to resolve this issue with `won't fix`. However I will keep it open for some time in case any requirements/discussions appear.

Having several connectors is as easy as the code below:

```
//This provides common configuration
//Shared event loop
HttpServer server = HttpServer.create()
                           .........; //some configuration

//This binds on port 7070
HttpServer server1 = server.port(7070);

//This binds on port 8080
HttpServer server1 = server.port(8080);

//This binds on port 9090 and enables security
HttpServer server1 = server.port(9090)
                           .secure(.....);
```

bclozel:
While discussing this in the context of spring-projects/spring-boot#12035, I was wondering how should we handling this particular case: two servers running on different ports, but sharing the same Micrometer meter registry. Given the metrics share the same name, we'd end up with two different servers publishing different values to the same metric name?

[...]

PR #133, opened 2017-07-11
multiple binds and shared handler lambda

simonbasle:
see #67.


simonbasle:
cc @bclozel. This is early stage prototyping and toying around, see `HttpMultiServerTest` for 2 alternative APIs.

simonbasle:
@smaldini any feedback on that? is it overlapping with your current API rework?

smaldini:
Will rework atop 0.8

bclozel:
@smaldini I'm trying to track those changes. Were those included or merged in any way?
Is there another issue that I can follow to track this?

PR #147, opened 2017-07-31
fix #146 Make sendFile work with ZipFileSystem, use FileChannel

### LINKED ISSUE #146: Use FileChannel for sendFile so that it works with ZipFs
While attempting to send a file from within a zip archive via a `ZipFileSystem`, the `sendFile` method fails. This is due to usage of `RandomAccessFile` since #112, where a `FileChannel` would previously be used.

`RandomAccessFile` needs a `File` but the `toFile()` method is unsupported on a `ZipFileSystem`:
```
java.lang.UnsupportedOperationException: null
    at com.sun.nio.zipfs.ZipPath.toFile(ZipPath.java:597)
    at reactor.ipc.netty.NettyOutbound.lambda$sendFile$0(NettyOutbound.java:206)
```

The solution is to go back to using `FileChannel`, and use `ChunkedNioFile` instead of `ChunkedFile` in the chunked strategy.

simonbasle:
This commit fixes the case of sending a file from within a zip archive
via a ZipFileSystem. This would fail since #112 due to the use of a
RandomAccessFile instead of a FileChannel. RandomAccessFile needs a File
but the toFile() method is unsupported on a ZipFileSystem

This commit goes back to using FileChannel, and uses ChunkedNioFile
instead of ChunkedFile in the chunked strategy.


simonbasle:
cc @mp911de

PR #172, opened 2017-09-28
Use port from HttpServer's builder also when options are present

lhotari:
This PR fixes a problem that the port given to `HttpServer.builder()` wasn't used after specifying some options. For example, in this case, a random port was used instead of the given port `12345`

```
        HttpServer httpServer = HttpServer.builder()
                 .options(options -> options.option(ChannelOption.TCP_NODELAY, false))
                 .port(12345)
                 .build();
```

### DIFF
--- a/src/main/java/reactor/ipc/netty/http/server/HttpServer.java
+++ b/src/main/java/reactor/ipc/netty/http/server/HttpServer.java
@@ -121,11 +121,10 @@ public final class HttpServer
 
 	private HttpServer(HttpServer.Builder builder) {
 		HttpServerOptions.Builder serverOptionsBuilder = HttpServerOptions.builder();
-		if (Objects.isNull(builder.options)) {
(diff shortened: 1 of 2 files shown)


violetagg:
Hi,

Thanks for the PR.
We are in a process of refactoring the API for creation of Clients/Servers, we will take into account this PR and will apply the change in the new implementation.

For the time being as a workaround you can use the following:
```		
HttpServer httpServer = HttpServer.builder()
        .options(options -> options.option(ChannelOption.TCP_NODELAY, false)
                                   .port(12345))
        .build();
```
or even the shortcut:
```
HttpServer httpServer =
        HttpServer.create(options -> options.option(ChannelOption.TCP_NODELAY, false)
                                            .port(12345));
```

Regards,
Violeta

violetagg:
Closing this one: the new HttpServer API is available in 0.8.x branch

PR #173, opened 2017-10-02
Prevent unwanted reverse DNS lookup calls

lhotari:
- use InetSocketAddress.getHostString instead of getHostName
  where applicable
- detect numeric IP addresses and construct them in a way that prevents
  reverse DNS lookup calls when InetSocketAddress.getAddress is called
- add possibility to pass InetSocketAddress instance to the builder
  of HttpServer and TcpServer
    - Binding to localhost can utilize NetUtil.LOCALHOST
      InetSocketAddress instance
    - Binding to wildcard address can be done by passing an instance
      of InetSocketAddress created with the "int port" constructor
- fixes some problems in tests when using IPv6 local address
  - NetUtil.LOCALHOST points to IPv6 local address when IPv6 is
    preferred
  - adds ability to run tests with java.net.preferIPv4Stack and
    java.net.preferIPv6Addresses settings enabled to compare
    differences in IPv4 and IPv6 handling

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -150,6 +150,14 @@ configure(rootProject) { project ->
   }
 
   project.tasks.withType(Test).all {
+	// run tests with IPv4 only when IPv6 is available
+	if (project.hasProperty('preferIPv4Stack')) {
+		systemProperty("java.net.preferIPv4Stack", "true")
+	}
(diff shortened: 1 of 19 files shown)


violetagg:
closed via 1bf036a02fe29855222ba8ffada0dd5fa683568d

PR #175, opened 2017-10-04
Use HttpClientCodec/HttpServerCodec

violetagg:
Instead of using separate `HttpRequestEncoder` and `HttpResponseDecoder`
in the client/server pipeline, use `HttpClientCodec`/`HttpServerCodec`.
`HttpClientCodec`/`HttpServerCodec` are a combination of `HttpRequestEncoder`
and `HttpResponseDecoder` which enables specific client/server side HTTP
implementation.
More information
https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpClientCodec.java
https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerCodec.java


violetagg:
Tested with Spring Framework master build (including all tests) and it is green.

violetagg:
closed via cce4d57c518ae0570b39f8254ecfd1cb0e71b58b

nebhale:
Just tested against our full system tests and everything looks green.

Issue #154, opened 2017-08-19
Client/Server "recursive" calling leads to error


ris58h:
I'm trying to implement [this ratpack fibonacci demo](https://github.com/setblackWs/ratpackFibb/blob/master/src/main/java/pl/setblack/fibbo/FibboServer.java) using reactor-netty.

My code is [here](https://gist.github.com/ris58h/9a3322c7e2989015e3dc09370b42ff7b).
It works on small numbers (up to 8 on my machine) but fails on 9 and greater with "Acquire operation took longer then configured maximum time". [Full stack trace](https://gist.github.com/ris58h/6c95b12a5e97f3c8c0a198f49034f43d). 

Reactor Netty version
0.7.0.M1
JVM version (e.g. `java -version`)
1.8.0_144
OS version (e.g. `uname -a`)
Linux 3.11.0-20-generic x86_64

smaldini:
This seems to be an issue that might have a short solution (default minimum pool connection is 8 where it should be 1000). More long term is that http client should support pipelining to allow such tail recurse situation via a request/response queue.

As a workaround until the default is changed, try doing ```HttpResources.set(PoolResources.fixed("http", 1000)``` before starting calling any thing (static initializer)

ris58h:
@smaldini Thank you for the response. Can I use PoolResources.elastic("http") instead of fixed for a workaround?

smaldini:
Sure that should work as well, usually client caps the number of connections tho. The problem will remain, we want to ultimately smartly limit the number of connections by pipelining http content.

violetagg:
@ris58h PoolResources.elastic is now the default strategy in Reactor Netty.

smaldini:
Closing for now we will specify in a new issue where / how to support http pipelining.

PR #176, opened 2017-10-05
Fix memory leak with keep-alive connections

lhotari:
- Channel close futures were registered for each new http request
  and not unregistered when the request handling was over.
    - this caused a memory leak when using keep-alive connections
- This adds an intermediate emitter to subscribe to the Channel's
  close future and cancels the subscription before disposing the
  emitter that is used for the duration of the http request
    - the emitter gets initialized on demand in
      ServerContextHandler.onCloseOrRelease and completed&disposed
      in ServerContextHandler.terminateChannel

### DIFF
--- a/src/main/java/reactor/ipc/netty/channel/ServerContextHandler.java
+++ b/src/main/java/reactor/ipc/netty/channel/ServerContextHandler.java
@@ -24,6 +24,11 @@ import io.netty.channel.socket.DatagramChannel;
 import io.netty.channel.socket.ServerSocketChannel;
 import io.netty.channel.socket.SocketChannel;
 import io.netty.handler.logging.LoggingHandler;
+import io.netty.util.AttributeKey;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscription;
+import reactor.core.CoreSubscriber;
+import reactor.core.publisher.DirectProcessor;
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.MonoSink;
 import reactor.ipc.netty.FutureMono;
@@ -37,6 +42,8 @@ import reactor.ipc.netty.options.ServerOptions;
 final class ServerContextHandler extends CloseableContextHandler<Channel>
 		implements NettyContext {
 
+	private static final AttributeKey<DirectProcessor<Void>> ON_TERMINATE_EMITTER = AttributeKey.newInstance("ON_TERMINATE_EMITTER");
+	private static final AttributeKey<Subscription> CLOSE_FUTURE_SUBSCRIPTION = AttributeKey.newInstance("CLOSE_FUTURE_SUBSCRIPTION");
 	final ServerOptions serverOptions;
 
 	ServerContextHandler(ChannelOperations.OnNew<Channel> channelOpFactory,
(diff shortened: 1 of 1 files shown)


lhotari:
I'm working on fixing the test that failed. It's a flaky one and I was now able to reproduce by adding 
```
    where:
    iteration << (1..100).toList()
```
to the test that failed (`reactor.ipc.netty.http.HttpSpec#http error with requests from clients`). Typically only 1 out of 100 fails so that's why I didn't notice this before pushing the PR.

src/main/java/reactor/ipc/netty/channel/ServerContextHandler.java
```diff
@@ -102,17 +115,59 @@ public boolean isDisposed() {
...
+		DirectProcessor<Void> terminateEmitter = channel.attr(ON_TERMINATE_EMITTER).getAndSet(null);
+		if (terminateEmitter != null) {
+			terminateEmitter.onComplete();
+		}
+		Subscription closeFutureSubscription = channel.attr(CLOSE_FUTURE_SUBSCRIPTION).getAndSet(null);
+		if (closeFutureSubscription != null) {
+			closeFutureSubscription.cancel();
+		}
+	}
+
```
violetagg: Just a note here - don't we need to `cancel` first the subscription and then to invoke `onComplete`on the emitter so that we are sure that we will not invoke `terminateEmitter.onComplete();` twice?

src/main/java/reactor/ipc/netty/channel/ServerContextHandler.java
```diff
@@ -102,17 +115,59 @@ public boolean isDisposed() {
...
+				@Override
+				public void onError(Throwable t) {
+					onTerminateEmitter.onError(t);
+				}
+
+				@Override
+				public void onComplete() {
+					onTerminateEmitter.onComplete();
+				}
+			});
```
violetagg: do we need to remove `CLOSE_FUTURE_SUBSCRIPTION` and `ON_TERMINATE_EMITTER ` when `onError` or `onComplete`?

src/main/java/reactor/ipc/netty/channel/ServerContextHandler.java
smaldini: double onComplete should be ignored by reactor operators. However yes it could make more sense to cancel first in that situation to avoid the later check.

smaldini:
First glance, what I'm not sure about is why not reusing the ChannelOperation itself which also a Void subscriber that listens for termination ?

src/main/java/reactor/ipc/netty/channel/ServerContextHandler.java
smaldini: Yes especially if its a persistent connection to avoid leaking references back.

smaldini:
@lhotari can you check but https://github.com/reactor/reactor-netty/commit/21fd7a996a019a06d3d3c33407286bad1884adb7 should solve this, in that case we can close this PR.

lhotari:
Thanks @smaldini ! I can confirm that commit 21fd7a9 fixes the memory leak. I created [a test app that reproduces the problem](https://github.com/lhotari/reactor-fibonacci) and there's also a semi-automated [script](https://github.com/lhotari/reactor-fibonacci/blob/master/profile_with_jfr.sh) to do Java Flight Recorder profiling (available on Oracle JDK) and heap dumps for the application. The test app is based on [the gist](https://gist.github.com/ris58h/9a3322c7e2989015e3dc09370b42ff7b) by @ris58h . That test app was referenced in issue #154 .

PR #183, opened 2017-10-11
fix #171: Do not set content length when HEAD request

### LINKED ISSUE #171: HttpSeverOperations overrides content-length to 0 on HTTP HEAD
For an HTTP HEAD request, I set the content-length to the actual number that would be written on a GET and do not write to the body. However `HttpSeverOperations` overrides that with content-length:0 when it calls `newFullEmptyBodyMessage` from `onOutboundComplete`.

It should probably realize that the request is HEAD and no body was expected to begin with, and/or also that an actual content-length was set and should be left alone.

Current workaround is to write with `Mono.empty()`. This is with 0.7.0.RELEASE.

violetagg:


violetagg:
Closed via db797d7544c90fb68c2368c180c76b6c93c10933

PR #195, opened 2017-10-24
Replace usage of java.net.URI with UriEndpoint in MonoHttpClientResponse

lhotari:
- reason is performance. using java.net.URI isn't optimal for the use case in MonoHttpClientResponse
  - using java.net.URI causes unnecessary String creation and parsing steps
- replace String[] in REDIRECT_ATTR_KEY with Supplier<String>[] so
  that redirected urls can be lazily created only when they are requested
- move uri endpoint parsing logic from HttpClientOptions to
  UriEndpointFactory class


smaldini:
That look nice and in the vein of the 0.8 refactor, I might continue working from this if you can target 0.8 instead. We can't use that in 0.7.x due to the breaking changes tho.
Way to go btw @lhotari thanks so much :D

lhotari:
@smaldini Thanks for the feedback. I changed the base to be `reactor:0.8.x`.

smaldini:
A few comments: 
- travis doesn't seem tolike the latest rebase, looks like an issue with Supplier[] vs Supplier<>[] :(
- we use package scope instead of private
- UriEndpoint could be final

lhotari:
@smaldini Thanks for the quick improvement suggestions. I fixed the immediate problems and the build is now passing.

quanticc:
While using this PR on 0.8.0.BUILD-SNAPSHOT I've discovered that the URL_MATCHER regex doesn't properly handle query parameters without a path.

Adding this case to the ``shouldParseUrls`` method inside UriEndpointFactoryTests file makes it fail.
```java
new String[]{"wss://localhost?query=value&query2=value2", "wss", "localhost", null, "?query=value&query2=value2"}
```

```
org.junit.ComparisonFailure: 
Expected :localhost
Actual   :localhost?query=value&query2=value2
```

The query parameter ends up in group 2 because of the missing the trailing slash due to the URL not having any path (which is permitted), therefore group 4 ends up being null.

This eventually cascades into an UnknownHostException when you call a method such as httpClient.ws() because the resulting URL adds a trailing slash like this: `localhost?query=value&query2=value2/` (due to not detecting anything as path + query)

Changing the regex to [/?] instead of just / under group 4 makes it work for this particular case.

lhotari:
Thanks for reporting the issue with this PR. I have created #228 to fix the issue.

PR #206, opened 2017-11-01
inner.justFlushed is not needed, hasPendingWriteBytes used instead.

violetagg:
DO NOT COMMIT: this change depends on the change in Netty: SslHandler and update of the pending data. Once 4.1.17.Final is out we can remove the `@Ignore` marker added to the tests and then commit the change.


smaldini:
I’d prefer to wait 4.1.17 before merging this one wdyt?

PR #207, opened 2017-11-01
Invoke pendingWriteOffer.test in event loop

violetagg:


smaldini:
Can increase the coverage by using send(flux.publishOn(Schedulers.single())

PR #228, opened 2017-11-28
Fix bug in UriEndpointFactory (0.8.x only): handle parsing URI with query and no path

lhotari:
- handle parsing URI with query and no path
- [fixes issue reported by @quanticc in the comments of the original PR](https://github.com/reactor/reactor-netty/pull/195#issuecomment-346952336) that introduced `UriEndpointFactory`.


violetagg:
Thanks for the fix
Violeta

PR #243, opened 2017-12-18
fix #232: Ensure write chunks are always in a correct order.

### LINKED ISSUE #232: HttpClient corrupts uploaded content
See test case in this repository https://github.com/rstoyanchev/spr16246.

HttpClient is used to send content from a large file. File content is loaded with `java.nio.channels.AsynchronousFileChannel` which fills one chunk of 4096 bytes at a time and invokes a callback to provide it. The chunk is then pushed into a `FluxSink`. Then we call read again and wait for the next callback, and so on until all chunks are processed. As a result chunks can be pushed in rapid succession from multiple threads but not concurrently since the next read doesn't start until `onNext` returns.

The server is the MockWebServer from OkHttp so it simply records the received content. The received content matches the file content by size but the content is out of order.

Also note that if the input Flux is cached and subscribed to before it is sent, it works fine. That proves the content is provided correctly.

Originally reported in [SPR-16246](https://jira.spring.io/browse/SPR-16246).

violetagg:
There are use cases when chunks of one write operation are written
either using event loop or external thread. The implementation should
ensure that if the previous chunks are still pending then the next chunks
will be scheduled for write regardless whether in event loop or not.

### DIFF
--- a/src/main/java/reactor/ipc/netty/channel/ChannelOperationsHandler.java
+++ b/src/main/java/reactor/ipc/netty/channel/ChannelOperationsHandler.java
@@ -505,6 +505,7 @@ final class ChannelOperationsHandler extends ChannelDuplexHandler
 		long           produced;
 		ChannelPromise promise;
 		ChannelFuture  lastWrite;
+		boolean        lastThreadInEventLoop;
 
 		PublisherSender(ChannelOperationsHandler parent) {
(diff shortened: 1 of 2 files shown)


smaldini:
We also suffer this issue in the other writeAndFlush used in the implementations - can be addressed separately or here

PR #252, opened 2018-01-04
fix #150 Expose Netty HttpServerCodec options in HttpServer

### LINKED ISSUE #150: Add support for configuring the HttpRequestDecoder when creating an HttpServer
A Spring Boot user has [raised an issue](https://github.com/spring-projects/spring-boot/issues/9955) that's caused by Netty's default configuration that limits the initial line of an HTTP request to 4096 bytes. As far as I can tell, Reactor Netty doesn't allow this to be configured as [the default `HttpRequestDecoder` constructor is used](https://github.com/reactor/reactor-netty/blob/f895546df5ecff24340366c8772d7b71a58c6195/src/main/java/reactor/ipc/netty/http/server/HttpServer.java#L210).

Would it be possible to expose the configuration options offered by `HttpRequestDecoder` to users?

simonbasle:
This commits exposes options from Netty's HttpServerCodec (that is
maxInitialLineLength, maxHeaderSize and maxChunkSize) through our
HttpServerOptions. This should allow the codec to decode requests
on very long URIs.

@smaldini: did that as a quick fix before the API changes in 0.8 are settled, would need introduction into 0.8.

### DIFF
--- a/src/main/java/reactor/ipc/netty/http/server/HttpServer.java
+++ b/src/main/java/reactor/ipc/netty/http/server/HttpServer.java
@@ -254,7 +254,10 @@ public final class HttpServer
 
         @Override
         public void accept(ChannelPipeline p, ContextHandler<Channel> c) {
-            p.addLast(NettyPipeline.HttpCodec, new HttpServerCodec());
+            p.addLast(NettyPipeline.HttpCodec, new HttpServerCodec(
+            		options.httpCodecMaxInitialLineLength(),
+		            options.httpCodecMaxHeaderSize(),
+		            options.httpCodecMaxChunkSize()));
 
             if(options.minCompressionResponseSize() >= 0) {
                     p.addLast(NettyPipeline.CompressionHandler, new CompressionHandler(options.minCompressionResponseSize()));
--- a/src/main/java/reactor/ipc/netty/http/server/HttpServerOptions.java
+++ b/src/main/java/reactor/ipc/netty/http/server/HttpServerOptions.java
@@ -38,10 +38,16 @@ public final class HttpServerOptions extends ServerOptions {
 	}
 
 	private final int minCompressionResponseSize;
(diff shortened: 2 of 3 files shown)


wilkinsona:
I've tried the proposed changes using the custom snapshots that @simonbasle kindly published. It does the trick from Boot's perspective, allowing a user to do the following:

```java
@Bean
public NettyReactiveWebServerFactory netty() {
    NettyReactiveWebServerFactory netty = new NettyReactiveWebServerFactory();
    netty.addServerCustomizers((builder) -> {
        builder.httpCodecOptions(8192, 0, 0);
    });
    return netty;
}
```

This increases the maximum initial line length to 8192 (and leaves the maximum header size and chunk size unchanged). FWIW, I think the API would produce more readable code if there were three separate methods. This would change the code above to something like the following:

```java
@Bean
public NettyReactiveWebServerFactory netty() {
    NettyReactiveWebServerFactory netty = new NettyReactiveWebServerFactory();
    netty.addServerCustomizers((builder) -> {
        builder.maxInitialLineSize(8192);
    });
    return netty;
}
```

IMO, this makes it much clearer that it's the max initial line size that's being set and doesn't leave you wondering what the two `0` parameters are going to do.

simonbasle:
@wilkinsona agreed this would be a better API 👍
@smaldini should I redo this on top of the 0.8 branch, with the above suggestion integrated? What does 0.8 changes for `HttpServerOptions`?

violetagg:
Do we want configuration for all settings?
```
        public HttpServerRequestDecoder(int maxInitialLineLength, int maxHeaderSize, int maxChunkSize,
                                        boolean validateHeaders, int initialBufferSize) {
            ...
        }

```

wilkinsona:
I don't think it would do any harm to offer all of them.

simonbasle:
closing this, superseded by #256

PR #258, opened 2018-01-10
fix #212: Add support for KQueue

### LINKED ISSUE #212: Add support for KQueue
### Expected behavior

Netty supports Kqueue and should de-facto use this when preferNative is true. We can discuss further down (0.8) if epoll or kqueue should be in dependencies by default (maybe not) to help user fine tune his preferred support.

### Actual behavior

Currently using DefaultLoopEpollDetector to fallback into epoll but we could repurpose the class to be more generic and support KQueue groups and channel factory. 

### Reactor Netty version

`0.7.1`

violetagg:


build.gradle
```diff
@@ -212,6 +212,7 @@ configure(rootProject) { project ->
 	compile "io.netty:netty-handler-proxy:${nettyVersion}"
 	compile "io.netty:netty-codec-http:${nettyVersion}"
 	compile "io.netty:netty-transport-native-epoll:${nettyVersion}"
+	compile "io.netty:netty-transport-native-kqueue:${nettyVersion}"
```
smaldini: will need to make that provided like slf4j in reactor-core.

violetagg: I made it optional

PR #261, opened 2018-01-16
fix #94 : Allow setting chunkedTransfer per HttpClient

### LINKED ISSUE #94: Add Global Option to Disable Chunked Transfer
Currently in order to disable chunked transfer encoding, each request must be modified setting `chunkedTransfer`.  If you want to code an application that is connecting to a network device that cannot support chunked transfer you must disable this option for each and every request.

Instead of requiring this for each request there should be a global option to disable this behavior for all uses of a specific `HttpClient`.

violetagg:
The support for chunkedTransfer per HttpClientRequest is removed.

### DIFF
--- a/src/main/java/reactor/ipc/netty/http/client/HttpClient.java
+++ b/src/main/java/reactor/ipc/netty/http/client/HttpClient.java
@@ -255,6 +255,15 @@ public abstract class HttpClient {
 	}
 
 	/**
+	 * Enable transfer-encoding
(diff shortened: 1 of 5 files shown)


poutsma:
Any chance of getting this back ported to 0.7.x? See https://jira.spring.io/browse/SPR-15914

vavimayor159:
Totally, it would be nice to have it on 0.7.x

PR #266, opened 2018-01-30
Fix IllegalArgumentException: Channel [...] was not acquired from this ChannelPool

violetagg:


smaldini:
Might want to add extra log for failure in debug mode.

violetagg:
> Might want to add extra log for failure in debug mode.

done

PR #267, opened 2018-01-30
On channelInactive event, do not fire error when the inbound has been already cancelled/disposed

violetagg:


smaldini:
I wonder if we should check for inbound cancel or disposed (cancel or done). The later case should not happen but better to be defensive maybe.

violetagg:
> I wonder if we should check for inbound cancel or disposed (cancel or done). The later case should not happen but better to be defensive maybe.

Added check for disposed also.

Issue #222, opened 2017-11-19
TLS Hostname verification is not performed


imgx64:
### Expected behavior
When connecting to a server and the server returns a valid certificate that doesn't correspond to its DNS name, the connection is accepted. For example: https://wrong.host.badssl.com/

### Actual behavior
Hostname verification should be done according to [RFC6125](https://tools.ietf.org/html/rfc6125) and the connection should be terminated.

### Steps to reproduce
```
HttpClient.create().get("https://wrong.host.badssl.com/").block();
```

### Reactor Netty version
0.7.1.RELEASE

### JVM version (e.g. `java -version`)
1.8.0_152

### OS version (e.g. `uname -a`)
Windows 10 version 1703

philwebb:
Isn't this one a bug, and quite important? Someone could do a man-in-the-middle attack. Should it not be targeted to 0.7.x?

philwebb:
I'm not sure if it's related, but I've just seen this error on Spring Boot

```
[ERROR] Tests run: 19, Failures: 0, Errors: 1, Skipped: 1, Time elapsed: 1.823 s <<< FAILURE! - in org.springframework.boot.web.embedded.tomcat.TomcatReactiveWebServerFactoryTests
[ERROR] sslNeedsClientAuthenticationSucceedsWithClientCertificate(org.springframework.boot.web.embedded.tomcat.TomcatReactiveWebServerFactoryTests)  Time elapsed: 0.159 s  <<< ERROR!
reactor.core.Exceptions$ReactiveException: javax.net.ssl.SSLException: SSLEngine closed already
	at reactor.core.Exceptions.propagate(Exceptions.java:326)
	at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:87)
	at reactor.core.publisher.Mono.block(Mono.java:1162)
	at org.springframework.boot.web.reactive.server.AbstractReactiveWebServerFactoryTests.testClientAuthSuccess(AbstractReactiveWebServerFactoryTests.java:202)
	at org.springframework.boot.web.reactive.server.AbstractReactiveWebServerFactoryTests.sslNeedsClientAuthenticationSucceedsWithClientCertificate(AbstractReactiveWebServerFactoryTests.java:213)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #275, opened 2018-02-07
Enable hostname verification in client SSL config

bclozel:
This is a fix for gh-222.

Netty client doesn't enable hostname verification in its client SSL Context by default (see [this warning](https://netty.io/4.1/api/io/netty/handler/ssl/SslContext.html#newHandler-io.netty.buffer.ByteBufAllocator-)).

I've just tested that change against badssl.com - but I'm willing to improve that PR in any way you'd like.

In order to make other tests pass, I had to use Netty's `InsecureTrustManagerFactory.INSTANCE` and I don't know if this is the right solution vs. using registering the self signed cert as previously. The self signed cert is generated with `"example.com"` as a domain. It looks like hostname verification will resolve the real cert associated with `"example.com"` and won't find a valid cert path between the generated and the real one.

It looks like Netty itself is [doing this when testing SSL-related scenarios](https://github.com/netty/netty/blob/4.1/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java).

### DIFF
--- a/src/main/java/reactor/ipc/netty/http/client/HttpClientOptions.java
+++ b/src/main/java/reactor/ipc/netty/http/client/HttpClientOptions.java
@@ -22,13 +22,19 @@ import java.net.URI;
 import java.util.Objects;
 import java.util.function.Function;
 
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLParameters;
+
 import io.netty.bootstrap.Bootstrap;
+import io.netty.buffer.ByteBufAllocator;
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslContextBuilder;
+import io.netty.handler.ssl.SslHandler;
 import io.netty.util.NetUtil;
 import reactor.ipc.netty.options.ClientOptions;
 import reactor.ipc.netty.options.ClientProxyOptions;
 import reactor.ipc.netty.options.ClientProxyOptions.Proxy;
+import reactor.util.function.Tuple2;
 
 /**
  * An http client connector builder with low-level connection options including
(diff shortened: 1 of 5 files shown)


violetagg:
@unlimitedsola In 0.8.x you can provide SslHandler configuration:

```
reactor.netty.http.client.HttpClient#secure(java.util.function.Consumer<? super reactor.netty.tcp.SslProvider.SslContextSpec>)
...
sslContextSpec
    .sslContext(SslContextBuilder)
    .defaultConfiguration(SslProvider.DefaultConfigurationType)
    .handlerConfigurator(handler -> {
        SSLEngine sslEngine = handler.engine();
        SSLParameters sslParameters = sslEngine.getSSLParameters();
        // extend/remove ssl params
        sslEngine.setSSLParameters(sslParameters);
    });
```

unlimitedsola:
@violetagg Thanks for your explanation, I'm wondering if Spring Boot 2.0.5.RELEASE compatible with reactor-netty 0.8.x? if not, is there any estimated time of being available?

violetagg:
> Thanks for your explanation, I'm wondering if Spring Boot 2.0.5.RELEASE compatible with reactor-netty 0.8.x? if not, is there any estimated time of being available?

No Spring Boot 2.0.5.RELEASE is not compatible with Reactor Netty 0.8.x
Spring Boot 2.1.x will be compatible with Reactor Netty 0.8.x. Spring Boot 2.1.0.M4 is scheduled for 24.09 (https://github.com/spring-projects/spring-boot/milestones)

If you need this for Reactor Netty 0.7.x please create and enhancement issue.

unlimitedsola:
I see, I'll wait for Spring Boot 2.1.0 coming out. Thank you again.

PR #288, opened 2018-02-17
POC for context transferring

OlegDokuka:
# The problem 

During preparation my demo project which is based on snapshot spring-boot + spring-webflux + spring-security I have found that Reactor 3.1.X feature, called context, is loosing somewhere between `org.springframework.web.reactive.socket.server.upgrade.ReactorNettyRequestUpgradeStrategy` and  actual `org.springframework.web.reactive.socket.WebSocketHandler` implementation. After the day of playing and debugging reactor-netty, I have found that actual problem is sitting in `reactor.ipc.netty.NettyOutbound`. The cause is that downstream (which is `WebFilter` chain + `WebHandler` (default impl is `DispatcherHandler`)) may produce `Context`, and which is populating to upstream (which is `WebSocketHandler`). During that path, at some point channels, responseHolders and data-producers are composing with-in one method which is https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/ipc/netty/NettyOutbound.java#L296, 
As it might be noted, downstream does not subscribe directly to upstream, so downstream context can't be passed to upstream `WebSocketHandler`. (which means that SecurityContext passed from `org.springframework.security.web.server.context.ReactorContextWebFilter` will not be accessible) 

# The solution

After some playing with NettyContext and Channels, I have found a naive solution which is passing downstream `Context` over Channels.attr(...).set() and consuming that attr in `reactor.ipc.netty.channel.ChannelOperationsHandler#drain`

So, I hope that this PR will produce productive discussion and we will found solution for that problem :)


OlegDokuka:
Here is reproducible example
[issue-example.zip](https://github.com/reactor/reactor-netty/files/1734264/issue-example.zip)

smaldini:
CC @rwinch / @rstoyanchev - 
Good catch @OlegDokuka, let me double check if we can simplify some of this but looks legit .

OlegDokuka:
Sure :)

OlegDokuka:
Also, if overall idea is good just comment on parts wich should be rewritten and then I will add some tests to cover all corener cases

src/main/java/reactor/ipc/netty/channel/ChannelOperationsHandler.java
```diff
@@ -458,7 +462,17 @@ else if (future instanceof ChannelPromise) {
 						else {
 							innerActive = true;
 							inner.promise = promise;
-							p.subscribe(inner);
+							AttributeKey<Object> attributeKey = AttributeKey.valueOf("subscriberContext");
+
+							if(ctx.channel()
+									.hasAttr(attributeKey)) {
+								p.subscribe(new ContextStartSubscriber<>(inner, inner.currentContext()
```
smaldini: instead of a new ContextStartSubscriber why not Flux.from(p).subscribe(inner, currentContext.putAll(attr.get()))

OlegDokuka: Hah, will fix it, I did not know about that method :)

src/main/java/reactor/ipc/netty/NettyOutbound.java
```diff
@@ -308,8 +313,12 @@ default NettyOutbound sendObject(Publisher<?> dataStream) {
 	 * any error during write
 	 */
 	default NettyOutbound sendObject(Object msg) {
-		return then(FutureMono.deferFuture(() -> context().channel()
-		                                                  .writeAndFlush(msg)));
+		return then(FutureMono.deferFuture((sctx) -> {
+			Channel channel = context().channel();
```
smaldini: I think its dangerous to use ctx to store the context here, maybe a wrapping PublisherEvent would suit this need better (with source and optional context). Further simplification could try if context is not empty before wrapping.

OlegDokuka: What is `PublisherEvent` my IDEA does not help me in searching for that class. Do you mean `writeAndFlush` something like 

```
class PublisherEvent {
Object source;
Context downstreamContext;
}
```

smaldini: It would be an immutable/ internal static inner final class. You can apply it only if Object is Publisher in addition to only if the context is not empty

smaldini:
Overall idea is def awesome

smaldini:
Do. My bad

OlegDokuka:
Dramatically simplified context passing without additional data structures to transfer it from downstream to upstream

OlegDokuka:
ping

smaldini:
Better, I think the PublisherContext can be local to channel package (I don't see the check in ChannelOperationHandler btw we need it to verify if instanceof PublisherContext and appropriately subscribe with the context then)

OlegDokuka:
Why do we need additional verification for `PublisherContext`?

OlegDokuka:
So in that case, `PublisherContext` plays utils-class role, so it allows wrapping incoming Publisher into Flux/Mono and using `.subscriberContext` transfer downstream context without additional wrapper's structure (`PublisherContext#withContext` returns the same `Publisher` but with additional `.sibscribeContext` operator) (https://github.com/reactor/reactor-netty/pull/288/files#diff-0eb319ea299bac0141a616c4735379e1R47), (https://github.com/reactor/reactor-netty/pull/288/files#diff-5e17fb740d162b382f10cedde81c13cdR299)

smaldini:
My bad its actually well simplified with the use of subscriberContext(Context) operator, didn't see that.
Alternatively (and I'm not saying that is better), it would have been checked for and unwrapped in the ChannelOpHandler. However we can still package-protect hide this util anyway for now.

Well done @OlegDokuka this might make it before our next maintenance release!

OlegDokuka:
made `PublisherContext` as package private. Increased code-coverage. Squashed everything as one commit

OlegDokuka:
ping

PR #290, opened 2018-02-19
fix #289 Register pool-releasing close listener only if acquired

### LINKED ISSUE #289: Regression: rejected client connection decrement pool counter
After fix for #285, spring-messaging has observed a failing test scenario with the following stack:
```
2018-02-18 14:30:27,578 WARN [tcp-client-loop-nio-1] [io.netty.util.concurrent.DefaultPromise] - An exception was thrown by io.netty.channel.pool.FixedChannelPool$4.operationComplete()
java.lang.AssertionError: null
	at io.netty.channel.pool.FixedChannelPool.decrementAndRunTaskQueue(FixedChannelPool.java:324) ~[netty-transport-4.1.21.Final.jar:4.1.21.Final]
	at io.netty.channel.pool.FixedChannelPool.access$600(c.java:38) ~[netty-transport-4.1.21.Final.jar:4.1.21.Final]
	at io.netty.channel.pool.FixedChannelPool$4.operationComplete(FixedChannelPool.java:305) ~[netty-transport-4.1.21.Final.jar:4.1.21.Final]
```

The stack points to an illegal counting occurring in `FixedChannelPool`. The scenario involves an immediately rejected connection to reproduce.

simonbasle:
Instead of setting up the close listener in the ChannelPoolHandler
channelCreated method, set it up in the PooledClientContextHandler.

This should prevent the listener to be invoked after a FixedChannelPool
has determined the channel was not healthy and has already decremented
its internal acquire counter (which would lead to a second decrement of
said counter and an assertion error).

### DIFF
--- a/src/test/java/reactor/ipc/netty/resources/DefaultPoolResourcesTest.java
+++ b/src/test/java/reactor/ipc/netty/resources/DefaultPoolResourcesTest.java
@@ -15,20 +15,32 @@
  */
 package reactor.ipc.netty.resources;
 
+import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import io.netty.bootstrap.Bootstrap;
 import io.netty.channel.Channel;
 import io.netty.channel.DefaultEventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.pool.ChannelPool;
+import io.netty.channel.socket.nio.NioSocketChannel;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.Promise;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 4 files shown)


simonbasle:
please heavily challenge the relevance of this fix. adding a little background on the issue itself.

src/test/java/reactor/ipc/netty/resources/DefaultPoolResourcesTest.java
```diff
@@ -114,4 +126,68 @@ public void disposeOnlyOnce() {
 		assertThat(closed.get()).as("pool closed only once").isEqualTo(1);
 	}
 
+	@Test
+	public void fixedPoolTwoAcquire()
+			throws ExecutionException, InterruptedException, IOException {
+		final ScheduledExecutorService service = Executors.newScheduledThreadPool(2);
+		int echoServerPort = SocketUtils.findAvailableTcpPort();
```
garyrussell: FYI, this is not entirely reliable and can cause problems on CI servers, if some other process grabs the socket. We've seen this happen once in a while, causing bogus CI build failures. 

In my projects, where possible, we've changed tests to listen on port 0 (so the OS selects the port) and then query the test server for which port was actually selected.

Not critical but worth consideration.

simonbasle: hard to do as the first part of the test checks what happens when the host is initially unreachable. crossing fingers for CI :(

PR #319, opened 2018-04-11
Don't share pipeline state in fields of the FileChunkedStrategy instance

lhotari:
- replace shared state with check each time
- remove redundant check in NettyOutbound.sendFileChunked


violetagg:
Thanks

PR #349, opened 2018-05-06
Enable errorprone compiler plugin in build

lhotari:
- enable [errorprone](http://errorprone.info/) compiler plugin in build
- for now, errorprone errors are converted to warnings with the [`-XepAllErrorsAsWarnings` flag](http://errorprone.info/docs/flags)


violetagg:
#350 will track the fixes for the warnings produced by this plugin

PR #356, opened 2018-05-14
 Update Netty version to 4.1.25.Final

lhotari:
- Contains a change to a test since it broke after upgrading
- Netty 4.1.25 contains https://github.com/netty/netty/pull/7891 , which changed the behaviour
  of HTTP status code 205 / Reset Content
  - the behaviour now conforms to [RFC 7231, section 6.3.6](https://tools.ietf.org/html/rfc7231#section-6.3.6)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,7 +58,7 @@ ext {
   assertJVersion = '3.6.1'
 
   // Libraries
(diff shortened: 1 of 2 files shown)


violetagg:
Thanks

Issue #376, opened 2018-06-19
Reactor Netty is slower to start than Tomcat


sdeleuze:
This is not obvious with Spring Boot but [Spring Fu minimal webapp benchmarks](https://github.com/spring-projects/spring-fu/tree/master/samples/minimal-webapp) seems to indicate than Reactor Netty is slower to start than Tomcat.

**Rector Netty:**
 Application started in 0.609 seconds (JVM running for 0.879)
 Application started in 0.676 seconds (JVM running for 0.97)
 Application started in 0.672 seconds (JVM running for 0.958)

**Tomcat:**
 Application started in 0.440 seconds (JVM running for 0.709)
 Application started in 0.440 seconds (JVM running for 0.723)
 Application started in 0.445 seconds (JVM running for 0.72)

This should maybe be explored and optimized.

To run Spring Fu minimal webapp with Reactor Netty:
```
git clone https://github.com/spring-projects/spring-fu.git
./gradlew clean build
java -XX:TieredStopAtLevel=1 -noverify -jar target/demo-startup-kotlin-0.0.1-SNAPSHOT.jar
```

To use Tomcat, update `samples/minimal-webapp/build.gradle.kts` with:
```
dependencies {
	implementation(project(":modules:webflux-tomcat"))
	testImplementation(project(":modules:webflux-netty"))
	testImplementation(project(":modules:test"))
}
```
And `Application.kt` with `server(tomcat())`.

To use Undertow, update `samples/minimal-webapp/build.gradle.kts` with:
```
dependencies {
	implementation(project(":modules:webflux-undertow"))
	testImplementation(project(":modules:webflux-netty"))
	testImplementation(project(":modules:test"))
}
```
And `Application.kt` with `server(undertow())`.

violetagg:
Is it correct that this is with Reactor Netty 0.7.x and not the current master?

sdeleuze:
Nope, this is using Reactor Netty `0.8.BUILD-SNAPSHOT` builds since Spring Fu depends on Spring Framework 5.1 snapshots.

smaldini:
Linked to #382

sdeleuze:
@violetagg I will be on PTO next week so I won't be able to do these test, but I suggest to test with `4.1.26.Final-SNAPSHOT` as it fix the thread started in initializer issue. Also in your tests Reactor Netty is still significantly slower than Tomcat, right?

violetagg:
With the latest snapshot these are the measurements:

[...]

PR #386, opened 2018-07-11
Related to #376: Do not create default SSL client provider when creating a server

violetagg:


violetagg:
With this change the times are
```
Application started in 0.463 seconds (JVM running for 0.788)
Application started in 0.289 seconds (JVM running for 0.604)
Application started in 0.288 seconds (JVM running for 0.598)
Application started in 0.299 seconds (JVM running for 0.608)
```

PR #396, opened 2018-07-23
fix #368 Throw a marker exception with local address information when server bind fails

### LINKED ISSUE #368: Expose read-only port/address when server binding errors
With the new `HttpServer` and `DisposableServer` API in Reactor Netty 0.8, we seem to be missing a way to get the port information from the server itself.

It is available from `DisposableServer`, but we can't get that information in case the server could not bind (in that case, `DisposableServer` is null and we're getting back an exception).

This issue is basically a new instance of #66, which was being used by Spring Boot so far.

violetagg:


src/main/java/reactor/netty/tcp/TcpServerBind.java
```diff
@@ -179,8 +182,12 @@ public final void operationComplete(ChannelFuture f) {
 					}
 					return;
 				}
-				if (f.cause() != null) {
-					sink.error(f.cause());
+				Throwable t = f.cause();
+				if (t != null) {
+					if (t instanceof BindException) {
+						t = new DisposableServerBindException(t.getMessage() + " " + localAddress, localAddress);
```
simonbasle: Would it make sense to craft a dedicated message for the DSBE that is more explicit and keep the original `BindException` as a `cause`?

violetagg:
Close in favour of #398

PR #408, opened 2018-08-02
Expose connect() method for websocket connections

rdegnan:
Having access to the underlying Connection for websocket/http2 as in TCP makes it more natural to implement DuplexConnection in RSocket. So instead of having to use the 'handle' method and cast:

`.<DuplexConnection>handle((in, out) -> Mono.just(new WebsocketDuplexConnection((Connection) in)))`

We can instead do:

`.connect().map(WebsocketDuplexConnection::new);`


violetagg:
Aren't you able to use `HttpClient.ResponseReceiver#responseConnection`?

rdegnan:
I was thinking more about the http/2 case, allowing access to the underlying bi-directional stream. This may not be possible yet in the current API, in which case perhaps only the websocket part of this applies.

Issue #390, opened 2018-07-12
Configure timeout options on HttpClient with new API


bclozel:
With the new `HttpClient`/`TcpClient` configuration APIs, I've tried customizing the timeout options for the HTTP client.

While I can change the connection timeout like this:

```java
TcpClient tcpClient = TcpClient.create();
tcpClient.configure()
	.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, 1000);
HttpClient httpClient = HttpClient.from(tcpClient);
```

I cannot find a way to configure the read and write timeouts, as it was previously explained in this [StackOverflow answer](https://stackoverflow.com/a/49140846/22982). Is there still a way (or an easier way) to configure that part of the `HttpClient`?

violetagg:
@bclozel You can configure the timeouts like the sample below:

```
TcpClient tcpClient =
        TcpClient.create()
                        .option(ChannelOption.CONNECT_TIMEOUT_MILLIS, 1000)
                        .doOnConnected(connection ->
                                connection.addHandlerLast(new ReadTimeoutHandler(10))
                                                   .addHandlerLast(new WriteTimeoutHandler(10)));
HttpClient httpClient = HttpClient.from(tcpClient);
```

bclozel:
Thanks a lot @violetagg !

PR #427, opened 2018-09-01
Add Netty based timeouts to HttpClient

dave-fl:


violetagg:
Hi,

What is the benefit of having this in Reactor Netty as addition to the Netty's handlers (`ReadTimeoutHandler`, `WriteTimeoutHandler`, `IdleStateHandler`)?
See more:

1. https://github.com/reactor/reactor-netty/issues/390
2. https://stackoverflow.com/questions/48992992/set-timeout-in-spring-webflux-webclient/49140846#49140846

Thanks,
Violeta

dave-fl:
Did not know that it could be don that way, very useful, thank you.  Will close the PR.

PR #432, opened 2018-09-15
Handle parsing URI with query

quanticc:
Changes introduced in #228 are currently not on master. This PR cherry picks ffd10eb.


smaldini:
@quanticc you are on fire !

PR #466, opened 2018-10-16
HttpClient.WebsocketSender#send returns when upgrade is done.

violetagg:


violetagg:
Closing this PR, we need to configure the request first and then to upgrade.

PR #514, opened 2018-11-20
Improve HttpClient send(Publisher) for GET/HEAD

smaldini:
- mark content length
- do not use direct memory to avoid leak


violetagg:
Merged with 6be19c16a05a972b0c1609b12eceff61cfb8e606

PR #542, opened 2018-12-01
Added invoking parent.drain if future/promise exists

segabriel:
Closes: #546

### DIFF
--- a/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
+++ b/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
@@ -645,10 +645,12 @@ final class ChannelOperationsHandler extends ChannelDuplexHandler
 					if (!future.isSuccess()) {
 						// Returned value is deliberately ignored
 						promise.setFailure(Exceptions.addSuppressed(future.cause(), t));
+						parent.drain();
 						return;
 					}
 					// Returned value is deliberately ignored
 					promise.setFailure(t);
+					parent.drain();
 				});
 			}
 			else {
@@ -747,6 +749,7 @@ final class ChannelOperationsHandler extends ChannelDuplexHandler
 				// Returned value is deliberately ignored
 				promise.setFailure(future.cause());
 			}
+			parent.drain();
 		}
 
 		@Override


segabriel:
For now, the following test is not passed:
```
reactor.netty.http.server.HttpSendFileTests > sendFileAsync4096Negative FAILED
    org.junit.ComparisonFailure: expected:<[101, 114, 114, 111, 114]> but was:<null>
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at reactor.netty.http.server.HttpSendFileTests.doTestSendFileAsync(HttpSendFileTests.java:345)
        at reactor.netty.http.server.HttpSendFileTests.sendFileAsync4096Negative(HttpSendFileTests.java:290)
```

but in the master branch, it's not passed too.

violetagg:
@segabriel Rebase on top the latest changes. Let's see whether this time the test will pass.
Regards,
Violeta

segabriel:
@violetagg done, but the test is still as the failed one.

violetagg:
@segabriel Try again (I made some changes)

smaldini:
I am still wondering, the fix itself is correct but do we want to have this behavior ? Shouldn't we close on first error and not try to pull more publisher ?
In the issue it is mentioned that you dispose processor, which creates a cancellation exception. Have you tried just onComplete the processor instead ?

segabriel:
@smaldini 
> In the issue it is mentioned that you dispose processor, which creates a cancellation exception. Have you tried just onComplete the processor instead ?

yes I tried and `onComplete` has the same behavior as `dispose`. 

> do we want to have this behavior ?

In my opinion yes, because we do it in other cases, for example, when we emit no message and immediately close the publisher. Or we have set `.timeout(Duration)` and we have no result it will invoke `close`.

smaldini:
Will close for now - #546 should be closed once we document an alternative but we already are planning to not have such queue for publishers anymore in the next major version.

PR #543, opened 2018-12-03
fix #522 New HttpClient API to allow configuring redirects via a predicate

### LINKED ISSUE #522: Allow configuration of redirects to follow
### Expected behavior

Currently, using using `followRedirect` on `HttpClient` (or `HttpClientRequest`) will follow `301` and `302` redirects. There are other forms of redirect however - for example `307` and `308`.

It would be useful to, rather than specifying a `boolean`, specify a list of redirect codes to honour in `followRedirect`.

### Actual behavior

`followRedirect` takes a `boolean` and follows only `30[1,2]` redirects

### Steps to reproduce

Connect to site that uses a`307` redirect, the request will not be redirected. I encountered this on the [Nest API][1]

### Reactor Netty version

0.8.2-RELEASE

Related to [#235][2]

[1]: https://developers.nest.com/guides/api/how-to-handle-redirects
[2]: https://github.com/reactor/reactor-netty/issues/235

violetagg:
`reactor.netty.http.client.HttpClient#followRedirect(boolean)` will redirect when
`301`|`302`|`307`|`308`

`reactor.netty.http.client.HttpClient#followRedirect(
BiPredicate<HttpClientRequest,HttpClientResponse>)` will redirect based on
the predicate

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClientOperations.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientOperations.java
@@ -27,6 +27,7 @@ import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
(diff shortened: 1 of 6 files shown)


src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -296,7 +305,15 @@ public HttpClientRequest headers(HttpHeaders headers) {
 
 	@Override
 	public boolean isFollowRedirect() {
-		return redirectable && redirectedFrom.length <= MAX_REDIRECTS;
+		return (redirectable ||
+				(followRedirectPredicate != null && followRedirectPredicate.test(this, this))) &&
```
smaldini: Shouldn't invoke on accessor like isFollowRedirect

violetagg: New update is provided. Please take a look.

PR #557, opened 2018-12-20
Use ByteBuf#(write|read)CharSequence when transforming from/to String

violetagg:
ByteBuf#(write|read)CharSequence methods have optimisations for
CharsetUtil.US_ASCII
CharsetUtil.ISO_8859_1
CharsetUtil.UTF_8

### DIFF
--- a/src/main/java/reactor/netty/ByteBufFlux.java
+++ b/src/main/java/reactor/netty/ByteBufFlux.java
@@ -84,10 +84,14 @@ public final class ByteBufFlux extends FluxOperator<ByteBuf, ByteBuf> {
 
 	public static ByteBufFlux fromString(Publisher<? extends String> source, Charset charset, ByteBufAllocator allocator) {
 		Objects.requireNonNull(allocator, "allocator");
-		return new ByteBufFlux(Flux.from(source)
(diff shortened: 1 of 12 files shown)


src/main/java/reactor/netty/ByteBufFlux.java
```diff
@@ -84,10 +84,14 @@ public static ByteBufFlux fromString(Publisher<? extends String> source) {
...
 	public static ByteBufFlux fromString(Publisher<? extends String> source, Charset charset, ByteBufAllocator allocator) {
 		Objects.requireNonNull(allocator, "allocator");
-		return new ByteBufFlux(Flux.from(source)
-		                           .map(s -> allocator.buffer()
-		                                              .writeBytes(s.getBytes(charset))),
-		                       allocator);
+		return new ByteBufFlux(
+				Flux.from(source)
+				    .map(s -> {
+				        ByteBuf buffer = allocator.buffer();
```
smaldini: I wonder if we need an allocator at all (or by default it should be unpooled) since we wrap bytes. I wonder about `Unpooled.wrappedBytes()`

PR #562, opened 2019-01-03
fix #558 Add CONFIGURING step to the lifecycle listener

### LINKED ISSUE #558: HttpClient#doOnRequest doesnt execute BEFORE request is sent
### Expected behavior
`HttpClient#doOnRequest` listener should be invoked _before_ the request is sent, allowing to mutate it (eg. add a basic set of headers on every request).

### Actual behavior
It seems to be invoked _AFTER_ the request is sent. In the listener:
 - doing `req.requestHeaders().set("foo", "bar")` mutates the headers but is never used
 - doing `req.header("foo", "bar")` throws an `java.lang.IllegalStateException: Status and headers already sent`

### Steps to reproduce
```java
@Test
public void postMutatesHeaders() {
			.doOnRequest((req, con) -> req.header("foo", "dufafa"))
			.baseUrl("https://postman-echo.com/");

	String resp =
			client.post()
					.uri("/post")
					.send((req, outbound) -> {
						String foo = req.requestHeaders().get("foo");
						if (foo != null) {
							req.header("foo", foo+"bar");
						}
						else {
							req.header("foo", "baz");
						}
						return outbound.sendString(Flux.just("post"));
					})
					.response((res, byteBufFlux) -> byteBufFlux.asString())
					.blockFirst();

	//check what the echo server saw as headers
	Assertions.assertThat(resp).contains("\"foo\":\"dufafabar\"");
}
```

### Analysis
`HttpClientConnect` is always added as the first observer, _after_ `doOnRequest`'s `HttpClientDoOn` has set itself as observer, resulting if I understand correctly in HCC calling the handler and sending the request before the doOnRequest is invoked.

Both react to the `CONFIGURED` state, and due to the order `HttpClientConnect` "wins".

The workaround would be to add a new intermediate state, eg. `CONFIGURING`, letting some observers alter the request before `HttpClientConnect` does its thing. For instance, `doOnRequest` would match on this one.

`CONFIGURING` would be fired right before `CONFIGURED` everywhere the later is currently fired.

### Reactor Netty version
0.8.x

simonbasle:
This additional step lets listeners react before a request body is sent,
since once the CONFIGURED state is reached, it triggers sending the
request over the wire.

### DIFF
--- a/src/main/java/reactor/netty/ConnectionObserver.java
+++ b/src/main/java/reactor/netty/ConnectionObserver.java
@@ -102,6 +102,12 @@ public interface ConnectionObserver {
 		State CONFIGURED = ReactorNetty.CONFIGURED;
 
 		/**
+		 * Propagated when a connection is bound to a channelOperation but still
+		 * allowing for observers' configuration.
+		 */
+		State CONFIGURED_ALMOST = ReactorNetty.CONFIGURED_ALMOST;
+
+		/**
 		 * Propagated when a connection has been reused / acquired
 		 * (keep-alive or pooling)
 		 */
--- a/src/main/java/reactor/netty/ReactorNetty.java
+++ b/src/main/java/reactor/netty/ReactorNetty.java
@@ -467,6 +467,13 @@ public final class ReactorNetty {
 		}
 	};
 
+	final static ConnectionObserver.State CONFIGURED_ALMOST = new ConnectionObserver.State() {
+		@Override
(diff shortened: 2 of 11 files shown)


smaldini:
We'll redesign if needed with one event only but I'd like to see explore other routes. I'm concerned with the possible confusion of "Almost configured" and adding extra handler traversal for an http specific case

PR #563, opened 2019-01-03
Support X-Forwarded-Host AND X-Forwarded-For

trecloux:
Simply removed an else to support X-Forwarded-Host and X-Forwarded-For
for the same request.

Thx @clement-buchart for the spot.


trecloux:
Should I update the license header year ?

violetagg:
> Should I update the license header year ?

No need I updated the license header year for the project

PR #574, opened 2019-01-15
Do not create aggregated ByteBuf when publisher empty or first element empty

violetagg:
The PR is scheduled for `0.9`

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClientOperations.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientOperations.java
@@ -775,21 +775,32 @@ class HttpClientOperations extends HttpOperations<NettyInbound, NettyOutbound>
 
 		@Override
 		public Mono<Void> then() {
-			ByteBufAllocator alloc = parent.channel().alloc();
+			ByteBufAllocator alloc = parent.channel()
+			                               .alloc();
 			return Flux.from(source)
-			           .collect(alloc::heapBuffer, ByteBuf::writeBytes)
-			           .flatMap(agg -> {
-				           if (!HttpUtil.isTransferEncodingChunked(request) && !HttpUtil.isContentLengthSet(request)) {
-					           request.headers()
-					                  .setInt(HttpHeaderNames.CONTENT_LENGTH, agg.readableBytes());
-				           }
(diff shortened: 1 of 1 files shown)


src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -796,16 +796,28 @@ public NettyOutbound sendObject(Object message) {
...
+			                   if (buf != null && buf.readableBytes() > 0) {
+			                       return flux.collect(alloc::heapBuffer, ByteBuf::writeBytes)
+			                                  .flatMap(agg -> {
+			                                      if (!HttpUtil.isTransferEncodingChunked(request) &&
+			                                              !HttpUtil.isContentLengthSet(request)) {
+			                                          request.headers()
+			                                                 .setInt(HttpHeaderNames.CONTENT_LENGTH, agg.readableBytes());
+			                                      }
+			                                      return parent.then()
+			                                                   .thenEmpty(FutureMono.disposableWriteAndFlush(parent.channel(), Mono.just(agg)));
```
bsideup: I think you can drop `.then()` here

src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -796,16 +796,28 @@ public NettyOutbound sendObject(Object message) {
...
+			                                  });
+			                   }
+			               }
+			               return parent.then();
```
bsideup: `.then()` here is redundant, there is `.then` after `switchOnFirst` (or maybe that `.then` can be dropped)

src/main/java/reactor/netty/http/client/HttpClientOperations.java
violetagg: `.then()` cannot be dropped - on this step Reactor Netty sends the headers for the outbound message.

src/main/java/reactor/netty/http/client/HttpClientOperations.java
violetagg: `.then()` cannot be dropped - on this step Reactor Netty sends the headers for the outbound message.

PR #621, opened 2019-02-27
fix connection header can contain other tokens

flyinprogrammer:
https://bugzilla.mozilla.org/show_bug.cgi?id=691300

I'm just the messenger dealing with a broken Firefox 😄 when using a `ws` route.


violetagg:
Fix applied with 151fc57a443c55f0c9ff299c0d052d236c2dc341

Issue #539, opened 2018-11-30
Add Tracing support


smaldini:
Right now, https://github.com/spring-cloud/spring-cloud-sleuth provides direct support for Reactor-Netty. Some of this support if not all could be directly integrated in reactor-netty in order to provide OoB optional tracing (OpenTracing, Zipkin...).

cc @marcingrzejszczak .

adriancole:
opentracing is not a stable library api. netty tends to be at the bottom of the stack so probably not the best coupling choice.

adriancole:
ps by stable I mean 0.31 was binary incompatible with 0.30 and 0.32 (soon maybe?) will likely be incompatible with 0.31

smaldini:
Do you have any recommendations @adriancole - I think I was confused in the issue I meant to say Open Zipkin or simply Zipkin 👍

violetagg:
#1307 #1317 #1324

PR #622, opened 2019-02-28
Add HttpServerTracing adapter for Brave tracer

dmexe:
Changes:

* Add `HttpServerTracing` which decorates existing HTTP handlers
* Add `routeName` property to `HttpServerResponse`

At this moment client tracing cannot be implemented due to #558 

Related to #539


pivotal-issuemaster:
@dmexe Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=622)!

[Click here](https://cla.pivotal.io/sync/reactor?repositoryId=reactor/reactor-netty&pullRequestId=622) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@dmexe Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=622)!

smaldini:
That's awesome, we'll look at it. Maybe some input from @adriancole as well ?

src/main/java/reactor/netty/instrumentation/zipkin/HttpServerTracing.java
```diff
@@ -0,0 +1,102 @@
...
+ *         .build();
+ *   HttpServerTracing serverTracing = HttpServerTracing.create(tracing);
+ *
+ *   DisposableServer c = HttpServer.create()
+ *     .port(0)
+ *     .handle(serverTracing.andThen(myHandler))
+ *     .bindNow();
+ * </pre></code>
+ */
+public class HttpServerTracing {
```
adriancole: Usually, what we do is something like this...

```java
ReactorNettyTracing reactorTracing = ReactorNettyTracing.create(httpTracing);

....
 DisposableServer c = HttpServer.create()
    .port(0)
    .handle(reactorTracing.serverHandler().andThen(myHandler))
    .bindNow();

```

It isn't required, but more conventional as the top-level type (in this case `ReactorNettyTracing`) is supplied by the object graph. If this library is is server-only, it is probably also fine what you've done.

src/main/java/reactor/netty/instrumentation/zipkin/TracingContext.java
```diff
@@ -0,0 +1,90 @@
...
+import com.sun.istack.internal.NotNull;
+import java.util.function.Consumer;
+import reactor.util.annotation.NonNull;
+import reactor.util.annotation.Nullable;
+import reactor.util.context.Context;
+
+/**
+ * Span related methods for {@link Context}.
+ */
+public class TracingContext {
```
adriancole: nit if this is not a user-type, hide it?

src/test/java/reactor/netty/instrumentation/zipkin/HttpServerTracingTest.java
```diff
@@ -0,0 +1,343 @@
...
+import reactor.netty.http.server.HttpServer;
+import reactor.netty.http.server.HttpServerRequest;
+import reactor.netty.http.server.HttpServerResponse;
+import reactor.netty.http.server.HttpServerRoutes;
+import reactor.test.StepVerifier;
+import zipkin2.Annotation;
+import zipkin2.Span;
+import zipkin2.Span.Kind;
+
+public class HttpServerTracingTest {
```
adriancole: it is easy to miss edge cases, eventhough ad-hoc tests are good for understanding things locally. It would be nice to add the tests we use for all frameworks, to help understand what works vs doesn't.

ex https://github.com/openzipkin/brave/tree/master/instrumentation/vertx-web/src/test/java/brave/vertx/web

which implies a test dependency on
```
    <dependency>
      <groupId>io.zipkin.brave</groupId>
      <artifactId>brave-instrumentation-http-tests</artifactId>
      <scope>test</scope>
</dependency>
```

dmexe: Thanks, `http-tests` found some mistakes

src/main/java/reactor/netty/instrumentation/zipkin/TracingContext.java
dmexe: It's a decorator for reactive `Context` which provides span related methods for user level code.

for example:

```
Mono.subscriberContext()
  .flatMap(ctx -> {
    TracingContext.of(ctx).span(span -> span.annotate("log message"));
    return ...;
  });

**Comment by smaldini** [comment]
That PR is pretty cool, it made me think about either providing a better extension option on client/server level as we are improving builder API for 0.9 like : 

```java
      server.extension(TracingExtension)
                .handle(normalHandler)
       //...
```

Or in that case directly offering a `trace` method with a default support for Brave. That would require us to extract the necessary hook used here. We do the same with metrics() in reactor-core `Flux#metrics and micrometers.

I'm also thinking about routeName and how this will be symmetric for a similar support on the client side. Should we use a generic concept of tag instead that could be reused multiple times and by various frameworks (e.g. micrometer too) ? We could tag any operation.

src/main/java/reactor/netty/instrumentation/zipkin/TracingContext.java
adriancole: cool. heh so, I'll nag you then :) javadoc the example? :D

adriancole:
@smaldini on tag.. depends on the goal.

route is a mildly generic term, but not as generic as operation name (aka span name)
https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpAdapter.java#L50

We use it where possible to create a route-based span name. This makes things "just work" with no custom logic.
https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpParser.java#L63

OTOH an aribtrary tag could also be a custom name parsing input, and/or also a normal tag

I really like this PR also, btw!

shakuzen:
> Should we use a generic concept of tag instead that could be reused multiple times and by various frameworks (e.g. micrometer too) ?

It definitely helps for correlation if metrics and tracing use the same value (to a lesser degree, name also) for like concepts like HTTP route. This was one of the reasons [Haystack](https://expediadotcom.github.io/haystack/) gave for generating metrics after-the-fact from tracing data - it ensures the tags match.

src/main/java/reactor/netty/instrumentation/zipkin/TracingContext.java
dmexe: I've added Javadoc examples for each user level methods

src/main/java/reactor/netty/instrumentation/zipkin/HttpServerTracing.java
dmexe: I think it isn't necessary, `brave.http.HttpTracing` is a good abstraction for both client and server configurations and customizations, so `HttpServerTracing` does nothing except wrap an existing hander to provide tracing functionality.

dmexe:
> it made me think about either providing a better extension option
It would be great

> Or in that case directly offering a trace method with a default support for Brave. That would require us to extract the necessary hook used here. We do the same with reactor-core Flux#metrics and micrometer.

Direct depending on brave isn't good, opposed to micrometer which is a facade library and provides an API interface. Also, micrometer allows end-user using various metrics libraries (or just disable it). 

> I'm also thinking about routeName and how this will be symmetric for a similar support on the client side.

routeName for clients and servers should be an internal property. In the future, it would be easily constructed for a server from `HttpServerRoutes` predicates and for a `HttpClient` by extending an `uri` method to accept URI templates, for example `httpClient.uri("/items/{itemId}", itemId).get()`

violetagg:
> Changes:
> 
> * Add `HttpServerTracing` which decorates existing HTTP handlers
> * Add `routeName` property to `HttpServerResponse`
> 
> At this moment client tracing cannot be implemented due to #558
> 
> Related to #539

@dmexe #558 is fixed and available with `0.8.7.BUILD-SNAPSHOT`

adriancole:
hi, one site is asking about this now..

build.gradle
```diff
@@ -62,6 +62,7 @@ ext {
   // Libraries
   nettyVersion = '4.1.34.Final'
   jacksonDatabindVersion = '2.5.1'
+  braveVersion = "5.6.1"
```
adriancole: switch to 5.8.0 for some more powerful things.. especially allows secondary sampling

src/main/java/reactor/netty/instrumentation/zipkin/HttpCarrier.java
```diff
@@ -0,0 +1,41 @@
...
+package reactor.netty.instrumentation.zipkin;
+
+import brave.propagation.Propagation.Getter;
+import brave.propagation.Propagation.Setter;
+import io.netty.handler.codec.http.HttpHeaders;
+
+/**
+ * HTTP headers injector/extractor
+ */
+class HttpCarrier implements Getter<HttpHeaders, String>, Setter<HttpHeaders, String> {
```
adriancole: In 5.8 this type is no longer needed

src/main/java/reactor/netty/instrumentation/zipkin/HttpClientAdapter.java
```diff
@@ -0,0 +1,39 @@
+package reactor.netty.instrumentation.zipkin;
+
+import reactor.netty.http.client.HttpClientRequest;
+import reactor.netty.http.client.HttpClientResponse;
+
+/**
+ * Brave HTTP adapter for client requests
+ */
+class HttpClientAdapter extends brave.http.HttpClientAdapter<HttpClientRequest, HttpClientResponse> {
```
adriancole: In 5.8, implement brave's HttpClientRequest and HttpClientResponse instead.

src/main/java/reactor/netty/instrumentation/zipkin/HttpServerAdapter.java
```diff
@@ -0,0 +1,80 @@
...
+
+import brave.Span;
+import java.net.InetSocketAddress;
+import reactor.netty.http.server.HttpServerRequest;
+import reactor.netty.http.server.HttpServerResponse;
+
+/**
+ * Brave HTTP adapter for server requests
+ */
+class HttpServerAdapter extends brave.http.HttpServerAdapter<HttpServerRequest, HttpServerResponse> {
```
adriancole: In 5.8, implement brave's HttpServerRequest and HttpServerResponse instead.

src/main/java/reactor/netty/instrumentation/zipkin/HttpServerTracing.java
```diff
@@ -0,0 +1,109 @@
...
+  private final HttpServerHandler<HttpServerRequest, HttpServerResponse> handler;
+
+  /**
+   * Create a new tracing adapter using preconfigured {@link HttpTracing} instance.
+   */
+  private HttpServerTracing(@NonNull HttpTracing httpTracing) {
+    Objects.requireNonNull(httpTracing, "httpTracing cannot be null");
+
+    this.extractor = httpTracing.tracing().propagation().extractor(GETTER);
+    this.handler = HttpServerHandler.create(httpTracing, new HttpServerAdapter());
```
adriancole: in 5.8, no extractor, just `HttpServerHandler.create(httpTracing) `

src/main/java/reactor/netty/instrumentation/zipkin/HttpServerTracing.java
```diff
@@ -0,0 +1,109 @@
...
+  > andThen(
+      @NonNull BiFunction<
+          ? super HttpServerRequest,
+          ? super HttpServerResponse,
+          ? extends Publisher<Void>
+      > next) {
+    Objects.requireNonNull(next, "next cannot be null");
+
+    return (request, response) -> {
+      final Span span = handler.handleReceive(extractor, request.requestHeaders(), request);
```
adriancole: in 5.8, no extractor, just `handler.handleReceive(new YourHttpRequest(request));`

src/main/java/reactor/netty/instrumentation/zipkin/TracingContext.java
```diff
@@ -0,0 +1,113 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package reactor.netty.instrumentation.zipkin;
+
+import brave.Span;
```
adriancole: If this is a user type, please don't expose Span, expose SpanCustomizer instead. We have code notes why.

violetagg:
This PR is superseded by #1307 
@dmexe @adriancole @smaldini PTAL

violetagg:
Closing this as superseded by #1307

PR #626, opened 2019-03-03
fix #625 Update ConnectionInfo#parseAddress behavior to handle ipv6 addresses without bracket

### LINKED ISSUE #625: ipv6 forwarded headers cause NumberFormatException in ConnectionInfo
When netty handles a request containing an x-forwarded-* header with ipv6 format, no port, and no brackets, the last segment is taken as port and Integer.parse throws NumberFormatException on the hex or compressed 0. 

parseAddress should infer when ipv6 but no brackets to prevent parsing last segment.

PR [here](https://github.com/reactor/reactor-netty/pull/626)

### Steps to reproduce

Send request to reactor netty with 'x-forwarded-for: 1fff:0:a88:85a3::ac1f', this should result in NumberFormatException -> ConnectionInfo.java:121

### Reactor Netty version
v0.8.5

1.8.0_201-b09

RayJSeth:
parseAddress changed to check number of ':' separators to infer ipv4 vs. ipv6. For ipv6 addresses, only parse last segment if following correct bracket notation required for ipv6 with ports.
This allows for valid, no-port ipv6 addresses that do not use bracket notation to be provided in forward headers without throwing NumberFormatException.
Added specific tests for these cases and updated tearDown to account for added tests not creating connection object.

### DIFF
--- a/src/main/java/reactor/netty/http/server/ConnectionInfo.java
+++ b/src/main/java/reactor/netty/http/server/ConnectionInfo.java
@@ -115,14 +115,15 @@ final class ConnectionInfo {
 	}
 
 	static InetSocketAddress parseAddress(String address, int defaultPort) {
-		int portSeparatorIdx = address.lastIndexOf(":");
-		if (portSeparatorIdx > address.lastIndexOf("]")) {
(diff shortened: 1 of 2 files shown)


violetagg:
Thanks

Issue #155, opened 2017-08-22
Documentation & Examples


magx2:
Hi,

Do you have some tutorial/examples how to use this plugin with Netty? I think that Java docs are not sufficient. 

Thanks

magx2:
Any update on this?

stavalfi:
Is it a pre-requirement to control `Netty` before learning this library? If yes, which topics in `Netty`?

violetagg:
In most cases there is no such prerequisite.

kweheliye:
@violetagg 

Is there a way to have early access of the new 0.8.0 release?, just to see and test the new up coming features and changes.

Appreciate if you can point out. Thanks

violetagg:
@kweheliye Hey, yes sure, try our 0.8.0.M2 from https://repo.spring.io/libs-milestone

PR #627, opened 2019-03-04
Initial documentation

violetagg:
#155


violetagg:
> We should start with this and continue iterating - also will setup integration on the website

+1

PR #665, opened 2019-03-23
Do not release the channel twice

violetagg:
When a channel is first released to the pool and then closed the
exception below will be created, that's because `release` will be
called twice. In order to prevent the exception creation
check for channel attribute `channelPool` when `closed channel` event
is received and only if it is present then release the channel.

```
java.lang.IllegalArgumentException.<init>(String) IllegalArgumentException.java:52
io.netty.channel.pool.SimpleChannelPool.doReleaseChannel(Channel, Promise) SimpleChannelPool.java:300
io.netty.channel.pool.SimpleChannelPool.release(Channel, Promise) SimpleChannelPool.java:281
io.netty.channel.pool.SimpleChannelPool.release(Channel) SimpleChannelPool.java:271
reactor.netty.resources.PooledConnectionProvider$Pool.release(Channel) PooledConnectionProvider.java:254
reactor.netty.resources.PooledConnectionProvider$DisposableAcquire.lambda$registerClose$0(Channel, Future) PooledConnectionProvider.java:583
```

### DIFF
--- a/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -591,7 +591,9 @@ final class PooledConnectionProvider implements ConnectionProvider {
 			}
 			c.closeFuture()
 			 .addListener(ff -> {
-			     pool.release(c);
+			     if (AttributeKey.exists("channelPool")) {
+			         pool.release(c);
+			     }
 			     pool.inactiveConnections.decrementAndGet();
 			     if (log.isDebugEnabled()) {
 			         log.debug(format(c, "Channel closed, now {} active connections and {} inactive connections"),


jim2paterson:
Will this fix will prevent a negative inactive connection count, which I've been seeing in messages like the following showing -144?

`12:18:22.915 [nioEventLoopGroup-9-2] DEBUG r.n.r.PooledConnectionProvider - [id: 0xdec7e6bc, L:/10.28.176.242:62748 - R:example.net/10.26.0.92:8080] Channel connected, now 2 active connections and -144 inactive connections`

As context, I've been trying to make a simple recreate for a problem we see in production that causes a fixed connection pool to believe there are no items available even though all physical connections have been closed and there are no longer any open sockets.  It seems to involve some combination of connection time-out errors and fixed connection pool exhausted (time-out on acquire operation for pool).  So far it has proven quite difficult to come up with a simple recreation, but one of the symptoms I've encountered is a negative inactive connection count.

violetagg:
@jim2paterson Reactor Netty version?

jim2paterson:
0.8.5

violetagg:
@jim2paterson I don't think it is related, this change is only for not creating `IllegalArgumentException`

PR #689, opened 2019-03-27
Calls ConnectionObserver.onStateChange on disconnecting

### LINKED ISSUE #688: disconnecting state is not called on ConnectionObserver of TcpServer
Only connection status of connected and configured is called on ConnectionObserver of TcpServer, not disconnecting.

while the [Javadoc](https://github.com/reactor/reactor-netty/blob/5a59344236ea6a04383766693d61fc15d23a58e3/src/main/java/reactor/netty/tcp/TcpServer.java#L354) of `TcpServer.observe` says it should be called after disconnected.
>called on or after Channel has been connected and after it has been disconnected.

[Here](https://github.com/chang-chao/reactor-netty-samples/blob/ff4db82398200206798c48ff9e9c73e437e33e67/src/main/java/me/changchao/reactor_netty_samples/tcp/server/EchoServer.java#L22) is the simple program that can reproduce this issue.
When the simple EchoServer is started, we can test it by a  `echo text | nc localhost 9999`.

chang-chao:
fixes #688

### DIFF
--- a/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
+++ b/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
@@ -120,9 +120,7 @@ final class ChannelOperationsHandler extends ChannelDuplexHandler
 			if (ops != null) {
 				ops.onInboundClose();
 			}
-			else {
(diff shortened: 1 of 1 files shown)


violetagg:
#690

PR #692, opened 2019-03-28
fix #653 When a channel is acquired and closed immediately re-acquire only once

### LINKED ISSUE #653: Spring stomp client does not always reconnect to broker if using SSL
It appears there may be a race in connection pooling logic when an SSL endpoint does not perform any handshake.  This causes spring's stomp client to fail to reconnect under certain conditions.

Details (logs, etc.): https://github.com/spring-projects/spring-framework/issues/22080#issuecomment-474952213

We have:

app -> haproxy -> rabbitmq.

If rabbitmq goes down (all backends), haproxy will still accept the TCP connection, but, quickly close the connection without doing any TLS handshake.

In debugging, it was observed that this block was being hit, and after that no reconnection attempts are made:

https://github.com/reactor/reactor-netty/blob/0.8.x/src/main/java/reactor/netty/resources/PooledConnectionProvider.java#L624

After removing that block https://github.com/krm1312/reactor-netty/commit/04ff03f3478bfdf871822fe3454f2ce9de9b19f5 I have been unable to reproduce the issue.  Not sure of all the implications.

spring boot 2.1.3
spring 5.1.5
reactor-netty 2.8.5
netty 4.1.34
Java 8

violetagg:
When a channel is acquired and closed immediately, re-acquire only once.
If the second attempt also fails return an error and let the user decide to retry
or not.
Ensure the channel is released to the pool so that the pool updates the counters,
this is in case of FixedChannelPool.

### DIFF
--- a/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -158,7 +158,7 @@ final class PooledConnectionProvider implements ConnectionProvider {
 				pool.close();
 			}
 
-			disposableAcquire(sink, obs, pool);
+			disposableAcquire(sink, obs, pool, false);
 
 		});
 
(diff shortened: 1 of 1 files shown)


smaldini:
Can we have a test for this ? I was missing a scenario when i left the comment and sounds like we have one here

violetagg:
yep with `FixedChannelPool` we need to release in order to update `acquiredChannelCount` which is critical for acquire operation.

https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/pool/FixedChannelPool.java#L298
https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/pool/FixedChannelPool.java#L329

PR #697, opened 2019-04-02
Add proxy protocol support

aftersss:
We use `haproxy` at the front of `reactor-netty` program, so real ip of users can not easily fetched, we found that `Proxy Protocol`(https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) is widely used by many softwares(like mysql, nginx, and so on), this PR try to add proxy protocol support to `reactor-netty`.

### DIFF
--- a/src/main/java/reactor/netty/http/server/ConnectionInfo.java
+++ b/src/main/java/reactor/netty/http/server/ConnectionInfo.java
@@ -21,7 +21,6 @@ import java.util.regex.Pattern;
 
 import io.netty.channel.Channel;
 import io.netty.channel.socket.SocketChannel;
-import io.netty.handler.codec.haproxy.HAProxyMessage;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.util.AttributeKey;
@@ -52,8 +51,8 @@ final class ConnectionInfo {
 	static final String XFORWARDED_PORT_HEADER = "X-Forwarded-Port";
 	static final String XFORWARDED_PROTO_HEADER = "X-Forwarded-Proto";
 
-	static final AttributeKey<HAProxyMessage> PROXY_PROTOCOL_MESSAGE =
-			AttributeKey.valueOf("proxyProtocolMessage");
+	static final AttributeKey<InetSocketAddress> REMOTE_ADDRESS_FROM_PROXY_PROTOCOL =
+			AttributeKey.valueOf("remoteAddressFromProxyProtocol");
 
 	final InetSocketAddress hostAddress;
 
@@ -159,13 +158,11 @@ final class ConnectionInfo {
 	}
 
 	private static InetSocketAddress getRemoteAddress(SocketChannel channel) {
-		Object attr = channel.attr(PROXY_PROTOCOL_MESSAGE).getAndSet(null);
+		InetSocketAddress remoteAddressFromProxyProtocol
+				= channel.attr(REMOTE_ADDRESS_FROM_PROXY_PROTOCOL).getAndSet(null);
 
-		if (attr != null) {
-			HAProxyMessage proxyMessage = (HAProxyMessage) attr;
-			if (proxyMessage.sourceAddress() != null && proxyMessage.sourcePort() != 0) {
-				return new InetSocketAddress(proxyMessage.sourceAddress(), proxyMessage.sourcePort());
-			}
+		if (remoteAddressFromProxyProtocol != null) {
+			return remoteAddressFromProxyProtocol;
 		}
 
 		return channel.remoteAddress();
--- a/src/main/java/reactor/netty/http/server/ConnectionInfo.java
+++ b/src/main/java/reactor/netty/http/server/ConnectionInfo.java
@@ -159,10 +159,13 @@ final class ConnectionInfo {
 	}
 
 	private static InetSocketAddress getRemoteAddress(SocketChannel channel) {
-		HAProxyMessage proxyMessage = channel.attr(PROXY_PROTOCOL_MESSAGE).getAndSet(null);
+		Object attr = channel.attr(PROXY_PROTOCOL_MESSAGE).getAndSet(null);
 
-		if (proxyMessage != null && proxyMessage.sourceAddress() != null && proxyMessage.sourcePort() != 0) {
-			return new InetSocketAddress(proxyMessage.sourceAddress(), proxyMessage.sourcePort());
+		if (attr != null) {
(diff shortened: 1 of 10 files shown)


violetagg:
@aftersss Just to clarify - do you need this feature for 0.8.x or 0.9.x?
0.9.x M1 is still not scheduled.

src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -238,6 +238,24 @@ public final HttpServer forwarded(boolean forwardedEnabled) {
...
+	 * for deriving information about the address of the remote peer is enabled.
+	 *
+	 * @param proxyProtocolEnabled if true support for the {@code "HAProxy proxy protocol"}
+	 *                         for deriving information about the address of the remote peer is enabled,
+	 *                         otherwise disabled.
+	 * @return a new {@link HttpServer}
+	 */
+	public final HttpServer proxyProtocol(boolean proxyProtocolEnabled) {
+		if (proxyProtocolEnabled) {
+			return tcpConfiguration(PROXY_PROTOCOL_ATTR_CONFIG);
```
violetagg: What about updating the bootstrap directly (see `wiretap`)?

src/main/java/reactor/netty/http/server/ConnectionInfo.java
```diff
@@ -152,6 +158,16 @@ static ConnectionInfo parseXForwardedInfo(HttpRequest request, SocketChannel cha
 		return new ConnectionInfo(hostAddress, remoteAddress, scheme);
 	}
 
+	static InetSocketAddress getRemoteAddress(SocketChannel channel) {
+		HAProxyMessage proxyMessage = channel.attr(PROXY_PROTOCOL_MESSAGE).get();
```
violetagg: Once obtaining the information, do we need this attribute or we can remove it?

aftersss:
> @aftersss Just to clarify - do you need this feature for 0.8.x or 0.9.x?
> 0.9.x M1 is still not scheduled.

We are now using `0.8.x`, and we resolve this problem by adding some ChannelHandler ourselves, so no hurry, add this feature to 0.9.x is just ok.

src/main/java/reactor/netty/http/server/ConnectionInfo.java
aftersss: Yes, it can be removed.

src/main/java/reactor/netty/http/server/HAProxyMessageReader.java
```diff
@@ -0,0 +1,41 @@
...
+package reactor.netty.http.server;
+
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.handler.codec.haproxy.HAProxyMessage;
+
+/**
+ * Consumes {@link io.netty.handler.codec.haproxy.HAProxyMessage}
+ * and set it into channel attribute for later use.
+ */
```
violetagg: Put your name as author

src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -238,6 +239,34 @@ public final HttpServer forwarded(boolean forwardedEnabled) {
...
+										.addFirst(NettyPipeline.ProxyProtocolDecoder, new HAProxyMessageDecoder());
+								channel.pipeline()
+										.addAfter(NettyPipeline.ProxyProtocolDecoder,
+												NettyPipeline.ProxyProtocolReader, new HAProxyMessageReader());
+							})));
+		}
+		else {
+			return tcpConfiguration(tcpServer ->
+					tcpServer.bootstrap(b -> BootstrapHandlers.removeConfiguration(b, NettyPipeline.ProxyProtocolDecoder)));
+		}
```
violetagg: Remove `NettyPipeline.ProxyProtocolReader` also

src/main/java/reactor/netty/http/server/HttpServerBind.java
```diff
@@ -704,5 +707,4 @@ protected void initChannel(Channel ch) {
 			addStreamHandlers(ch, listener, forwarded, cookieEncoder, cookieDecoder);
 		}
 	}
-
```
violetagg: Revert all changes in this class

violetagg:
> And if this feature is better added to 0.8.x, I will be pleasure to create a new PR to replace this PR, should I do this?

I'll apply it to 0.8.x no worries

src/main/java/reactor/netty/http/server/HttpServer.java
aftersss: NettyPipeline.ProxyProtocolDecoder used here is just a Configuration name, not channelHander name, so NettyPipeline.ProxyProtocolReader is no need to remove here.

violetagg: oh yes, correct.
Thanks for checking that.

smaldini:
Are we ok with ConnectionInfo import on HAMessage if the dependency is optional (should we use a separate private class ?)

violetagg:
@smaldini @aftersss 
- What about if the `InetSocketAddress` is created in `HAProxyMessageReader` and set as an attribute there. Then in the `ConnectionInfo` we will check for that attribute and if it is there we will use it.
- When the dependency is missing what we want to do just warn the user or fail to create the server?

@aftersss 
- Use `InetSocketAddressUtil.createUnresolved` when creating the `InetSocketAddress`.
- In the `reactor.netty.http.server.HttpServer#proxyProtocol` catch a potential `ClassNotFoundException` if the dependency is not there.

aftersss:
OK, done.

src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -238,6 +240,42 @@ public final HttpServer forwarded(boolean forwardedEnabled) {
...
+	 *                         otherwise disabled.
+	 * @return a new {@link HttpServer}
+	 */
+	public final HttpServer proxyProtocol(boolean proxyProtocolEnabled) {
+		if (proxyProtocolEnabled) {
+			try {
+				Class.forName("io.netty.handler.codec.haproxy.HAProxyMessageDecoder");
+			}
+			catch (ClassNotFoundException e) {
+				throw new MissingDependencyException(
```
smaldini: Can we just use UnsupportedOperationException to avoid a new public type for now ?

violetagg:
Committed 334379a05b847e837b489f82354720c4eff5152f
Will be available in `0.8.7.RELEASE` and `0.9.0.M1`

Thanks,
Violeta

PR #709, opened 2019-04-09
Override reactor.netty.http.HttpDecoderSpec methods to keep compatibility in 0.8.x

violetagg:
This change is only for 0.8.x it will not be ported to 0.9.x


smaldini:
The pr targets master ?

PR #712, opened 2019-04-11
Clean up ServerBootstrap default options

violetagg:


src/main/java/reactor/netty/tcp/TcpServerBind.java
```diff
@@ -124,14 +124,11 @@ static void convertLazyLocalAddress(ServerBootstrap b) {
 	}
 
 	ServerBootstrap createServerBootstrap() {
-		return new ServerBootstrap().option(ChannelOption.ALLOCATOR, PooledByteBufAllocator.DEFAULT)
+		return new ServerBootstrap()
 				.option(ChannelOption.SO_REUSEADDR, true)
-				.option(ChannelOption.SO_BACKLOG, 1000)
 				.childOption(ChannelOption.ALLOCATOR, PooledByteBufAllocator.DEFAULT)
 				.childOption(ChannelOption.AUTO_READ, false)
```
smaldini: could this be moved in our default bootstrap handler (setting channel.config().setAutoRead(false) here

PR #720, opened 2019-04-19
fix #719 Filter the empty buffers emitted by the Publisher

### LINKED ISSUE #719: POSTing empty request body over HTTPS hangs on second attempt
If I submit two HTTPS POST requests using `HttpClient`, and the second request has an empty request body (content-length: 0), then the client hangs, and does not emit a response.

When I do the same over HTTP, it works. So, this appears isolated to HTTPS.
I only tried this with the JDK SSL Provider.  Did not try with the OpenSSL provider.

I originally reported this against `WebClient` as https://github.com/spring-projects/spring-framework/issues/22819, but I later determined it was related to empty request bodies (instead of empty response bodies as originally reported), and I was able to reproduce using `HttpClient` without using `WebClient`.  So it appears to be reactor-netty related.

I have a created a unit test ([project attached](https://github.com/reactor/reactor-netty/files/3096529/httpclient_empty_request_body.zip)) that executes requests against an okhttp3 MockWebServer to demonstrate the issue.

The unit test has three methods that each execute two requests against the mock server
* The HTTP test passes.
* The HTTPS test with non-empty request bodies passes
* The HTTPS test with empty request bodies fails with a timeout exception waiting on the second response.

### Expected behavior

POST requests sent over HTTPS with empty bodies return a response.

### Actual behavior

POST requests sent over HTTPS with empty bodies can hang.

If chunked encoding is used (which is what the unit test does), the request is never fully sent (and the server never responds)

If chunked encoding is _not_ used (by explicitly providing `Content-Length` header), then the server will respond, but the client still doesn't emit the response.

### Steps to reproduce

Execute attached unit test.

The failing test case (testHttps_emptyRequestBody) does the following:
1. Submits a POST request over HTTPS
2. Submits another POST request over HTTPS with an empty response body

### Reactor Netty version

| module | version |
|----|----|
| reactor-core | 3.2.8.RELEASE|
| reactor-netty | 0.8.6.RELEASE|
| netty |4.1.34.Final |

Reproduced on both 1.8.0_172 and 11.0.2

### OS version (e.g. `uname -a`)

Windows 10 Version 1709

violetagg:
Mark the promise as successful and do not proceed with writing.

### DIFF
--- a/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
+++ b/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
@@ -694,6 +694,12 @@ final class ChannelOperationsHandler extends ChannelDuplexHandler
 
 			produced++;
 
+			if ((t instanceof ByteBuf && ((ByteBuf) t).readableBytes() == 0) ||
+					(t instanceof FileRegion && ((FileRegion) t).count() == 0)) {
+				promise.setSuccess();
+				return;
+			}
+
 			// Returned value is deliberately ignored
 			parent.doWrite(t, promise, this);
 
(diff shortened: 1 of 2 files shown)


smaldini:
Superseeded by #725

PR #723, opened 2019-04-22
fix #717 Switch from Netty Connection Pool to Reactor Pool

### LINKED ISSUE #717: Switch from Netty Connection Pool to Reactor Pool

violetagg:


src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -79,11 +80,12 @@ static ConnectionProvider newConnection() {
 	 */
 	static ConnectionProvider elastic(String name) {
 		return new PooledConnectionProvider(name,
-				(bootstrap, handler, checker) -> new SimpleChannelPool(bootstrap,
-						handler,
-						checker,
-						true,
-						false));
+				(allocator, destroyHandler, evictionPredicate) ->
+				        (InstrumentedPool<PooledConnectionProvider.PooledConnection>)
```
simonbasle: maybe just double check the `InstrumentedPool` instanceof? for now all vanilla implementations from reactor-pool are meant to implement this, but in the future somebody else could decide to implement `Pool` and not `InstrumentedPool` (although `reactor-netty` is currently 100% in charge of choosing such concrete implementation)

simonbasle: note that in WIP https://github.com/reactor/reactor-pool/pull/55 we're moving towards the builder returning an `InstrumentedPool` explicitly.

Issue #654, opened 2019-03-21
Latency Increase as Connections Increase


dave-fl:
I am seeing that latency at the 75%, 90% and 99% is increasing compared to some other Netty based frameworks - total throughput here about 30% less.  This seems to be directly related to an increase in load.  As connections increase - the average latency doubles relative to Vert.x.

My setup is not perfect - better hardware would reveal more.  But I believe it is enough to demonstrate.

Container 1 is a  Docker container that has wrk.
Container 2 is one of Vert.X, Reactory Netty or Webflux

All running on PCF - no HTTPS.

Each app accepts a parameter that specifies how many characters to return.  Once this value is computed it is cached in a concurrent map so subsequent retrieval times are constant.  I am using a 10kb string for GETs only.

Below are the results.  The repot with these back ends can be found [here](https://github.com/dave-fl/gateway-performance-test.git)

Also, I don't known why but the results for Webflux aren't even in the same league.  About 20%-30% the throughput.  Maybe @rstoyanchev can comment here.

Do you consider  these results to be as expected or is there room for tuning here?

	# Vert-X 200 Connections

	Running 1m test @ vertx/demo?length=10240
	  16 threads and 200 connections
	  Thread Stats   Avg      Stdev     Max   +/- Stdev
		Latency    15.46ms    5.51ms  61.43ms   82.83%
		Req/Sec   779.30    219.36     3.37k    69.11%
	  Latency Distribution
		 50%   13.31ms
		 75%   14.51ms
		 90%   27.11ms
		 99%   30.10ms
	  745196 requests in 1.00m, 7.14GB read
	Requests/sec:  12400.12
	Transfer/sec:    121.59MB

	Running 1m test @ vertx/demo?length=10240
	  16 threads and 200 connections
	  Thread Stats   Avg      Stdev     Max   +/- Stdev
		Latency    16.45ms    6.49ms  55.82ms   75.52%
		Req/Sec   732.69    240.14     1.40k    54.57%
	  Latency Distribution
		 50%   13.27ms
		 75%   17.66ms
		 90%   27.78ms
		 99%   31.21ms
	  700642 requests in 1.00m, 6.71GB read
	Requests/sec:  11664.57
	Transfer/sec:    114.38MB

	# Netty Reactor 200 Connections

	Running 1m test @ reactor/demo/10240
	  16 threads and 200 connections
	  Thread Stats   Avg      Stdev     Max   +/- Stdev
[...]

PR #725, opened 2019-05-01
Revisit write operations

### LINKED ISSUE #572: Revisit ChannelOperationHandler
Change to flush on burst by default, stop supporting "unrelated" multiple concurrent send calls (but sequential send should remain fine). 

In recent RSocket tests we have observed important improvement of overall request throughput with some experiment.

smaldini:
Fix #572 - #654 
- send(Publisher) : supports fusion, Mono, flush on write burst. Flush API will be revisited in 0.9
- Interleaving send(Publisher) is now possible, avoiding livelock with recursive send/write. Use a single send with a Processor to serialize multiple sends. 
- HTTP send(Publisher) if Mono, a content length header will be automatically used and the response will be flushed as once by default, consistent with the client Get and Head requests as of now. If chunked transfer is still expected, use of Flux instead of Mono is necessary.
- default TCP options have been tuned down and smaller sndbuf/rcvbuf will result


smaldini:
There will be further improvement in the next iteration, notably around single send, http message creation and more send aliases (encoder, source cleanup)

PR #726, opened 2019-05-01
Http client should ignore malformed cookies

fred84:
Netty's `ClientCookieDecoder` may return null if cookie's name or value is malformed. 

Current behavior
====
Http client throws NPE on malformed cookie

Modified behavior
====
Http client ignores malformed cookie.


smaldini:
Cool, and yup feels more resilient

smaldini:
@violetagg do you think we can target 0.8 directly ?

violetagg:
The fix will be available in 0.8.7
Thanks

PR #727, opened 2019-05-02
Use shadow plugin for Netty's HAProxy, Proxy, Http/2 support

violetagg:


smaldini:
Do we need to change the HAProxy code detection we have now ?
Also what about native epoll lib ?

smaldini:
Superseded by #730

PR #728, opened 2019-05-02
Fix ProxyProvider hash function leading to memory leakage

shrykull:
it seems i triggered a memory leak by registering a proxy with `reactor.netty.http.client.HttpClient.create().tcpConfiguration(this::configureProxy)` as the lambda is called for every connection due to the ChannelHandler being a "different" one in every call of `reactor/netty/resources/PooledConnectionProvider.java:137`

i guess the problem is that the used hash function `reactor.netty.tcp.ProxyProvider#hashCode` includes the nonProxyHosts as a regex pattern which gets his hash function from `java.lang.Object#hashCode` and therefore fails to recognize equal regexes. The tests in `src/test/java/reactor/netty/tcp/ProxyProviderTest.java` did not break because they get null as test fixture for nonProxyHosts.

This ultimately causes the `reactor.netty.resources.PooledConnectionProvider#channelPools` to grow indefinitely once a proxy with defined `nonProxyHosts` is used, no matter if the connection is actually proxied or not.

### DIFF
--- a/src/main/java/reactor/netty/tcp/ProxyProvider.java
+++ b/src/main/java/reactor/netty/tcp/ProxyProvider.java
@@ -227,7 +227,7 @@ public final class ProxyProvider {
 		return Objects.equals(username, that.username) &&
 				Objects.equals(getPasswordValue(), that.getPasswordValue()) &&
 				Objects.equals(getAddress().get(), that.getAddress().get()) &&
-				Objects.equals(getNonProxyHosts(), that.getNonProxyHosts()) &&
+				Objects.equals(getNonProxyHostsValue(), that.getNonProxyHostsValue()) &&
 				Objects.equals(httpHeaders.get(), that.httpHeaders.get()) &&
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@shrykull Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=728)!

[Click here](https://cla.pivotal.io/sync/reactor?repositoryId=reactor/reactor-netty&pullRequestId=728) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@shrykull Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=728)!

violetagg:
Fix with e39d2c1a61509162f2cca4307006ae3382d41f9f

PR #732, opened 2019-05-07
fix shading config (exclude compile dependencies from `shaded`)

bsideup:
The structure now:
Screen Shot 2019-05-07 at 08 51 59


bsideup:
I will re-submit shading PR for M2, with this PR included

PR #734, opened 2019-05-07
 Shade Netty Proxy, HTTP/2 and HAProxy

bsideup:
- for now, we strip the native image metadata
- OSGI excludes could be generated automatically, but it requires an additional effort and can be done separately

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,6 +29,7 @@ buildscript {
 plugins {
   id "com.github.hierynomus.license" version "0.14.0"
   id 'org.asciidoctor.convert' version '1.5.11'
+  id 'com.github.johnrengelman.shadow' version '5.0.0'
 }
 
 description = 'Reactive Streams Netty driver'
@@ -97,6 +98,9 @@ configure(rootProject) { project ->
 
   ext.bundleImportPackages = [ '!javax.annotation',
 							   'io.netty.channel.kqueue;resolution:=optional;version="[4.1,5)"',
+							   '!io.netty.handler.codec.haproxy',
+							   '!io.netty.handler.codec.http2',
+							   '!io.netty.handler.proxy',
 							   '*']
 
   [compileJava, compileTestJava]*.options*.compilerArgs = ["-Xlint:varargs",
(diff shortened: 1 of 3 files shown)


sdeleuze:
I tend to think shading only some Netty dependencies and not others will cause much trouble on dependency management side. And since Netty is exposed on public API, you can't shade it totally. What is the problem you are trying to resolve by doing that?

philwebb:
+1 to what @sdeleuze said. A half shaded approach seems quite risky to me and also prevents the user from upgrading the shaded parts at all without a related reactor-netty release. I'm especially concerned that an upgrade in the non-shaded part will cause some binary incompatibly with the shaded part. I don't expect that Netty themselves would expect such a split in their codebase, and I doubt they will have tests for such a scenario.

mp911de:
I'd also like to understand the rationale behind that. It is easier to adopt a netty bugfix by just incrementing the netty dependency version instead of waiting for reactor-netty to adapt the netty version and ship a release.

smaldini:
There are multiple reasons we wanted to experiment with this:
- Going from 15 back to 5 dependencies (keeping the spec api and bytebuf, shading the impl codec)
- Controlling the codec lifecycle if we need to apply patches/workaround without duplicating them
- Getting ready for the codec breaking changes planned in netty 5
- According to users reports, the implementations sometimes conflict when 2 libs use netty, e.g. if one lib pulls netty 4.x.y and we pull 4.u.v, especially if one library has to build a workaround until a next netty version (which is a semi regular case).

mp911de:
Probably reducing mandatory dependencies to just a few ones is a good move and making the other ones optional. 

> Controlling the codec lifecycle if we need to apply patches/workaround without duplicating them

This option opens a can of worms as compensating for other libraries isn't the right thing to do even it looks appealing in the first moment. Users will get used to you fixing netty bugs in reactor-netty.

> Getting ready for the codec breaking changes planned in netty 5

Typically we increment the version number when switching to newer (incompatible) dependency. E.g. reactory-netty 0.9 could stick with netty 4.1.x and reactor-netty 0.10 with netty 5.0. Alternatively having a reactor-netty5 artifact/branch with its own lifecycle.

smaldini:
We will pause this for the time being after many discussions around. However we might provide a separation of projects like reactor-netty-core and reactor-netty-http. Maybe a do it all reactor-netty separately for end users will also be helpful.

PR #743, opened 2019-05-17
use WireMock to locally simulate HTTP server

simonbasle:
The server is exposed as HttpEchoTestingServer utility class, which
is a JUnit ClassRule and can also be used standalone (has start and
stop methods). It exposes the port it runs on, as well as the base
HTTP url.

Other approaches were explored and discarded:
 - httpin.org in a Docker container (via TestContainers): a bit
 overkill on resources, and not all dev / ci machine will have Docker
 - MockServer: uses Netty AND doesn't shade the dependency, resulting
 in potential Netty version conflics.

WireMock doesn't use Netty dependency like MockServer, so no clash.
Furthermore, there is a shaded standalone version of WireMock.

This commit also polishes some tests to run on local HTTP server:
  - polish tests that don't BAD_REQUEST anymore due to GET having a body
  (not enforced on MockServer, contrary to example.com and google.com)
 - polish one TcpClientTests to also avoid HTTP remote, using the
 HttpEchoTestingServer standalone.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,6 +61,7 @@ ext {
   // Testing
   mockitoVersion = '1.10.19'
   hoverflyJavaVersion = '0.11.1'
+  wireMockVersion = '2.23.2'
 
   javadocLinks = ["https://docs.oracle.com/javase/7/docs/api/",
 				  "https://docs.oracle.com/javaee/6/api/",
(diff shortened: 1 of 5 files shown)


simonbasle:
superseded by #748

PR #745, opened 2019-05-20
Extend `EventLoop` from Netty's `FastThreadLocalThread`

bsideup:
As was suggested by @normanmaurer, we should extend from `FastThreadLocalThread`.

`FastThreadLocal` is a special optimization in Netty that is faster than JVM's ThreadLocals in Netty's use cases


normanmaurer:
@bsideup yes! Thanks for listening ;)

bsideup:
@normanmaurer conference-driven development rocks 😎 Thanks for pointing out!
We need to run the benchmarks, but the explanation you had sounds solid :)

normanmaurer:
@bsideup lol... you are welcome :)

smaldini:
Thanks @normanmaurer for the suggestion it was quite a miss actually since we moved to use our marker interface for blocking code detection.

PR #748, opened 2019-05-21
Use Tomcat as HTTP server in order to test Reactor Netty HTTP client

violetagg:
In some test scenarios HTTP server different that Reactor Netty is needed.


src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -490,19 +549,21 @@ public void contentHeader() {
 
 		HttpResponseStatus r =
```
simonbasle: I'm not sure what this test ascertains... but I made it check the `Content-Length` header in _both_ responses in the other PR.

src/test/java/reactor/netty/TomcatServer.java
```diff
@@ -0,0 +1,161 @@
...
+	public void createDefaultContext() {
+		Context ctx = this.tomcat.addContext("", null);
+
+		addServlet(ctx, new DefaultServlet(), "/");
+		addServlet(ctx, new StatusServlet(), "/status/*");
+		addServlet(ctx, new MultipartServlet(), "/multipart")
+				.setMultipartConfigElement(new MultipartConfigElement(""));
+	}
+
+	public void createContext(HttpServlet servlet, String mapping) {
```
simonbasle: this looks unused?

src/test/java/reactor/netty/TomcatServer.java
```diff
@@ -0,0 +1,161 @@
...
+
+	public void stop() throws Exception {
+		if (this.started) {
+			this.tomcat.stop();
+			this.tomcat.destroy();
+			this.started = false;
+		}
+	}
+
+	public void createDefaultContext() {
```
simonbasle: perhaps the `createDefaultContext()` call could be made in `start()` so that an extra call is avoided in each `HttpClientTest` that make use of the `TomcatServer`?

Have you tried to start the server once in a `@ClassRule` btw? this would at least avoid the ceremony around starting and stopping the server (see below).

src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -341,26 +347,40 @@ public void postUpload() throws IOException {
 
 		assertThat(res).as("response").isNotNull();
 		assertThat(res.getT1()).as("status code").isEqualTo(200);
-		assertThat(res.getT2()).as("response body reflecting request").contains("\"form\": {\n    \"attr1\": \"attr2\", \n    \"test\": \"This is an UTF-8 file that is smaller than 1024 bytes.\\nIt contains accents like \\u00e9.\\nEnd of File\", \n    \"test2\": \"\"\n  },");
+		assertThat(res.getT2()).as("response body reflecting request").contains("test: 95\nattr1: 5\ntest2: 0\n");
+
+		tomcat.stop();
```
simonbasle: this won't happen in case of an assertion error (or any exception really)

simonbasle: all the tests below are vulnerable to this caveat

violetagg:
@simonbasle I'll extract all tests that use Tomcat in a separate class and then will use @ClassRule.
I just don't want to do it here in order to have clean history what was changed in every test.

src/test/java/reactor/netty/http/client/HttpClientTest.java
violetagg: The idea here is to test that content will be send with GET request and the server will respond with 400 Bad request because it does not expect content with GET request.

src/test/java/reactor/netty/TomcatServer.java
violetagg: This I will need for future usage.

simonbasle:
> @simonbasle I'll extract all tests that use Tomcat in a separate class and then will use @ClassRule.
> I just don't want to do it here in order to have clean history what was changed in every test.

OK, so in order to make the PR mergeable, I'd add that commit with the `@ClassRule` here and merge this PR with rebase in order to keep the clearer history.

violetagg:
tests are extracted in a separate class

PR #755, opened 2019-06-04
Fix unintended stack trace suppression despite DEBUG level being enab…

ginkel:
…led, expose exception message

The order of checks `log.isWarnEnabled()` and `log.isDebugEnabled()` is wrong. If debug-logging is enabled, `log.isWarnEnabled()` is also true causing the debug branch to be never executed. This change fixes that by swapping the branches.

In addition, I changed the warning-level logging to not just output the exception class name, but also the message (using `Exception#toString()`) as the exception class name alone often is not sufficient to infer what went wrong.

### DIFF
--- a/src/main/java/reactor/netty/channel/FluxReceive.java
+++ b/src/main/java/reactor/netty/channel/FluxReceive.java
@@ -341,7 +341,7 @@ final class FluxReceive extends Flux<Object> implements Subscription, Disposable
 	final void onInboundError(Throwable err) {
 		if (isCancelled() || inboundDone) {
 			if (log.isDebugEnabled()) {
-				log.error(format(channel, "An exception has been observed post termination"), err);
+				log.warn(format(channel, "An exception has been observed post termination"), err);
 			} else if (log.isWarnEnabled()) {
 				log.warn(format(channel, "An exception has been observed post termination, use DEBUG level to see the full stack: {}"), err.toString());
 			}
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@ginkel Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=755)!

[Click here](https://cla.pivotal.io/sync/reactor?repositoryId=reactor/reactor-netty&pullRequestId=755) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ginkel Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=755)!

src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -340,11 +340,10 @@ final void onInboundComplete() {
 
 	final void onInboundError(Throwable err) {
 		if (isCancelled() || inboundDone) {
-			if (log.isWarnEnabled()) {
-				log.warn(format(channel, "An exception has been observed post termination, use DEBUG level to see the full stack: {}"), err.getClass().getSimpleName());
-			}
-			else if (log.isDebugEnabled()) {
+			if (log.isDebugEnabled()) {
 				log.error(format(channel, "An exception has been observed post termination"), err);
```
violetagg: @smaldini I think we should switch this to `log.debug`. Wdyt?

ginkel: If I may join the discussion: Changing the severity as a consequence of the log level change is somewhat unexpected. IMHO, keeping the level consistent for both branches (e.g., WARN) would come up with the least surprise.

violetagg:
Thanks

src/main/java/reactor/netty/channel/FluxReceive.java
violetagg: agreed

PR #767, opened 2019-06-25
Remove send options

smaldini:


violetagg:
what about deprecating this in 0.8.x?

PR #773, opened 2019-07-01
Clean resources of long living connections to reduce gc pressure

aftersss:
When Using `HttpClient` with connection pool enabled, We found that there is some resources not released after some http request is completed and before the keep-alive connection is closed. When there are many long living connections, the java gc will be very bad.

The following code can reproduce this issue:

```
public class HttpClientBugTest {

    private static class MyConsumer implements Consumer<HttpClientResponse>{
        @Override
        public void accept(HttpClientResponse httpClientResponse) {
            httpClientResponse.receiveContent().map(httpContent -> httpContent.content())
                    .subscribe(byteBuf -> System.out.println(byteBuf.toString(Charset.defaultCharset())));
        }
    }

    public static void main(String[] args) throws IOException {
        HttpClient httpClient = HttpClient.create(opts -> {
            opts.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, 1000);
            PoolResources poolResources = PoolResources.fixed("a",
                    100, 1000);
            opts.poolResources(poolResources);
        });

        Mono<HttpClientResponse> resp = httpClient.get("https://www.google.com/");
        resp.subscribe(new MyConsumer());

        System.in.read();
    }

}
```

Execute the above code, and as soon as the console shows the http response string, use shell `jmap -dump:live,format=b,file=a.heap ${pid}` to generate the heap dump file(this shell must be executed before the keep-alive connection is closed!), then use `Eclipse Memory Analyzer` to open it, the following image shows the memory leak:

image

This PR tries to fix this issue.

### DIFF
--- a/src/main/java/reactor/ipc/netty/channel/ChannelOperationsHandler.java
+++ b/src/main/java/reactor/ipc/netty/channel/ChannelOperationsHandler.java
@@ -572,6 +572,7 @@ final class ChannelOperationsHandler extends ChannelDuplexHandler
 					else {
 						// Returned value is deliberately ignored
 						promise.setFailure(new AbortedException("Connection has been closed"));
+						clean();
 						return;
 					}
 				}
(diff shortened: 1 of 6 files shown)


violetagg:
@aftersss The PR is made for 0.7.x. We stopped supporting this version. Can't you migrate to 0.8.x?

aftersss:
Close this and move to #755

PR #775, opened 2019-07-02
Clean resources of long living connections to reduce gc pressure

aftersss:
When Using `HttpClient` with connection pool enabled, We found that there is some resources not released after some http request is completed and before the keep-alive connection is closed. When there are many long living connections, the java gc will be very bad.

The following code can reproduce this issue:

```

public class HttpClientBugTest {

    private static class MyConsumer implements Consumer<String> {
        @Override
        public void accept(String str) {
            System.out.println(str);
        }
    }

    public static void main(String[] args) throws IOException {
        HttpClient httpClient = HttpClient.create(
                ConnectionProvider.fixed("a", 100, 100))
        .tcpConfiguration(opts -> {
            return opts.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, 1000);
        });

        while(true) {
            httpClient.request(HttpMethod.GET).uri("http://172.16.209.199:1028/monitor/check")
                    .responseContent().aggregate().asString(Charset.defaultCharset())
                    .subscribe(new MyConsumer());

            System.in.read();
        }
    }
}
```

Execute the above code, and as soon as the console shows the http response string, use shell `jmap -dump:live,format=b,file=a.heap ${pid}` to generate the heap dump file(this shell must be executed before the keep-alive connection is closed!), then use `Eclipse Memory Analyzer` to open it, the following images shows the memory leak:

memory leak type 1:
image

memory leak type 2:
image

When using spring-cloud-gateway there is another type of memory leak issue:
image

This PR tries to fix this issue.

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClientConnect.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientConnect.java
@@ -151,8 +151,10 @@ final class HttpClientConnect extends HttpClient {
 			}
 
 			HttpClientConfiguration conf = HttpClientConfiguration.getAndClean(b);
+			ClientCookieEncoder cookieEncoder = conf.cookieEncoder;
+			ClientCookieDecoder cookieDecoder = conf.cookieDecoder;
 			BootstrapHandlers.channelOperationFactory(b,
-					(ch, c, msg) -> new HttpClientOperations(ch, c, conf.cookieEncoder, conf.cookieDecoder));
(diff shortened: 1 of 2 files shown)


violetagg:
Thanks for the PR. It is applied with e2a234dd1b447e8468a89286672a3945b45320b3.
It will be very helpful if you continue exploring in this direction (whether there are such issues on the server side)

PR #792, opened 2019-07-29
fix #612 API for configuring pooled connection idle time

### LINKED ISSUE #612: how to configure pooled connection idle timeout
I am using reactor-netty http client (0.7.X series) with connection pooling and would like to configure pooled connection's idle timeout but don't know where.

More precisely, I need to configure reactor-netty's connection pool in such a way that it will automatically close connections that did not see any activity within configurable timeout. These connections are open but no bytes were transferred in or out for some (configurable) amount of time.

As an example, Jetty's http client has [a configuration option with the above semantics branded as connectionIdleTimeout](https://www.eclipse.org/jetty/javadoc/current/org/eclipse/jetty/client/HttpClient.html#setIdleTimeout-long-).

Is there an analogous setting in reactor-netty that allows me to set a connection's idle timeout?
How can I configure reactory-netty http client to close idle connections preemptively?

We are getting `Connection prematurely closed` errors described in #413 and #498 because of this.

### Expected behavior
Pool automatically closes a connection which was idle for a given time interval
### Actual behavior
Don't know how to force the pool to close the inactive connection 
### Steps to reproduce
N/A
### Reactor Netty version
0.7.13
N/A
### OS version (e.g. `uname -a`)
N/A

violetagg:


src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -1828,4 +1828,58 @@ private void doTestIssue777_2(HttpClient client, String uri, String expectation,
...
+				          .bindNow();
+
+		Flux<ChannelId> id = createHttpClientForContextWithAddress(server, provider)
+		                       .get()
+		                       .uri("/")
+		                       .responseConnection((res, conn) -> Mono.just(conn.channel().id())
+		                                                              .delayUntil(ch -> conn.inbound().receive()));
+
+		ChannelId id1 = id.blockLast(Duration.ofSeconds(30));
+		Mono.delay(Duration.ofMillis(30)).block();
```
simonbasle: is that on purpose? that feels weirder than using `Thread.sleep(30)` or `Awaitility` 😉

violetagg: ;)
changed to `Thread.sleep(30)`

PR #827, opened 2019-08-30
Disable auto-redirect when the follow redirect predicate throws RuntimeException

violetagg:


violetagg:
this will be merged in #826

PR #841, opened 2019-09-11
Detect whether secure or not when the first http request comes to support  SniHandler added by user.

aftersss:
reactor-netty supports users to add some netty handlers themselves, So we add a SniHandler to the handler chain to let our server to support sni, this worked well, but recently when we upgrade the reactor-netty to `0.8.11.RELEASE`, we found that it will  lost `https` schema when the http request is routed to upstream. 

This PR tries to fix this problem, detect whether secure or not when the first http request comes, so  users can add SniHandler or SslHandler themselves, without using the reactor-netty's api to support more customized requirement.

### DIFF
--- a/src/main/java/reactor/netty/http/server/Http2StreamBridgeHandler.java
+++ b/src/main/java/reactor/netty/http/server/Http2StreamBridgeHandler.java
@@ -29,6 +29,7 @@ import io.netty.handler.codec.http2.Http2Headers;
 import io.netty.handler.codec.http2.Http2HeadersFrame;
 import io.netty.handler.codec.http2.Http2StreamChannel;
 import io.netty.handler.codec.http2.HttpConversionUtil;
+import io.netty.handler.ssl.SslHandler;
 import reactor.netty.Connection;
 import reactor.netty.ConnectionObserver;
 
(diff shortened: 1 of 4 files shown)


violetagg:
@aftersss Can you provide a test case for this use case?

aftersss:
> @aftersss Can you provide a test case for this use case?

Ok, PLS review it.

violetagg:
@aftersss One last question. Do you expect `SslHandler` to be added after creation of the `HttpTrafficHandler`? The current test that you provided still works if I move `secure = ctx.channel().pipeline().get(SslHandler.class) != null;` to the `HttpTrafficHandler` constructor.

aftersss:
> @aftersss One last question. Do you expect `SslHandler` to be added after creation of the `HttpTrafficHandler`? The current test that you provided still works if I move `secure = ctx.channel().pipeline().get(SslHandler.class) != null;` to the `HttpTrafficHandler` constructor.

Actually we added a `SniHandler` to the handler chain, not `SslHandler` directly, When channel is created, if the SniHandler receives a ssl handshake with a `sni` info, we will replace SniHandler with a SslHandler which match that `sni` domain to let this channel support https protocol. if no `sni` is detected, we think it uses http protocol(modern web browsers supports sni),  so we will remove SniHandler and not add any `SslHandler` to let the channel support http protocol. So we can use reactor-netty to support http/https requests using the same port(but each channel can only choose to support http or https).

So I think detect whether it is secure in the `HttpTrafficHandler` constructor is not enough.

violetagg:
Thanks

smaldini:
Ultimately we will probably get rid of that extra check and provide SniHandler support directly. In recent benchmarks I saw lookup in the pipelines flagged so if we know at configuration time there will be such handler it will be more efficient. More on that in the next major release.

PR #842, opened 2019-09-12
Update Netty version to 4.1.41.Final

violetagg:
Adapt to a change in Netty connection pool
https://github.com/netty/netty/issues/9542

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,7 @@ ext {
   assertJVersion = '3.6.1'
 
   // Libraries
-  nettyVersion = '4.1.39.Final'
+  nettyVersion = '4.1.41.Final'
   jacksonDatabindVersion = '2.5.1'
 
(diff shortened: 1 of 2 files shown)


violetagg:
#844 is still not resolved

barclay-reg:
is there any chance to update directly to netty  4.1.42 ... needed to see https://github.com/netty/netty/issues/9571 included, as this exposes a security issue

simonbasle:
(NB: as of today the UDP issue with epoll is still not resolved in `4.1.42.Final` nor in `4.1.43.Final-SNAPSHOT`)

smaldini:
Looking at this fix I wonder abuot the AttributeKey - it feels like this is only checking if the key has been ever used in "any pool". I suspect the intent was to check if the pool key is currently set on the channel ?

PR #851, opened 2019-09-18
fix #850 Specify fire-and-forget dispose() behaviour for LoopResources, ConnectionProvider

### LINKED ISSUE #850: Specify fire-and-forget `dispose()` behavior in javadoc of classes that have a `disposeLater()`
The `Disposable.dispose` javadoc currently mentions:

> Cancel or dispose the underlying task or resource.

Several classes inheriting that javadoc are implementing this method in a way that:
1. resources are marked as disposed
2. the method returns
3. resources are actually disposed asynchronously, later

This can be misleading as the javadoc and the `void` return type might lead users to think that resources are disposed right away. See `ConnectionProvider` and `LoopResources` in Reactor Netty, which also implement `disposeLater` variants.

violetagg:
Specify fire-and-forget disposeLoopsAndConnections() behaviour for
TcpResources, HttpResources
Specify fire-and-forget shutdown() behaviour for UdpResources

### DIFF
--- a/src/main/java/reactor/netty/http/HttpResources.java
+++ b/src/main/java/reactor/netty/http/HttpResources.java
@@ -72,6 +72,10 @@ public final class HttpResources extends TcpResources {
 	/**
 	 * Shutdown the global {@link HttpResources} without resetting them,
 	 * effectively cleaning up associated resources without creating new ones.
+	 * This method is NOT blocking. It is implemented as fire-and-forget.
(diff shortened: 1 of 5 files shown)


violetagg:
> I would mention combining disposeLater with blocking in these new docs, since otherwise it might lead people to think simply replacing `dispose()` with `disposeLater()` would work

done

PR #853, opened 2019-09-19
HAProxyMessage should be released to avoid memory leak.

aftersss:
HAProxyMessage is sub class of `ReferenceCounted`, so it should be released to avoid memory leak.

And `REMOTE_ADDRESS_FROM_PROXY_PROTOCOL` attribute should not be removed from  channel, or else the second http request and the subsequent requests in the same connection will lose the removeAddress info.

### DIFF
--- a/src/main/java/reactor/netty/http/server/ConnectionInfo.java
+++ b/src/main/java/reactor/netty/http/server/ConnectionInfo.java
@@ -22,7 +22,6 @@ import java.util.regex.Pattern;
 import io.netty.channel.Channel;
 import io.netty.channel.socket.SocketChannel;
 import io.netty.handler.codec.http.HttpRequest;
-import io.netty.util.AttributeKey;
 import reactor.netty.tcp.InetSocketAddressUtil;
 /**
  * Resolve information about the current connection, including the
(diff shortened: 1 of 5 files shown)


src/main/java/reactor/netty/http/server/ConnectionInfo.java
```diff
@@ -168,7 +168,7 @@ static ConnectionInfo parseXForwardedInfo(HttpRequest request, SocketChannel cha
 	private static InetSocketAddress getRemoteAddress(SocketChannel channel) {
 		if (HAProxyMessageReader.hasProxyProtocol()) {
 			InetSocketAddress remoteAddressFromProxyProtocol =
-					channel.attr(REMOTE_ADDRESS_FROM_PROXY_PROTOCOL).getAndSet(null);
+					channel.attr(REMOTE_ADDRESS_FROM_PROXY_PROTOCOL).get();
```
violetagg: Why don't we cache this in the `HttpTrafficHandler`/`Http2StreamBridgeHandler`?

aftersss: OK

PR #855, opened 2019-09-20
Add proxy protocol auto detect feature.

aftersss:
We are using the following architecture: 
```
User's Http Request -> HaProxy(with proxy protocol enabled) -> spring-cloud-gateway -> realServer.
```

We configured the spring-cloud-gateway to enable proxy protocol support, but this bring us a new problem: it is hard to check if `spring-cloud-gateway` is alive,  most http health check tools does not support proxy protocol, so if `spring-cloud-gateway` can support proxy-protocol-enabled/proxy-protocol-disabled requests at the same time, it should be better.

So I create this PR to support proxy-protocol `AUTO` mode, this mode can auto detect whether the client(or reverse proxy server) has enabled proxy protocol, so the `HttpServer` can support proxy-protocol-enabled and proxy-protocol-disabled clients at the same time.

The proxy-protocol auto detect feature should also be convenient for this case: maybe in production environment `spring-cloud-gateway` runs behind HaProxy Server, but in development environment `spring-cloud-gateway` need to be accessed directly, users can set proxy-protocol to `AUTO` mode, and won't worry about the environment any more.

### DIFF
--- a/src/main/java/reactor/netty/http/server/HAProxyMessageDetector.java
+++ b/src/main/java/reactor/netty/http/server/HAProxyMessageDetector.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright (c) 2011-2019 Pivotal Software Inc, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package reactor.netty.http.server;
 
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/main/java/reactor/netty/http/server/HAProxyMessageDetector.java
```diff
@@ -0,0 +1,38 @@
+package reactor.netty.http.server;
```
violetagg: Copyright header

src/main/java/reactor/netty/http/server/ProxyProtocolSupportType.java
```diff
@@ -0,0 +1,26 @@
+package reactor.netty.http.server;
```
violetagg: Copyright header

src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -243,13 +243,41 @@ public final HttpServer forwarded(boolean forwardedEnabled) {
...
+	/**
+	 * Specifies whether support for the {@code "HAProxy proxy protocol"}
+	 * for deriving information about the address of the remote peer is enabled.
+	 *
+	 * @param proxyProtocolSupportType
+	 * 		<ul>
+	 *         <li>choose {@link ProxyProtocolSupportType#ON} to enable support for the {@code "HAProxy proxy protocol"}
+	 *             for deriving information about the address of the remote peer.</li>
+	 *         <li>choose {@link ProxyProtocolSupportType#OFF} to disable the proxy protocol support.</li>
+	 *         <li>choose {@link ProxyProtocolSupportType#AUTO} then each connection of the same `HttpServer` will auto detect whether there is proxy protocol, so HttpServer can accept requests with or without proxy protocol.</li>
```
violetagg: This line is too long

violetagg:
Thanks

PR #858, opened 2019-09-23
avoid redundant regex compile

Liyixin95:
The method `replaceAll` will compile a regex when invoked, which will cause unnecessary overhead.


violetagg:
The change was committed to 0.8.x branch and merged in master - 3555eb0e543bed348d6a7fe3fffa51232003283c

Thanks

PR #875, opened 2019-10-24
Handle decoding failures with LastHttpContent

violetagg:


violetagg:
> Can we do the instanceof DecorderResultProvider after the if(msg instanceof HttpRequest) as a micro optimization. Ideally we shouldn't even do instanceof HttpRequest more than once if the request is already read but thats another story.

done

Issue #876, opened 2019-10-24
Memory leak if a timeout is triggered while still reading response body


anilgursel:
On a stream that uses `timeout` operator with `HttpClient`, if timeout is triggered while `FluxReceiver` is still reading the response bytes, intermittent leaks happen.  Please see below reproducer and the logs:

```java
package org.sample;


import io.netty.buffer.ByteBuf;
import io.netty.buffer.ByteBufAllocator;
import io.netty.channel.ChannelOption;
import io.netty.handler.codec.http.HttpMethod;
import io.netty.handler.timeout.ReadTimeoutHandler;
import io.netty.util.ResourceLeakDetector;
import org.junit.AfterClass;
import org.junit.BeforeClass;
import org.junit.Test;
import reactor.core.publisher.Flux;
import reactor.netty.DisposableServer;
import reactor.netty.http.HttpProtocol;
import reactor.netty.http.client.HttpClient;
import reactor.netty.http.server.HttpServer;
import reactor.netty.resources.ConnectionProvider;

import java.io.ByteArrayOutputStream;
import java.io.InputStream;
import java.io.PrintStream;
import java.nio.charset.Charset;
import java.time.Duration;
import java.util.concurrent.ExecutionException;
import java.util.concurrent.TimeUnit;

import static org.hamcrest.MatcherAssert.assertThat;
import static org.hamcrest.core.IsNot.not;
import static org.hamcrest.core.StringContains.containsString;

public class ResourceReleaseTest {

    private static final Duration CLIENT_TIMEOUT_DURATION = Duration.ofMillis(50);
    private static final int CONNECTION_POOL_SIZE = 10;
    private static final Duration SERVER_DELAY_BETWEEN_ELEMENTS_WHILE_STREAMING = Duration.ofMillis(10);
    private static int SERVER_ELEMENT_COUNT = 10;

    private static final DisposableServer mockServer = setupMockServer();
    private static final HttpClient httpClient = setupClient();

    private static final PrintStream SYSTEM_ERROR = System.err;
    private static final ByteArrayOutputStream errContent = new ByteArrayOutputStream();

    @BeforeClass
    public static void setup() {
        ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.PARANOID);
        System.setErr(new PrintStream(errContent));
    }

    @AfterClass
    public static void cleanup() {
        System.setErr(SYSTEM_ERROR);
[...]

PR #882, opened 2019-10-30
Fixed memory leak in ByteBufFlux#aggregate

violetagg:
Fixed memory leak when `timeout` occurs.

Related to #876

### DIFF
--- a/src/main/java/reactor/netty/ByteBufFlux.java
+++ b/src/main/java/reactor/netty/ByteBufFlux.java
@@ -28,6 +28,7 @@ import java.util.function.Function;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.ByteBufHolder;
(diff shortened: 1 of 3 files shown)


src/main/java/reactor/netty/ByteBufFlux.java
```diff
@@ -259,23 +260,31 @@ public static ByteBufFlux fromPath(Path path,
...
-				        b.release();
-				    }
-				})
-				.as(ByteBufMono::maybeFuse);
+		return Mono.defer(() -> {
+		               CompositeByteBuf output = alloc.compositeBuffer();
+		               return doOnNext(ByteBuf::retain)
+		                       .collectList()
+		                       .doOnDiscard(ByteBuf.class, ByteBuf::release)
+		                       .flatMap(list -> {
```
smaldini: could it just be map ?

violetagg: changed to `handle`

PR #901, opened 2019-11-17
Add responseUrl to HttpClientResponse

Tankika:
Hello,

I would like to ask your help to implement this feature we need in our project. In our application we would need to know the last URL of a redirect chain, because we're downloading content from various sites across the internet and we would need to know where we've been before. I'm not sure this is the correct implementation, but this is something I have in mind. The other option that came to my mind (similar to Apache HttpClient's implementation) is adding the last url to redirectedFrom array, but this implementation seems cleaner to me.

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClientConnect.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientConnect.java
@@ -523,6 +523,8 @@ final class HttpClientConnect extends HttpClient {
 
 		Publisher<Void> requestWithBody(HttpClientOperations ch) {
 			try {
+				ch.resourceUrl = toURI.toExternalForm();
+
 				UriEndpoint uri = toURI;
 				HttpHeaders headers = ch.getNettyRequest()
 				                        .setUri(uri.getPathAndQuery())
--- a/src/main/java/reactor/netty/http/client/HttpClientDoOnError.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientDoOnError.java
@@ -117,6 +117,11 @@ final class HttpClientDoOnError extends HttpClientOperator {
 			return EMPTY;
 		}
 
+		@Override
+		public String resourceUrl() {
+			return null;
+		}
+
 		final static String[] EMPTY = new String[0];
 
 		@Override
--- a/src/main/java/reactor/netty/http/client/HttpClientInfos.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientInfos.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2011-2019 Pivotal Software Inc, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package reactor.netty.http.client;
+
+import reactor.netty.http.HttpInfos;
+import reactor.util.context.Context;
+
+public interface HttpClientInfos extends HttpInfos {
+
+    /**
+     * Return the current {@link Context} associated with the Mono/Flux exposed
+     * via {@link HttpClient.ResponseReceiver#response()} or related terminating API.
+     *
+     * @return the current user {@link Context}
+     */
+    Context currentContext();
+
+    /**
+     * Return the previous redirections or empty array
+     *
+     * @return the previous redirections or empty array
+     */
+    String[] redirectedFrom();
+
+    /**
+     * Return the fully qualified URL of the requested resource. In case of redirects, return the URL the last
+     * redirect led to.
+     *
+     * @return The URL of the retrieved resource. This method can return null in case there was an error before the
+     * client could create the URL
+     */
+    String resourceUrl();
+}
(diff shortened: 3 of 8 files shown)


violetagg:
@Tankika `reactor.netty.http.client.HttpClientResponse#redirectedFrom` contains all the redirects not only the last one. Can't you use this API? https://projectreactor.io/docs/netty/release/api/reactor/netty/http/client/HttpClientResponse.html#redirectedFrom--

Tankika:
@violetagg the redirectedFrom array is almost what we need, but not exactly. It contains all the PREVIOUS urls we got redirected from. What we neee is the very last url we visit. In case of no redirections, this is same as the request url. In case of redirections, this is basically the url of the last request we make.

The term response Url might not be the best, but I've found other libraries referring to this url as such. Of course I'm open to any other name, as long as the functionality is the same

violetagg:
@Tankika Actually you need `HttpClientResponse#uri` but with schema, host and port?

Tankika:
@violetagg Yes, I would say. The `HttpClientResponse#uri` as a fully qualified url. I could not find this information in the `nettyRequest` object, but my implementation gives you exactly this information.

violetagg:
@Tankika So with your recent PR we will keep the schema so you know what the schema is. And the address you can get from the connection - `Connection#address`. Isn't this enough?

Tankika:
@violetagg Well, unfortunately, `HttpClientOperations#isSecure` is a package-private property of `HttpClientOperations`. Even if it were not, I would still need a cast from `Connection` to `HttpClientOperations`, which is ugly. And even if it were not the case, I would still need to manually put together the URL, which is still not that pretty, considering that the `HttpClientHandler` already has what I need, in exactly the format I need. I believe the cleanest solution is to offer this information through the API, because I don't think this is some internal data, that the clients of the library should not be aware of, similar to `HttpClientResponse#redirectedFrom`

May I ask what is the problem with my offered solution, so I might be able to think about alternatives?

violetagg:
@Tankika What about deprecating `reactor.netty.http.HttpInfos#uri` and adding `reactor.netty.http.HttpInfos#url`?

/cc @smaldini

Tankika:
@violetagg Actually I was thinking about something similar, but keeping the method as something like `HttpClientResponse#fullUri`. I don't know much about the library, but I can see, that `HttpServerRequest` and `HttpServerResponse` also extend the `HttpInfos` interface. My problem with it is that this information does not seem to be valid on those classes.

Also, as I understand, there can be cases, when this new method returns null, when the request terminates before it gets to the `reactor.netty.ConnectionObserver.State.CONFIGURED` state. In those cases, even in the request, it can be confusing, why the `url` field is null. In an error response, maybe this can be understood.

Last, but not least, I was wondering about the return type of this method. Is String good enough or maybe it could return a more versatile `java.net.URI`?

violetagg:
> @violetagg Actually I was thinking about something similar, but keeping the method as something like `HttpClientResponse#fullUri`. I don't know much about the library, but I can see, that `HttpServerRequest` and `HttpServerResponse` also extend the `HttpInfos` interface. My problem with it is that this information does not seem to be valid on those classes.

This is also valid for the server side

Let's do the following
- deprecate `uri()` and introduce `requestUri()`
- add new method `requestUrl()`

both will return String

Wdyt?

Tankika:
@violetagg 

> This is also valid for the server side. On the server side you might need info for the URL the client used to make the request.

And what should be the implementation on the server side? In the `HttpServerOperations`, the `ConnectionInfo` class holds information about the host. I assume we should use the `hostAdress` field, that contains the original `host` header value the client used, even if there was a proxy in between.

> When this happen, are HttpClientRequest/HttpClientResponse available for users?

Yes, they are. Please see: `reactor.netty.http.client.HttpClientDoOnError` and `reactor.netty.http.client.HttpClient#doOnError`.

What would be the difference between the `requestUri()` and the `requestUrl()` you mentioned?

To be honest, I was proposing to add this field to the `HttpClientResponse`, because there is a clean implementation for it, the interface is only slightly changed. When we add this field to the `HttpInfos`, too many changes are needed for my taste. My main reason for adding it to the `HttpClientResponse`, instead of the `HttpInfos` is that, if it turns out, that there is a problem with the field in any of the descendants of `HttpInfos`, then it is hard to fix the issue. When we put this field now only into the `HttpClientResponse` it will be very easy to introduce the field into other interfaces descending from `HttpInfos`, when we find out, that we want to use it there too and we have a clean idea how to do so.

violetagg:
> Yes, they are. Please see: `reactor.netty.http.client.HttpClientDoOnError` and `reactor.netty.http.client.HttpClient#doOnError`.

We always have either HttpClientOperations or HttpClientConfiguration

> To be honest, I was proposing to add this field to the `HttpClientResponse`, because there is a clean implementation for it, the interface is only slightly changed. When we add this field to the `HttpInfos`, too many changes are needed for my taste. My main reason for adding it to the `HttpClientResponse`, instead of the `HttpInfos` is that, if it turns out, that there is a problem with the field in any of the descendants of `HttpInfos`, then it is hard to fix the issue. When we put this field now only into the `HttpClientResponse` it will be very easy to introduce the field into other interfaces descending from `HttpInfos`, when we find out, that we want to use it there too and we have a clean idea how to do so.

OK add only `requestUrl()` to the HttpClientResponse. Also please consider setting the url info in `reactor.netty.http.client.HttpClientConnect.HttpClientHandler#requestWithBody`

Tankika:
@violetagg 
Sorry for being silent for a while. I was thinking a lot about what you wrote and I've changed my PR the following way:
 - Change `responseUrl` to `resourceUrl` (instead of the suggested `requestUrl`): I feel neither `responseUrl`, nor `requestUrl` was truly the best name for this method. I believe the URL of the resource is really, in essence, what we get from this method, so I feel this is the most appropriate name.
 - Moved setting of the `resourceUrl` field into the  `reactor.netty.http.client.HttpClientConnect.HttpClientHandler#requestWithBody`, as suggested.
 - I did NOT put any logic into `reactor.netty.http.client.HttpClientDoOnError.PreparingHttpClientRequest#resourceUrl`, it simply returns a null. This is, because I do not feel comfortable with replicating the complex URL creation logic from `reactor.netty.http.client.HttpClientConnect.HttpClientHandler`'s constructor. I also feel that it could lead to more problems, for example, if there was an error during the URL creation part.

What do you think, is it OK now?

violetagg:
The change was committed to 0.9.x branch - 06c147c83252ad0d805737afb266cdef23cb700b
and then merged into master

Thanks for the PR

PR #923, opened 2019-12-04
fix #922 Add ability to configure maxLifeTime for connection pools

### LINKED ISSUE #922: Add ability to configure maxLifeTime of connections in connection pool
## Motivation
Currently it's possible to configure max idle time of connections in "elastic" and "fixed" pools, but there doesn't seem to be a way to configure max life time (equivalent of TTL in Apache HttpClient).
Without it we're not able to use AWS Elastic Beanstalk CNAME swaps, as both old and new environments are always running and connections to the old environment are never closed (regardless of DNS cache settings).

## Desired solution
Extend ConnectionProvider.elastic and fixed methods with nullable maxLifeTime parameter.

mgr32:


violetagg:
@mgr32 master branch currently is `0.10.x` which is under development. Are you targeting `0.9.x`?

mgr32:
Right, sorry. I created pull request #924 for 0.9.x.

PR #927, opened 2019-12-06
fix #663 Allow client to proxy ping frames instead of respond

JacobASeverson:
Fix for the situation in https://github.com/spring-cloud/spring-cloud-gateway/issues/729 where Spring Cloud Gateway responds to ping frames from a websocket server instead of proxying them to the client. I used the project from that issue to verify the functionality.

Please see the following commits in spring-framework and spring-cloud-gateway that will make this configurable by a new `spring.cloud.gateway.httpclient.websocket.proxy-ping` property:

https://github.com/spring-projects/spring-framework/compare/master...JacobASeverson:proxy-ping

https://github.com/spring-cloud/spring-cloud-gateway/compare/master...JacobASeverson:issue-729

PR #945, opened 2020-01-07
fix #663 Allow server to proxy ping frames instead of respond

### LINKED ISSUE #663: Add option to allow WebsocketServerOperations to not respond to Ping frames
This is a use case for Spring Cloud Gateway that uses Spring's WebSockets support to proxy WebSockets. The problem is that the ping frame doesn't get proxied to the backend WebSocket server.

See https://github.com/spring-cloud/spring-cloud-gateway/issues/729 for the original issue.

See https://github.com/reactor/reactor-netty/compare/master...spencergibb:spencergibb-proxy-ping for a POC that works (it just uses a System Property as a flag, probably want a different mechanism).

violetagg:


spencergibb:
How would that be enabled? Gateway doesn't use any of those classes directly.

violetagg:
> How would that be enabled? Gateway doesn't use any of those classes directly.

We need a change here
https://github.com/spring-projects/spring-framework/blob/master/spring-webflux/src/main/java/org/springframework/web/reactive/socket/client/ReactorNettyWebSocketClient.java#L109
https://github.com/spring-projects/spring-framework/blob/master/spring-webflux/src/main/java/org/springframework/web/reactive/socket/server/upgrade/ReactorNettyRequestUpgradeStrategy.java#L81

/cc @rstoyanchev

spencergibb:
Thanks

Fetsivalen:
Hi @violetagg, do you have plans to backport this change to the 9xx branch, or it will be available only with 1xx release?
And the same question for https://github.com/reactor/reactor-netty/pull/927
I think that could be a lot of users willing to see this in the spring cloud gateway.

violetagg:
ok I'll back port that
however Spring Gateway will need a change in Spring Framework

violetagg:
Backported with 3edbbf6c 8aa8846c and will be available in 0.9.3

Fetsivalen:
Thank you @violetagg!

spencergibb:
See https://github.com/spring-projects/spring-framework/issues/24331

PR #947, opened 2020-01-09
Optimize metrics implementation

### LINKED ISSUE #931: Optimize MicrometerHttpMetricsRecorder
##  Motivation
When profiling with a reactor-netty based app under load, with http metrics enabled, I've noticed a significant amount of time spent in `MicrometerHttpMetricsRecorder` and `MicrometerHttpClientMetricsRecorder`.  In concrete terms, metric recording takes about 1/3 of the time spent inside `AbstractChannel.writeAndFlush`.

Here is a flamegraph generated from [yourkit async cpu profiling](https://www.yourkit.com/docs/java/help/async_sampling_cpu.jsp) that shows the relative time spent inside of the metric recorder, when compared to the overall time spent in writeAndFlush.  The full width represents the length of time for `writeAndFlush`, and you can see that roughly 1/3 of that is for metrics.

http_metrics_flamegraph

I believe a lot of that time spent is due to:
* object allocation (specifically for `DistributionSummary.Builder`, `Timer.Builder`, `ImmutableTag`, and `Tag`)
* metric registration

## Desired solution
Would it be possible to cache the `DistributionSummary` and `Timer` objects inside of the metrics recorder, so that the builders and tags do not need to be instantiated and metrics registered with the registry on every network write?

For example, cache the return value of `register` on line 78, so that the only thing that happens when recording is a cache lookup, and the invocation of `.record`:

https://github.com/reactor/reactor-netty/blob/576850e2cf6b6401d7e0798c83d2b8d8e41914c8/src/main/java/reactor/netty/http/MicrometerHttpMetricsRecorder.java#L72-L80

This seems like it would reduce the overhead of metric recording significantly,

violetagg:
Fix #931

### DIFF
--- a/src/main/java/reactor/netty/channel/MicrometerChannelMetricsRecorder.java
+++ b/src/main/java/reactor/netty/channel/MicrometerChannelMetricsRecorder.java
@@ -20,11 +20,13 @@ import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Metrics;
 import io.micrometer.core.instrument.Timer;
+import io.netty.util.internal.PlatformDependent;
 
-import javax.annotation.Nullable;
 import java.net.SocketAddress;
 import java.time.Duration;
+import java.util.concurrent.ConcurrentMap;
 
+import static reactor.netty.Metrics.ADDRESS_RESOLVER;
 import static reactor.netty.Metrics.CONNECT_TIME;
 import static reactor.netty.Metrics.DATA_RECEIVED;
 import static reactor.netty.Metrics.DATA_SENT;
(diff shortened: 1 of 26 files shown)


src/main/java/reactor/netty/channel/MicrometerChannelMetricsRecorder.java
```diff
@@ -76,94 +79,73 @@ protected MicrometerChannelMetricsRecorder(String name, @Nullable String remoteA
...
-			this.dataSent = dataSentBuilder.register(registry);
+		this.tlsHandshakeTimeBuilder =
+				Timer.builder(name + TLS_HANDSHAKE_TIME)
+				     .description("Time that is spent for TLS handshake");
 
-			this.errorCountBuilder.tag(REMOTE_ADDRESS, remoteAddress);
-			this.errorCount = errorCountBuilder.register(registry);
-		}
+		this.addressResolverTimeBuilder =
+				Timer.builder("reactor.netty.address.resolver")
```
jessicadeery: Should **_name_** be set for the **_reactor.netty.address.resolver_** metric (_Timer.builder(**_name_** + "address.resolver")_) or is it deliberately not set for some reason?

violetagg: ok let's use the name also for the DNS resolver

src/main/java/reactor/netty/http/MicrometerHttpMetricsRecorder.java
```diff
@@ -21,82 +21,126 @@
...
 	@Override
 	public void recordDataReceived(SocketAddress remoteAddress, String uri, long bytes) {
-		DistributionSummary.builder(name + DATA_RECEIVED)
-		                   .baseUnit("bytes")
-		                   .description("Amount of the data that is received, in bytes")
-		                   .tags(REMOTE_ADDRESS, Metrics.formatSocketAddress(remoteAddress), URI, uri)
-		                   .register(registry)
-		                   .record(bytes);
+		String address = Metrics.formatSocketAddress(remoteAddress);
+		MeterKey meterKey = MeterKey.builder().uri(uri).remoteAddress(address).build();
```
philsttr: Since we're trying to optimize this particularly hot path, it seems like a direct instantiation of a MeterKey object would be better than using a temporary builder

violetagg: agree

PR #949, opened 2020-01-14
fix #948 Fix IllegalArgumentException while resolving the path

### LINKED ISSUE #948: Failure to process URL where path is nothing but "//"
After upgrading to Dysprosium snapshots, there is one test failure in the Spring Framework in a scenario where the path is nothing but a double slash, e.g. `"http://localhost:8080//"`.

It results in the following:

```
java.lang.IllegalArgumentException: Expected authority at index 2: //
	at java.net.URI.create(URI.java:852) ~[?:1.8.0_222]
	at reactor.netty.http.HttpOperations.resolvePath(HttpOperations.java:300) ~[reactor-netty-0.9.3.BUILD-SNAPSHOT.jar:0.9.3.BUILD-SNAPSHOT]
	at reactor.netty.http.server.HttpServerOperations.<init>(HttpServerOperations.java:128) ~[reactor-netty-0.9.3.BUILD-SNAPSHOT.jar:0.9.3.BUILD-SNAPSHOT]
	at reactor.netty.http.server.HttpTrafficHandler.channelRead(HttpTrafficHandler.java:157) ~[reactor-netty-0.9.3.BUILD-SNAPSHOT.jar:0.9.3.BUILD-SNAPSHOT]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:377) [netty-transport-4.1.44.Final.jar:4.1.44.Final]
...
```

This looks like it might be a side effect of refactoring for #926 in https://github.com/reactor/reactor-netty/commit/dc3fb4b1fe1538e1ac1249412a272a7204912c27.

violetagg:


src/main/java/reactor/netty/http/HttpOperations.java
```diff
@@ -297,15 +297,32 @@ protected final boolean markSentHeaderAndBody() {
...
+				char c = path.charAt(i);
+				if (c == '?' || c == '#') {
+					path = path.substring(0, i);
+					break;
+				}
+			}
+		}
+		else {
+			path = URI.create(uri).getPath();
+		}
```
rstoyanchev: The if does not decode while the else does, which seems inconsistent:
```java
assertEquals("a b", HttpOperations.resolvePath("http://localhost:8080/a%20b"));
assertEquals("a%20b", HttpOperations.resolvePath("/a%20b"));
```

rstoyanchev: The if does not decode while the else does, which seems inconsistent:
```java
assertEquals("a b", HttpOperations.resolvePath("http://localhost:8080/a%20b"));
assertEquals("a%20b", HttpOperations.resolvePath("/a%20b"));
```

One option might be something like this instead?
```java
uri = (uri.charAt(0) == '/' ? "http://abc.org" + uri : uri); 
String path = URI.create(uri).getPath();
```

src/test/java/reactor/netty/http/HttpOperationsTest.java
```diff
@@ -90,4 +94,28 @@ public void httpAndJsonDecoders() {
...
+		assertEquals("a", HttpOperations.resolvePath("http://localhost:8080/a/"));
+		assertEquals("", HttpOperations.resolvePath("http://localhost:8080/?b"));
+		assertEquals("a", HttpOperations.resolvePath("http://localhost:8080/a?b"));
+		assertEquals("", HttpOperations.resolvePath("http://localhost:8080/#b"));
+		assertEquals("a", HttpOperations.resolvePath("http://localhost:8080/a#b"));
+		assertEquals("a", HttpOperations.resolvePath("http://localhost:8080/a?b#c"));
+
+		assertEquals("", HttpOperations.resolvePath(""));
+		assertEquals("", HttpOperations.resolvePath("/"));
+		assertEquals("/", HttpOperations.resolvePath("//"));
```
rstoyanchev: I realize this is preexisting behavior but the Javadoc says the leading and trailing slash are removed if present, but here only the leading slash is removed. Is the second slash not considered a trailing slash? It also means the following two have the same result :

```java
assertEquals("/", HttpOperations.resolvePath("//"));
assertEquals("/", HttpOperations.resolvePath("///"));
```

violetagg:
@rstoyanchev Please check this #952

rstoyanchev:
LGTM

PR #956, opened 2020-01-16
fix #955 Deprecate the methods for creating ConnectionProvider in favour of ConnectionProvider.Builder

### LINKED ISSUE #955: Deprecate the methods for creating ConnectionProvider in favour of ConnectionProvider builder

violetagg:


src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -240,13 +242,29 @@ default int maxConnections() {
...
+	 *         <p>It will never wait before opening a new connection.
+	 *         The reuse window is limited but it cannot starve an undetermined volume of clients using it.
+	 *     </li>
+	 *     <li>
+	 *         Fixed {@link ConnectionProvider} will cache and reuse a fixed maximum number of {@link Connection}.
+	 *         <p>It will open up to the given max number of processors observed by this jvm (minimum 4).
+	 *         Further connections will be pending acquisition depending on acquireTimeout.
+	 *     </li>
+	 * </ul>
+	 */
```
rstoyanchev: This would become simpler if there was no fixed vs elastic.

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -277,23 +295,23 @@ public final Builder name(String name) {
 		 * Set the options to use for configuring {@link ConnectionProvider} acquire timeout.
 		 * Default to DEFAULT_POOL_ACQUIRE_TIMEOUT.
```
rstoyanchev: This could be a `{@link ...}` to the constant.

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -277,23 +295,23 @@ public final Builder name(String name) {
...
+		public final Builder acquireTimeout(Duration acquireTimeout) {
+			this.acquireTimeout = Objects.requireNonNull(acquireTimeout, "acquireTimeout");
 			return this;
 		}
 
 		/**
 		 * Set the options to use for configuring {@link ConnectionProvider} maximum connections.
 		 * Default to DEFAULT_POOL_MAX_CONNECTIONS.
+		 * When invoked with MAX_CONNECTIONS_ELASTIC an elastic ConnectionProvider will be created
+		 * and acquireTimeout will be set automatically to ACQUIRE_TIMEOUT_NEVER_WAIT.
```
rstoyanchev: These could  `{@link ...}` to the constants.

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -240,13 +242,29 @@ default int maxConnections() {
 		return -1;
 	}
 
+	/**
+	 * Create a {@link ConnectionProvider}.
```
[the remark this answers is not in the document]

src/main/java/reactor/netty/resources/ConnectionProvider.java

src/main/java/reactor/netty/resources/ConnectionProvider.java

src/main/java/reactor/netty/resources/ConnectionProvider.java

src/main/java/reactor/netty/resources/ConnectionProvider.java

violetagg:
> I think a couple of the `fixed` methods (with name and maxConnections) are useful shortcuts still. I can see that also from the refactored usages. Perhaps a `create(String)` and `create(String, int)` could be added as shortcut alternatives, and the builder is for anything else (acquire, idle time, etc). It would also look better next to `newConnection` which would otherwise remain as the only static factory method.

Added `create(String)` and `create(String, int)`

> It would be nice to have `builder(String)` directly on `ConnectionProvider` and deprecate `newInstance` on the Builder:
> 
> ```java
> ConnectionProvider.builder("fixed")...
> // vs.
> ConnectionProvider.Builder.newInstance("fixed")...
> ```

Added `builder(String)`, no need to deprecate `Builder.newInstance` as we still do not have released version with this method.

violetagg:
Committed with 71af605daf5d6b8d41060725d64edb351e6c048a

PR #957, opened 2020-01-16
fix #951 Add API for configuring max pending acquire requests

### LINKED ISSUE #951: Connection pool maxPending size unbounded by default and cannot be configured
In plain netty 4.1.x it is possible to configure and query pool maxPending size in addition to acquireTimeout.

When reactor-netty fixed Connection pool is used maxPending size cannot be configured and it is set internally to unbound. Only the following parameters can be configured
- pool name
- maxConnections
- acquireTimeout
- MaxIdleTime

Having unbound is problematic because service can run out of memory or out of resource just like elastic connection pool might result out of sockets.

reactor-netty should have configuration option to set/get pool maxPending value and have bound default value as default.

violetagg:


violetagg:
superseded with #960

PR #963, opened 2020-01-19
Fix #962 Add configuration for switching between FIFO and LIFO order for the channels in the pool

### LINKED ISSUE #962: Add configuration for switching between FIFO and LIFO order for the channels in the pool
The current implementation uses FIFO and this cannot be changed.
https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/resources/PooledConnectionProvider.java#L108

This feature request is about adding a configuration for switching between FIFO and LIFO.
https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/resources/ConnectionProvider.java#L168

jdaru:


src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -274,6 +275,19 @@ public final Builder maxLifeTime(Duration maxLifeTime) {
...
 
+		/**
+		 * Switch order for the channels in the pool to LIFO.
+		 * <br>
+		 * Default FIFO.
+		 *
+		 * @param lifo flag for switching order for the channel to LIFO
+		 * @return {@literal this}
+		 */
+		public final Builder lifo(boolean lifo) {
```
violetagg: Let's change this method to two methods -> `fifo()` and `lifo()`

jdaru: @violetagg ,what do think about this approach? should i use primitive or Object boolean?
i mean, if we'll use primitive boolean then we don't need to use (e.g.) fifo(), because it will use by default.
or do we need to explicitly use one of the methods?

violetagg: The point is that at the moment we have only two types of leasing strategies, so using the approach that you implemented is OK for the moment. But if we decide to add a third type of leasing strategy we will not be able to do it.

jdaru: aaa, ok, thanks :)

jdaru:
@violetagg , something like that?

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -111,8 +115,7 @@ static ConnectionProvider create(String name) {
 	 * number of {@link Connection}
 	 */
 	static ConnectionProvider create(String name, int maxConnections) {
-		return builder(name).maxConnections(maxConnections)
-		                    .build();
+		return builder(name).maxConnections(maxConnections).lifo();
```
violetagg: Why did you change the strategy to LIFO, we still want it FIFO by default?

jdaru: @violetagg if lifo is the default, then we won’t need an explicit switch to lifo.
we can just use .build(), or fifo() if need.

violetagg: @jdaru We say that FIFO is by default, so I would expect that the methods that do not provide Builder (the two create methods) will always create FIFO, if somebody wants LIFO then he/she needs to use the Builder and not these two methods.

violetagg:
@jdaru I squashed the commits and applied the changes to 0.9.x branch with 0c50da819f766ef4897b654f11f4c9472ba5c754
also forward merged the changes to master branch

Thanks for the contribution

PR #966, opened 2020-01-21
ConnectionProvider API refinements

violetagg:


src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -274,7 +274,7 @@ static ConnectionProvider fixed(String name, int maxConnections, long acquireTim
 	@Deprecated
 	static ConnectionProvider fixed(String name, int maxConnections, long acquireTimeout, @Nullable Duration maxIdleTime, @Nullable Duration maxLifeTime) {
 		return builder(name).maxConnections(maxConnections)
-		                    .acquireTimeout(Duration.ofMillis(acquireTimeout))
+		                    .pendingAcquireTime(Duration.ofMillis(acquireTimeout))
```
simonbasle: maybe `pendingAcquireTimeout`?

violetagg:
> One thing I still wonder about is a default value for pendingAcquireMaxCount. Currently maxConnections and pendingAcquireTimeout have bounded, default values. However pendingAcquireMaxCount is unbounded by default and while the timeout of 45 secs helps, it still leaves plenty of room for connections to pile up, and that could be surprising. I think out of the box this should behave in a more bounded way that errs on the side of failing at some upper limit.
> 
> Perhaps outside of the deprecated elastic factory methods, pendingAcquireMaxCount should also default to some upper bound?

I changed `pendingAcquireMaxCount` to have a limit for the new functionality. However for the deprecated one (0.9.x) it will be unbounded. For 1.0.x it will have limit by default - no more unbounded.

PR #971, opened 2020-01-24
fix #970 Ensure WebSocket compression is enabled on the server in cases there is a compressionPredicate

### LINKED ISSUE #970: WebSocket compression extension should not depends on SimpleCompressionHandler
## Expected Behavior
If compression is enabled on the server side then WebSocket compression extensions should be applied according to https://tools.ietf.org/html/rfc7692#section-5 and not depend on `SimpleCompressionHandler` present in pipeline or not.

## Actual Behavior
Configure `HttpServer` with enable compression and set `maxResponseSize` to value grate than zero, WebSocket compression never applied.

## Steps to Reproduce

```java
public class EchoServer {

    public static void main(String[] args) {
        HttpServer server =
            HttpServer.create()
                .port(8080)
                .compress(true)
                .compress(2048)
                .wiretap(true)
                .route(r -> r.ws("/ws", (in, out) ->  out.sendObject(in.receiveFrames().doOnNext(WebSocketFrame::retain))));

        server.bindNow()
            .onDispose()
            .block();
    }
}
```

## Possible Solution
Change logic inside `WebsocketServerOperations` constructor.

## Your Environment

* Reactor version(s) used: 1.0.0.BUILD-SNAPSHOT
* Other relevant libraries versions (eg. `netty` ):  4.1.45.Final
* JVM version (`javar -version`): 1.8.0_232
* OS and version (eg `uname -a`): Darwin Kernel Version 18.7.0: Sun Dec  1 18:59:03 PST 2019; root:xnu-4903.278.19~1/RELEASE_X86_64 x86_64

violetagg:


src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
```diff
@@ -93,7 +93,7 @@
 			       .set(replaced.nettyRequest.headers());
 
 			if (channel().pipeline()
-			             .get(NettyPipeline.CompressionHandler) != null) {
+			             .get(NettyPipeline.CompressionHandler) != null || compressionPredicate != null) {
```
rstoyanchev: Shouldn't the predicate be tested like in `HttpServerOperations#preSendHeadersAndStatus`?

```java
if (channel().pipeline().get(NettyPipeline.CompressionHandler) != null ||
		(compressionPredicate != null && compressionPredicate.test(this, this))) {
    // ...
}
```

amizurov:
@violetagg thx for reactive fix and in common case it works, but if we change configuration to
 ```
 HttpServer server =
            HttpServer.create()
                .port(8080)
                .compress(false)
                .compress(2048)
                .wiretap(true)
                .route(r -> r.ws("/ws", (in, out) -> out.sendObject(in.receiveFrames().doOnNext(WebSocketFrame::retain))));
```
the WebSocket compression is applied because predicate present, looks unobvious. WDYT ?

src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
violetagg: We cannot test the predicate as it makes sense only for HTTP.
We use just the existence of the predicate as an indicator that the user wants `compression` on the server.
I would never use `compress(2048)` when I use the HTTP client for WebSocket, but when you are on the spring gateway where the HTTP client serves both HTTP and WebSocket it make sense.

violetagg:
> @violetagg thx for reactive fix and in common case it works, but if we change configuration to
> 
> ```
> HttpServer server =
>            HttpServer.create()
>                .port(8080)
>                .compress(false)
>                .compress(2048)
>                .wiretap(true)
>                .route(r -> r.ws("/ws", (in, out) -> out.sendObject(in.receiveFrames().doOnNext(WebSocketFrame::retain))));
> ```
> 
> the WebSocket compression is applied because predicate present, looks unobvious. WDYT ?

There is kind of misunderstanding here
`.compress(true)` means compress always
then you override this setting with
`.compress(2048)` which mean stop compressing always (because this was the previous setting) and start compressing when this condition is met.

then the second example is 
`.compress(false)` means DO NOT compress
then you override this setting with
`.compress(2048)` which mean start compressing when this condition is met.

amizurov:
> > @violetagg thx for reactive fix and in common case it works, but if we change configuration to
> > ```
> > HttpServer server =
> >            HttpServer.create()
> >                .port(8080)
> >                .compress(false)
> >                .compress(2048)
> >                .wiretap(true)
> >                .route(r -> r.ws("/ws", (in, out) -> out.sendObject(in.receiveFrames().doOnNext(WebSocketFrame::retain))));
> > ```
> > 
> > 
> > the WebSocket compression is applied because predicate present, looks unobvious. WDYT ?
> 
> There is kind of misunderstanding here
> `.compress(true)` means compress always
> then you override this setting with
> `.compress(2048)` which mean stop compressing always (because this was the previous setting) and start compressing when this condition is met.
> 
> then the second example is
> `.compress(false)` means DO NOT compress
> then you override this setting with
> `.compress(2048)` which mean start compressing when this condition is met.

Yes i understand this, but compression for http response and webSocket compression is not depend to each other, also i can use simultaneously `ws` and `http` and i want to configure each of them independently IMHO.

violetagg:
> Yes i understand this, but compression for http response and webSocket compression is not depend to each other, also i can use simultaneously ws and http and i want to configure each of them independently IMHO.

We do not support that. If you want I can revert this change and we can treat your issue as a feature request.
wdyt?

amizurov:
> > Yes i understand this, but compression for http response and webSocket compression is not depend to each other, also i can use simultaneously ws and http and i want to configure each of them independently IMHO.
> 
> We do not support that. If you want I can revert this change and we can treat your issue as a feature request.
> wdyt?

Thx again for your time. I'm ok with this PR but we change behavior for `org.springframework.boot.web.embedded.netty.NettyWebServer` because before it websocket compression never work :). I agree that is a good idea to create separate feature request where we can improve current logic and clear api. Also we can improve websocket compression to selective approach (Netty suported this https://github.com/netty/netty/pull/8910/files/6bdf7ad8bf6c6f23111671fbaba7403ab38c53fe)  and apply only to websocket frames by predicate like http does.

violetagg:
> Thx again for your time. I'm ok with this PR but we change behavior for org.springframework.boot.web.embedded.netty.NettyWebServer because before it websocket compression never work :). I agree that is a good idea to create separate feature request where we can improve current logic and clear api. Also we can improve websocket compression to selective approach (Netty suported this https://github.com/netty/netty/pull/8910/files/6bdf7ad8bf6c6f23111671fbaba7403ab38c53fe) and apply only to websocket frames by predicate like http does.

@rstoyanchev wdyt? My opinion I that we should separate both configurations.

rstoyanchev:
Yes I think it would be a good idea to separate the two. Unless only `compress(true)` is used, the presence of any other compress predicate (size or custom), since it can't be evaluated properly, that makes it ambiguous whether it should even apply to WebSocket.

violetagg:
I'm closing this PR - we will NOT apply this fix.

There will be a WebSocket configuration for compression on/off.
For the WebSocketExtensionFilterProvider support I created a separate feature request #977

violetagg:
new PR #980

Issue #950, opened 2020-01-14
Add option for changing the max URI tags


violetagg:
https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/http/MicrometerHttpMetricsRecorder.java#L96

michaelmcfadyen:
I'm looking to pick this up but looking for some advice on the design of the solution. Looking through the code, I code see a few ways to implement this.

 - overload the `metrics()` function on HttpServer to accept a new parameter - `maxUriTags`
 - add new method on HttpServer specifically for setting the maxUriTags ie. `maxUriTags(int maxTags)`
 - add a parameter to `MicrometerHttpMetricsRecorder` constructor 
`MicrometerHttpMetricsRecorder(String name, String protocol, int maxUriTags)` 
which is passed up from both `MicrometerHttpServerMetricsRecorder` and `MicrometerHttpClientMetricsRecorder`. Then `metrics(boolean metricsEnabled, HttpServerMetricsRecorder recorder)` can be used to override the default max uri tags value.

However, I think for the third option both `MicrometerHttpClientMetricsRecorder` and `MicrometerHttpServerMetricsRecorder` classes would have to change scope from `default` to `public`.

violetagg:
I was thinking for something like this:

`metrics(boolean metricsEnabled, int maxUriTags)`
`BootstrapHandlers#updateMetricsSupport(io.netty.bootstrap.ServerBootstrap, reactor.netty.channel.ChannelMetricsRecorder, int maxUriTags)`
`MetricsSupportConsumer(ChannelMetricsRecorder recorder, int maxUriTags, boolean onServer)`

so when we add `NettyPipeline.ChannelMetricsHandler` to the pipeline if the recorder is `MicrometerHttpMetricsRecorder` we can add this filter

wdyt?

parviz-93:
Hi.
I would like to join your project. Can I start with this task?

violetagg:
@parviz-93 @michaelmcfadyen already stated that he is willing to take this issue.
@michaelmcfadyen Are you still going to work on this?

parviz-93:
@violetagg got it. 
Maybe you have some simple task for me? I would be happy to help 😊

violetagg:
@parviz-93 let's move the discussion here https://gitter.im/reactor/reactor-netty

michaelmcfadyen:
[...]

PR #972, opened 2020-01-25
Add option for changing the max URI tags

michaelmcfadyen:
#950

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -830,6 +830,20 @@ public abstract class HttpClient {
 	 * @return a new {@link HttpClient}
 	 */
 	public final HttpClient metrics(boolean metricsEnabled) {
+		return metrics(metricsEnabled, reactor.netty.Metrics.MAX_URI_TAGS);
+	}
+
+	/**
+	 * Specifies whether the metrics are enabled on the {@link HttpClient}
+	 * and the maximum number of uri tags allowed on the http server metrics.
+	 * All generated metrics are registered in the Micrometer MeterRegistry,
+	 * assuming Micrometer is on the classpath.
+	 *
+	 * @param metricsEnabled if true enables the metrics on the client.
+	 * @param maxUriTags maximum number of uri tags allowed on metrics
+	 * @return a new {@link HttpClient}
+	 */
+	public final HttpClient metrics(boolean metricsEnabled, int maxUriTags) {
 		if (metricsEnabled) {
 			if (!Metrics.isInstrumentationAvailable()) {
 				throw new UnsupportedOperationException(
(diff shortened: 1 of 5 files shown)


violetagg:
@michaelmcfadyen Currently we do not expose an API for creating metrics with different basic name and arbitrary repository. Having only on instance of the recorders is deliberate. #947 

Let's say we have this

```
HttpClient basicClient = HttpClient.create();

HttpClient client1 = basicClient.metrics(true);

HttpClient client2 = basicClient.metrics(true, 50);
```

at the end we will have max uri tags -> 50, but we will have two instances of the recorder and we will record one and the same metrics twice in the repository.

michaelmcfadyen:
I guess it really doesn't make sense to pass in the maxUriTags when building the HttpClient/HttpServer as the meter filter should be a singleton and is applied across all HttpClients. 
Having the maxUriTags param on the HttpClient is actually misleading to the person using the API as it suggests you can have different maxUriTags per HttpClient, which isn't the case.

src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -830,6 +830,20 @@ public final HttpClient wiretap(boolean enable) {
 	 * @return a new {@link HttpClient}
 	 */
 	public final HttpClient metrics(boolean metricsEnabled) {
+		return metrics(metricsEnabled, reactor.netty.Metrics.MAX_URI_TAGS);
+	}
```
rstoyanchev: Now that there is an overloaded method, the Javadoc for this one should probably mention the default value used for maxUriTags.

src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -839,7 +853,7 @@ public final HttpClient metrics(boolean metricsEnabled) {
 
 			return tcpConfiguration(tcpClient ->
 					tcpClient.bootstrap(b -> BootstrapHandlers.updateMetricsSupport(b,
-							MicrometerHttpClientMetricsRecorder.INSTANCE)));
+							new MicrometerHttpClientMetricsRecorder(maxUriTags))));
```
rstoyanchev: The `INSTANCE` constant is no longer used and could be removed. Or it could still be applied when `maxUriTags == Metrics.MAX_URI_TAGS` but these methods aren't called often enough to benefit from a constant.

src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -318,7 +332,7 @@ public final HttpServer metrics(boolean metricsEnabled) {
 
 			return tcpConfiguration(tcpServer ->
 					tcpServer.bootstrap(b -> BootstrapHandlers.updateMetricsSupport(b,
-							MicrometerHttpServerMetricsRecorder.INSTANCE)));
+							new MicrometerHttpServerMetricsRecorder(maxUriTags))));
 		}
```
rstoyanchev: Same here about `INSTANCE` as for the client side.

src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -309,6 +309,20 @@ else if (proxyProtocolSupportType == ProxyProtocolSupportType.OFF) {
 	 * @return a new {@link HttpServer}
 	 */
 	public final HttpServer metrics(boolean metricsEnabled) {
+		return metrics(metricsEnabled, reactor.netty.Metrics.MAX_URI_TAGS);
+	}
```
rstoyanchev: Same comment here about the Javadoc as for the client side.

violetagg:
Superseded by #985

PR #978, opened 2020-01-29
fix #975 Add HttpClient.doAfterResponseSuccess that will be triggered only when the response is fully received

### LINKED ISSUE #975: Confusing javadoc for `doAfterResponse` and `doOnResponseError`
The [Javadoc](https://github.com/reactor/reactor-netty/blob/fbd1c30f0da8edb9b81a88ecf540f55d1983db84/src/main/java/reactor/netty/http/client/HttpClient.java#L576) for `doAfterResponse` suggests the callback will fire whenever the response was 'fully received'; however, the callback is described as an observer of disconnect events.  Are these equivalent things?  If the connection is persistent, would the `doAfterResponse` not fire for each response?  Our testing suggests the callback does fire upon each response; however, it doesn't seem tied to a 'fully received' response (see code below).

In our case, we are trying to use `doAfterResponse` and `doOnResponseError` to trigger different behavior.  It requires them to be mutually exclusive (i.e. we didn't expect both to fire for the same request/response).  I think the Javadoc suggests they are mutually exclusive.

Expectations:

- `doAfterResponse` should not fire if `doOnResponseError` fires
- Javadoc for `doAfterResponse` should be cleaned up a bit.

```java
    public static void main(final String[] args) throws Exception {
        final DisposableServer server =
            HttpServer.create().host("localhost")
                .route(routes -> routes
                    .get("/kill", (request, response) -> {
                        return response.sendString(
                            Flux.range(0, 10_000)
                                .map(i -> {
                                    if (i == 1_000) {
                                        response.withConnection(Connection::disposeNow);
                                    }
                                    return "a";
                                })
                        );
                    }))
                .bindNow();


        try {
            final String respBody = HttpClient.create()
                .doAfterResponse((resp, conn) -> {
                    System.out.println("I did not expect to be called:(");
                })
                .doOnResponseError((resp, exc) -> {
                    System.out.println("I expected to be called:(");
                })
                .get()
                .uri("http://localhost:" + server.port() + "/kill")
                .responseSingle((resp, bytes) -> bytes.asString())
                .block();

            // Should not get here!
            System.out.println(respBody);
        } finally {
            server.disposeNow();
        }

    }
```

* Reactor version(s) used: 0.9.3.RELEASE

violetagg:
Fix #975

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -494,7 +494,7 @@ public abstract class HttpClient {
 	 * {@link #doOnRequest(BiConsumer)} or {@link RequestSender#send(BiFunction)} might
 	 * not be visible if the error results from a connection failure.
 	 *
-	 * @param doOnRequest a consumer observing connected events
+	 * @param doOnRequest a consumer observing request failures
 	 * @param doOnResponse a consumer observing response failures
 	 *
 	 * @return a new {@link HttpClient}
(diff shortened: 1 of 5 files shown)


src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -573,15 +575,36 @@ public final HttpClient doOnResponseError(BiConsumer<? super HttpClientResponse,
...
+	 * returned to the pool or closed.
 	 *
-	 * @param doAfterResponse a consumer observing disconnected events
+	 * @param doAfterResponse a callback called after
+	 * {@link HttpClientState#RESPONSE_RECEIVED} has been emitted and the connection is
+	 * returned to the pool or closed.
 	 *
 	 * @return a new {@link HttpClient}
+	 * @deprecated as of 0.9.5. This callback is invoked in both cases when the response is successful/there are errors.
+	 * Consider using {@link #doAfterResponseSuccess(BiConsumer)} or {@link #doOnResponseError(BiConsumer)}
```
rstoyanchev: Maybe this can be in the main paragraph above, i.e. "Setup a callback called after... This callback is invoked in both cases when the response is successful/there..." and the then deprecation just saying in favor of...

PR #983, opened 2020-01-31
fix #976 Support for "X-Forwarded-Proto" headers Multiple Values

### LINKED ISSUE #976: Support "X-Forwarded-Proto" headers Multiple Values
Recently i found issue in an environment using proxies.
my proxy server sends traffic to netty server with following headers.

> X-Forwarded-Proto: https,https

however, reactor-netty doesn't support "x-forwarded-proto" headers multiple values. (only one value can be parsed, so, it causes a 'scheme' error.)
https://github.com/reactor/reactor-netty/blob/fbd1c30f0da8edb9b81a88ecf540f55d1983db84/src/main/java/reactor/netty/http/server/ConnectionInfo.java#L174

as far as i know, many frameworks can handle multiple values.
and "[Tornnado](https://github.com/tornadoweb/tornado)"web server framework supports it as well.

> [X-Forwarded-Proto Multiple Values](https://github.com/tornadoweb/tornado/issues/2161)

I hope reactor-netty also supports this feature.

devsonic:
#976 

changes:
- Added splitting for "x-forwarded-proto" multiple values. 
- Added test codes.

### DIFF
--- a/src/main/java/reactor/netty/http/server/ConnectionInfo.java
+++ b/src/main/java/reactor/netty/http/server/ConnectionInfo.java
@@ -173,7 +173,7 @@ final class ConnectionInfo {
 		}
 		String protoHeader = request.headers().get(XFORWARDED_PROTO_HEADER);
 		if (protoHeader != null) {
-			scheme = protoHeader.trim();
+			scheme = protoHeader.split(",", 2)[0].trim();
 		}
 		return new ConnectionInfo(hostAddress, remoteAddress, scheme);
 	}
--- a/src/test/java/reactor/netty/http/server/ConnectionInfoTests.java
+++ b/src/test/java/reactor/netty/http/server/ConnectionInfoTests.java
@@ -178,6 +178,40 @@ public class ConnectionInfoTests {
 	}
 
 	@Test
+	public void xForwardedForHostAndPortAndProto() {
+		testClientRequest(
+				clientRequestHeaders -> {
+					clientRequestHeaders.add("X-Forwarded-For", "192.168.0.1");
+					clientRequestHeaders.add("X-Forwarded-Host", "a.example.com");
+					clientRequestHeaders.add("X-Forwarded-Port", "8080");
+					clientRequestHeaders.add("X-Forwarded-Proto", "http");
+				},
+				serverRequest -> {
+					Assertions.assertThat(serverRequest.remoteAddress().getHostString()).isEqualTo("192.168.0.1");
+					Assertions.assertThat(serverRequest.hostAddress().getHostString()).isEqualTo("a.example.com");
+					Assertions.assertThat(serverRequest.hostAddress().getPort()).isEqualTo(8080);
+					Assertions.assertThat(serverRequest.scheme()).isEqualTo("http");
+				});
+	}
+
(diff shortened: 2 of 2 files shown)


violetagg:
@devsonic Can you elaborate more why you decided to take into consideration the first entry from the comma separated header value?

devsonic:
@violetagg 

As far as i know, reactor-netty use 'ConnectorInfo.java' to write requested client connection infos.
In it, "x-forwarded-proto" is used to make 'scheme' and "x-forwded-host" is used to make 'host address'. and then finally these properties would create base url to call api. (
[spring reference](https://github.com/spring-projects/spring-framework/blob/cbb037bc681e85517aee31ce623478ce0a9b3189/spring-web/src/main/java/org/springframework/http/server/reactive/ReactorServerHttpRequest.java#L73))

**In the reactor-netty code, "x-forwarded-host" was supposed to get the first element, so i decided to follow rule. (Similarly, 'proto' of forwarded specification also gets the first value.)**
https://github.com/reactor/reactor-netty/blob/fbd1c30f0da8edb9b81a88ecf540f55d1983db84/src/main/java/reactor/netty/http/server/ConnectionInfo.java#L166

Interestingly, other project(tornade) handled it differently.
they decided to use last entry of the 'x-forwarded-proto'. (refer to [link](https://github.com/tornadoweb/tornado/pull/2162))
they seem to think last entry would be more reliable.

I think it can be different for different server frameworks.
And also not all proxies set this header or support the multi value format.
the main reason is that it is de-factor.

if you have any opinion, please let me know. thank you.

violetagg:
@devsonic What do you think if we expose a new API?

`reactor.netty.http.server.HttpServer#forwarded(boolean, Function<?, ConnectionInfo>)`

so if you just use `reactor.netty.http.server.HttpServer#forwarded(boolean)` it will work as it is now,
if you use the new API, you will receive the request, etc. and as a result you will return `ConnectionInfo`

devsonic:
@violetagg 
good idea!

but i have slight worries about that.
when using a new api, 'ConnectionInfo' requires too much information to create.
and i just need to focus on how to implement the forwarded information.
so, it might be better to just add a generating rule for forwarded connection, rather than create a whole thing. 
otherwise i think it could be better to make ConnectionInfo into abstract class, and then you would override the methods that add forward rules.
```
reactor.netty.http.server.HttpServer#forwardedConnection(Function<?, ForwardedConnection>)
 // it will work if 'forwarded' options is true
```

and as you mentioned, 
in current api, It's a good idea to use the first entry just like PR in order to be consistent with the rest.

violetagg:
@devsonic I'll be happy to review your idea when it is ready
👍

violetagg:
@devsonic should I/you create a feature request for this

`reactor.netty.http.server.HttpServer#forwardedConnection(Function<?, ForwardedConnection>)
 // it will work if 'forwarded' options is true`

or we will cover it with this PR?

devsonic:
@violetagg 
would you please create the feature request for this? 
then i will follow the issue you create.

violetagg:
@devsonic thanks for the contribution
here is the new feature request #990

PR #985, opened 2020-01-31
fix #950 Remove maximum allowable tags configuration

### LINKED ISSUE #950: Add option for changing the max URI tags

violetagg:
As the global composite repository is used Reactor Netty users can
configure maximum allowable tags outside of Reactor Netty.
Fix #950


violetagg:
@michaelmcfadyen @rstoyanchev PTAL

src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -848,8 +848,14 @@ public final HttpClient wiretap(boolean enable) {
...
 	 * Specifies whether the metrics are enabled on the {@link HttpClient}.
-	 * All generated metrics are registered in the Micrometer MeterRegistry,
+	 * All generated metrics are registered in the global composite registry
+	 * provided by Micrometer with name prefix {@code "reactor.netty.http.client"},
 	 * assuming Micrometer is on the classpath.
+	 * <p>
+	 * Note: As the the global composite registry is used for registering the meters,
+	 * all MeterFilter (for example for limiting the maximum allowable tags should be applied
+	 * outside Reactor Netty.
+	 * </p>
```
rstoyanchev: The note seems incomplete (and a closing parenthesis is missing). Perhaps:

```java
/**
 * Whether to enable metrics to be collected and registered in Micrometer's
 * {@link io.micrometer.core.instrument.Metrics#globalRegistry globalRegistry}
 * under the name {@code "reactor.netty.http.client"}. Applications can
 * separately register their own
 * {@link io.micrometer.core.instrument.Meter meters} associated with this name.
 * For example, to put an upper bound on the number of tags produced:
 * <pre class="code">
 * String name = "reactor.netty.http.client"
 * MeterFilter filter = ... ;
 * Metrics.globalRegistry.config().meterFilter(
 * 		MeterFilter.maximumAllowableTags(name, 100, filter));
 * </pre>
 * @param metricsEnabled true enables metrics collection; false disables it
 */
```

src/main/java/reactor/netty/Metrics.java
```diff
@@ -20,6 +20,18 @@
...
+	public static final String HTTP_CLIENT_NAME_PREFIX = "reactor.netty.http.client";
+
+	public static final String TCP_SERVER_NAME_PREFIX = "reactor.netty.tcp.server";
+
+	public static final String TCP_CLIENT_NAME_PREFIX = "reactor.netty.tcp.client";
+
+	public static final String UDP_SERVER_NAME_PREFIX = "reactor.netty.udp.server";
+
+	public static final String UDP_CLIENT_NAME_PREFIX = "reactor.netty.udp.client";
+
```
rstoyanchev: These look like could be shortened a bit, minus the `_NAME_` part. `Metrics.HTTP_SERVER_PREFIX` still reads pretty well to me with "prefix" and the type `String` + some Javadoc perhaps could be a good balance between readability and brevity.

violetagg: done 69ac2d5607282f3e95320f40ce6f12362790792a

PR #994, opened 2020-02-12
Adds failing test for POST request parsing

adriancole:
See https://github.com/spring-cloud/spring-cloud-sleuth/pull/1554

### DIFF
--- a/src/test/java/reactor/netty/http/client/HttpClientTest.java
+++ b/src/test/java/reactor/netty/http/client/HttpClientTest.java
@@ -171,6 +171,32 @@ public class HttpClientTest {
 		return r;
 	}
 
+	/** This ensures that non-default values for the HTTP request line are visible for parsing. */
+	@Test
+	public void postVisibleToOnRequest() {
+		disposableServer = HttpServer.create().port(0)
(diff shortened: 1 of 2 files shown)


violetagg:
@adriancole thanks
see PR #998

src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -171,6 +171,31 @@ private DefaultFullHttpResponse response() {
...
+					path.set(req.path());
+				})
+				.post()
+				.uri("/foo")
+				.send(ByteBufFlux.fromString(Mono.just("foo")))
+				.response().block();
+
+		assertThat(response.status()).isEqualTo(HttpResponseStatus.OK);
+		assertThat(method.get()).isEqualTo(HttpMethod.POST);
+		assertThat(path.get()).isEqualTo("/foo");
```
violetagg: Remove `/` in front of `foo` because `req.path()` returns `the decoded path portion from the uri() without the leading and trailing '/' if present`

adriancole: gotcha. done!

PR #996, opened 2020-02-12
Backfills passing test about redirect behavior

adriancole:
By default auto-redirect is not enable and it should NOT redirect.
https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/http/client/HttpClientOperations.java#L536
https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/http/client/HttpClientOperations.java#L597

this backfills the test around this


violetagg:
thanks

PR #998, opened 2020-02-13
fix #997 Ensure the actual request data is available when HttpClient#doOnRequest

### LINKED ISSUE #997: HttpClient#doOnRequest does not provide the actual request data

violetagg:
The event will be trigger on `HttpClientState#REQUEST_PREPARED` and not on
`ConnectionObserver.State#CONFIGURED`

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -509,7 +509,7 @@ public abstract class HttpClient {
 
 	/**
 	 * Setup a callback called when {@link HttpClientRequest} is about to be sent
-	 * and {@link HttpClientState#CONFIGURED} has been emitted.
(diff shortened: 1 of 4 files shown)


adriancole:
unit test passes now. will do a quick integration test with sleuth and ping back shortly

adriancole:
it is unusual for the path to strip the leading slash, as this will cause garbage that's likely unnecessary here. The only library I'm aware of that does this also is jersey. Anyway, it is a separate issue and probably too late to change in reactor.

This should be merged, because it brings the failed test count in sleuth down from 12 to 9. Thanks!

adriancole:
moved the leading and trailing slash issue to a new issue https://github.com/reactor/reactor-netty/issues/999

adriancole:
is this something we can merge? I think regardless of other work, it makes sense right?

violetagg:
Fixed with 40552bc52b97aea79acdfa31b30cc91ece494964

PR #1000, opened 2020-02-13
Adds test showing implicit redirects are invisible to metrics/tracing

adriancole:
To complete https://github.com/spring-cloud/spring-cloud-sleuth/pull/1554, we need to allow visibility into implicit handling such as redirects or auto-retry on failure. This could be most easily accomplished on our side, if it uses the onXXX hooks. Alternatives could be evaluated as well.

### DIFF
--- a/src/test/java/reactor/netty/http/client/HttpRedirectTest.java
+++ b/src/test/java/reactor/netty/http/client/HttpRedirectTest.java
@@ -122,6 +122,41 @@ public class HttpRedirectTest {
 		server.disposeNow();
 	}
 
+	/** This ensures functionality such as metrics and tracing can accurately count requests. */
+	@Test
+	public void redirect_issuesOnRequestForEachAttempt() {
+		DisposableServer server =
+				HttpServer.create()
+						.port(0)
+						.host("localhost")
+						.wiretap(true)
+						.route(r -> r.get("/1", (req, res) -> res.sendRedirect("/3"))
+								.get("/3", (req, res) -> res.status(200)
+										.sendString(Mono.just("OK"))))
+						.wiretap(true)
+						.bindNow();
+
+		AtomicInteger onRequestCount = new AtomicInteger();
+		AtomicInteger onResponseCount = new AtomicInteger();
+
+		HttpClientResponse response = HttpClient.create()
+				.addressSupplier(server::address)
+				.wiretap(true)
+				.followRedirect(true)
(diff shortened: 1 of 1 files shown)


src/test/java/reactor/netty/http/client/HttpRedirectTest.java
```diff
@@ -122,6 +122,41 @@ public void redirectDisabledByDefault() {
...
+				.followRedirect(true)
+				.doOnRequest((r, c) -> onRequestCount.incrementAndGet())
+				.doOnResponse((r, c) -> onResponseCount.incrementAndGet())
+				.get()
+				.uri("/1")
+				.response()
+				.block(Duration.ofSeconds(30));
+
+		assertThat(response.status()).isEqualTo(HttpResponseStatus.OK);
+		assertThat(onRequestCount.get()).isEqualTo(2);
```
adriancole: this fails with 1, showing we can't see redirect handling with `doOnRequest`

violetagg:
@adriancole With the #998 I do see doOnRequest invoked twice, however doOnResponse is invoked once which is kind of by design
https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/http/client/HttpClientOperations.java#L536-L545

adriancole:
do you know a way to get an accurate duration if there isn't parity between onRequest and onResponse? Ex it would seem we would have to make a guess here..

adriancole:
the other problem is we won't be able to guess the correct redirect code even if we assume re-entrance is due to redirect handling.. seems hacky

adriancole:
also seems re-entrance of `doOnRequest()` is also happening on implicit retry on connect fail

ex when this is present:
```
2020-02-13 13:07:49.682  WARN   --- [ctor-http-nio-7] r.netty.http.client.HttpClientConnect    : [id: 0x62c406bc, L:/127.0.0.1:57634 - R:/127.0.0.1:57632] The connection observed an error

io.netty.handler.timeout.ReadTimeoutException: null
```

adriancole:
rebased and update the code with the impact I noted.. hoping there's a way out of this. cc also @shakuzen as this impacts metrics, not just tracing.

violetagg:
@adriancole PTAL #1020

violetagg:
@adriancole Thanks for the test

adriancole:
Thanks for keeping on after 1000 issues :)

PR #1007, opened 2020-02-19
fix #973 Add connection pool configuration per remote host

### LINKED ISSUE #973: Add connection pool configuration per remote host
Currently all ConnectionProvider's configurations are applied to all connection pools (a connection pool is related to a concrete remote host).
We want to have a configuration per remote host.

violetagg:


src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -98,7 +102,7 @@ static ConnectionProvider create(String name) {
 		return builder(name).maxConnections(DEFAULT_POOL_MAX_CONNECTIONS)
 		                    .pendingAcquireMaxCount(500)
 		                    .pendingAcquireTimeout(Duration.ofMillis(DEFAULT_POOL_ACQUIRE_TIMEOUT))
-		                    .fifo();
+		                    .build();
```
simonbasle: the builder has not yet been released, right?

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -64,45 +65,26 @@
...
-	final String      name;
-	final int         maxConnections;
-	final int         pendingAcquireMaxCount;
-	final long        pendingAcquireTimeout;
-	final long        maxIdleTime;
-	final long        maxLifeTime;
-	final boolean     metricsEnabled;
-	final PoolFactory poolFactory;
+	final String                          name;
+	final Map<SocketAddress, PoolFactory> poolFactoryPerRemoteHost = new HashMap<>();
```
simonbasle: I'm wondering about the implication of keeping `SocketAddress` references in this map's keyset, but no limited resource should leak from that, I think?

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -349,23 +355,15 @@ default int maxConnections() {
 	 * {@link Connection}. Further connections will be pending acquisition depending on
 	 * pendingAcquireTime. The maximum number of connections is for the connections in a single
 	 * connection pool, where a connection pool corresponds to a concrete remote host.
+	 * The configuration can be either default for all connection pools, or per a connection pool.
```
simonbasle: I'd say "The configuration can be either `global` for all connection pools or `be tuned for each individual pools, per host`."

the per pool/host aspect can be combined at will, right? (ie if I configure for 1 host in particular, all other hosts would get treated by the global config?)

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -385,6 +383,52 @@ public final Builder name(String name) {
...
+		 * Builds new ConnectionProvider
+		 *
+		 * @return builds new ConnectionProvider
+		 */
+		public ConnectionProvider build() {
+			return new PooledConnectionProvider(this);
+		}
+	}
+
+	class ConnectionPoolSpec<SPEC extends ConnectionPoolSpec<SPEC>> implements Supplier<SPEC> {
```
simonbasle: At first I was not sure the generic self-type helps with code readability here since the only two values of `SPEC` are `Builder` and `?` in the current code. A simple `ConnectionPoolSpec` class that is extended by `Builder` to limit the `name()` `build()` and `forRemoteHost()` methods to the global-level builder could have sufficed, but at the price of duplicating the methods in `ConnectionPoolSpec` to change the return type... So I guess it helps ;)

Maybe add a `forRemoteHost`-specific implementation of `ConnectionPoolSpec`, like `HostSpecificSpec extends ConnectionPoolSpec<HostSpecificSpec>`, and replace `ConnectionPoolSpec<?>` with `HostSpecificSpec` in signatures, for clarity? (and could especially be useful if we later want to add forRemoteHost-specific builder methods)

src/main/java/reactor/netty/resources/ConnectionProvider.java
simonbasle: I see the `Builder` was introduced in #955, and that issue is in `0.9.5.RELEASE` milestone, but the commit is from Jan 16, and we released `0.9.4.RELEASE` on Jan 20 🤔

violetagg: yes 0.9.4 was release only with a fix that covers an incompatible change in 0.9.3
https://github.com/reactor/reactor-netty/compare/v0.9.3.RELEASE...v0.9.4.RELEASE

src/main/java/reactor/netty/resources/ConnectionProvider.java
violetagg: > the per pool/host aspect can be combined at will, right? (ie if I configure for 1 host in particular, all other hosts would get treated by the global config?)

yes

src/main/java/reactor/netty/resources/ConnectionProvider.java
violetagg: > Maybe add a forRemoteHost-specific implementation of ConnectionPoolSpec, like HostSpecificSpec extends ConnectionPoolSpec<HostSpecificSpec>, and replace ConnectionPoolSpec<?> with HostSpecificSpec in signatures, for clarity? (and could especially be useful if we later want to add forRemoteHost-specific builder methods)

Agree

Issue #938, opened 2019-12-19
Netty connection in pool reset by peer in Spring Cloud Gateway Application


hanscrg:
We use Spring Cloud Gateway with Spring Security OAuth2. 

SpringBoot version: 2.2.1
Spring Cloud version: Hoxton.Release
Netty: reactor-netty-0.9.2.RELEASE

The Spring security OAuth2 filter post request to UAA server, there is load balancer in between. While initially connection all good but after about 10-15 min, Netty used Spring OAuth2 filter got connection reset by peer and the other end is load balancer. If try again, then it will work. But after another 10-15 min, the problem happens again.

The log shows reactor.netty.resources.PooledConnectionProvider uses some active channel but it is actually not active.  We are seeking solution to work around it. So far we tried 
to customize Netty ClientHttpConnector (code as below) to disable pool, but no success. It seems that bean has no effect on Netty used by Spring security. 

I uploaded demo project at GitHub link below,
https://github.com/spring-cloud/spring-cloud-gateway/issues/1493

That project cannot produce problem at local as there is no load balancer between but it shows the way to customize ClientHttpConnector used by Spring Security to disable Netty Pool does not work.

```
@bean
public ClientHttpConnector clientHttpConnector() {
return new ReactorClientHttpConnector(HttpClient.create(ConnectionProvider.newConnection()));
}
```

From Netty side, is there any solution or work around to this issue?  Since our project is due soon in production, any work around we are considering. If you could point out where in Netty code, we can disable the pool as a temporary solutions so that we can move forward and revisit this issue later on.

Here is the logs:

```
2019-12-19 18:07:53.460 DEBUG 7 --- [reactor-http-epoll-1] o.s.w.r.f.client.ExchangeFunctions       : [1b45ced7] HTTP POST https://server-name/oauth/token
2019-12-19 18:07:53.462 DEBUG 7 --- [reactor-http-epoll-4] r.n.resources.PooledConnectionProvider   : [id: 0x2a4d88b0, L:/xxx.xxx.xxx.xxx:38628 - R:xxx.xxx.xxx.xxx:443] Channel acquired, now 1 active connections and 1 inactive connections
[...]

PR #1011, opened 2020-02-21
Add System properties for connection pool - max idle time and leasing strategy

violetagg:
Related to #938

### DIFF
--- a/src/main/java/reactor/netty/resources/ConnectionProvider.java
+++ b/src/main/java/reactor/netty/resources/ConnectionProvider.java
@@ -31,6 +31,7 @@ import javax.annotation.Nullable;
 import java.net.SocketAddress;
 import java.time.Duration;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeoutException;
@@ -63,6 +64,33 @@ public interface ConnectionProvider extends Disposable {
 	long DEFAULT_POOL_ACQUIRE_TIMEOUT = Long.parseLong(System.getProperty(
 			ReactorNetty.POOL_ACQUIRE_TIMEOUT,
 			"" + 45000));
+
+	/**
+	 * Default max idle time, fallback - max idle time is not specified.
+	 */
+	long DEFAULT_POOL_MAX_IDLE_TIME = Long.parseLong(System.getProperty(
+			ReactorNetty.POOL_MAX_IDLE_TIME,
+			"-1"));
+
+	/**
+	 * The connection selection is first in, first out.
+	 */
+	String LEASING_STRATEGY_FIFO = "fifo";
+
+	/**
+	 * The connection selection is last in, first out.
+	 */
+	String LEASING_STRATEGY_LIFO = "lifo";
+
+	/**
+	* Default leasing strategy (fifo, lifo), fallback to fifo.
+	* <ul>
+	*     <li>fifo - The connection selection is first in, first out</li>
+	*     <li>lifo - The connection selection is last in, first out</li>
+	* </ul>
+	*/
+	String DEFAULT_POOL_LEASING_STRATEGY = System.getProperty(ReactorNetty.POOL_LEASING_STRATEGY, LEASING_STRATEGY_FIFO)
+					.toLowerCase(Locale.ENGLISH);
 
 	/**
 	 * Creates a builder for {@link ConnectionProvider}
@@ -369,6 +397,7 @@ public interface ConnectionProvider extends Disposable {
 		 * @param name {@link ConnectionProvider} name
 		 */
 		private Builder(String name) {
+			super();
 			name(name);
 		}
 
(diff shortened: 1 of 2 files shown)


src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -422,16 +437,25 @@ public ConnectionProvider build() {
...
+				InstrumentedPool<PooledConnectionProvider.PooledConnection>> leasingStrategy;
 
 		/**
 		 * Returns {@link ConnectionPoolSpec} new instance with default properties.
 		 */
 		private ConnectionPoolSpec() {
+			if (DEFAULT_POOL_MAX_IDLE_TIME != null) {
+				maxIdleTime(Duration.ofMillis(Long.parseLong(DEFAULT_POOL_MAX_IDLE_TIME)));
+			}
+			if("lifo".equals(DEFAULT_POOL_LEASING_STRATEGY)) {
```
simonbasle: maybe extract that as a public constant in `ReactorNetty`? at least as a private constant next to `DEFAULT_POOL_LEASING_STRATEGY` above.

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -64,6 +64,20 @@
...
+	String DEFAULT_POOL_MAX_IDLE_TIME = System.getProperty(ReactorNetty.POOL_MAX_IDLE_TIME);
+
+	/**
+	 * Default leasing strategy (fifo, lifo), fallback to fifo.
+	 * <ul>
+	 *     <li>fifo - The connection selection is first in, first out</li>
+	 *     <li>lifo - The connection selection is last in, first out</li>
+	 * </ul>
+	 */
+	String DEFAULT_POOL_LEASING_STRATEGY = System.getProperty(ReactorNetty.POOL_LEASING_STRATEGY, "fifo");
```
simonbasle: use the extracted constant for `"fifo"`

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -64,6 +64,20 @@
 			ReactorNetty.POOL_ACQUIRE_TIMEOUT,
 			"" + 45000));
 
+	/**
+	 * Default max idle time, fallback - max idle time is not specified.
+	 */
+	String DEFAULT_POOL_MAX_IDLE_TIME = System.getProperty(ReactorNetty.POOL_MAX_IDLE_TIME);
```
simonbasle: I would maybe do the conversion to long here, with a default value of `"-1"` for the system property for instance, so that a misconfiguration is detected way earlier than the first time a resource is acquired from a pool.

simonbasle:
a quick side note on `maxIdleTime`: this is NOT really strictly enforced by the pool by a reaper job. Instead, the idle time of a resource is evaluated during `acquire()` and the resource is destroyed if it has been too idle (which at that point could be way over the maxIdleTime), and `acquire()` will look at the next resource.

src/main/java/reactor/netty/resources/ConnectionProvider.java
violetagg: the conversion is done when constructing the Builder

> a quick side note on `maxIdleTime`: this is NOT really strictly enforced by the pool (eg. by a regular reaper job). Instead, the idle time of a resource is evaluated during `acquire()` and the resource is destroyed if it has been too idle (which at that point could be way over the maxIdleTime), and `acquire()` will look at the next resource.

that's OK

violetagg:
> a quick side note on `maxIdleTime`: this is NOT really strictly enforced by the pool (eg. by a regular reaper job). Instead, the idle time of a resource is evaluated during `acquire()` and the resource is destroyed if it has been too idle (which at that point could be way over the maxIdleTime), and `acquire()` will look at the next resource.

that's OK

src/main/java/reactor/netty/resources/ConnectionProvider.java

src/main/java/reactor/netty/resources/ConnectionProvider.java

src/main/java/reactor/netty/resources/ConnectionProvider.java
violetagg: The conversion happens when constructing the Builder https://github.com/reactor/reactor-netty/pull/1011/files#diff-838a952a6538b5173f40436cdfacae56R459

simonbasle: yes, that's what I saw and what prompted me to make this comment. better to fail fast at JVM start than at a seemingly random point of app execution because the configuration error is detected late, the first time the `Builder` is constructed

simonbasle: so I propose to detect the bad value for system property earlier than later, by doing the `Long.parseLong` here

simonbasle: (if the property is not set, have the `String` be `"-1"` so that the `parseLong` doesn't fail but return `-1L`. down in `Builder` constructor, check `DEFAULT_POOL_MAX_IDLE_TIME != -1` to activate max idle time)

violetagg: I would say `DEFAULT_POOL_MAX_IDLE_TIME > -1`

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -64,6 +65,33 @@
...
+
+	/**
+	* Default leasing strategy (fifo, lifo), fallback to fifo.
+	* <ul>
+	*     <li>fifo - The connection selection is first in, first out</li>
+	*     <li>lifo - The connection selection is last in, first out</li>
+	* </ul>
+	*/
+	String DEFAULT_POOL_LEASING_STRATEGY = System.getProperty(ReactorNetty.POOL_LEASING_STRATEGY, LEASING_STRATEGY_FIFO)
+					.toLowerCase(Locale.ENGLISH);
```
simonbasle: good thinking on the lowercasing 😄

PR #1014, opened 2020-02-24
Fix compilation errors when using Java 13+

PyvesB:
Java 13 introduced a new overload for `FileSystems.newFileSystem`. See the [documentation](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/nio/file/FileSystems.html#newFileSystem(java.nio.file.Path,java.util.Map)) for more details.

Several unit tests were longer compiling as `FileSystems.newFileSystem(path, null)` has now become ambiguous.

This pull request fixes the problem by adding an explicit cast.


violetagg:
Committed to 0.8.x (704a82323bda66f44287f8861ba83b12b13f462a) and then forward merge to 0.9.x and master.

Thanks for the PR

PR #1018, opened 2020-03-05
Abort send* operation if the connection is closed

violetagg:


violetagg:
the failures are not related to this change

violetagg:
> This completes channel active checks in all `send` methods, which makes them consistent, and probably makes the problem in [spring-projects/spring-framework#24596](https://github.com/spring-projects/spring-framework/issues/24596) go away, in the way it's simulated at least, via breakponit before the write.
> 
> However I don't see what will give the same `AbortedException` if the connection is lost at any time after that active channel check and during the send, e.g. subscribing to the write source and the actual writing.
> 
> In the case of [spring-projects/spring-framework#24596](https://github.com/spring-projects/spring-framework/issues/24596), `ClosedChannelException` is wrapped with `InternalNettyException` which happens in places like `FluxReceive` and `FutureMono`. Aren't those also cases of `AbortedException`, i.e. unexpectedly closed inbound?

I would expect some other exception here and not `ClosedChannelException`. Something related to write/flush operation.

rstoyanchev:
I guess I'm wondering if there is a way to recognize cases when the connection is closed, after the start of `send`?  Regarding `ClosedChannelException` what about [this here](https://github.com/reactor/reactor-netty/blob/9a5e1be2f1255834b5bca3fb5cd455aba58b83e3/src/main/java/reactor/netty/FutureMono.java#L184), couldn't that be outcome of a pending write?

We could also provide a snapshot with this PR and see if the reporter under https://github.com/spring-projects/spring-framework/issues/24596 would still see the error but if it's production only, it might have to be released first.

violetagg:
Let's commit this and then we will see the other cases.

PR #1019, opened 2020-03-05
fix #995 Add API for disabling retry support for HTTP client

### LINKED ISSUE #995: opt-out of connection retry
As a part of http tracing interop, I noticed reactor-netty's HttpClient uniquely has a retry-once on connection-reset behavior, which is not possible to disable. I'd like to be able to disable this.

## Motivation
I'd like to improve the quality of trace instrumentation without disabling or special-casing the part that tests traces produced in connection error https://github.com/spring-cloud/spring-cloud-sleuth/pull/1554

## Desired solution
I'd like a way to disable the retry on connection fail.

## Considered alternatives
We could rewrite the behavior of Brave's test framework to work around this, but it seems odd especially as several new and old clients are able to control retry behavior.

## Additional context
According to @violetagg, HttpClient will retry once ONLY in cases of Connection reset by peer

https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L666

violetagg:
Fixes #995

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -737,6 +737,25 @@ public abstract class HttpClient {
 	}
 
 	/**
+	 * Option to disable {@code retry once} support for the outgoing requests that fail with
+	 * {@link reactor.netty.channel.AbortedException#isConnectionReset(Throwable)}.
+	 * <p>By default this is set to false in which case {@code retry once} is enabled.
+	 *
+	 * @param disableRetry true to disable {@code retry once}, false to enable it
+	 *
+	 * @return a new {@link HttpClient}
+	 * @since 0.9.6
+	 */
+	public final HttpClient disableRetry(boolean disableRetry) {
+		if (disableRetry) {
+			return tcpConfiguration(RETRY_ATTR_DISABLE);
+		}
+		else {
+			return tcpConfiguration(RETRY_ATTR_CONFIG);
+		}
+	}
+
+	/**
 	 * Specifies whether HTTP status 301|302|307|308 auto-redirect support is enabled.
 	 *
 	 * <p><strong>Note:</strong> The sensitive headers {@link #followRedirect(boolean, Consumer) followRedirect}
@@ -1149,6 +1168,12 @@ public abstract class HttpClient {
 
 	static final Function<TcpClient, TcpClient> KEEPALIVE_ATTR_DISABLE =
 			tcp -> tcp.bootstrap(HttpClientConfiguration.MAP_NO_KEEPALIVE);
+
+	static final Function<TcpClient, TcpClient> RETRY_ATTR_CONFIG =
+			tcp -> tcp.bootstrap(HttpClientConfiguration.MAP_RETRY);
+
+	static final Function<TcpClient, TcpClient> RETRY_ATTR_DISABLE =
+			tcp -> tcp.bootstrap(HttpClientConfiguration.MAP_NO_RETRY);
 
 	static final Function<TcpClient, TcpClient> FOLLOW_REDIRECT_ATTR_DISABLE =
 			tcp -> tcp.bootstrap(HttpClientConfiguration.MAP_NO_REDIRECT);
--- a/src/main/java/reactor/netty/http/client/HttpClientConfiguration.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientConfiguration.java
@@ -55,6 +55,7 @@ final class HttpClientConfiguration {
 	String                        websocketSubprotocols          = null;
 	int                           websocketMaxFramePayloadLength = 65536;
 	boolean                       websocketProxyPing             = false;
+	boolean                       retryDisabled                  = false;
 	int                           protocols                      = h11;
 	HttpResponseDecoderSpec       decoder                        = new HttpResponseDecoderSpec();
 
@@ -86,6 +87,7 @@ final class HttpClientConfiguration {
 		this.websocketSubprotocols = from.websocketSubprotocols;
 		this.websocketMaxFramePayloadLength = from.websocketMaxFramePayloadLength;
 		this.websocketProxyPing = from.websocketProxyPing;
+		this.retryDisabled = from.retryDisabled;
 		this.body = from.body;
 		this.protocols = from.protocols;
 		this.deferredConf = from.deferredConf;
@@ -148,6 +150,17 @@ final class HttpClientConfiguration {
 			c.headers = new DefaultHttpHeaders();
 		}
 		HttpUtil.setKeepAlive(c.headers, HttpVersion.HTTP_1_1, false);
+		return b;
+	};
+
+	static final Function<Bootstrap, Bootstrap> MAP_RETRY = b -> {
+		getOrCreate(b).retryDisabled = false;
+		return b;
+	};
+
+
+	static final Function<Bootstrap, Bootstrap> MAP_NO_RETRY = b -> {
+		getOrCreate(b).retryDisabled = true;
 		return b;
 	};
 
--- a/src/main/java/reactor/netty/http/client/HttpClientConnect.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientConnect.java
@@ -380,10 +380,14 @@ final class HttpClientConnect extends HttpClient {
 					log.debug(format(connection.channel(), "The request will be redirected"));
 				}
 			}
-			else if (AbortedException.isConnectionReset(error)) {
+			else if (AbortedException.isConnectionReset(error) && handler.shouldRetry) {
+				HttpClientOperations ops = connection.as(HttpClientOperations.class);
+				if (ops != null) {
+					ops.retrying = true;
+				}
 				if (log.isDebugEnabled()) {
-					log.debug(format(connection.channel(), "The connection observed an error, " +
-							"the request will be retried"), error);
+					log.debug(format(connection.channel(),
+							"The connection observed an error, the request will be retried"), error);
 				}
 			}
 			else if (log.isWarnEnabled()) {
@@ -457,7 +461,7 @@ final class HttpClientConnect extends HttpClient {
 		volatile UriEndpoint        toURI;
 		volatile UriEndpoint        fromURI;
 		volatile Supplier<String>[] redirectedFrom;
-		volatile boolean            retried;
+		volatile boolean            shouldRetry;
 
 		@SuppressWarnings("unchecked")
 		HttpClientHandler(HttpClientConfiguration configuration, @Nullable SocketAddress address,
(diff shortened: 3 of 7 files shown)


violetagg:
the failing test is not related

src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -716,6 +716,24 @@ public final HttpClient keepAlive(boolean keepAlive) {
...
+	/**
+	 * Enable or disable {@code retry once} support for the outgoing request when
+	 * {@link reactor.netty.channel.AbortedException#isConnectionReset(Throwable)}.
+	 * By default {@code retry once} is enabled.
+	 *
+	 * @param retryEnabled true if retry should be enabled, false otherwise
+	 *
+	 * @return a new {@link HttpClient}
+	 */
+	public final HttpClient retry(boolean retryEnabled) {
```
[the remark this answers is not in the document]
violetagg: changed the method name to `retryEnabled`

violetagg:
Added new callback `doOnRetryEnabled`

src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -657,7 +657,22 @@ public final HttpClient doAfterResponse(BiConsumer<? super HttpClientResponse, ?
...
+	 * Setup a callback called when {@link HttpClientResponse} failed with
+	 * {@link reactor.netty.channel.AbortedException#isConnectionReset(Throwable)}
+	 * and {@link HttpClientState#RETRY_ENABLED} has been emitted.
+	 *
+	 * @param doOnRetryEnabled a callback called when {@link HttpClientResponse} failed with
+	 * {@link reactor.netty.channel.AbortedException#isConnectionReset(Throwable)}
+	 *
+	 * @return a new {@link HttpClient}
+	 */
+	public final HttpClient doOnRetryEnabled(BiConsumer<? super HttpClientResponse, ? super Connection> doOnRetryEnabled) {
```
adriancole: I suppose this is intended to allow tagging of the error that resulted in repeating the other events again? Or when this is called, it is not going to call something else maybe you can look at this state diagram and comment expectations here https://github.com/square/okhttp/blob/master/docs/events.md#events-with-retries-and-follow-ups

src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -657,7 +657,22 @@ public final HttpClient doAfterResponse(BiConsumer<? super HttpClientResponse, ?
...
 	public final HttpClient doAfterResponseSuccess(BiConsumer<? super HttpClientResponse, ? super Connection> doAfterResponseSuccess) {
 		Objects.requireNonNull(doAfterResponseSuccess, "doAfterResponseSuccess");
-		return new HttpClientDoOn(this, null, null, null, null, doAfterResponseSuccess);
+		return new HttpClientDoOn(this, null, null, null, null, doAfterResponseSuccess, null);
+	}
+
+	/**
+	 * Setup a callback called when {@link HttpClientResponse} failed with
+	 * {@link reactor.netty.channel.AbortedException#isConnectionReset(Throwable)}
+	 * and {@link HttpClientState#RETRY_ENABLED} has been emitted.
```
rstoyanchev:
```suggestion
	 * Setup a callback called when {@link HttpClientResponse} failed with
	 * {@link reactor.netty.channel.AbortedException#isConnectionReset(Throwable)}
	 * and a retry to connect is about to be performed.
```

src/main/java/reactor/netty/http/client/HttpClientState.java
```diff
@@ -42,6 +42,15 @@ public String toString() {
...
 	},
+	/**
+	 * The request is about to be retried
+	 */
+	RETRY_ENABLED() {
+		@Override
+		public String toString() {
+			return "[retry_enabled]";
+		}
+	},
```
rstoyanchev: Could this be called `RETRYING` perhaps? Again `RETRY_ENABLED` sounds like a config time callback.

For source code ordering, it's a bit strange to see this state in the middle of the request related ones. Perhaps it can be ordered at the top of this class since it belongs more with the base class (connection related).

src/main/java/reactor/netty/http/client/HttpClient.java
adriancole: hard to elaborate my question without a test.

I suppose I mean.. should we expect in retry this?

doOnRequest
doOnResponseError
doOnRetry
doOnRequest
doOnResponse

or

doOnRequest
doOnRetry
doOnRequest
doOnResponse

or something else... if the former I agree we can kill the doOnRetry as long as we can see the doOnResponseError

violetagg: My idea is to have this flow

```
doOnRequest
doOnRetry
doOnRequest
doOnResponse
```

violetagg:
> This description for #995 only asks for a way to disable retry. Just double checking whether the doOnRetry callback is actually needed.

The discussion for `doOnRetry` was started here
https://github.com/reactor/reactor-netty/pull/1020#issuecomment-595633152

src/main/java/reactor/netty/http/client/HttpClient.java
violetagg: Retry is not performed for any exception so using this callback retry can be distinguish from error

src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -716,6 +731,24 @@ public final HttpClient keepAlive(boolean keepAlive) {
...
+	/**
+	 * Enable or disable {@code retry once} support for the outgoing request when
+	 * {@link reactor.netty.channel.AbortedException#isConnectionReset(Throwable)}.
+	 * By default {@code retry once} is enabled.
+	 *
+	 * @param retryEnabled true if retry should be enabled, false otherwise
+	 *
+	 * @return a new {@link HttpClient}
+	 */
+	public final HttpClient retryEnabled(boolean retryEnabled) {
```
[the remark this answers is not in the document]
violetagg: Changed the method name to `disableRetry`

src/main/java/reactor/netty/http/client/HttpClient.java
violetagg: done, method name changed to `doOnRetry`

src/main/java/reactor/netty/http/client/HttpClientState.java
violetagg: done, changed to `RETRYING`

src/main/java/reactor/netty/http/client/HttpClient.java
adriancole: technically it is an exception :P just a matter if the callback sees it or not. Is the response object useful here? Would it not be better to provide the underlying throwable and/or replace with doOnError? Maybe I'm reading too much into the javadoc, but it seems like this is literally in response to a Throwable.

src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -683,6 +640,43 @@ public boolean test(Throwable throwable) {
...
+		}
+
+		@Override
+		public void onUncaughtException(Connection connection, Throwable error) {
+			if (error instanceof RedirectClientException) {
+				if (log.isDebugEnabled()) {
+					log.debug(format(connection.channel(), "The request will be redirected"));
+				}
+			}
+			else if (AbortedException.isConnectionReset(error) && !retried) {
```
adriancole: basically it looks like a special case of error. I might say this is like doOnError, except we know it will be retried, then pass the `error` to it.

Code-wise, brave/zipkin will do nothing different that the hook is named doOnRetry vs doOnError as the behavior for any follow-up request would be the same, regardless of if reason is retry or redirect.

violetagg:
@adriancole @rstoyanchev PTAL
I removed `doOnRetry` there will be only `doOnError` event

src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -512,6 +468,7 @@ public void onStateChange(Connection connection, State newState) {
 			this.websocketProtocols = configuration.websocketSubprotocols;
 			this.maxFramePayloadLength = configuration.websocketMaxFramePayloadLength;
 			this.websocketProxyPing = configuration.websocketProxyPing;
+			this.retried = !configuration.retryEnabled;
```
rstoyanchev: I think this would be easier to read as `shouldRetry`.

violetagg: done

src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -1329,18 +1338,43 @@ public void testRetryNotEndlessIssue587() throws Exception {
...
+				HttpClient.create()
+				          .port(serverPort)
+				          .wiretap(true)
+				          .doOnRequest((req, conn) -> doOnRequest.getAndIncrement())
+				          .doOnError((req, t) -> doOnRequestError.getAndIncrement(),
+				                     (res, t) -> doOnResponseError.getAndIncrement());
+
+		if (retryDisabled) {
+			client = client.disableRetry(true);
+		}
```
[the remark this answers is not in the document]
violetagg: done

src/main/java/reactor/netty/http/client/HttpClientConfiguration.java
```diff
@@ -55,6 +55,7 @@
 	String                        websocketSubprotocols          = null;
 	int                           websocketMaxFramePayloadLength = 65536;
 	boolean                       websocketProxyPing             = false;
+	boolean                       retryEnabled                   = true;
```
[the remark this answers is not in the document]
violetagg: done

src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -736,6 +736,25 @@ public final HttpClient keepAlive(boolean keepAlive) {
 		}
 	}
 
+	/**
+	 * Enable or disable {@code retry once} support for the outgoing request when
+	 * {@link reactor.netty.channel.AbortedException#isConnectionReset(Throwable)}.
+	 * By default {@code retry once} is enabled.
+	 *
+	 * @param disableRetry true if {@code retry once} should be disabled, false otherwise
```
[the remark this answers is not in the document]

violetagg:
> This looks better without the extra state.
> 
> One minor comment, `HttpObserver` in `HttpClientConnect` could have remained a separate decorator class, since it has access to `handler.retried`.

agree

adriancole:
somehow this seems to trip up redirect.. Ex when I set disableRetry my redirect tests fail that otherwise work when followRedirect(true)

```java
				.disableRetry(true)
				.followRedirect(true)
```

PR #1020, opened 2020-03-06
Add HttpClient#doOnRedirect callback that will be invoked when the request is about to be redirected

violetagg:
Related to #1000
Tested with #1000


adriancole:
Cool. Does this allow visibility of automated retry? (ex when enabled)

violetagg:
> Cool. Does this allow visibility of automated retry? (ex when enabled)

No for retry functionality I want to think more about whether we want to expose it with `doOnResponseError` or something else.

adriancole:
ok lets go with this, as it is progress. thanks!

PR #1025, opened 2020-03-10
fix #999 Add HttpInfos#fullPath for not removing leading/trailing '/' symbol

### LINKED ISSUE #999: Allow retrieval of the http path including the "/" prefix
It would be great to include the full HTTP path, not the path except the leading slashes, as that's the data used in trace tags, parsing and sampling. spring-cloud/spring-cloud-sleuth#1554

## Motivation
We are trying to reduce the overhead of reactor instrumentation in sleuth. While most of this is propagation related, the most commonly requested tag, "http.path" is the entire path, not the path except the leading and trailing slash. When reactor strips these, it causes us to add them back, allocating new arrays to fix it, increasing overhead vs other HTTP libraries.

## Desired solution
Either the .path() returns the entire path, no stripping, or something else is provided that does without excess allocations implied.

## Considered alternatives
We will have to allocate strings meanwhile

## Additional context
it is unusual for the path to strip the leading slash, as this will cause garbage that's likely unnecessary here. The only library I'm aware of that does this also is jersey, though that's the server.. I'm not sure what their client does.

violetagg:
Fixes #999

### DIFF
--- a/src/main/java/reactor/netty/http/HttpInfos.java
+++ b/src/main/java/reactor/netty/http/HttpInfos.java
@@ -65,7 +65,28 @@ public interface HttpInfos {
 	 *
 	 * @return the decoded path portion from the {@link #uri()} without the leading and trailing '/' if present
 	 */
-	String path();
+	default String path() {
+		String path = fullPath();
+		if (!path.isEmpty()) {
+			if (path.charAt(0) == '/') {
+				path = path.substring(1);
+				if (path.isEmpty()) {
+					return path;
+				}
+			}
+			if (path.charAt(path.length() - 1) == '/') {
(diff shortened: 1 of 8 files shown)


src/main/java/reactor/netty/http/HttpInfos.java
```diff
@@ -65,7 +65,28 @@
 	 *
 	 * @return the decoded path portion from the {@link #uri()} without the leading and trailing '/' if present
 	 */
-	String path();
+	default String path() {
```
adriancole: thanks! prefer special casing here, vs trying to reverse engineer after the fact if there was a trailing slash or not in fullPath!

PR #1026, opened 2020-03-11
Add release of HttpContent when redirecting

Tankika:
Currently, there is a leak error on redirection, in case the redirection response has a body and not just headers. This is because `HttpClientOperations#onInboundNext` won't pass forward the HttpContent as there is no receiver yet, but it won't release it either, because it usually expects downstream to release it.

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClientOperations.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientOperations.java
@@ -104,6 +104,7 @@ class HttpClientOperations extends HttpOperations<NettyInbound, NettyOutbound>
 	volatile ResponseState responseState;
 
 	boolean started;
+	boolean redirected;
 
 	BiPredicate<HttpClientRequest, HttpClientResponse> followRedirectPredicate;
 	Consumer<HttpClientRequest> redirectRequestConsumer;
@@ -111,6 +112,7 @@ class HttpClientOperations extends HttpOperations<NettyInbound, NettyOutbound>
 	HttpClientOperations(HttpClientOperations replaced) {
 		super(replaced);
 		this.started = replaced.started;
+		this.redirected = replaced.redirected;
 		this.redirectedFrom = replaced.redirectedFrom;
 		this.isSecure = replaced.isSecure;
 		this.nettyRequest = replaced.nettyRequest;
@@ -558,13 +560,22 @@ class HttpClientOperations extends HttpOperations<NettyInbound, NettyOutbound>
 					ReferenceCountUtil.release(msg);
 					return;
 				}
+			} else {
+				redirected = true;
 			}
+
 			if (msg instanceof FullHttpResponse) {
 				super.onInboundNext(ctx, msg);
 				terminate();
 			}
 			return;
 		}
+
+		if(redirected) {
+			ReferenceCountUtil.release(msg);
+			return;
+		}
+
 		if (msg instanceof LastHttpContent) {
 			if (!started) {
 				if (log.isDebugEnabled()) {
--- a/src/test/java/reactor/netty/http/client/HttpRedirectTest.java
+++ b/src/test/java/reactor/netty/http/client/HttpRedirectTest.java
@@ -16,9 +16,14 @@
 
 package reactor.netty.http.client;
 
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import java.security.cert.CertificateException;
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import io.netty.handler.codec.http.HttpHeaderNames;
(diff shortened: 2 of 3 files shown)


src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -558,13 +560,22 @@ protected void onInboundNext(ChannelHandlerContext ctx, Object msg) {
...
 			if (msg instanceof FullHttpResponse) {
 				super.onInboundNext(ctx, msg);
 				terminate();
 			}
 			return;
 		}
+
+		if(redirected) {
+			ReferenceCountUtil.release(msg);
+			return;
```
violetagg: Why don't you release the response when `notRedirected` is invoked, just before emitting the RedirectClientException?

src/test/java/reactor/netty/http/client/HttpRedirectTest.java
```diff
@@ -583,4 +583,34 @@ public void testLastLocationSetToResourceUrlOnRedirect() throws CertificateExcep
...
+										.header(HttpHeaderNames.LOCATION, redirectPath)
+										.sendString(Mono.just("Redirect response content!")))
+								.get(redirectPath, (req, res) -> res.send()))
+						.wiretap(true)
+						.bindNow();
+
+		HttpClient.create()
+				.wiretap(true)
+				.followRedirect(true)
+				.get()
```
violetagg: You may try to add `ChannelInboundHandler` so that you will have access to the incoming message.
After that you can test that message with `assertThat(ReferenceCountUtil.refCnt(msg)).isEqualTo(0);`

src/main/java/reactor/netty/http/client/HttpClientOperations.java
TamasPerlakiMW: Because `HttpResponse` is not implementing `ByteBufHolder`, so we don't need to release it. What caused the leak is the `HttpContent` coming in a separate call (but belonging to the `HttpResponse` with the redirect), as it implements `ByteBufHolder`, so we need to release it. But it currently is not released when there is a redirect, and the content is not empty.

src/test/java/reactor/netty/http/client/HttpRedirectTest.java
Tankika: Good idea, thank you. I've updated the test, so it fails, when there is a leak.

src/main/java/reactor/netty/http/client/HttpClientOperations.java
violetagg: just wondering whether it can be FullHttpResponse

src/test/java/reactor/netty/http/client/HttpRedirectTest.java
```diff
@@ -583,4 +590,57 @@ public void testLastLocationSetToResourceUrlOnRedirect() throws CertificateExcep
...
+		final String redirectPath = "/redirect";
+
+		DisposableServer server =
+				HttpServer.create()
+						.port(0)
+						.route(r -> r
+								.get(initialPath, (req, res) -> res
+										.status(HttpResponseStatus.MOVED_PERMANENTLY)
+										.header(HttpHeaderNames.LOCATION, redirectPath)
+										.sendString(Mono.just(redirectResponseContent)))
```
violetagg: What will happen with the connection if you send `Flux.just(redirectResponseContent, redirectResponseContent)`?

violetagg: I think we have issue with the connection staying active when we do a redirection.

src/main/java/reactor/netty/http/client/HttpClientOperations.java
Tankika: I have no idea in what case we receive a FullHttpResponse. According to the code, I guess it's possible that we receive it for a redirect with content, so probably I should update the code to take care of this case.

src/test/java/reactor/netty/http/client/HttpRedirectTest.java
violetagg: Should I create a issue for the connection staying active so that we do not hijact this PR?

Tankika: Yes, I think that would be the best. Thank you.

violetagg: #1027

src/main/java/reactor/netty/http/client/HttpClientOperations.java
Tankika: Well, I've looked through the code and I can see no way that this piece of code is reached. In the channel pipeline, `HttpResponseDecoder`, which is the previous handler produces a `FullHttpResponse` in only one case, when there is an invalid response, but in that case there is no content free up. I've also ran all the tests and not a single one covers this branch. 

So, to conclude, on second thought I would leave the PR as it is, because I see no point in handling this case.

violetagg: ok

violetagg:
@Tankika The change was committed to 0.8.x (6ca1683e846c94cdf34efd252dfa37ecf6ce9978) and then forward merge to 0.9.x and master

Thanks for the contribution.

Issue #1016, opened 2020-02-28
Backpressure for HTTP requests doesn't prevent OutOfMemoryError


evgenyvsmirnov:
Please have a look at the following test. It is comprised of the REST controller and the test which sends a request
with a generated 2GB text payload using java.net.http.HttpClient.
The REST controller (WebfluxUploadController) incorporates the following classes:
* RequestBodySubscriber – establishes a subscription to a request body, collects body fragments (DataBuffers) into a concurrent list. The list is processed by a separate thread which is started during the subscription. The thread simulates a buffer processing delay via 30Ms sleep (it might be a low disk throughput or a thread starvation),
* ResponsePublisher – responds with a 200 response whatever happens,
* ConsumptionProgressLogger – logs the body consumption total.

The test (WebfluxUploadTest) incorporates the following classes:
* RequestBodyPublisher – java.net.http.HttpRequest.BodyPublisher implementation. Alongside _0123456789nPayloadProvider produces a number of lines "012345678\n" (in accordance with the test configuration),
* _0123456789nPayloadProvider – stateful payload provider,
* SendProgressLogger – same as ConsumptionProgressLogger but for the client.

**Expected behaviour:** the server accepts data gradually without a significant memory consumption increase.
**Actual behaviour:** the server doesn't prevent a client from sending additional data when the consumption rate lowers (eventually OutOfMemoryError is thrown).

WebfluxUploadController.java
```java
package com.example.backpressure;

import java.util.concurrent.ConcurrentLinkedQueue;
import java.util.concurrent.CountDownLatch;
import java.util.concurrent.atomic.AtomicLong;
import java.util.function.LongConsumer;

import org.reactivestreams.Publisher;
import org.reactivestreams.Subscriber;
import org.reactivestreams.Subscription;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.core.io.buffer.DataBuffer;
import org.springframework.core.io.buffer.DataBufferUtils;
import org.springframework.http.HttpStatus;
import org.springframework.web.bind.annotation.PostMapping;
import org.springframework.web.bind.annotation.RestController;
import org.springframework.web.server.ServerWebExchange;

import reactor.core.publisher.Mono;

@RestController
public class WebfluxUploadController
{
    private static final Logger LOGGER = LoggerFactory.getLogger(WebfluxUploadController.class);

    @PostMapping("/upload")
    public Mono<Void> upload(ServerWebExchange exchange)
    {
[...]

PR #1030, opened 2020-03-18
When reading the incoming data enable auto-read when the queue is empty

violetagg:
Related to #1016

### DIFF
--- a/src/main/java/reactor/netty/channel/FluxReceive.java
+++ b/src/main/java/reactor/netty/channel/FluxReceive.java
@@ -44,6 +44,8 @@ import static reactor.netty.ReactorNetty.format;
  */
 final class FluxReceive extends Flux<Object> implements Subscription, Disposable {
 
+	static final int QUEUE_LOW_LIMIT = 32;
+
 	final Channel           channel;
 	final ChannelOperations<?, ?> parent;
 	final EventLoop         eventLoop;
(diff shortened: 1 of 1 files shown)


src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -254,7 +254,7 @@ final void drainReceiver() {
 				receiverFastpath = true;
 			}
 
-			if ((receiverDemand -= e) > 0L || e > 0L) {
+			if ((receiverDemand -= e) > 0L || (e > 0L && q.isEmpty())) {
```
rstoyanchev:
```suggestion
			if ((receiverDemand -= e) > 0L || (e > 0L && q.size() > 8)) {
```

Rather than let the queue go empty, it could proactively switch to reading below a certain level.

evgenyvsmirnov: By the look of it the receive window will be opened and closed on and off – we will face  a jerky transmission which will push back the overall transfer rate. I agree with @rstoyanchev it is better to have something in the wire constantly. Notwithstanding as a quick fix its ok and I would prefer to get it. 

One thought if I may: from my point of view the step change would be to control the memory consumption grand total among all active channels. In this case FluxReceive could degrade gracefully restraining the transmission if there are multiple simultaneous uploads or accept data at the highest rate. It implies the utilization of the overall quota for all concurrent channels – the implementation may be tricky.

violetagg: I created an issue about revising the FluxReceive queue low limit #1064

evgenyvsmirnov: Great! Thanks!

PR #1033, opened 2020-03-19
Adds failing test for unexpected timing in `HttpClient.doOnResponse()`

adriancole:
For metrics and timing, we should be able to get a hook at headers time.
`HttpClient.doOnResponse()` docs say this is the hook for that, except
it clarifies as *after*, not at the moment of the headers time. Small
discrepencies are ok, but right now, the implementation runs the hook
after the `Mono.block()` completes. This makes tracing look strange as
a blocking client should never finish after the next span starts.

There are a couple ways to deal with this, but perhaps only state hooks
need adjusted, as this only breaks consistently on a redirect follow-up
request. Other tests seem to pass consistently.

### DIFF
--- a/src/test/java/reactor/netty/http/client/HttpRedirectTest.java
+++ b/src/test/java/reactor/netty/http/client/HttpRedirectTest.java
@@ -21,6 +21,9 @@ import java.security.cert.CertificateException;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import io.netty.channel.ChannelHandlerContext;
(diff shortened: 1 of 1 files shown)


violetagg:
@simonbasle @rstoyanchev 

Currently we add all connection state observers in-between HTTP observer and HTTP IO Handler observer.

In Flux for example `doOnNext` is emitted before `onNext`.

What do you think should we move the connection state observers before HTTP observer?

https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L314-L317

Currently it will affect states `RESPONSE_RECEIVED` and `CONFIGURED`, and exceptions. We will not be able to provide some redirect and retry information to the connection state observers.
For example the `redirectedFrom` information. 
We can think of separating the functionality also.

src/test/java/reactor/netty/http/client/HttpRedirectTest.java
```diff
@@ -172,6 +175,43 @@ public void redirect_issuesOnRequestForEachAttempt() {
...
+						.followRedirect(true)
+						.doOnResponse((r, c) -> doOnResponseNanos.add(System.nanoTime()))
+						.get()
+						.uri("/1")
+						.response().map(r -> System.nanoTime())
+						.block(Duration.ofSeconds(30));
+
+		assertThat(responseNanos)
+				.isGreaterThan(doOnResponseNanos.poll(5, TimeUnit.SECONDS))
+				.isGreaterThan(doOnResponseNanos.poll(5, TimeUnit.SECONDS));
```
violetagg: There will be only one `doOnResponse` and only one `response` when you have redirect.

simonbasle:
@violetagg I think that idea of splitting `RESPONSE_RECEIVED` out of `HttpObserver` makes sense. Aligning the `doOnResponse` behavior with how core's `doOnXxx` operators behave (ie. the listeners execute _before_ the signal is passed down) also makes total sense.

rstoyanchev:
Sounds good indeed.

violetagg:
@adriancole Committed the test with f5898fea378262bd89d4bd5e966cddb78e1e423e
Removed the second expectation with 702012a268acda13d77d422878e65b389f3751f3 - doOnResponse/response are called once

PR #1034, opened 2020-03-20
fix #1022 Add support for graceful shutdown

### LINKED ISSUE #1022: Support Graceful Shutdown
The current solution is

Stoping the `ServerSocketChannel` so that no more connections are accepted

```
DisposableServer#disposeNow()
```

Disposing the `LoopResources` with a delay in order to close the current active connections.

```
HttpResources.disposeLoopsAndConnectionsLater(quietPeriod, timeout)
    .delaySubscription(Duration.ofSeconds(20)) --> this delay is just an example
    .block();
```

or 

```
LoopResources.disposeLater(quietPeriod, timeout)
    .delaySubscription(Duration.ofSeconds(20)) --> this delay is just an example
    .block();
```

We need a solution to be able to dispose the `LoopResources` as soon as there is no active connections instead of waiting a fixed time.

violetagg:
Fixes #1022

### DIFF
--- a/src/main/java/reactor/netty/http/server/HttpServer.java
+++ b/src/main/java/reactor/netty/http/server/HttpServer.java
@@ -25,6 +25,7 @@ import java.util.function.Function;
 import javax.annotation.Nullable;
 
 import io.netty.bootstrap.ServerBootstrap;
+import io.netty.channel.group.ChannelGroup;
 import io.netty.handler.codec.haproxy.HAProxyMessageDecoder;
 import io.netty.handler.codec.http.cookie.ServerCookieDecoder;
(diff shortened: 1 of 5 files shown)


justinesymphony:
Hi @violetagg, how can we configure a timeout for the server shutdown in this case?

violetagg:
@justinesymphony Please take a look at the javadoc (the note for graceful shutdown)
https://projectreactor.io/docs/netty/release/api/reactor/netty/http/server/HttpServer.html#channelGroup-io.netty.channel.group.ChannelGroup-
https://projectreactor.io/docs/netty/release/api/reactor/netty/tcp/TcpServer.html#channelGroup-io.netty.channel.group.ChannelGroup-

PR #1040, opened 2020-03-25
[build] Backport releaser.gradle additional tasks to 3.2.x

simonbasle:


simonbasle:
Note this doesn't include a change that applies `releaser.gradle`, since a lightweight version of it already existed and was applied

simonbasle:
merged forward to `0.9.x` and `master`

Issue #991, opened 2020-02-06
Too much `Bootstrap` and Tcp configuration related objects created for each request


anilgursel:
The `configure()` method gets called from in `TcpClient#connect` with every single request.  Calling `configure()` causes too much of `BootStrap` cloning also other setup to be done again and again.  When connection pool is in use, is there a specific reason to keep running configuration with each request?  For instance, for the below code example with max 10 connections in pool, I see `TcpUtils.updatePort` keeps getting called with every request:

```java
import io.netty.handler.codec.http.HttpMethod;
import reactor.core.publisher.Flux;
import reactor.netty.http.HttpProtocol;
import reactor.netty.http.client.HttpClient;
import reactor.netty.resources.ConnectionProvider;

public class Sample {

    public static void main(final String[] args) throws InterruptedException {
        System.out.println("Start...");

        final HttpClient httpClient =
                HttpClient.create(ConnectionProvider.fixed("dummy", 10, 100))
                        .baseUrl("http://localhost:8080")
                        .protocol(HttpProtocol.HTTP11)
                        .keepAlive(true);

        Flux.range(1, 1000000)
                .flatMap(i ->
                        httpClient.request(HttpMethod.GET)
                                .uri("/hello")
                                .responseSingle((resp, bytes) -> bytes.asString())
                                .doOnNext(s -> System.out.println(s)), 10)
                .blockLast();

        System.out.println("End..");
    }
}
```

And here is the mock service:

```java
import com.github.tomakehurst.wiremock.WireMockServer;
import com.github.tomakehurst.wiremock.core.WireMockConfiguration;

import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
import static com.github.tomakehurst.wiremock.client.WireMock.get;
import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.wireMockConfig;

public class MockServer {


    public static void main(final String[] args) {
        final WireMockConfiguration config =
                wireMockConfig()
                        .port(8080)
                        .containerThreads(10)
                        .jettyAcceptQueueSize(10);

        final WireMockServer wireMockServer = new WireMockServer(config);
        wireMockServer.start();

        wireMockServer.stubFor(get(urlPathEqualTo("/hello"))
                .willReturn(aResponse()
                        .withBody("Hello World!")
                        .withStatus(200)));
    }
}
```

[...]

Issue #586, opened 2019-01-24
Common abstraction for Client (and possibly Servers)


smaldini:
Add a common parent contract for client/server creations with default lifecycle observers and netty options.
- add channel group

violetagg:
Common abstraction is committed with #1046

Issue #568, opened 2019-01-11
Transition out from using Bootstrap attributes/options for configuration passing


smaldini:
Benchmark here (https://github.com/sharathsrinivasa/httpclientbenchmark) has revealed issues when subscribing or just creating many Client (HttpClient).
Root causing the top offenders showed Bootstrap#option/attr with cloning were very impacting.

violetagg:
Bootstrap is not used anymore #1046

PR #1046, opened 2020-03-27
Use the new Transport API for TCP/UDP/HTTP implementations

violetagg:
Related to #568, #586, #991


src/main/java/reactor/netty/resources/NewConnectionProvider.java
```diff
@@ -81,6 +86,41 @@
 		});
 	}
 
+	@Override
+	@SuppressWarnings("unchecked")
+	public <CONF extends TransportConfig, CONN extends Connection> Mono<CONN> acquire(TransportClientConfig<CONF, CONN> config) {
+		return Mono.create(sink -> {
```
simonbasle: do we need a `create`? (ie. would `defer` work?)

src/main/java/reactor/netty/resources/NewConnectionProvider.java
```diff
@@ -81,6 +86,41 @@
...
+				remote = Objects.requireNonNull(((Supplier<? extends SocketAddress>) remote).get(),
+						"Remote Address supplier returned null");
+			}
+
+			ChannelInitializer<Channel> channelInitializer =
+					config.channelInitializer(new NewConnectionObserver((MonoSink<Connection>) sink, config.connectionObserver()), remote);
+			TransportConnector.connect(config, remote, config.resolver(), channelInitializer).log()
+			                  .subscribe(
+			                          channel -> {
+			                              sink.onCancel(channel::close);
```
simonbasle: I wonder if that functionality can be integrated directly in the `TransportConnector`-generated `Mono`?

src/main/java/reactor/netty/transport/TransportConnector.java
```diff
@@ -0,0 +1,417 @@
...
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
+import static reactor.netty.ReactorNetty.format;
+
+/**
+ * @author Stephane Maldini
+ * @author Violeta Georgieva
+ * @since 1.0.0
+ */
+public final class TransportConnector {
```
simonbasle: Can you elaborate on what this class does (and describe it in the javadoc, since it is `public`)? Could it be package-private ? (ie do we expect users to directly call the only public static method of the class, `connect`?)

src/main/java/reactor/netty/transport/Transport.java
```diff
@@ -0,0 +1,318 @@
...
+
+	/**
+	 * Specifies whether the metrics are enabled on the {@link Transport}.
+	 * All generated metrics are provided to the specified recorder.
+	 *
+	 * @param enable if true enables the metrics on the {@link Transport}.
+	 * @param recorder the {@link ChannelMetricsRecorder}
+	 * @return a new {@link Transport} reference
+	 */
+	public final T metrics(boolean enable, ChannelMetricsRecorder recorder) {
```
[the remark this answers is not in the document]
violetagg: The idea is that you might have system property or env variable that controls whether metrics are enabled or not.

`metrics(boolean)` - enables/disables the default implementation with Micrometer while
`metrics(boolean, ChannelMetricsRecorder)`- enables/disables the metrics and uses a custom ChannelMetricsRecorder not the default with Micrometer

src/main/java/reactor/netty/resources/NewConnectionProvider.java
violetagg: This method returns `Mono<CONN>` where `CONN` extends `Connection` while `reactor.netty.transport.TransportConnector#connect` returns `Mono<Channel>`

The special `ConnectionObserver` will provide the `Connection` when it is available

src/main/java/reactor/netty/resources/NewConnectionProvider.java
```diff
@@ -81,6 +86,41 @@
...
+		return Mono.create(sink -> {
+			SocketAddress remote = config.remoteAddress();
+			if (remote instanceof Supplier) {
+				remote = Objects.requireNonNull(((Supplier<? extends SocketAddress>) remote).get(),
+						"Remote Address supplier returned null");
+			}
+
+			ChannelInitializer<Channel> channelInitializer =
+					config.channelInitializer(new NewConnectionObserver((MonoSink<Connection>) sink, config.connectionObserver()), remote);
+			TransportConnector.connect(config, remote, config.resolver(), channelInitializer).log()
```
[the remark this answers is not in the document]
violetagg: It was for a debugging purposes. I removed it.

src/main/java/reactor/netty/transport/TransportConnector.java
violetagg: The different `ConnectionProvider`'s implementations will call it so it cannot be package-private. I added javadoc.

src/main/java/reactor/netty/transport/Transport.java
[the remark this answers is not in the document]
simonbasle: but wouldn't it be weird to have a (nonNull) `recorder` parameter along with a `boolean false` value for `enable`? I kinda see the usage, but I fear it might get confusing. For instance, say I do `.metrics(false, new MyFancyMetricsRecorder())`. Then later on, I call `.metrics(true)` with the sense that it will activate the fancy metrics recorder I prepared earlier. But instead, it activates the micrometer one (because the `MyFancyMetricsRecorder` was never saved).

even if it makes it a tiny bit more complicated to use with a sysprop, we might instead go this way:
```java
//activate metrics, use the provided recorder
metricsRecordedWith(MetricsRecorder recorder);
//activate metrics, use the default recorder
metricsRecordedWithMicrometer();
//deactivate metrics
metricsDisabled();
```
wdyt @violetagg @rstoyanchev ?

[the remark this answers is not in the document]
violetagg: Javadoc clearly indicates the usage and the difference

```
	/**
	 * Whether to enable metrics to be collected and registered in Micrometer's
	 * {@link io.micrometer.core.instrument.Metrics#globalRegistry globalRegistry}.
	 * Applications can separately register their own
	 * {@link io.micrometer.core.instrument.config.MeterFilter filters}.
	 * For example, to put an upper bound on the number of tags produced:
	 * <pre class="code">
	 * MeterFilter filter = ... ;
	 * Metrics.globalRegistry.config().meterFilter(MeterFilter.maximumAllowableTags(prefix, 100, filter));
	 * </pre>
	 *
	 * @param enable true enables metrics collection; false disables it
	 * @return a new {@link Transport} reference
	 */
	public final T metrics(boolean enable) {
```

```
	/**
	 * Specifies whether the metrics are enabled on the {@link Transport}.
	 * All generated metrics are provided to the specified recorder.
	 *
	 * @param enable if true enables the metrics on the {@link Transport}.
	 * @param recorder the {@link ChannelMetricsRecorder}
	 * @return a new {@link Transport} reference
	 */
	public final T metrics(boolean enable, ChannelMetricsRecorder recorder) {
```

Till now there were no complains about the methods usage.

src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -96,10 +88,9 @@
  * @author Stephane Maldini
  * @author Violeta Georgieva
  */
-public abstract class HttpClient {
+public abstract class HttpClient extends TransportClient<HttpClient, HttpClientConfig> {
```
rstoyanchev, line 91: It is now possible to call `httpClient.connect()` it seems which doesn't make sense. I can suggest a separate interface `ConnectableTransport` interface that `HttpClient` does not implement.

Related to that, `HttpClientConnect` doesn't need anything from `HttpClient` and I don't see why it extends it, and likewise `HttpClient` could be self-sufficient, and only needs its own `config` field.

So then `HttpClient` could extend `ClientTransport` (without any `connect` methods), while `HttpClientConnect` doesn't extend anything and only implements `ConnectableTransport`.

If we do that then the names `HttpClientConnect`, `HttpClientFinalizer`, and `WebSocketFinalizer` no longer make sense. They could be `HttpConnectableTransport`, `HttpRequestSender`, and `WebSocketRequestSender`.

smaldini: @rstoyanchev I like the thinking here I wondered as much and we discussed that over a call with @violetagg .  The intent was to avoid multiplying the contracts but I can see how it deteriorates readability in that case.

An alternative would be to keep that connect() to provide an `HttpConnection` (duplex) with sendRequest/receiveResponse methods. The way `WebClient` works would easily align with this and gives an easier lifecycle management for users building on top of the client where users would still prefer the guided api (and the finalizers should change name as you suggested).

src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -476,7 +448,15 @@ public final HttpClient mapConnect(BiFunction<? super Mono<? extends Connection>
 	 * @return a new {@link HttpClient}
 	 */
 	public final HttpClient cookie(Cookie cookie) {
-		return new HttpClientCookie(this, cookie);
+		Objects.requireNonNull(cookie, "cookie");
+		if (!cookie.value().isEmpty()) {
+			HttpClient dup = duplicate();
+			HttpHeaders headers = configuration().headers.copy();
+			headers.add(HttpHeaderNames.COOKIE, dup.configuration().cookieEncoder.encode(cookie));
```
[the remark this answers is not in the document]
violetagg: Currently the strategy is once we duplicate we guarantee deep copy.
For example
`reactor.netty.transport.Transport#attr`
`reactor.netty.transport.Transport#option`
`reactor.netty.transport.Transport#observe`
etc.

src/main/java/reactor/netty/transport/TransportClient.java
```diff
@@ -0,0 +1,218 @@
...
+ *
+ * @param <T> TransportClient implementation
+ * @param <CONF> Client Configuration implementation
+ * @author Stephane Maldini
+ * @author Violeta Georgieva
+ * @since 1.0.0
+ */
+public abstract class TransportClient<T extends TransportClient<T, CONF>,
+		CONF extends TransportClientConfig<CONF>>
+		extends Transport<T, CONF> {
```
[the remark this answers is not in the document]
violetagg: done

src/main/java/reactor/netty/transport/TransportClientConfig.java
```diff
@@ -0,0 +1,239 @@
...
+
+/**
+ * Encapsulate all necessary configuration for client transport. The public API is read-only.
+ *
+ * @param <CONF> Configuration implementation
+ * @author Stephane Maldini
+ * @author Violeta Georgieva
+ * @since 1.0.0
+ */
+public abstract class TransportClientConfig<CONF extends TransportConfig> extends TransportConfig {
```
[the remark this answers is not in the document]
violetagg: done

src/main/java/reactor/netty/http/client/HttpClient.java
violetagg: in addition to the @smaldini comment: 
We have public API (used for example by RSocket)
`reactor.netty.http.client.HttpClient.WebsocketReceiver#connect`

so the only use case is pure HTTP requests

src/main/java/reactor/netty/http/client/HttpClient.java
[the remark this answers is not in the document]
violetagg: Added to javadoc that `duplicate` is a shallow copy.
We want to have deep copy only for a particular field only.

src/main/java/reactor/netty/http/client/HttpClient.java
violetagg: I made `ClientTransport#connect*` methods protected so only the transports that provide direct access to `connect` can expose that functionality.

src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -2093,11 +2089,11 @@ public void testIssue988() {
 		ConnectionProvider provider = ConnectionProvider.create("testIssue988", 1);
 		HttpClient client =
 				createHttpClientForContextWithAddress(provider)
-				        .tcpConfiguration(tcpClient -> tcpClient.wiretap("testIssue988", LogLevel.INFO)
-				                                                .metrics(true));
+				        .wiretap("testIssue988", LogLevel.INFO)
+				        .metrics(true, s -> s);
```
simonbasle, line 2093: I wonder how often we expect users to change the representation of the URI in the metrics tags... if the answer is "not often", maybe it would be worth it to add an overload that doesn't take the `Function`?

violetagg: This is https://github.com/reactor/reactor-netty/pull/1066 the PR from 0.9.x
The method `metrics(boolean, Function)` is not related to this PR.
The PR has adaptation for the fix in 0.9.x as I rebase on top of the master branch and adapt all changed that are merged to master.

simonbasle: thanks for the reminder 👍

src/main/java/reactor/netty/http/client/FailedHttpClientRequest.java
```diff
@@ -0,0 +1,137 @@
...
+import reactor.netty.http.HttpOperations;
+import reactor.util.context.Context;
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * @author Violeta Georgieva
+ */
+final class FailedHttpClientRequest implements HttpClientRequest {
```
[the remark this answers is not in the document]
violetagg: done

src/main/java/reactor/netty/ReactorNetty.java
```diff
@@ -391,6 +392,64 @@ else if (log.isDebugEnabled()) {
 		}
 	}
 
+	static ChannelPipelineConfigurer compositeChannelPipelineConfigurer(
```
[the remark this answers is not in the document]
violetagg: done

violetagg:
@simonbasle @rstoyanchev Are we good with this change?

PR #1047, opened 2020-03-30
Enable WebSocket connection to external services.

Unthrottled:
## Description

Ensures that the websocket connection url is the same as the websocket channel, as the host and open channel are not always the same. Such as when proxying websocket requests to external services. Example using spring cloud gateway to proxy websocket connections to other services.

## Motivation

Spring cloud gateway is currently unable to proxy websocket connections.

## Example project 

See sample error project for more information about the issue this pull request fixes:
https://github.com/QDivision/external-websocket-error-sample


Unthrottled:
Looks like I have some failing tests. I'll get it fixed.

Unthrottled:
I am not sure how I can write a test to exercise this bug.
All the tests start a server on localhost. I want it to open a connection to a foreign host and not open a connection to itself.

violetagg:
@Unthrottled 
`reactor.netty.http.client.HttpClientOperations#withWebsocketSupport` is always used from
`reactor.netty.http.client.HttpClientConnect.HttpClientHandler#requestWithBody`

so instead of calculating the URL you can provide it immediately as it is known

https://github.com/reactor/reactor-netty/blob/99e483b44e22aec42ade109111bd007653160e49/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L537-L544

About testing, can't you play with the HOST header?

Unthrottled:
@violetagg 
> so instead of calculating the URL you can provide it immediately as it is known
I think I did what you wanted, let me know if you actually wanted something different.

I also wrote a test that exercised the issue. It does not replicate the issue the same way as we found it, but it is close enough.

src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -534,7 +535,7 @@ public SocketAddress get() {
 
 		Publisher<Void> requestWithBody(HttpClientOperations ch) {
 			try {
-				ch.resourceUrl = toURI.toExternalForm();
+				ch.resourceUrl = new URI(toURI.toExternalForm());
```
violetagg: Please keep this as String, because `resourceUrl()` might be called many times while the web socket URI is needed just once for the handshake

src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -16,9 +16,7 @@
 
 package reactor.netty.http.client;
 
-import java.net.InetSocketAddress;
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.net.*;
```
violetagg: do not use `.*` for the imports

src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -410,27 +408,7 @@ public NettyOutbound send(Publisher<? extends ByteBuf> source) {
...
-		}
-		catch (URISyntaxException e) {
-			throw new IllegalArgumentException(e);
-		}
-		return uri;
-	}
-
-
-	@Override
+  @Override
```
violetagg: keep the formatting

src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -707,6 +685,7 @@ final void withWebsocketSupport(String protocols, int maxFramePayloadLength, boo
 				                WebSocketClientCompressionHandler.INSTANCE);
 			}
 
+			log.debug(String.format("Attempting to perform websocket handshake with %s", url));
```
violetagg:
```suggestion
			if (log.isDebugEnabled()) {
			    log.debug(format(channel(), "Attempting to perform websocket handshake with {}"), url);
			}
```

src/main/java/reactor/netty/http/client/HttpClientConnect.java
Unthrottled: Fixed!

src/main/java/reactor/netty/http/client/HttpClientOperations.java
Unthrottled: My IDE is trying to be helpful, but in this case it was not. I fixed it!

src/main/java/reactor/netty/http/client/HttpClientOperations.java
Unthrottled: Fixed!

violetagg:
Ignore the GitHub Actions failure, the build is OK

violetagg:
@Unthrottled The changes are squashed and committed to 0.9.x - bd19e12875dfad2e469b532b25c2e105f9390273
then forward merge to master

Thanks for the contribution

PR #1055, opened 2020-04-07
Deprecate #addressSupplier(Supplier) in favour of server#localAddress…

violetagg:
…(Supplier)/client#remoteAddress(Supplier)

Motivation:
It is clear on API level whether it is local or remote address configuration.
The API can provide on client both local and remote address configuration.


violetagg:
> LGTM. I wonder if instead of `localAdress` using the `bind` terminology would be helpful. I guess people are accustomed to `remote` vs `local`, so using `bindAdress` could maybe break expectations (in which case, principle of least surprise, use the most common terminology)

Ok `bindAddress` sounds better

PR #1056, opened 2020-04-07
Ensure HttpClient#doAfterResponseSuccess is sent after the inbound is completed

violetagg:
Related to https://github.com/spring-projects/spring-framework/issues/24846


PyvesB:
I've tested this branch with the unit tests from https://github.com/PyvesB/spring-webclient-connection-close, I can confirm that they now pass. 👍

Issue #981, opened 2020-01-30
SSLEngine closed already


dkulig:
I have a SpringBoot WebFlux app which Is talking to an external http server. 
All looks good expect the fact that my tracing library is reporting `SSLEngine closed already` errors randomly after a request to my app is already finished.

`javax.net.ssl.SSLException: SSLEngine closed already at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:837) at io.netty.handler.ssl.SslHandler.wrapAndFlush(SslHandler.java:800) at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:781) at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:1886) at io.netty.handler.ssl.SslHandler.closeOutboundAndChannel(SslHandler.java:1854) at `

The weird thing is that this is happening 10 - 20 - 30 **minutes** after the request completes, I happens only periodically.

`
Gradle: io.projectreactor.netty:reactor-netty:0.9.2.RELEASE
Gradle: io.projectreactor:reactor-core:3.3.1.RELEASE
Gradle: io.projectreactor:reactor-test:3.3.1.RELEASE
`

Spring set-up:

```
    @Bean
    fun webClient(httpClient: HttpClient): WebClient = WebClient.builder()
        .clientConnector(ReactorClientHttpConnector(httpClient))
        .build()

    @Bean
    fun httpClient(tcpClient: TcpClient): HttpClient = HttpClient
        .from(tcpClient)
        .compress(true)

    @Bean
    fun tcpClient(@Value("\${app.httpClient.requestTimeoutMs}") httpRequestTimeout: Int): TcpClient =
        TcpClient.create(connectionProvider())
            .option(ChannelOption.CONNECT_TIMEOUT_MILLIS, httpRequestTimeout)
            .doOnConnected { connection ->
                connection.addHandlerLast(ReadTimeoutHandler(httpRequestTimeout.toLong(), TimeUnit.MILLISECONDS))
                    .addHandlerLast(WriteTimeoutHandler(httpRequestTimeout.toLong(), TimeUnit.MILLISECONDS))
            }

    private fun connectionProvider(): ConnectionProvider {
        return ConnectionProvider.fixed(
            "fixed-tpc-connection-provider",
            DEFAULT_NETTY_MAX_CONNECTIONS,
            DEFAULT_NETTY_ACQUIRE_TIMEOUT_MS,
[...]

PR #1065, opened 2020-04-15
Ensure re-acquire in case of channel was active when acquiring but cl…

violetagg:
…osed when PooledConnectionProvider.DisposableAcquire#onNext

Ensure the "owner" will be set in the channel only if the channel is active

Related to #782, #981

### DIFF
--- a/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -45,6 +45,7 @@ import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscription;
 import reactor.core.CoreSubscriber;
 import reactor.core.Disposable;
+import reactor.core.Disposables;
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.MonoProcessor;
 import reactor.core.publisher.MonoSink;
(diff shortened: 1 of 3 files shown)


violetagg:
the failing tests are not related

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -462,6 +462,16 @@ public String toString() {
...
 		}
 
+		DisposableAcquire(DisposableAcquire parent) {
+			this.cancellations = parent.cancellations;
+			this.sink = parent.sink;
+			this.pool = parent.pool;
+			this.obs = parent.obs;
+			this.opsFactory = parent.opsFactory;
+			this.pendingAcquireTimeout = parent.pendingAcquireTimeout;
+			this.retried = !parent.retried;
```
simonbasle, line 472: is it possible that the `DisposableAcquire` is created twice this way, meaning that the third instance would be back to `retried == false` ? ie here the state transition from "retried" to "not retried" is possible, but that doesn't sound relevant and all disposableAcquires derived from an original one should probably be considered as "retried", right?

violetagg: agree I will change it

Issue #1010, opened 2020-02-20
Netty metrics OutOfMemoryError


deci260:
Using reactor-netty metrics causes a java OutOfMemory error when dealing with multiple requests at the same time.
I'm assuming it's a metrics issue because when disabling netty metrics it works fine.

## Steps to Reproduce (using spring-boot-starter-reactor-netty)

- Activate reactor-netty metrics:
```java
@Bean 
public NettyWebServerFactoryCustomizer customizer(Environment environment, ServerProperties serverProperties) {
	return new NettyWebServerFactoryCustomizer(environment, serverProperties) {
		@Override
		public void customize(NettyReactiveWebServerFactory factory) {
			factory.addServerCustomizers(httpServer -> httpServer.metrics(true));
		}
	};
}
```

- Perform multiple requests at the same time (In my tests I wrote a python script with 10 threads, each one of them performs infinite gets on netty).

## Your Environment

I'm using spring cloud gateway 2.2.2.BUILD-SNAPSHOT and spring cloud dependencies Hoxton.BUILD-SNAPSHOT with other metrics enabled. 

* Reactor version(s) used: 0.9.4.RELEASE
* Netty version: 4.1.45.Final
* JVM version: 1.8.0_191 
* OS and version: Windows 10 Insider Preview (10.0.19564 Build 19564)

This happens only when netty metrics are enabled, spring cloud gateway metrics or jvm metrics provided by spring do not cause this.

violetagg:
@deci260 Can you give us more information - how is configured Spring Gateway, do you have a lot of URLs or just simultaneous requests to one and the same URL? Can you give us reproducible example?

deci260:
@violetagg Spring cloud gateway is configured to use different static routes and uses discovery locator. It is also configured to generate custom tags, which has caused no problem as of now. I first found out about this issue in-browser, while on a page that uses spring cloud gateway to load different resources (all requests are routed to the same route). I've also been able to reproduce it by doing simultaneous requests to the same url over and over again. Unfortunately I haven't been able to reproduce the issue at home and I cannot give you the project on which this happens, but I'll be able to give you some more info tomorrow.

deci260:
After further analysis,  might not be a problem on your side. I am able to reproduce this issue when a certain number of tags are used. I also get this warning from reactor before the application starts dying:
[...]

PR #1066, opened 2020-04-16
Add HttpServer/HttpClient#metrics(boolean, Function<String, String>)

violetagg:
The new API is a variant of HttpServer/HttpClient#metrics(boolean) that also accepts
a function that receives the actual uri and returns the uri tag value
that will be used for the metrics with Metrics#URI tag.

Related to #1010


violetagg:
> Looks good overall but maybe `metrics()` should be deprecated in order to require consideration of a `uriTagValue`. It seems that should be part of the decision when enabling metrics, and it's easy enough to provide a no-op in case it's not relevant.

done

PR #1070, opened 2020-04-23
fix #935 Add separate builder for Websocket client configuration

### LINKED ISSUE #935: Add builder for configuring WebSocket on the client

violetagg:
fixes #935

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -454,6 +454,13 @@ public abstract class HttpClient {
 	/**
 	 * Specifies whether GZip compression/websocket compression
 	 * extension is enabled.
+	 * <p>
+	 * Note: Using this method for enabling websocket compression is strongly discouraged.
+	 * As of 0.9.7, use {@link WebsocketClientSpec#builder()} for providing websocket compression configuration.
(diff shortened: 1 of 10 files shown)


src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -454,6 +454,10 @@ public final HttpClient remoteAddress(Supplier<? extends SocketAddress> remoteAd
 	/**
 	 * Specifies whether GZip compression/websocket compression
 	 * extension is enabled.
+	 * <p>
+	 * Note: Using this method for enabling websocket compression is strongly discouraged.
+	 * As of 0.9.7, use {@link WebsocketClientSpec#builder()} for providing websocket configuration.
+	 * </p>
```
rstoyanchev, line 460: Allowing the HTTP `compress` property to be inherited by default for WebSocket would match the current behavior and avoid breaking backwards compatibility eventually when the this `compress` flag is no longer checked for WebSocket.

In any case, explaining a little more here in this note would be helpful, i.e. that setting this currently applies to both, but eventually it won't, etc.

violetagg: javadoc is extended

PR #1072, opened 2020-04-23
fix #1071 Add HttpServer configuration for specifying the max content…

### LINKED ISSUE #1071: Support for h2c upgrade with body
In https://tools.ietf.org/html/rfc7540#section-3.2 in order to perform an http2 clear text upgrade, requests that contain a payload body MUST be sent in their entirety before the client can send HTTP/2 frames.

## Expected Behavior
When a client execute a request without prior knowledge of h2c support by the server it send the same request with Upgrade and HTTP2-Settings. If the request contains a body the full body is sent along. This is what hapend when you use curl:
```bash
curl --http2 -v --request POST \
  --url http://localhost:8080/echo \
  --header 'content-type: application/json' \
  --data '{"message": "Hello"}'
```
or use Java 11 HttpClient
```java
final HttpResponse<String> response = HttpClient.newHttpClient()
                .send(HttpRequest.newBuilder()
                        .uri(URI.create("http://localhost:8080/echo"))
                        .POST(BodyPublishers.ofString("Hello world"))
                        .build(), BodyHandlers.ofString());
```

## Actual Behavior
Currently an HttpServer configured to support H2C and HTTP11 will respond with status 413.

## Steps to Reproduce
Create an HttpServer with H2C support like:

```java
import reactor.netty.DisposableServer;
import reactor.netty.http.HttpProtocol;
import reactor.netty.http.server.HttpServer;

public class Application {

    public static void main(String[] args) {
        DisposableServer server =
                HttpServer.create()
                        .protocol(HttpProtocol.H2C, HttpProtocol.HTTP11)
                        .route(routes ->
                                routes.post("/echo", (request, response) -> response.send(request.receive().retain()))
                        )
                        .port(8080)
                        .bindNow();

        server.onDispose().block();
    }

}
```
and call the echo endpoint with an http2 request without prior knowledge using curl or Java HttpClient (see Expected Behavior)

## Possible Solution
This is caused by the HttpServerUpgradeHandler used in Http1OrH2CleartextInitializer:
```java
final CleartextHttp2ServerUpgradeHandler h2cUpgradeHandler = new CleartextHttp2ServerUpgradeHandler(
					httpServerCodec,
					new HttpServerUpgradeHandler(httpServerCodec, upgrader),
					http2ServerHandler);
```

The HttpServerUpgradeHandler initialized like this expect the upgrade to be a GET request. Another constructor with a maximum content-lenght can be used and could be configured at the HttpServer builder.

Or another specific implementation for h2c upgrade that will skip the content-length check can be used.

## Your Environment

* Reactor version used: 3.3.4.RELEASE
* Reactor netty: 0.9.6.RELEASE
* Netty: 4.1.48.Final
* JVM version (`javar -version`): openjdk version "11.0.6" 2020-01-14
* OS and version (eg `uname -a`): Ubuntu 18.04 64bit 4.15.0-96-generic

violetagg:
… length of the H2C upgrade request

fixes #1071

### DIFF
--- a/src/main/java/reactor/netty/http/server/HttpRequestDecoderSpec.java
+++ b/src/main/java/reactor/netty/http/server/HttpRequestDecoderSpec.java
@@ -22,17 +22,44 @@ import reactor.netty.tcp.TcpServer;
 
 /**
  * A configuration builder to fine tune the {@link io.netty.handler.codec.http.HttpServerCodec}
- * (or more precisely the {@link io.netty.handler.codec.http.HttpServerCodec.HttpServerRequestDecoder}).
+ * (or more precisely the {@link io.netty.handler.codec.http.HttpServerCodec.HttpServerRequestDecoder}) for HTTP/1.1
+ * or {@link io.netty.handler.codec.http.HttpServerUpgradeHandler} for H2C.
  * <p>
  * Defaults are accessible as constants {@link #DEFAULT_MAX_INITIAL_LINE_LENGTH}, {@link #DEFAULT_MAX_HEADER_SIZE},
- * {@link #DEFAULT_MAX_CHUNK_SIZE}, {@link #DEFAULT_INITIAL_BUFFER_SIZE} and {@link #DEFAULT_VALIDATE_HEADERS}.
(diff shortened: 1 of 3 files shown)


violetagg:
the failing tests are not related

PR #1076, opened 2020-04-27
fix #1075 Remove REMOTE_ADDRESS tag from the metrics reported by Micr…

### LINKED ISSUE #1075: Server metrics cardinality explosion
## Expected Behavior
A reasonable amount of series for every reactor_netty_http_server_* metric.

## Actual Behavior
There is a cardinality explosion with one per `client ip` * `client random port` * `other tags`

metrics-cardinality

## Steps to Reproduce
- Enable HttpServer.metrics(true)
- Curl some endpoints

## Possible Solution
Remove remote_address tag from all server metrics

## Your Environment

* Reactor version(s) used: 0.9.4
* Other relevant libraries versions (eg. `netty`, ...): Spring Boot 2.2.4.RELEASE
* JVM version (`javar -version`): OpenJDK 11
* OS and version (eg `uname -a`): MacOS Catalina

violetagg:
…ometerHttpServerMetricsRecorder

MicrometerHttpServerMetricsRecorder does not record metrics on protocol level

Fixes #1075

### DIFF
--- a/src/main/java/reactor/netty/http/MicrometerHttpMetricsRecorder.java
+++ b/src/main/java/reactor/netty/http/MicrometerHttpMetricsRecorder.java
@@ -51,14 +51,14 @@ public class MicrometerHttpMetricsRecorder extends MicrometerChannelMetricsRecor
 	protected final Timer.Builder responseTimeBuilder;
 	protected final ConcurrentMap<MeterKey, Timer> responseTimeCache = PlatformDependent.newConcurrentHashMap();
 
-	final DistributionSummary.Builder dataReceivedBuilder;
-	final ConcurrentMap<MeterKey, DistributionSummary> dataReceivedCache = PlatformDependent.newConcurrentHashMap();
+	protected final DistributionSummary.Builder dataReceivedBuilder;
+	protected final ConcurrentMap<MeterKey, DistributionSummary> dataReceivedCache = PlatformDependent.newConcurrentHashMap();
 
-	final DistributionSummary.Builder dataSentBuilder;
-	final ConcurrentMap<MeterKey, DistributionSummary> dataSentCache = PlatformDependent.newConcurrentHashMap();
+	protected final DistributionSummary.Builder dataSentBuilder;
+	protected final ConcurrentMap<MeterKey, DistributionSummary> dataSentCache = PlatformDependent.newConcurrentHashMap();
 
-	final Counter.Builder errorsBuilder;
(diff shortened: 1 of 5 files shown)


src/main/java/reactor/netty/http/MicrometerHttpMetricsRecorder.java
```diff
@@ -19,28 +19,23 @@
...
-import static reactor.netty.Metrics.REMOTE_ADDRESS;
 import static reactor.netty.Metrics.RESPONSE_TIME;
-import static reactor.netty.Metrics.URI;
 
 /**
  * @author Violeta Georgieva
  * @since 0.9
  */
-public class MicrometerHttpMetricsRecorder extends MicrometerChannelMetricsRecorder implements HttpMetricsRecorder {
+public abstract class MicrometerHttpMetricsRecorder extends MicrometerChannelMetricsRecorder implements HttpMetricsRecorder {
```
simonbasle, line 38: I _think_ that is binary compatible, the `abstract` + removal of overrides in this base class, but can you confirm JAPICMP doesn't find any issue here?

src/main/java/reactor/netty/http/server/HttpServerBind.java
```diff
@@ -380,6 +380,10 @@ public void accept(ConnectionObserver listener, Channel channel) {
 				if (channelMetricsRecorder instanceof HttpServerMetricsRecorder) {
 					p.addAfter(NettyPipeline.HttpTrafficHandler, NettyPipeline.HttpMetricsHandler,
 							new HttpServerMetricsHandler((HttpServerMetricsRecorder) channelMetricsRecorder, uriTagValue));
+					if (channelMetricsRecorder instanceof MicrometerHttpServerMetricsRecorder) {
+						// MicrometerHttpServerMetricsRecorder does not implement metrics on protocol level
+						p.remove(handler);
```
simonbasle, line 385: nitpick: reading the code it was not immediately clear `p` was the channel-level metrics and was different from the `HttpServerMetricsHandler` (applies to 3 sites)

src/main/java/reactor/netty/http/MicrometerHttpMetricsRecorder.java
violetagg: There is an issue if this class is inherited.
Fixed

src/main/java/reactor/netty/http/server/HttpServerBind.java
violetagg: fixed

PR #1078, opened 2020-04-27
fix #453 Deprecate LoopResources #onChannel/onServerChannel/onDatagramChannel

### LINKED ISSUE #453: Rework global resources in a single registry
Given how much they overlap, we should refactor TcpResources and UdpResources into a single GlobalResource (or similar).

violetagg:
In favour of LoopResources#onChannel(Class, EventLoopGroup)
This is in preparation for Unix Domain Sockets support

Fixes #453

### DIFF
--- a/src/main/java/reactor/netty/resources/DefaultLoopNIO.java
+++ b/src/main/java/reactor/netty/resources/DefaultLoopNIO.java
@@ -19,73 +19,49 @@ import java.util.concurrent.ThreadFactory;
 
 import io.netty.channel.Channel;
 import io.netty.channel.EventLoopGroup;
-import io.netty.channel.epoll.Epoll;
-import io.netty.channel.epoll.EpollDatagramChannel;
-import io.netty.channel.epoll.EpollEventLoopGroup;
-import io.netty.channel.epoll.EpollServerSocketChannel;
-import io.netty.channel.epoll.EpollSocketChannel;
 import io.netty.channel.socket.DatagramChannel;
 import io.netty.channel.socket.ServerSocketChannel;
 import io.netty.channel.socket.SocketChannel;
-import reactor.util.Logger;
-import reactor.util.Loggers;
+import io.netty.channel.socket.nio.NioDatagramChannel;
+import io.netty.channel.socket.nio.NioServerSocketChannel;
+import io.netty.channel.socket.nio.NioSocketChannel;
 
 /**
- * {@link DefaultLoop} that uses {@code Epoll} transport.
+ * {@link DefaultLoop} that uses {@code NIO} transport.
  *
  * @author Stephane Maldini
  * @author Violeta Georgieva
+ * @since 0.9.8
  */
-final class DefaultLoopEpoll implements DefaultLoop {
+final class DefaultLoopNIO implements DefaultLoop {
 
 	@Override
 	@SuppressWarnings("unchecked")
 	public <CHANNEL extends Channel> CHANNEL getChannel(Class<CHANNEL> channelClass) {
 		if (channelClass.equals(SocketChannel.class)) {
-			return (CHANNEL) new EpollSocketChannel();
+			return (CHANNEL) new NioSocketChannel();
 		}
 		if (channelClass.equals(ServerSocketChannel.class)) {
-			return (CHANNEL) new EpollServerSocketChannel();
+			return (CHANNEL) new NioServerSocketChannel();
 		}
 		if (channelClass.equals(DatagramChannel.class)) {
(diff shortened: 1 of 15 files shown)


violetagg:
the failing test is not related

src/main/java/reactor/netty/tcp/TcpResources.java
```diff
@@ -210,6 +210,7 @@ public void disposeWhen(@NonNull SocketAddress address) {
 	}
 
 	@Override
+	@SuppressWarnings("deprecation")
```
simonbasle, line 213: is the method going to stay in `TcpResources` after being removed from `LoopResources` interface? If not I'd say suppressing warnings here don't make much sense

simonbasle: (also valid for the 2 occurrences below)

src/main/java/reactor/netty/udp/UdpResources.java
```diff
@@ -167,6 +167,7 @@ public boolean isDisposed() {
 	}
 
 	@Override
+	@SuppressWarnings("deprecation")
```
simonbasle, line 170: same question as in `TcpResources`

src/main/java/reactor/netty/resources/LoopResources.java
```diff
@@ -206,18 +225,18 @@ default EventLoopGroup onClient(boolean useNative) {
...
 	 */
+	@Deprecated
 	default Class<? extends ServerChannel> onServerChannel(EventLoopGroup group) {
-		return preferNative() ? DefaultLoopNativeDetector.getInstance().getServerChannel(group) :
+		return preferNative() ? onChannel(ServerSocketChannel.class, group).getClass() :
 				NioServerSocketChannel.class;
 	}
 
 	/**
 	 * Create a server select {@link EventLoopGroup} for servers to be used
```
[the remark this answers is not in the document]
violetagg: There is a difference

`onServerSelect` - Create a **server select** {@link EventLoopGroup}, which means the event loop group for the server acceptor channel, in Netty this is the `boss` group.

`onServer` - Callback for **server** {@link EventLoopGroup}, which means the event loop group for the child channel that will process the request, in Netty this is the `worker` group.

see here
https://github.com/netty/netty/blob/387e451c8289f05bf553821170bde30052101f8a/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java#L218
https://github.com/netty/netty/blob/387e451c8289f05bf553821170bde30052101f8a/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java#L154

In `LoopResources`, `onClient` and `onServerSelect` are aliases for `onServer`, but in the actual implementation - 
`DefaultLoopResources`, the implementation is different depending on the configuration i.e. number of the worker and selector threads (that's for the server), while for the client we create `ColocatedEventLoopGroup` so that when in one JVM the server and the client will share the workers.

src/main/java/reactor/netty/tcp/TcpResources.java
violetagg: they will be removed in master
and it is suppressed because otherwise `errorprone` plugin complains about it

```
/.../reactor-netty/src/main/java/reactor/netty/tcp/TcpResources.java:225: warning: [deprecation] onDatagramChannel(EventLoopGroup) in LoopResources has been deprecated
        public Class<? extends DatagramChannel> onDatagramChannel(EventLoopGroup group) {
                                                ^
/.../reactor-netty/src/main/java/reactor/netty/tcp/TcpResources.java:226: warning: [deprecation] onDatagramChannel(EventLoopGroup) in LoopResources has been deprecated
                return defaultLoops.onDatagramChannel(group);
                                   ^
```

src/main/java/reactor/netty/resources/DefaultLoopNIO.java
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	@Override
+	public EventLoopGroup newEventLoopGroup(int threads, ThreadFactory factory) {
+		throw new IllegalStateException("Missing Epoll/KQueue on current system");
+	}
+
+	@Override
+	public boolean supportGroup(EventLoopGroup group) {
+		return true;
```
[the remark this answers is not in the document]
violetagg: changed

src/main/java/reactor/netty/resources/LoopResources.java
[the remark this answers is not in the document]
violetagg: I extended the javadoc

Issue #403, opened 2018-07-26
Add UnixDomain support


smaldini:
Would be good for IPC and could mirror most of TcpClient/Server under UnixClient/UnixServer

mp911de:
Domain sockets kind-of work when supplying a custom `LoopResources` implementation that sets the appropriate `Class<? extends Channel>`.

With the default setup, connect fails with `Address family not supported by protocol` as the socket channel class defaults to a regular `[Epoll|KQueue]SocketChannel`.

mp911de:
See also mirromutth/r2dbc-mysql#80 and r2dbc/r2dbc-postgresql#181 as pre-requisites for GCP MySQL/Postgres usage via R2DBC.

violetagg:
UDS for TCP transport #1080

PR #1080, opened 2020-04-29
Add UDS support

violetagg:
Covers TCP transport

Related to #403

### DIFF
--- a/src/main/java/reactor/netty/resources/DefaultLoopEpoll.java
+++ b/src/main/java/reactor/netty/resources/DefaultLoopEpoll.java
@@ -21,12 +21,16 @@ import io.netty.channel.Channel;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.epoll.Epoll;
 import io.netty.channel.epoll.EpollDatagramChannel;
+import io.netty.channel.epoll.EpollDomainSocketChannel;
 import io.netty.channel.epoll.EpollEventLoopGroup;
+import io.netty.channel.epoll.EpollServerDomainSocketChannel;
 import io.netty.channel.epoll.EpollServerSocketChannel;
 import io.netty.channel.epoll.EpollSocketChannel;
 import io.netty.channel.socket.DatagramChannel;
 import io.netty.channel.socket.ServerSocketChannel;
 import io.netty.channel.socket.SocketChannel;
+import io.netty.channel.unix.DomainSocketChannel;
+import io.netty.channel.unix.ServerDomainSocketChannel;
 import reactor.util.Logger;
 import reactor.util.Loggers;
 
@@ -49,6 +53,12 @@ final class DefaultLoopEpoll implements DefaultLoop {
 		}
 		if (channelClass.equals(DatagramChannel.class)) {
 			return (CHANNEL) new EpollDatagramChannel();
+		}
+		if (channelClass.equals(DomainSocketChannel.class)) {
+			return (CHANNEL) new EpollDomainSocketChannel();
+		}
+		if (channelClass.equals(ServerDomainSocketChannel.class)) {
+			return (CHANNEL) new EpollServerDomainSocketChannel();
 		}
 		throw new IllegalArgumentException("Unsupported channel type: " + channelClass.getSimpleName());
 	}
--- a/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -236,9 +236,9 @@ final class PooledConnectionProvider implements ConnectionProvider {
 
 				pooledConnection.bind();
 
+				ch.pipeline().remove(this);
 				ch.pipeline()
 				  .addFirst(config.channelInitializer(pooledConnection, remoteAddress, false));
-				ch.pipeline().remove(this);
 			}
 
 			@Override
@@ -634,7 +634,7 @@ final class PooledConnectionProvider implements ConnectionProvider {
 
 		PoolKey(SocketAddress holder, int pipelineKey) {
 			this.holder = holder;
-			this.fqdn = holder instanceof InetSocketAddress ? holder.toString() : "null";
+			this.fqdn = holder.toString();
 			this.pipelineKey = pipelineKey;
 		}
 
--- a/src/main/java/reactor/netty/transport/TransportConfig.java
+++ b/src/main/java/reactor/netty/transport/TransportConfig.java
@@ -237,11 +237,32 @@ public abstract class TransportConfig {
 	}
 
 	/**
+	 * Return the channel type this configuration is associated with, it can be one of the following:
+	 * <ul>
+	 *   <li>{@link io.netty.channel.socket.SocketChannel}</li>
+	 *   <li>{@link io.netty.channel.socket.ServerSocketChannel}</li>
+	 *   <li>{@link io.netty.channel.unix.DomainSocketChannel}</li>
+	 *   <li>{@link io.netty.channel.unix.ServerDomainSocketChannel}</li>
+	 *   <li>{@link io.netty.channel.socket.DatagramChannel}</li>
+	 * </ul>
+	 *
+	 * @param isDomainSocket true if {@link io.netty.channel.unix.DomainSocketChannel} or
+	 * {@link io.netty.channel.unix.ServerDomainSocketChannel} is needed, false otherwise
+	 * @return the channel type this configuration is associated with
+	 */
+	protected abstract Class<? extends Channel> channelType(boolean isDomainSocket);
+
+	/**
 	 * Return the {@link ChannelFactory} which is used to create {@link Channel} instances.
 	 *
+	 * @param elg the {@link EventLoopGroup}
+	 * @param isDomainSocket true if {@link io.netty.channel.unix.DomainSocketChannel} or
+	 * {@link io.netty.channel.unix.ServerDomainSocketChannel} is needed, false otherwise
 	 * @return the {@link ChannelFactory} which is used to create {@link Channel} instances.
 	 */
-	protected abstract ChannelFactory<? extends Channel> connectionFactory(EventLoopGroup elg);
+	protected ChannelFactory<? extends Channel> connectionFactory(EventLoopGroup elg, boolean isDomainSocket) {
+		return () -> loopResources().onChannel(channelType(isDomainSocket), elg);
+	}
 
 	/**
 	 * Return the configured default {@link ConnectionObserver}.
--- a/src/main/java/reactor/netty/udp/UdpClientConfig.java
+++ b/src/main/java/reactor/netty/udp/UdpClientConfig.java
@@ -74,15 +74,24 @@ public final class UdpClientConfig extends ClientTransportConfig<UdpClientConfig
 	}
 
 	@Override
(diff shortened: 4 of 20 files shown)


src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -634,7 +635,7 @@ void registerClose(PooledRef<PooledConnection> pooledRef, InstrumentedPool<Poole
 
 		PoolKey(SocketAddress holder, int pipelineKey) {
 			this.holder = holder;
-			this.fqdn = holder instanceof InetSocketAddress ? holder.toString() : "null";
+			this.fqdn = holder instanceof InetSocketAddress || holder instanceof DomainSocketAddress ? holder.toString() : "null";
```
ericbottard, line 638: Should this condition be inverted and `toString()` invoked for all non-null cases instead?

src/main/java/reactor/netty/udp/UdpClientConfig.java
```diff
@@ -74,15 +74,26 @@ public final InternetProtocolFamily family() {
...
+
+	@Override
+	protected ChannelFactory<? extends Channel> connectionFactory(EventLoopGroup elg, boolean isDomainSocket) {
+		if (isDomainSocket) {
+			throw new UnsupportedOperationException();
+		}
+		DatagramChannel channel;
 		if (isPreferNative()) {
-			channelFactory = () -> loopResources().onChannel(DatagramChannel.class, elg);
+			channel = loopResources().onChannel(DatagramChannel.class, elg);
```
ericbottard, line 91: This changes the semantics of when objects are created (once vs at factory invocation). Is this intended?

src/test/java/reactor/netty/tcp/TcpClientTests.java
```diff
@@ -1029,4 +1030,35 @@ private void checkReference(WeakReference<ByteBuf> ref) throws Exception {
...
+		         .connectNow();
+	}
+
+	@Test(expected = IllegalArgumentException.class)
+	public void testTcpClientWithDomainSocketsWithPort() {
+		TcpClient.create()
+		         .remoteAddress(() -> new DomainSocketAddress("/tmp/test.sock"))
+		         .port(1234)
+		         .connectNow();
+	}
```
ericbottard, line 1063: Should there be new tests for things that *are* supported?

src/test/java/reactor/netty/tcp/TcpServerTests.java
```diff
@@ -934,6 +936,77 @@ public void testGracefulShutdown() throws Exception {
...
+	public void testTcpServerWithDomainSocketsWithPort() {
+		TcpServer.create()
+		         .bindAddress(() -> new DomainSocketAddress("/tmp/test.sock"))
+		         .port(1234)
+		         .bindNow();
+	}
+
+	@Test
+	public void testTcpServerWithDomainSockets() throws Exception {
+		if (!LoopResources.hasNativeSupport()) {
```
ericbottard, line 972: Seems like a could candidate for JUnit `assumeTrue()`

src/test/java/reactor/netty/tcp/TcpServerTests.java
```diff
@@ -934,6 +936,77 @@ public void testGracefulShutdown() throws Exception {
...
+		    .sendString(Flux.just("1", "2", "3"))
+		    .then()
+		    .subscribe();
+
+		CountDownLatch latch = new CountDownLatch(1);
+		conn.inbound()
+		    .receive()
+		    .asString()
+		    .doOnNext(s -> {
+		        if (s.endsWith("3")) {
```
ericbottard, line 998: Why `endsWith` and not `equals`?

src/test/java/reactor/netty/tcp/TcpClientTests.java
ericbottard: I see the server tests covers the client part, nvm

src/test/java/reactor/netty/udp/UdpServerTests.java
```diff
@@ -269,4 +270,27 @@ public void portBindingException() {
...
+		         .bindNow();
+	}
+
+	@Test(expected = IllegalArgumentException.class)
+	public void testUdpServerWithDomainSocketsWithPort() {
+		UdpServer.create()
+		         .bindAddress(() -> new DomainSocketAddress("/tmp/test.sock"))
+		         .port(1234)
+		         .bindNow();
+	}
```
ericbottard, line 295: Should there be a test for a valid server+client interaction, similar to the TCP case?

violetagg: no support for UDP protocol

src/test/java/reactor/netty/tcp/TcpServerTests.java
violetagg: There is no guarantee how the messages will be received on the server side, this depends on the buffers on network level and the flushes that will be performed.


For example here with this test
The client sends
```
16:18:28.996 [reactor-tcp-kqueue-2] DEBUG reactor.netty.tcp.TcpClient - [id: 0xe6610674, L:null - R:/tmp/test.sock] WRITE: 1B
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 31                                              |1               |
+--------+-------------------------------------------------+----------------+
16:18:28.998 [reactor-tcp-kqueue-2] DEBUG reactor.netty.tcp.TcpClient - [id: 0xe6610674, L:null - R:/tmp/test.sock] WRITE: 1B
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 32                                              |2               |
+--------+-------------------------------------------------+----------------+
16:18:28.999 [reactor-tcp-kqueue-2] DEBUG reactor.netty.tcp.TcpClient - [id: 0xe6610674, L:null - R:/tmp/test.sock] WRITE: 1B
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 33                                              |3               |
+--------+-------------------------------------------------+----------------+
```

on the server this is the incoming and outgoing data:

```
16:18:29.002 [reactor-tcp-kqueue-3] DEBUG reactor.netty.tcp.TcpServer - [id: 0x2b5371dd] READ: 3B
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 31 32 33                                        |123             |
+--------+-------------------------------------------------+----------------+
16:18:29.004 [reactor-tcp-kqueue-3] DEBUG reactor.netty.tcp.TcpServer - [id: 0x2b5371dd] WRITE: 3B
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 31 32 33                                        |123             |
+--------+-------------------------------------------------+----------------+
```

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
violetagg: let's call `toString` in all cases

src/main/java/reactor/netty/udp/UdpClientConfig.java
violetagg: thanks for spotting this -> changed

src/test/java/reactor/netty/tcp/TcpServerTests.java
violetagg: changed

violetagg:
> Tried to review as much as I could, given my limited understanding of the architecture for now.
> 
> One thing that strikes me though is the numerous use of instanceof decisions. Maybe this warrants introduction of a strategy pattern?

Sure. Please propose a change.

src/main/java/reactor/netty/resources/DefaultLoopEpoll.java
```diff
@@ -50,6 +54,12 @@
 		if (channelClass.equals(DatagramChannel.class)) {
 			return (CHANNEL) new EpollDatagramChannel();
 		}
+		if (channelClass.equals(DomainSocketChannel.class)) {
+			return (CHANNEL) new EpollDomainSocketChannel();
+		}
+		if (channelClass.equals(ServerDomainSocketChannel.class)) {
+			return (CHANNEL) new EpollServerDomainSocketChannel();
```
simonbasle, line 61: I think that is orthogonal to this PR, but I'm wondering about the throwing in both `getChannel` and `channelType`. AFAICS when producing a `ChannelFactory` in `xxxConfig#connectionFactory`, the lambda representing the factory calls `channelType`, which can throw. The result is then wrapped by a call to `getChannel`, which can also throw. But that exception comes late in the cycle, _after_ having called the configuration methods, right ? Or is the exception surfacing right away when building the client / binding the server?

I was wondering if a null-object pattern could be used instead to signal an illegal combination of channel type and transport earlier. The `channelType` would produce a special class and the factory  could check it eagerly, in order to throw from the configuration phase?

of course this is all moot if the factory MUST be lazy (ie, the channel type can be changed BETWEEN a call to `xxxConfig#connectionFactory` and the final build/bind step. This is also less of an issue if the throwing `ChannelFactory` can never make it past the build/bind step, ie the resulting client or server could never start.

src/main/java/reactor/netty/transport/TransportConfig.java
```diff
@@ -236,12 +236,33 @@ protected TransportConfig(TransportConfig parent) {
...
+	 *   <li>{@link io.netty.channel.unix.DomainSocketChannel}</li>
+	 *   <li>{@link io.netty.channel.unix.ServerDomainSocketChannel}</li>
+	 *   <li>{@link io.netty.channel.socket.DatagramChannel}</li>
+	 * </ul>
+	 *
+	 * @param isDomainSocket true if {@link io.netty.channel.unix.DomainSocketChannel} or
+	 * {@link io.netty.channel.unix.ServerDomainSocketChannel} is needed, false otherwise
+	 * @return the channel type this configuration is associated with
+	 */
+	protected abstract Class<? extends Channel> channelType(boolean isDomainSocket);
```
simonbasle, line 253: can we expect new type of sockets to be supported at some point? (see if an enum would make sense). It seems there's a lot of dimensions now: client vs server + transport protocol + isNative + isDomainSocket. Is there a simplification possible due to mutually exclusive dimensions? (eg. can we collapse the two booleans into a single enum?)

src/main/java/reactor/netty/resources/DefaultLoopEpoll.java
violetagg: By default `LoopResources#onChannel` method invokes the default implementation used by Reactor Netty.
It throws exception to signal that the default implementation does not know what to do with other class types. Of course the user can provide his/her own implementation that will be able to handle other class types.

src/main/java/reactor/netty/transport/TransportConfig.java
violetagg: I'll think about that and address on a later stage.

PR #1084, opened 2020-04-30
[build] Officially use new versioning scheme for 1.0 Europium

simonbasle:
This commit introduces the new versioning scheme, using -SNAPSHOT
instead of .BUILD-SNAPSHOT as the qualifier for snapshots.
The whole scheme is detailed in the reference documentation, same as
in core, and adaptations are made to the build script to accommodate it.

See reactor/reactor#683

### DIFF
--- a/src/docs/asciidoc/getting-started.adoc
+++ b/src/docs/asciidoc/getting-started.adoc
@@ -27,33 +27,46 @@ It has transitive dependencies on:
 * Netty v4.1.x
 
 [[getting-started-understanding-bom]]
-== Understanding the BOM
+== Understanding the BOM and versioning scheme
 
 `Reactor Netty` is part of the `Project Reactor BOM` (since the `Aluminium` release train).
 This curated list groups artifacts that are meant to work well together, providing
 the relevant versions despite potentially divergent versioning schemes in these artifacts.
 
-The `BOM` (Bill of Materials) is itself versioned, using a release train scheme
-with a codename followed by a qualifier. The following list shows a few examples:
+Note the versioning scheme has changed between 0.9.x and 1.0.x (Dysprosium and Europium).
 
-[verse]
-Aluminium-RELEASE
-Californium-BUILD-SNAPSHOT
-Aluminium-SR1
-Bismuth-RELEASE
-Californium-SR32
+Artifacts follow a versioning scheme of `MAJOR.MINOR.PATCH-QUALIFIER` while the BOM is versioned using a CalVer inspired scheme of `YYYY.MINOR.PATCH-QUALIFIER`, where:
 
-The codenames represent what would traditionally be the `MAJOR.MINOR` number. They (mostly)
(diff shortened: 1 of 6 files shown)


simonbasle:
Note: the github milestone will need to be renamed `1.0.0-M1`

simonbasle:
@violetagg should I use this PR to nuke OSGI support as well?

violetagg:
> @violetagg should I use this PR to nuke OSGI support as well?

yes

violetagg:
> Note: the github milestone will need to be renamed `1.0.0-M1`

I renamed the milestone

src/docs/asciidoc/getting-started.adoc
```diff
@@ -27,33 +27,46 @@ It has transitive dependencies on:
...
-Aluminium-RELEASE
-Californium-BUILD-SNAPSHOT
-Aluminium-SR1
-Bismuth-RELEASE
-Californium-SR32
+Artifacts follow a versioning scheme of `MAJOR.MINOR.PATCH-QUALIFIER` while the BOM is versioned using a CalVer inspired scheme of `YYYY.MINOR.PATCH-QUALIFIER`, where:
 
-The codenames represent what would traditionally be the `MAJOR.MINOR` number. They (mostly)
+ * `MAJOR` is the current generation of Reactor, where each new generation can bring fundamental changes to the structure of the project (which might imply a more significant migration effort)
+ * `YYYY` is the year of the first GA release in a given release cycle (like 3.4.0 for 3.4.x)
```
violetagg, line 41: let's use Reactor Netty version as an example and not Reactor Core i.e.

simonbasle: yeah good idea

swimmesberger:
@simonbasle why was OSGi support removed for reactor-netty and reactor-core is still officially supporting OSGi? I mean when you do not want to maintain OSGi, why are you maintaining it for reactor-core? I would expect some consistency in this library ecosystem.
We are using reactor-netty in an OSGi environment and expected a longer commitment to OSGi metadata for reactor-netty.

simonbasle:
We had already decided to move to a new versioning scheme that introduces better consistency in both Maven and Gradle, which have always been our primary targets. For OSGI, it appeared that the automatic conversion to a `Bundle-Version` wasn't going to be that great, and so the team discussed potentially dropping support for OSGI altogether vs cobbling together yet another hacky and semi-understood solution around BND (which is never simple for us in terms of maintenance). The formal decision wasn't made though, so it wasn't tracked in an issue. When BND broke here, it was the straw that broke the camel's back for me, so I removed OSGI support thinking we'd do that across the board anyway... but it fell through the cracks.

We've found a way to force a `Bundle-Version` that is consistent with what we had in the previous scheme (I think), so we'll reintroduce OSGI support for reactor-netty. Expect some version comparison issues in 3.4.0 / BOM 2020.0.0 and 2020.0.1, with a fix starting with BOM 2020.0.2

simonbasle:
created reactor/reactor#689 to track this, along #1392 for specifics in reactor-netty

PR #1094, opened 2020-05-14
When HttpServer in H2 mode, check the parent channel for SslHandler presence

violetagg:
When the server is configured to support H2, SslHandler is configured
on the connection level and not on the stream level.

A test will be provided as part of `Http2 for HttpClient` implementation.


violetagg:
fixed with 2eaf6c4a3784ed31a9d7d90c44c93e9d9b51f486

PR #1129, opened 2020-05-30
Add @sharable to ChannelHandler

IndiraFinish:
ChannelOperationsHandler is stateless, and @sharable reduces memory usage.


violetagg:
@IndiraFinish Can you elaborate more why you think it is stateless?
In order to be stateless one handler it is not enough to annotate it but also to use one and the same instance across all channels.
http://normanmaurer.me/presentations/2014-facebook-eng-netty/slides.html#30.0

violetagg:
@IndiraFinish I'm closing this. We can reopen it if you want to discuss this in more details.

PR #1147, opened 2020-06-11
Support HTTP/2 for HttpClient

### LINKED ISSUE #639: Http/2 Client Support
Are there any examples of connection to a server with http/2 support?

Thanks

violetagg:
Extended negative testing depends on https://github.com/reactor/reactor-pool/issues/82
Fixes #639

### DIFF
--- a/src/main/java/reactor/netty/http/Http2StreamBridgeHandler.java
+++ b/src/main/java/reactor/netty/http/Http2StreamBridgeHandler.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2011-Present VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty.http;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPromise;
+import io.netty.handler.codec.http.DefaultHttpContent;
+import io.netty.handler.codec.http2.Http2StreamFrameToHttpObjectCodec;
+
+import static reactor.netty.ReactorNetty.format;
+
+/**
+ * This handler is intended to work together with {@link Http2StreamFrameToHttpObjectCodec}
+ * it converts the outgoing messages into objects expected by
+ * {@link Http2StreamFrameToHttpObjectCodec}.
+ *
+ * @author Violeta Georgieva
+ * @since 1.0.0
+ */
+public class Http2StreamBridgeHandler extends ChannelDuplexHandler {
+
+	@Override
+	public void handlerAdded(ChannelHandlerContext ctx) {
+		if (HttpOperations.log.isDebugEnabled()) {
+			HttpOperations.log.debug(format(ctx.channel(), "New HTTP/2 stream"));
+		}
+	}
+
+	@Override
+	@SuppressWarnings("FutureReturnValueIgnored")
+	public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
+		if (msg instanceof ByteBuf) {
+			//"FutureReturnValueIgnored" this is deliberate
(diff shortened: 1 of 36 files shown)


src/main/java/reactor/netty/http/Http2StreamBridgeHandler.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * This handler is intended to work together with {@link Http2StreamFrameToHttpObjectCodec}
+ * it converts the outgoing messages into objects expected by
+ * {@link Http2StreamFrameToHttpObjectCodec}.
+ *
+ * @author Violeta Georgieva
+ * @since 1.0.0
+ */
+public class Http2StreamBridgeHandler extends ChannelDuplexHandler {
```
smaldini, line 35: Nit picking: Good reuse between client/server, Unfortunate it makes it in the public api tho. I wonder if separately you should think about using an "internal" subpackage for some of these or an annotation cc @rstoyanchev

src/main/java/reactor/netty/http/HttpDecoderSpec.java
```diff
@@ -131,6 +134,32 @@ public int initialBufferSize() {
 		return initialBufferSize;
 	}
 
+	/**
+	 * Configure the maximum length of the content of the H2C upgrade request.
```
smaldini, line 138: Maybe make it more clear since its a builder spec with the full name of h2c: http2ClearText

src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -229,6 +230,21 @@ public final HttpServer host(String host) {
 		return super.host(host);
 	}
 
+	/**
+	 * Apply HTTP/2 configuration
+	 *
+	 * @param http2Settings configures {@link Http2Settings} before requesting
+	 * @return a new {@link HttpServer}
+	 */
+	public final HttpServer http2Setting(Consumer<Http2Settings> http2Settings) {
```
smaldini, line 239: We should choose a constitent way to expose builders, most of the code use spec, I think this will be a unique one. WDYT

violetagg: That one in particular I want to discuss because at the moment `Http2Settings` is an API provided by Netty. Do we want to use this directly from Netty or it is better to wrap it with our API.

src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -384,18 +385,37 @@
...
+	/**
+	 * Prepare an {@link HttpClient}. {@link UriConfiguration#uri(String)} or
+	 * {@link #baseUrl(String)} should be invoked before a verb
+	 * {@link #request(HttpMethod)} is selected.
+	 *
+	 * @param connectionProvider the {@link ConnectionProvider} to be used
+	 * @param maxHttp2Connections the max number of connections that will be used for HTTP/2 requests
+	 * @return a {@link HttpClient}
+	 */
+	public static HttpClient create(ConnectionProvider connectionProvider, int maxHttp2Connections) {
```
smaldini, line 413: This might be a bit limiting - is there any other configuration that might apply later ? E.g. connection timeout etc. Should this be a Supplier of PoolProvider or similar ? or a Consumer of a spec?

src/main/java/reactor/netty/http/server/HttpServer.java
violetagg: I added `Http2SettingsSpec`

src/main/java/reactor/netty/http/client/HttpClient.java
violetagg: we will think about this in M2

src/main/java/reactor/netty/http/HttpDecoderSpec.java
violetagg: fixed

src/main/java/reactor/netty/http/Http2StreamBridgeHandler.java
violetagg: we will think about this in M2

src/main/java/reactor/netty/http/client/HttpTrafficHandler.java
```diff
@@ -0,0 +1,117 @@
...
+		else if (evt == UPGRADE_SUCCESSFUL) {
+			if (log.isDebugEnabled()) {
+				log.debug(format(channel, "The upgrade to H2C protocol was successful."));
+			}
+			sendNewState(Connection.from(channel), HttpClientState.UPGRADE_SUCCESSFUL);
+			ctx.pipeline().remove(this);
+		}
+		else if (evt == UPGRADE_REJECTED) {
+			if (log.isDebugEnabled()) {
+				log.debug(format(channel, "The upgrade to H2C protocol was rejected continue using HTTP/1.x protocol."));
```
Inego, line 99: rejected**,** continue

violetagg: Thanks
I fixed that

PR #1156, opened 2020-06-15
fix #1153 Ensure only Content-Length is set when creating full response

### LINKED ISSUE #1153: Transfer-Encoding & Content-Length headers are being sent in response headers.
When a string is sent in the response body, `GET` request has only `Content-Length` whereas  `HEAD` request has both `Transfer-Encoding` and `Content-Length`.

## Expected Behavior
Should remove `Transfer-Encoding` header for HEAD request too, as it is being done for `GET` request.

## Actual Behavior
Both `Transfer-Encoding` and `Content-Length`  are being sent.
### Worth noting
[https://tools.ietf.org/html/rfc7230#section-3.3.2](https://tools.ietf.org/html/rfc7230#section-3.3.2) 
> A sender MUST NOT send a Content-Length header field in any message
   that contains a Transfer-Encoding header field.

## Steps to Reproduce
* Controller code
```java
@GetMapping("/message")
public ResponseEntity<String> getMessage() {
    return  ResponseEntity.ok("foo");
}
```
* Result of GET API
```
curl -v -X GET http://localhost:8080/message
Note: Unnecessary use of -X or --request, GET is already inferred.
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /message HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Content-Type: text/plain;charset=UTF-8
< Content-Length: 3
<
* Connection #0 to host localhost left intact
foo* Closing connection 0
```
* Result of HEAD API
```
 curl -v -X HEAD http://localhost:8080/message
Warning: Setting custom HTTP method to HEAD with -X/--request may not work the
Warning: way you want. Consider using -I/--head instead.
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> HEAD /message HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200 OK
< transfer-encoding: chunked
< Content-Type: text/plain;charset=UTF-8
< Content-Length: 3
<
```

## Possible Solution
https://github.com/reactor/reactor-netty/blob/abd6ac4e2e8dd2d129929f9451e7276903d53564/src/main/java/reactor/netty/http/server/HttpServerOperations.java#L161
has a condition to remove transfer-encoding only if it is not a HEAD request.

## Your Environment
Environment:
spring-boot:  2.2.6.RELEASE
Using spring-boot-starter-webflux with Netty.

violetagg:
Fixes #1153

### DIFF
--- a/src/main/java/reactor/netty/http/server/HttpServerOperations.java
+++ b/src/main/java/reactor/netty/http/server/HttpServerOperations.java
@@ -158,13 +158,11 @@ class HttpServerOperations extends HttpOperations<HttpServerRequest, HttpServerR
 		HttpResponse res =
 				new DefaultFullHttpResponse(version(), status(), body);
 
-		if (!HttpMethod.HEAD.equals(method())) {
-			responseHeaders.remove(HttpHeaderNames.TRANSFER_ENCODING);
-			if (!HttpResponseStatus.NOT_MODIFIED.equals(status())) {
+		responseHeaders.remove(HttpHeaderNames.TRANSFER_ENCODING);
+		if (!HttpResponseStatus.NOT_MODIFIED.equals(status())) {
 
-				if (HttpUtil.getContentLength(nettyResponse, -1) == -1) {
-					responseHeaders.setInt(HttpHeaderNames.CONTENT_LENGTH, body.readableBytes());
-				}
+			if (HttpUtil.getContentLength(nettyResponse, -1) == -1) {
+				responseHeaders.setInt(HttpHeaderNames.CONTENT_LENGTH, body.readableBytes());
 			}
 		}
 
--- a/src/test/java/reactor/netty/http/server/HttpServerTests.java
+++ b/src/test/java/reactor/netty/http/server/HttpServerTests.java
@@ -64,6 +64,7 @@ import io.netty.handler.ssl.util.SelfSignedCertificate;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.ReferenceCounted;
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Test;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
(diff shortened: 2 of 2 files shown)


src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -554,38 +558,95 @@ public void testContentLengthHeadRequest() {
...
+		doTestContentLengthHeadRequest("/8", disposableServer.address(), HttpMethod.HEAD, sentHeaders, false, false);
+		doTestContentLengthHeadRequest("/9", disposableServer.address(), HttpMethod.HEAD, sentHeaders, false, false);
+		doTestContentLengthHeadRequest("/10", disposableServer.address(), HttpMethod.HEAD, sentHeaders, false, false);
+		doTestContentLengthHeadRequest("/11", disposableServer.address(), HttpMethod.HEAD, sentHeaders, false, false);
+		doTestContentLengthHeadRequest("/12", disposableServer.address(), HttpMethod.HEAD, sentHeaders, false, false);
+		doTestContentLengthHeadRequest("/13", disposableServer.address(), HttpMethod.HEAD, sentHeaders, false, false);
+	}
+
+	@Test
+	public void testIssue() {
```
simonbasle, line 632: that would benefit from a more descriptive name :)

violetagg: oppps :)

src/main/java/reactor/netty/http/server/HttpServerOperations.java
```diff
@@ -158,13 +158,11 @@ protected HttpMessage newFullBodyMessage(ByteBuf body) {
 		HttpResponse res =
 				new DefaultFullHttpResponse(version(), status(), body);
 
-		if (!HttpMethod.HEAD.equals(method())) {
```
kaladhar-mummadi, line 161: @violetagg , do you have a context of why this condition was written in the first place?

violetagg: it is just a bug

violetagg:
@mrbrown-09 The release is planned for Monday 20.07

mrbrown-09:
great, ty very much!

PR #1162, opened 2020-06-16
fix #1161 Clean the queue when onInboundError/onInboundClose is received

### LINKED ISSUE #1161: FluxReceive queue not cleaned in case of inbound error after all data read
Originally reported in https://github.com/spring-projects/spring-framework/issues/25216. 

In the sample provided there https://github.com/EtienneMiret/reactor-netty-oom the server closes connections at random points with `socket.setSoLinger (true, 0)`. Meanwhile the client delays the start of reading with:

```java
webClient.get ()
    .uri ("/{index}", i)
    .exchange ()
    .delayElement (Duration.ofMillis (200))
    .flatMapMany (response -> ...)
```

So in some cases `FluxReceive#onInboundError` is called before a subscription and when `inboundDone` is already true, which simply exits because it expects the cancellation or the completion to have done it.

violetagg:
There are two use cases for HttpClient:
- a channelInactive is received, the response was fully received but there is no subscriber
- a error is receive while receiving the response and there is no subscriber

### DIFF
--- a/src/main/java/reactor/netty/channel/FluxReceive.java
+++ b/src/main/java/reactor/netty/channel/FluxReceive.java
@@ -178,8 +178,11 @@ final class FluxReceive extends Flux<Object> implements Subscription, Disposable
 					cleanQueue(q);
 					return;
 				}
+				Throwable ex = inboundError;
+				if (ex != null) {
+					cleanQueue(q);
+				}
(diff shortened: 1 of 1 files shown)


violetagg:
Closing as this is not the correct fix for the issue

PR #1172, opened 2020-06-22
fixes memory-leaks when onNext races with cancellation signal

OlegDokuka:
This PR fixes memory leaks which might appear because of racing between onNext and cancel signals in case of the `MonoSend` operator.

Please note:

1. It is impossible to guarantee exactly-once releasing since we have `SpScQueue` which can not be drained by multiple threads. Hence if we have `onInterruptionCleanup` racing with `onNext` the double releasing might have a place. To prevent that `CONSUMER_NOCHECK_CLEANUP` and `CONSUMER_BB_NOCHECK_CLEANUP` are instrumented with extra try-ignore statements

2. In order to reduce objects allocations per `MonoSendMany`, this class extends the `Context` interface and provides extra the `onDiscard` consumer implemented in the same class for the same reason. This done exactly for fusion case, when it is unsafe to do draining of `Queue` as usually (again, because SpSc characteristics) so we delegate that operation to the `Queue.clear` since all fuseable must guard that method according to the internal `Queue` behavior (for more info about guarding see the following https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/UnicastProcessor.java#L503). It leads to the fact that the only way to handle and release discarded messages in that cases is to offer own onDiscardHook which implement the same business logic as for the non-fusion case. 

### DIFF
--- a/src/main/java/reactor/netty/channel/MonoSend.java
+++ b/src/main/java/reactor/netty/channel/MonoSend.java
@@ -25,6 +25,7 @@ import io.netty.buffer.ByteBufHolder;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.FileRegion;
+import io.netty.util.IllegalReferenceCountException;
 import io.netty.util.ReferenceCountUtil;
 import reactor.core.publisher.Mono;
 import reactor.netty.ReactorNetty;
@@ -70,9 +71,24 @@ abstract class MonoSend<I, O> extends Mono<Void> {
 
 	static final Function<Object, Object>   TRANSFORMATION_FUNCTION    = Function.identity();
 
-	static final Consumer<ByteBuf> CONSUMER_BB_NOCHECK_CLEANUP = ByteBuf::release;
+	static final Consumer<ByteBuf> CONSUMER_BB_NOCHECK_CLEANUP = buf -> {
+		try {
+			buf.release();
+		}
+		catch (IllegalReferenceCountException e) {
+			// ignored
+		}
+	};
 
-	static final Consumer<Object>  CONSUMER_NOCHECK_CLEANUP    = ReferenceCountUtil::release;
+	static final Consumer<Object>  CONSUMER_NOCHECK_CLEANUP    =
+			msg -> {
+				try {
+					ReferenceCountUtil.release(msg);
+				}
+				catch (IllegalReferenceCountException e) {
+					// ignored
+				}
+			};
 
 	static final ToIntFunction<ByteBuf> SIZE_OF_BB  = ByteBuf::readableBytes;
 
--- a/src/main/java/reactor/netty/channel/MonoSend.java
+++ b/src/main/java/reactor/netty/channel/MonoSend.java
@@ -25,7 +25,6 @@ import io.netty.buffer.ByteBufHolder;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.FileRegion;
-import io.netty.util.IllegalReferenceCountException;
 import io.netty.util.ReferenceCountUtil;
 import reactor.core.publisher.Mono;
 import reactor.netty.ReactorNetty;
@@ -71,24 +70,9 @@ abstract class MonoSend<I, O> extends Mono<Void> {
 
 	static final Function<Object, Object>   TRANSFORMATION_FUNCTION    = Function.identity();
 
-	static final Consumer<ByteBuf> CONSUMER_BB_NOCHECK_CLEANUP = buf -> {
-		try {
-			buf.release();
-		}
-		catch (IllegalReferenceCountException e) {
-			// ignored
-		}
-	};
+	static final Consumer<ByteBuf> CONSUMER_BB_NOCHECK_CLEANUP = ByteBuf::release;
 
-	static final Consumer<Object>  CONSUMER_NOCHECK_CLEANUP    =
-			msg -> {
-				try {
-					ReferenceCountUtil.release(msg);
-				}
-				catch (IllegalReferenceCountException e) {
-					// ignored
-				}
-			};
+	static final Consumer<Object>  CONSUMER_NOCHECK_CLEANUP    = ReferenceCountUtil::release;
 
 	static final ToIntFunction<ByteBuf> SIZE_OF_BB  = ByteBuf::readableBytes;
 
--- a/src/main/java/reactor/netty/channel/MonoSendMany.java
+++ b/src/main/java/reactor/netty/channel/MonoSendMany.java
@@ -17,6 +17,7 @@
 package reactor.netty.channel;
 
 import java.nio.channels.ClosedChannelException;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Queue;
 import java.util.concurrent.TimeUnit;
@@ -26,6 +27,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.ToIntFunction;
+import java.util.stream.Stream;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufHolder;
@@ -51,6 +53,8 @@ import reactor.util.concurrent.Queues;
 import reactor.util.context.Context;
 
 final class MonoSendMany<I, O> extends MonoSend<I, O> implements Scannable {
+
+	static final String KEY_ON_DISCARD = "reactor.onDiscard.local";
 
 	static MonoSendMany<ByteBuf, ByteBuf> byteBufSource(Publisher<? extends ByteBuf> source,
 			Channel channel,
@@ -83,14 +87,14 @@ final class MonoSendMany<I, O> extends MonoSend<I, O> implements Scannable {
 
 	@Override
 	@Nullable
-	@SuppressWarnings("rawtypes")
 	public Object scanUnsafe(Attr key) {
 		if (key == Attr.PREFETCH) return MAX_SIZE;
 		if (key == Attr.PARENT) return source;
 		return null;
 	}
 
-	static final class SendManyInner<I, O> implements CoreSubscriber<I>, Subscription, Fuseable,
+	static final class SendManyInner<I, O> implements CoreSubscriber<I>, Subscription,
+	                                                  Fuseable, Context, Consumer<I>,
 	                                                  ChannelFutureListener, Runnable, Scannable, ChannelPromise {
 
 		final ChannelHandlerContext        ctx;
@@ -132,7 +136,7 @@ final class MonoSendMany<I, O> extends MonoSend<I, O> implements Scannable {
 
 		@Override
 		public Context currentContext() {
-			return actual.currentContext();
+			return this;
 		}
 
 		@Override
@@ -176,14 +180,13 @@ final class MonoSendMany<I, O> extends MonoSend<I, O> implements Scannable {
 				return;
 			}
 
-			if (terminalSignal != null) {
+			if (terminalSignal != null || this.s == Operators.cancelledSubscription()) {
 				parent.sourceCleanup.accept(t);
 				Operators.onDiscard(t, actual.currentContext());
 				return;
 			}
 
 //			ReferenceCountUtil.touch(t);
-			//FIXME check cancel race
 			if (!queue.offer(t)) {
 				onError(Operators.onOperatorError(s,
 						Exceptions.failWithOverflow(Exceptions.BACKPRESSURE_ERROR_QUEUE_FULL),
@@ -201,7 +204,7 @@ final class MonoSendMany<I, O> extends MonoSend<I, O> implements Scannable {
 					@SuppressWarnings("unchecked") QueueSubscription<I> f =
 							(QueueSubscription<I>) s;
 
-					int m = f.requestFusion(Fuseable.ANY/* | Fuseable.THREAD_BARRIER*/);
+					int m = f.requestFusion(Fuseable.ANY | Fuseable.THREAD_BARRIER);
 
 					if (m == Fuseable.SYNC) {
 						sourceMode = Fuseable.SYNC;
@@ -358,15 +361,23 @@ final class MonoSendMany<I, O> extends MonoSend<I, O> implements Scannable {
 			if (queue == null) {
 				return;
 			}
-			Context context = null;
-			while (!queue.isEmpty()) {
-				I sourceMessage = queue.poll();
-				if (sourceMessage != null) {
-					parent.sourceCleanup.accept(sourceMessage);
-					if (context == null) {
-						context = actual.currentContext();
+
+			if (sourceMode == ASYNC) {
+				// delegates discarding to the queue holder to ensure there is no racing on draining from the SpScQueue
+				// fused queue should discard elements on clear
+				queue.clear();
+			}
+			else {
+				Context context = null;
+				while (!queue.isEmpty()) {
+					I sourceMessage = queue.poll();
+					if (sourceMessage != null) {
+						parent.sourceCleanup.accept(sourceMessage);
+						if (context == null) {
+							context = actual.currentContext();
+						}
+						Operators.onDiscard(sourceMessage, context);
 					}
-					Operators.onDiscard(sourceMessage, context);
 				}
 			}
 		}
@@ -375,7 +386,7 @@ final class MonoSendMany<I, O> extends MonoSend<I, O> implements Scannable {
 			return pending == 0 && terminalSignal != null;
 		}
 
-		void trySchedule(@Nullable Object data) {
+		void trySchedule(@Nullable I data) {
 			if (WIP.getAndIncrement(this) == 0) {
 				try {
 					if (eventLoop.inEventLoop()) {
@@ -391,10 +402,24 @@ final class MonoSendMany<I, O> extends MonoSend<I, O> implements Scannable {
 					}
 				}
 			}
+			else {
+				if (this.s == Operators.cancelledSubscription()) {
+					if (sourceMode == ASYNC) {
+						// delegates discarding to the queue holder to ensure there is no racing on draining from the SpScQueue
+						queue.clear();
+					}
+					else {
+						// discard given dataSignal since no more is enqueued (spec guarantees serialised onXXX calls)
+						if (data != null) {
+							parent.sourceCleanup.accept(data);
+							Operators.onDiscard(data, actual.currentContext());
+						}
+					}
+				}
+			}
 		}
 
 		@Override
-		@SuppressWarnings("rawtypes")
 		public Object scanUnsafe(Attr key) {
 			if (key == Attr.PARENT) return s;
 			if (key == Attr.ACTUAL) return actual;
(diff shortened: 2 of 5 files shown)


src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -201,7 +204,7 @@ public void onSubscribe(Subscription s) {
 					@SuppressWarnings("unchecked") QueueSubscription<I> f =
 							(QueueSubscription<I>) s;
 
-					int m = f.requestFusion(Fuseable.ANY/* | Fuseable.THREAD_BARRIER*/);
+					int m = f.requestFusion(Fuseable.ANY | Fuseable.THREAD_BARRIER);
```
OlegDokuka, line 207: Thread barrier should be here since if we have `publishOn` on the upstream, this will break fusion (we are rescheduling on the `eventLoop` and that might lead to incorrect fusion model and subsequent leaks since two threads can clean queue again)

src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -52,6 +54,8 @@
 
 final class MonoSendMany<I, O> extends MonoSend<I, O> implements Scannable {
 
+	static final String KEY_ON_DISCARD = "reactor.onDiscard.local";
```
violetagg, line 57: Can we name this `reactor.netty.onDiscard.local`

src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -618,6 +643,64 @@ public Void get(long timeout, TimeUnit unit) {
...
+			if (KEY_ON_DISCARD.equals(key)) {
+				return (T) this;
+			}
+
+			return actual.currentContext().get(key);
+		}
+
+		@Override
+		public boolean hasKey(Object key) {
+			if (KEY_ON_DISCARD.equals(key)) {
```
violetagg, line 667: Can we use reference equality

src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -618,6 +643,64 @@ public Void get(long timeout, TimeUnit unit) {
...
+			parent.sourceCleanup.accept(i);
+			// propagates discard to the downstream
+			Operators.onDiscard(i, actual.currentContext());
+		}
+
+		// Context interface impl
+		@Override
+		@SuppressWarnings("unchecked")
+		public <T> T get(Object key) {
+			if (KEY_ON_DISCARD.equals(key)) {
```
violetagg, line 658: Can we use reference equality

src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -618,6 +643,64 @@ public Void get(long timeout, TimeUnit unit) {
...
+		public void accept(I i) {
+			parent.sourceCleanup.accept(i);
+			// propagates discard to the downstream
+			Operators.onDiscard(i, actual.currentContext());
+		}
+
+		// Context interface impl
+		@Override
+		@SuppressWarnings("unchecked")
+		public <T> T get(Object key) {
```
violetagg, line 657: Please resolve this warning

```
MonoSendMany.java:658: warning: [TypeParameterUnusedInFormals] Declaring a type parameter that is only used in the return type is a misuse of generics: operations on the type parameter are unchecked, it hides unsafe casts at invocations of the method, and it interacts badly with method overload resolution.
                public <T> T get(Object key) {
                             ^
    (see https://errorprone.info/bugpattern/TypeParameterUnusedInFormals)
```

src/test/java/reactor/netty/channel/MonoSendManyTest.java
```diff
@@ -16,18 +16,35 @@
...
 
 import java.lang.ref.WeakReference;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.IdentityHashMap;
 import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedDeque;
```
violetagg, line 26: Remove unused imports

src/main/java/reactor/netty/channel/MonoSendMany.java
OlegDokuka: I guess not. What I need here is access to package-private reactor discard hook since in case of fusion This operator delegates discarding from the queue to the upstream. It means that for example https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/UnicastProcessor.java#L503 unicastProcessor.clear will be called and it will be discarding elements. In order to ensure their release, we have to intercept that and mimic `onDiscard` acceptor.

As you can see, we delegate discarding to the downstream subscriber. Here is the question which we should think about - is there anybody downstream to handle discarded elements? Should we that onDiscard propagation to downstream at all like at the https://github.com/reactor/reactor-netty/pull/1172/files#diff-34c9e1ba9859e522329366ebbb5ffb56R379. I mean who can be interested in released `ByteBuf`? If we remove that further delegation we can use standard functionality and create a static context with onDiscard hook to offer it to the upstream.

violetagg: So at some point if this `reactor.onDiscard.local` is changed/removed, the functionality will stop working. (this is not a public API)

src/main/java/reactor/netty/channel/MonoSendMany.java
OlegDokuka: Not in that case. The reference will be different https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/Hooks.java#L602 but we need to return an acceptor in case of the same content

violetagg: yep

src/main/java/reactor/netty/channel/MonoSendMany.java
OlegDokuka: Correct. One of the options is to hack it is:

```java

static final Object ON_DISCARD_HOOK;

static {
	Context c = reactor.core.publisher.Operators#enableOnDiscard(null, (v) -> {});
    
    Map.Entry<Object,Object> entry = c.stream().findFirst().orElse(null);

    if (entry != null) {
       ON_DISCARD_HOOK = entry.getKey();
    }
    else {
       ON_DISCARD_HOOK = null;
    }
}
```

src/main/java/reactor/netty/channel/MonoSendMany.java
simonbasle: @OlegDokuka you could remove the `KEY_ON_DISCARD` by ensuring `currentContext` doesn't return `this` but `Operators.enableOnDiscard(actual.currentContext(), this)`, which would simplify things IMHO. The class can continue implementing `Consumer<?>` as the discarding hook.

OlegDokuka: This is object overhead. The goal is to avoid this overhead. I can use `enabledOnDiscard` if we agree to not propagate released buffers do downstream hooks

src/main/java/reactor/netty/channel/MonoSendMany.java
OlegDokuka: Cc / @simonbasle

violetagg:
@smaldini PTAL

src/main/java/reactor/netty/channel/MonoSendMany.java
simonbasle: but on the other hand you're effectively creating a new `Context` instance every time you invoke `put`, `hasKey`, `delete`, `size` or `stream`... might as well pay that price upfront, which would additionally protect from the case where a new `Context` API is introduced with default method and you miss the `KEY_ON_DISCARD` in it.

OlegDokuka: > which would additionally protect from the case where a new Context API is introduced with default method and you miss the KEY_ON_DISCARD in it.

https://github.com/reactor/reactor-netty/pull/1172#discussion_r445528928
Here is the hack. 


> but on the other hand you're effectively creating a new Context instance every time you invoke put, hasKey, delete, 

The probability of one using context is significantly lower compares to normal usage of `MonoSend`. Unless the spring framework does it somewhere above like  

```
outbound.send(
	flux.doOnDiscard()
)
```

and not 

```
outbound.send(flux)
        .doOnDiscard()
```

violetagg: There is no direct `doOnDiscard` for NettyOutbound

`NettyOutbound extends Publisher<Void>`

OlegDokuka: Then. It makes no sense to support `Operator.onDiscard(v, actual.currentContext()` since `actual.currenctContext()` will always return empty context. Is not that true?

OlegDokuka: if that is true, then I can simplify my changes

src/main/java/reactor/netty/channel/MonoSendMany.java
smaldini: I commented that one specifically as it reduces a lot of the fusion scope unfortunately and I couldn't see at the time the impact. But that seems to be on a the safer side yes.

src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -176,14 +180,13 @@ public void onNext(I t) {
 				return;
 			}
 
-			if (terminalSignal != null) {
+			if (terminalSignal != null || this.s == Operators.cancelledSubscription()) {
```
smaldini, line 183: Why ? shouldn't cancellation be checked in the drainLoop for safety anyway ?

src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -618,6 +643,64 @@ public Void get(long timeout, TimeUnit unit) {
...
+			             .size();
+		}
+
+		@Override
+		public Stream<Map.Entry<Object, Object>> stream() {
+			return actual.currentContext()
+			             .put(KEY_ON_DISCARD, this)
+			             .stream();
+		}
+
```
smaldini, line 703: The overall put context is a bit convoluted although i understand why you inline Context. Size could be simplified (`size + 1`), stream too (concat),  but for delete and put we might just need to add a putAll or a multi put (varargs ?) in reactor. Other solutions are probably more complicated, but its giving some good idea for Context support like should we support lazy merge to avoid looking up currentContext.

smaldini:
Overall its good finding and mindful fix regarding efficiency (inlining context). 
Initially, I was wondering if it was possible to just consume the source to discard its data but that might cause other problems and will still be exposed to race conditions. It's really an annoying problem and I wish there was an extension to Reactive Streams spec that allowed circular cancellation (source would error with a specific signal or complete once it finishes cleanup, allowing child to gracefully cleanup)

simonbasle:
> extension to Reactive Streams spec

Yeah between cancellation, discarding and `Context` in general I'm thinking more and more about proposing an extension to the spec (probably as a secondary spec for resource management or something...). I'll keep that idea in mind.

src/main/java/reactor/netty/channel/MonoSendMany.java
OlegDokuka: The answer is - if we will not do that: 

The element goes into a queue - and then will be discarded later in else branch of trySchedule since WIP will be 1. Why 1 - because by the moment, drainLoop can exit, discarded what was seen in the queue, and simply return leaving WIP == 1. (yeah, this is basically the issue which leaves some numbers of elements undiscarded). Thus, the question is - do we wanna discard element earlier, or do we wanna have enqueued with volatile write and then 2 volatile reads in order to say - we have to discard?

src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -391,10 +402,24 @@ void trySchedule(@Nullable Object data) {
 					}
 				}
 			}
+			else {
+				if (this.s == Operators.cancelledSubscription()) {
```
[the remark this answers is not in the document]
OlegDokuka: because we will never go in the loop if loop has already exit seeing cancel. This is a racing case which causes memory leaks. The same problems were detected in FluxPublishOn, UnicastProcessor, FluxFlatMap, FLuxConcatMap and the only solution to ensure all the elements are discarded either - use SpMcQueue or do those checks

OlegDokuka:
> note, between cancellation, discarding and `Context` in general I'm thinking more and more about proposing an extension to the spec (probably as a secondary spec for resource management or something...). I'll keep that idea in mind.

Well, reactive-streams-contrib waits for us. QueueSubscription, ConditionalSubscriber, Context - all that awesome stuff should be standardized :D

src/main/java/reactor/netty/channel/MonoSendMany.java
OlegDokuka: My question is - does any downstream supply a context at all? Is it useful to do `Operators.onDiscrd(releasedByteBuf, actual.currectContext())`.

The main question - which subscriber is interested in ByteBuf memory of which is already freed? What one should do with such? We already take care of releasing. Also, we are the last handler in the chain, so should one do anything with such messages after all? (except such messages are Strings)

src/main/java/reactor/netty/channel/MonoSendMany.java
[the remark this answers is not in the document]
OlegDokuka: of we will not check for cancel, then it will not be possible to guess at that point whether it is somebody doing work or it is 1 because it has already exited. In theory, I can use WIP field to do that check (e.g. if wip is negative == hurray, the stream is canceled - let's do a discard right immediately)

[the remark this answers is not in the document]
OlegDokuka: if that question is optimizations matter - I can experiment with WIP as a cancel state, so we will get one less volatile read but may get few more volatile tries to writes in the drainLoop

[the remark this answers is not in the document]
OlegDokuka: In theory that should be a good win.

src/main/java/reactor/netty/channel/MonoSend.java
```diff
@@ -70,9 +71,24 @@
...
 
-	static final Consumer<Object>  CONSUMER_NOCHECK_CLEANUP    = ReferenceCountUtil::release;
+	static final Consumer<Object>  CONSUMER_NOCHECK_CLEANUP    =
+			msg -> {
+				try {
+					ReferenceCountUtil.release(msg);
+				}
+				catch (IllegalReferenceCountException e) {
+					// ignored
+				}
```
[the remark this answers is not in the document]
OlegDokuka: ✅

src/main/java/reactor/netty/channel/MonoSend.java
```diff
@@ -70,9 +71,24 @@
...
 	static final Function<Object, Object>   TRANSFORMATION_FUNCTION    = Function.identity();
 
-	static final Consumer<ByteBuf> CONSUMER_BB_NOCHECK_CLEANUP = ByteBuf::release;
+	static final Consumer<ByteBuf> CONSUMER_BB_NOCHECK_CLEANUP = buf -> {
+		try {
+			buf.release();
+		}
+		catch (IllegalReferenceCountException e) {
+			// ignored
+		}
```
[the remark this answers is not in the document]
OlegDokuka: it is package-private. Not accessible at the MonoSend Level

src/main/java/reactor/netty/channel/MonoSendMany.java
OlegDokuka: fixed

src/main/java/reactor/netty/channel/MonoSendMany.java
OlegDokuka: should be fixed

OlegDokuka:
Just FYI. The `contextShouldBeTransferredFromDownStreamToUpStream` test is flacky when pooled connection is used.

The problem appears when the pooled `Channel` is reused and the `OWNER` property is set to `emptyListener` -> https://github.com/reactor/reactor-netty/blob/41e73261a5935dac3b5773ab3668d0dd99d4cf42/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java#L393

So once the context is resolved in https://github.com/reactor/reactor-netty/blob/41e73261a5935dac3b5773ab3668d0dd99d4cf42/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java#L364

the `owner()` of in `channel.attr` is EmptyListener instance

violetagg:
@OlegDokuka Is this ready for a review? Thanks

violetagg:
@smaldini @simonbasle PTAL
Please note that the usage of UnicastProcessor is OK as I'm going to introduce this in 0.9.x and then forward merge to master.

src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -358,38 +383,48 @@ void onInterruptionCleanup() {
...
-				try {
-					if (eventLoop.inEventLoop()) {
-						run();
-						return;
+		void trySchedule() {
+			int wip = wipIncrement(WIP, this);
+			if (wip != 0) {
+				if (wip == Integer.MIN_VALUE) {
+					if (sourceMode == ASYNC) {
+						queue.clear();
```
simonbasle, line 407: can you expand on `clear()` vs discard-draining the queue in ASYNC mode ?

src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -618,6 +654,98 @@ public Void get(long timeout, TimeUnit unit) {
...
+			}
+			catch (IllegalReferenceCountException e) {
+				// FIXME: should be removed once fusion is fixed in reactor-core
+				//        for now we have double releasing issue
+			}
+			// propagates discard to the downstream
+			Operators.onDiscard(i, actual.currentContext());
+		}
+
+		// Context interface impl
```
simonbasle, line 671: this seems to be all done to avoid overhead on `get`, but now we potentially create a `Context` object _per write operation_ and _per iteration_. Plus the implementation is convoluted. Wouldn't it be saner to instantiate a `Context thisContext = actual.currentContext().put(KEY_DISCARD, this)` once at construction ?

src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -794,6 +795,7 @@ public void testConnectionCloseOnServerError() {
 		FutureMono.from(ch.get().closeFuture()).block(Duration.ofSeconds(30));
 	}
 
+	@Ignore
```
simonbasle, line 798: isn't that a pretty important feature @violetagg @OlegDokuka @smaldini ?

OlegDokuka: #1208 fixes that. But it is a separate PR, so this one goes with the ignored test

src/main/java/reactor/netty/channel/MonoSendMany.java
OlegDokuka: I don't think so. Having `actual.currentContext().put(KEY_DISCARD, this)` per operation becomes common and sometimes redundant (since the normal usage without discarding is a general case whilst discarding and releasing is a small percentage of cases. So now if we do that all the time we create a new object which decreases performance significantly (usually it is around 10% of normal throughput).

On the other hand, we have this implementation of the context which is not harmful but improves performance

src/main/java/reactor/netty/channel/MonoSendMany.java
OlegDokuka: We have that pattern in a reactor where the queue owner discards messages over `queue.clear()`. In the case of fusion, queue is a proxy subscription that refers to another queue in, e.g. UnicastProcessor or publishOn.

OlegDokuka: https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/UnicastProcessor.java#L503

simonbasle: ah sorry of course. `queue` is the `QueueSubscription` 🤦

src/main/java/reactor/netty/channel/MonoSendMany.java
simonbasle: I'm not sure I followed all your argument. If discarding is a small percentage of cases, then it must mean that there are a lot of other cases where data is put in the `Context` other that discard hook. So there's a high chance something puts data in the `Context` upstream of this operator, and thus calls `MonoSendMany#put` and thus creates up to TWO intermediary `Context` objects ?

This implementation "does something redundant all the time" (in write operations).
Yes, granted, read operations are guaranteed to happen more frequently that writes.

Another issue is that every single operation on the `MonoSendMany#currentContext` will now invoke `actual.currentContext()`.

If your concern is about code that never calls `currentContext`, then have a `volatile Context` field that gets lazily initialized upon first call to `currentContext()`, wdyt ?

OlegDokuka: I'm saying that `actual.currentContext().put(KEY_DISCARD, this)` allocates an object. If that object is stored as a field - the long-lived object will be produced, so GC will not be able to easily eliminate it.

Also, usually, context modified at the very bottom, not at the top. So it is likely that `actual.currentContext()` has values rather than later will be `MonoSendMany#put` added something.

> and thus calls MonoSendMany#put and thus creates up to TWO intermediary Context objects?

In contrast to long-lived objects, objects allocated within a stack is nothing for GC since JIT is smart enough to escape redundant objects, especially when they are immutable, so this will have at the end only a single `Context` which is identical to how it was before. 

So, the main difference of having `context` field and implementing `Context` interface is in the fact that one allocation will be avoided and no extra longlived objects will be produced.

simonbasle: ok yeah, the argument that this is upstream is a good one 👍

violetagg:
@OlegDokuka I squashed and committed the change to 0.9.x branch (7632da1), then forward merge to master.
Thanks for the PR.

PR #1176, opened 2020-06-23
Update javadoc with the version when the deprecated functionality will be removed

violetagg:


violetagg:
@rstoyanchev @simonbasle @mp911de PTAL

Issue #1092, opened 2020-05-13
reactor.netty.http.client.PrematureCloseException: Connection prematurely closed BEFORE response


yantaowu:
I use spring cloud gateway as our gateway, this exception occurs in our production environment randomly when the request arrived.  I search the related issue on the github and found it had been fixed, but my reactor-netty is the latest release(0.9.7). 

## Steps to Reproduce
It is very difficult to reproduce，which only occurs in production environment randomly.
This is the log when the exception occurs:

2020-05-12 17:31:22.948 DEBUG [reactor-http-epoll-11] [PooledConnectionProvider] [id: 0xcbc4c890, L:/172.18.206.6:23016 - R:172.18.207.3/172.18.207.3:8151] Channel acquired, now 1 active connections and 0 inactive connections
2020-05-12 17:31:22.948 DEBUG [reactor-http-epoll-11] [HttpClientConnect] [id: 0xcbc4c890, L:/172.18.206.6:23016 - R:172.18.207.3/172.18.207.3:8151] Handler is being applied: {uri=http://172.18.207.3:8151/route, method=POST}
2020-05-12 17:31:22.948 DEBUG [reactor-http-epoll-11] [PooledConnectionProvider] [id: 0xcbc4c890, L:/172.18.206.6:23016 - R:172.18.207.3/172.18.207.3:8151] onStateChange(POST{uri=/route, connection=PooledConnection{channel=[id: 0xcbc4c890, L:/172.18.206.6:23016 - R:172.18.207.3/172.18.207.3:8151]}}, [request_prepared])
2020-05-12 17:31:22.948 DEBUG [reactor-http-epoll-11] [PooledConnectionProvider] [id: 0xcbc4c890, L:/172.18.206.6:23016 - R:172.18.207.3/172.18.207.3:8151] onStateChange(POST{uri=/route, connection=PooledConnection{channel=[id: 0xcbc4c890, L:/172.18.206.6:23016 - R:172.18.207.3/172.18.207.3:8151]}}, [request_sent])
2020-05-12 17:31:22.948 DEBUG [reactor-http-epoll-11] [PooledConnectionProvider] [id: 0xcbc4c890, L:/172.18.206.6:23016 ! R:172.18.207.3/172.18.207.3:8151] Channel closed, now 0 active connections and 0 inactive connections
[...]

PR #1183, opened 2020-06-25
Retry the request once if the connection was closed before sending the actual request

violetagg:
With this change the request will be retried if for some reason the connection close event
appears after acquiring the connection from the pool and before sending the actual request.

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClientOperations.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientOperations.java
@@ -267,7 +267,10 @@ class HttpClientOperations extends HttpOperations<NettyInbound, NettyOutbound>
 		}
 		listener().onStateChange(this, HttpClientState.RESPONSE_INCOMPLETE);
 		if (responseState == null) {
-			if (markSentBody()) {
+			if (markSentHeaderAndBody()) {
+				listener().onUncaughtException(this, AbortedException.beforeSend());
+			}
+			else if (markSentBody()) {
 				listener().onUncaughtException(this, new PrematureCloseException("Connection has been closed BEFORE response, while sending request body"));
 			}
 			else {


violetagg:
@simonbasle @rstoyanchev @smaldini PTAL

Please ignore GitHub Actions failures.

violetagg:
> can this be demonstrated in a test?

No, it is not a reliable execution.

lovingcodesk:
> > can this be demonstrated in a test?
> 
> No, it is not a reliable execution.
> This fix is based on a code review because of the recent issues that happen from time to time. For example [#1092 (comment)](https://github.com/reactor/reactor-netty/issues/1092#issuecomment-648651826)

could you tell me how does it retry for some reason the connection close events,i just only see you threw a AbortedExcetion,Where does the code catch this exception and retry, thank you ,we need to solve this problem, but we don't want to upgrade version

violetagg:
> > > can this be demonstrated in a test?
> > 
> > 
> > No, it is not a reliable execution.
> > This fix is based on a code review because of the recent issues that happen from time to time. For example [#1092 (comment)](https://github.com/reactor/reactor-netty/issues/1092#issuecomment-648651826)
> 
> could you tell me how does it retry for some reason the connection close events,i just only see you threw a AbortedExcetion,Where does the code catch this exception and retry, thank you ,we need to solve this problem, but we don't want to upgrade version

Here it is throwing the exception
https://github.com/reactor/reactor-netty/blob/03c22c8fe0c29a204de298f0c16b852d9c7d223a/src/main/java/reactor/netty/http/client/HttpClientOperations.java#L281

Here it is propagated
https://github.com/reactor/reactor-netty/blob/03c22c8fe0c29a204de298f0c16b852d9c7d223a/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L407

and here is the actual test for retry
https://github.com/reactor/reactor-netty/blob/03c22c8fe0c29a204de298f0c16b852d9c7d223a/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L708

lovingcodesk:
> > > > can this be demonstrated in a test?
> > > 
> > > 
> > > No, it is not a reliable execution.
> > > This fix is based on a code review because of the recent issues that happen from time to time. For example [#1092 (comment)](https://github.com/reactor/reactor-netty/issues/1092#issuecomment-648651826)
> > 
> > 
> > could you tell me how does it retry for some reason the connection close events,i just only see you threw a AbortedExcetion,Where does the code catch this exception and retry, thank you ,we need to solve this problem, but we don't want to upgrade version
> 
> Here it is throwing the exception
> 
> https://github.com/reactor/reactor-netty/blob/03c22c8fe0c29a204de298f0c16b852d9c7d223a/src/main/java/reactor/netty/http/client/HttpClientOperations.java#L281
> 
> Here it is propagated
> 
> https://github.com/reactor/reactor-netty/blob/03c22c8fe0c29a204de298f0c16b852d9c7d223a/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L407
> 
> and here is the actual test for retry
> 
> https://github.com/reactor/reactor-netty/blob/03c22c8fe0c29a204de298f0c16b852d9c7d223a/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L708

thank you very much

lovingcodesk:
> > > > can this be demonstrated in a test?
> > > 
> > > 
> > > No, it is not a reliable execution.
> > > This fix is based on a code review because of the recent issues that happen from time to time. For example [#1092 (comment)](https://github.com/reactor/reactor-netty/issues/1092#issuecomment-648651826)
> > 
> > 
> > could you tell me how does it retry for some reason the connection close events,i just only see you threw a AbortedExcetion,Where does the code catch this exception and retry, thank you ,we need to solve this problem, but we don't want to upgrade version
> 
> Here it is throwing the exception
> 
> https://github.com/reactor/reactor-netty/blob/03c22c8fe0c29a204de298f0c16b852d9c7d223a/src/main/java/reactor/netty/http/client/HttpClientOperations.java#L281
> 
> Here it is propagated
> 
> https://github.com/reactor/reactor-netty/blob/03c22c8fe0c29a204de298f0c16b852d9c7d223a/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L407
> 
> and here is the actual test for retry
> 
> https://github.com/reactor/reactor-netty/blob/03c22c8fe0c29a204de298f0c16b852d9c7d223a/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L708

I have a question,Why only throw errors after mark

violetagg:
> I have a question,Why only throw errors after markSentHeaderAndBody ，why not markSentBody or marksentheader，could you tell me the reason,thank you

Because fro Reactor Netty point of view it is safe to retry only in case the headers are not sent.
Reactor Netty cannot decide whether retry is OK in other cases, this can be done only by the Reactor Netty's users as they know whether a particular request can be retried in case headers and/or body were already sent.

lovingcodesk:
> > I have a question,Why only throw errors after markSentHeaderAndBody ，why not markSentBody or marksentheader，could you tell me the reason,thank you
> 
> Because from Reactor Netty point of view it is safe to retry only in case the headers are not sent.
> Reactor Netty cannot decide whether retry is OK in other cases, this can be done only by the Reactor Netty's users as they know whether a particular request can be retried in case headers and/or body were already sent.

I want to contribute to this project,Can I only submit code to the latest branch？or I can submit to some specified branch for example 0.9.10 ？

violetagg:
@lovingcodesk We do not produces patches for a particular version. For example if there is an issue in 0.9.10 the fix will be available in the next scheduled release for 0.9.x version.
For PRs if the issue is in 0.9.x version it will be good if you can create the patch against 0.9.x branch, but this is not a must. If it is more convenient for you to work against master then do the change against master and we will back port it to 0.9.x version.

lovingcodesk:
> @lovingcodesk We do not produces patches for a particular version. For example if there is an issue in 0.9.10 the fix will be available in the next scheduled release for 0.9.x version.
> For PRs if the issue is in 0.9.x version it will be good if you can create the patch against 0.9.x branch, but this is not a must. If it is more convenient for you to work against master then do the change against master and we will back port it to 0.9.x version.

ok

PR #1185, opened 2020-06-26
Enhances FluxReceiver internals

OlegDokuka:
This PR fixes the improper behavior of FluxReceiver when subscription sending might be offloaded to a separate thread (which contradicts to the ReactiveStreams Spec, for more info see -> https://github.com/reactive-streams/reactive-streams-jvm/issues/486#issuecomment-638797014)

Also, since all actions of the `FluxReceive` class (includes `drainReceiver`, `cancelReceiver`, `cleanQueue`, `onInboundXXX`) happens within an `EventLoop`, it makes no sense to use volatile `WIP` which brings extra overhead. The same applies to `SpScUnboundedQueue` which can be replaced with the same non-thread-safe analog since there is no racing between offering and draining of data

  
### DIFF
--- a/src/main/java/reactor/netty/channel/FluxReceive.java
+++ b/src/main/java/reactor/netty/channel/FluxReceive.java
@@ -65,6 +65,12 @@ final class FluxReceive extends Flux<Object> implements Subscription, Disposable
 	static final AtomicIntegerFieldUpdater<FluxReceive> ONCE =
 		AtomicIntegerFieldUpdater.newUpdater(FluxReceive.class, "once");
 
+	// Please note, in this specific case WIP is non-volatile since all operation that
+	// involves work-in-progress pattern is within Netty Event-Loops which guarantees
+	// serial, thread-safe behaviour.
+	// However, we need that flag in order to preserve work-in-progress guarding that
+	// prevents stack overflow in case of onNext -> request -> onNext cycling on the
+	// same stack
 	int wip;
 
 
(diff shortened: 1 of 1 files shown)


violetagg:
@simonbasle PTAL

src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -60,10 +60,13 @@
...
 	volatile Disposable receiverCancel;
-	volatile int wip;
 
-	final static AtomicIntegerFieldUpdater<FluxReceive> WIP = AtomicIntegerFieldUpdater.newUpdater
-			(FluxReceive.class, "wip");
+	volatile int once;
+	static final AtomicIntegerFieldUpdater<FluxReceive> ONCE =
+		AtomicIntegerFieldUpdater.newUpdater(FluxReceive.class, "once");
+
+	int wip;
```
simonbasle, line 68: Nitpick: I would just add the same kind of comment than in this PR's body next to `wip`, or in the class' javadoc (ie. why there's no need for volatile `wip` nor MPSC queues, due to running on event loop)

OlegDokuka: Sure

OlegDokuka: ✅

violetagg:
@OlegDokuka I squashed and applied the change to 0.9.x branch (5e4bc03) and then forward merge to master.
Thanks for the PR.

PR #1186, opened 2020-06-29
Add proxy connectTimeoutMillis

lee-jjs:
I think it would be very helpful if we can set proxy connect timeout.

### DIFF
--- a/src/main/java/reactor/netty/transport/ProxyProvider.java
+++ b/src/main/java/reactor/netty/transport/ProxyProvider.java
@@ -248,7 +248,7 @@ public final class ProxyProvider {
 		String nonProxyHosts;
 		Supplier<? extends HttpHeaders> httpHeaders;
 		Proxy type;
-		long connectTimeoutMillis;
+		long connectTimeoutMillis = 10000;
 
 		Build() {
 		}
@@ -407,9 +407,11 @@ public final class ProxyProvider {
 		Builder httpHeaders(Consumer<HttpHeaders> headers);
 
 		/**
-		 * The proxy connect timeout in millis.
-		 * @param connectTimeoutMillis
-		 * @return
+		 * The proxy connect timeout in millis. Default to 10000 ms.
+		 * If this value set as non positive value, 0 ms will be applied.
+		 *
+		 * @param connectTimeoutMillis The proxy connect timeout in millis.
+		 * @return {@code this}
 		 */
 		Builder connectTimeoutMillis(long connectTimeoutMillis);
 
--- a/src/main/java/reactor/netty/transport/ProxyProvider.java
+++ b/src/main/java/reactor/netty/transport/ProxyProvider.java
@@ -408,7 +408,7 @@ public final class ProxyProvider {
 
 		/**
 		 * The proxy connect timeout in millis. Default to 10000 ms.
-		 * If this value set as non positive value, 0 ms will be applied.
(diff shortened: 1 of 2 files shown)


src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -238,6 +248,7 @@ private String getPasswordValue() {
 		String nonProxyHosts;
 		Supplier<? extends HttpHeaders> httpHeaders;
 		Proxy type;
+		long connectTimeoutMillis;
```
violetagg, line 251: Please set a default value
https://github.com/netty/netty/blob/662e0b4b81833a835b290e60982c9de3a1878656/handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java#L46

Otherwise with this change you effectively disable the default implementation in `ProxyHandler`

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -389,6 +406,13 @@ public ProxyProvider build() {
 		 */
 		Builder httpHeaders(Consumer<HttpHeaders> headers);
 
+		/**
+		 * The proxy connect timeout in millis.
+		 * @param connectTimeoutMillis
+		 * @return
+		 */
+		Builder connectTimeoutMillis(long connectTimeoutMillis);
```
violetagg, line 414: Please provide javadoc explaining the default value, what happens when it is a non positive value.

lee-jjs:
> The change is made against `master` branch which is in development.
> Do you need this for 0.9.x?

yes. i think it would be better.

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -389,6 +406,15 @@ public ProxyProvider build() {
 		 */
 		Builder httpHeaders(Consumer<HttpHeaders> headers);
 
+		/**
+		 * The proxy connect timeout in millis. Default to 10000 ms.
+		 * If this value set as non positive value, 0 ms will be applied.
```
violetagg, line 411: > If this value set as non positive value, 0 ms will be applied.

May be something like this is better `If this value set as non positive value, there is no connect timeout.`

violetagg:
@simonbasle PTAL

src/test/java/reactor/netty/transport/ProxyProviderTest.java
```diff
@@ -105,4 +114,12 @@ private ProxyProvider createHeaderProxy(InetSocketAddress address, Consumer<Http
...
 	}
 
+	private ProxyProvider createConnectTimeoutProxy(long connectTimeoutMillis) {
+		return ProxyProvider.builder()
+												.type(ProxyProvider.Proxy.SOCKS5)
+												.address(ADDRESS_1)
+												.connectTimeoutMillis(connectTimeoutMillis)
+												.build();
+	}
+
```
simonbasle, line 124: please add tests to validate the documented default, the behavior when `connectTimeoutMillis` is set to `0` and when it is set to `-1`

lee-jjs: I did. 
thanks.

violetagg:
@lee-jjs I squashed and applied the change to 0.9.x branch (5446678) and then forward merge to master.
Thanks for the PR.

PR #1192, opened 2020-07-03
fix #1031 Request headers are available from HttpClientRequest and HttpClientResponse

### LINKED ISSUE #1031: Provide HttpClientResponse.request() api
Let's add an api that makes it easy to tell which request of caused the current response.

## Motivation
On implicit retry and redirect, which request caused the response could be somewhat unclear. Also, some code the user is unaware of could have modified the request. In tracing, it is helpful to know the exact request line and properties a response hook received. This is not just useful at reactor-netty tier, but also for clarifying abstractions above, such as WebClient

## Desired solution
Provide HttpClientResponse.request() api

## Considered alternatives
The best current alternative is to map the cherry-picked properties available on `HttpClientResponse`, such as method, to a fake request object. However, this lacks certain properties, unless I'm mistaken, such as the request headers.

Another alternative is to add an atomic reference to the reactor context to track this. However, it is expensive and also doesn't guarantee what you've tracked was definitely the same object as what caused the request (due to immutability), iiuc.

## Additional context
https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpResponse.java#L40

violetagg:
All request settings are now available also from `HttpClientResponse`

Fixes #1031

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClientInfos.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientInfos.java
@@ -16,6 +16,7 @@
 
 package reactor.netty.http.client;
 
+import io.netty.handler.codec.http.HttpHeaders;
 import reactor.netty.http.HttpInfos;
 import reactor.util.context.Context;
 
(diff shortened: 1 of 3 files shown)


violetagg:
@simonbasle @adriancole PTAL

Issue #1190, opened 2020-07-02
PatternSyntaxException on nonProxyHosts


nhomble:
I am porting my existing application that defines nonProxyHosts with the patterns described in [Oracle: Networking Properties](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html) but the netty client is throwing parsing errors.

## Expected Behavior

The clients proxy selector should understand ```*``` as a wildcard despite not being a normal regex like ```.*```

An example value is ```"*.some.domain"```

## Actual Behavior

```PatternSyntaxException``` is thrown in ```ProxyProvider``` when ```Pattern.compile(builder.nonProxyHosts, 2)``` is executed.

## Steps to Reproduce
https://github.com/nhomble/repro-netty-nonproxyhosts-pattern

## Possible Solution

~~We could massage the string before compiling and replace ```*``` with ```.*```~~
We can't do a simple replace on wildcards without also considering the periods in the domain name.

Given their exist users who depend on the regex approach, we could either break compatibility or introduce another method on the builder to accept the selector ourselves rather than take the pattern as a string and let reactor-netty compile it internally

**Comment by violetagg** [comment]
@nhomble 
https://projectreactor.io/docs/netty/release/api/reactor/netty/tcp/ProxyProvider.Builder.html#nonProxyHosts-java.lang.String-

```
Regular expression (using java.util.regex) for a configured list of hosts that should be reached directly, bypassing the proxy.
```

This is not just a pattern but gives you the possibility to create a regular expression for hosts that will bypass the proxy.
Reactor Netty will use the regular expression that is provided by the user.

**Comment by nhomble** [comment]
I agree the docs make it clear that netty expects a regular expression, so maybe this is less of a bug in the code but I think a bug in the expectations many people would have when configuring their proxies. To quote from the oracle doc
```text
For example -Dhttp.nonProxyHosts=”*.foo.com|localhost” will indicate that every hosts in the foo.com domain and the localhost should be accessed directly even if a proxy server is specified.
```

won't work because of the dangling metacharacter ```*``` in the beginning

violetagg:
@nhomble yep as it is not a valid regular expression.
If we provide an API like this will it solve your issue `Builder nonProxyHosts(Predicate<SocketAddress> nonProxyHostsPredicate);`

nhomble:
[...]

PR #1193, opened 2020-07-04
support predicates to check for non proxied addresses

nhomble:
@violetagg open to any suggestions on how to safely add this new method and maintain backwards compatibility. With this change, I'll follow up on our other issue and use this new method when loading configurations from system variables. 

- A new method in ```Builder``` is added to set the predicate directly
- The existing nonProxyHost string is now encapsulated in a predicate object for consistency
  - An implementation is added to support the string presented in #1190

### DIFF
--- a/src/main/java/reactor/netty/transport/ProxyProvider.java
+++ b/src/main/java/reactor/netty/transport/ProxyProvider.java
@@ -36,7 +36,6 @@ import io.netty.handler.proxy.Socks4ProxyHandler;
 import io.netty.handler.proxy.Socks5ProxyHandler;
 import io.netty.util.internal.StringUtil;
 import reactor.netty.NettyPipeline;
-import reactor.util.annotation.NonNull;
 import reactor.util.annotation.Nullable;
 
 /**
@@ -118,7 +117,7 @@ public final class ProxyProvider {
 	 * @return Regular expression (<code>using java.util.regex</code>) for
 	 * a configured list of hosts that should be reached directly, bypassing the
 	 * proxy.
-	 * @deprecated as of 0.9.10. Use {@link #getNonProxyHostsPredicate()} ()}. This method will be removed in version 1.0.0.
+	 * @deprecated as of 0.9.10. Use {@link #getNonProxyHostsPredicate()}. This method will be removed in version 1.0.0.
 	 */
 	@Nullable
 	@Deprecated
@@ -173,7 +172,7 @@ public final class ProxyProvider {
 	}
 
 	/**
-	 * Returns true when the given {@link SocketAddress} should be reached via the configured proxy. When the methods
+	 * Returns true when the given {@link SocketAddress} should be reached via the configured proxy. When the method
 	 * returns false, the client should reach the address directly and bypass the proxy
 	 *
 	 * @param address the address to test
@@ -254,7 +253,7 @@ public final class ProxyProvider {
 		String host;
 		int port;
 		Supplier<? extends InetSocketAddress> address;
-		Predicate<SocketAddress> nonProxyHostPredicate = RegexShouldProxyPredicate.ALWAYS_PROXY;
+		Predicate<SocketAddress> nonProxyHostPredicate = new AlwaysProxyPredicate();
 		String nonProxyHosts;
 		Supplier<? extends HttpHeaders> httpHeaders;
 		Proxy type;
@@ -303,7 +302,7 @@ public final class ProxyProvider {
 		@Override
 		public final Builder nonProxyHosts(String nonProxyHostsPattern) {
 			this.nonProxyHosts = nonProxyHostsPattern;
-			return nonProxyHostsPattern == null ? this : nonProxyHostsPredicate(new RegexShouldProxyPredicate(nonProxyHostsPattern));
+			return StringUtil.isNullOrEmpty(nonProxyHostsPattern) ? this : nonProxyHostsPredicate(new RegexShouldProxyPredicate(nonProxyHostsPattern));
 		}
 
 		@Override
@@ -340,9 +339,23 @@ public final class ProxyProvider {
 		}
 	}
 
+	static final class AlwaysProxyPredicate implements Predicate<SocketAddress> {
+
+		/**
+		 * The test always returns false which means for any {@link SocketAddress} do not bypass the proxy. Instead,
+		 * we should always go through the proxy.
+		 *
+		 * @param socketAddress the address we are choosing to connect via proxy or not
+		 * @return always return false
+		 */
+		@Override
+		public boolean test(SocketAddress socketAddress) {
+			return false;
+		}
+	}
+
 	static final class RegexShouldProxyPredicate implements Predicate<SocketAddress> {
 
-		public static final RegexShouldProxyPredicate ALWAYS_PROXY = RegexShouldProxyPredicate.fromWildcardedPattern("");
 		public static final RegexShouldProxyPredicate DEFAULT_NON_PROXY = RegexShouldProxyPredicate.fromWildcardedPattern("localhost|127.*|[::1]|0.0.0.0|[::0]");
 
 		private final String regex;
@@ -390,7 +403,7 @@ public final class ProxyProvider {
 				parts[2] = ".*";
 				in = in.substring(0, in.length() - 1);
 			}
-			parts[1] = in;
+			parts[1] = Pattern.quote(in);
 			return String.join("", parts);
 		}
 
@@ -412,7 +425,8 @@ public final class ProxyProvider {
 				return false;
 			}
 			InetSocketAddress isa = (InetSocketAddress) socketAddress;
-			return isa.getHostString() != null && pattern.matcher(isa.getHostString()).matches();
+			String hostString = isa.getHostString();
+			return hostString != null && pattern.matcher(hostString).matches();
 		}
 
 		@Override
--- a/src/main/java/reactor/netty/transport/ProxyProvider.java
+++ b/src/main/java/reactor/netty/transport/ProxyProvider.java
@@ -508,8 +508,8 @@ public final class ProxyProvider {
 		 * A standard predicate expression for a configured list of hosts that should be reached directly, bypassing
 		 * the proxy.
 		 *
-		 * @param nonProxyHostsPredicate A Predicate {@link Predicate} on {@link SocketAddress} to determine if the
-		 *                               address should not go through the configured proxy
+		 * @param nonProxyHostsPredicate A Predicate {@link Predicate} on {@link SocketAddress} that returns
+		 * {@literal true} if the host should bypass the configured proxy
 		 * @return {@code this}
 		 */
 		Builder nonProxyHostsPredicate(Predicate<SocketAddress> nonProxyHostsPredicate);
--- a/src/main/java/reactor/netty/transport/ProxyProvider.java
+++ b/src/main/java/reactor/netty/transport/ProxyProvider.java
@@ -322,12 +322,32 @@ public final class ProxyProvider {
 		private final String regex;
 		private Pattern pattern;
 
-		public RegexShouldProxyPredicate(String pattern) {
+        /**
+         * Create a {@link RegexShouldProxyPredicate} based off the provided regular expression.
+         * @param regex The string regular expression
+         * @return a predicate whether we should direct to proxy
+         */
+		public static RegexShouldProxyPredicate fromRegularExpression(String regex) {
+            return new RegexShouldProxyPredicate(regex);
+        }
+
+        /**
+         * Creates a {@link RegexShouldProxyPredicate} based off the provided pattern with possible wildcards as
+         * described in https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html
+         *
+         * @param pattern The string wildcarded expression
+         * @return a predicate whether we should direct to proxy
+         */
+        public static RegexShouldProxyPredicate fromWildcardedPattern(String pattern) {
+            return new RegexShouldProxyPredicate(pattern.replaceAll("\\.", "\\\\.").replaceAll("\\*", "\\.*"));
+		}
+
+		private RegexShouldProxyPredicate(String pattern) {
 			this.regex = pattern;
 		}
 
-		private Pattern getPattern(){
-			if(pattern == null){
+		private Pattern getPattern() {
+			if (pattern == null) {
 				pattern = Pattern.compile(regex, Pattern.CASE_INSENSITIVE);
 			}
 			return pattern;
@@ -335,7 +355,7 @@ public final class ProxyProvider {
 
 		@Override
 		public boolean test(SocketAddress socketAddress) {
-			if(!(socketAddress instanceof InetSocketAddress)){
+			if (!(socketAddress instanceof InetSocketAddress)) {
 				return false;
 			}
 			InetSocketAddress isa = (InetSocketAddress) socketAddress;
--- a/src/main/java/reactor/netty/transport/ProxyProvider.java
+++ b/src/main/java/reactor/netty/transport/ProxyProvider.java
@@ -35,6 +35,7 @@ import io.netty.handler.proxy.ProxyHandler;
 import io.netty.handler.proxy.Socks4ProxyHandler;
 import io.netty.handler.proxy.Socks5ProxyHandler;
 import reactor.netty.NettyPipeline;
+import reactor.util.annotation.NonNull;
 import reactor.util.annotation.Nullable;
 
 /**
@@ -43,6 +44,8 @@ import reactor.util.annotation.Nullable;
  * @author Violeta Georgieva
  */
 public final class ProxyProvider {
+
+	private static final Predicate<SocketAddress> ALWAYS_PROXY = a ->true;
 
 	/**
 	 * Creates a builder for {@link ProxyProvider ProxyProvider}
@@ -59,6 +62,7 @@ public final class ProxyProvider {
 	final String username;
 	final Function<? super String, ? extends String> password;
 	final Supplier<? extends InetSocketAddress> address;
+	final Pattern nonProxyHosts;
 	final Predicate<SocketAddress> shouldProxy;
 	final Supplier<? extends HttpHeaders> httpHeaders;
 	final Proxy type;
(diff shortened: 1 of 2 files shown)


src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -324,6 +317,57 @@ public ProxyProvider build() {
...
+
+        /**
+         * Creates a {@link RegexShouldProxyPredicate} based off the provided pattern with possible wildcards as
+         * described in https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html
+         *
+         * @param pattern The string wildcarded expression
+         * @return a predicate whether we should direct to proxy
+         */
+        public static RegexShouldProxyPredicate fromWildcardedPattern(String pattern) {
+            return new RegexShouldProxyPredicate(pattern.replaceAll("\\.", "\\\\.").replaceAll("\\*", "\\.*"));
```
violetagg, line 342: Here https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html there is a little bit more specific about `http.nonProxyHosts` 

```
http.nonProxyHosts:a list of hosts that should be reached directly, bypassing the proxy. This is a list of patterns separated by '|'. The patterns may start or end with a '*' for wildcards. Any host matching one of these patterns will be reached through a direct connection instead of through a proxy.
```

So `*` is accepted only as a start/end of the pattern.

Let's add tests like `localhost|127.*|[::1]`, `127.*`, `127.*.0.1`, `some.foo.*`, `some.*.com`

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -161,18 +160,7 @@ public final ProxyHandler newProxyHandler() {
...
-	/**
-	 * Should proxy the given hostname
-	 *
-	 * @param hostName the hostname to test
-	 * @return true if candidate to proxy
-	 */
-	public boolean shouldProxy(@Nullable String hostName) {
-		return nonProxyHosts == null || hostName == null || !nonProxyHosts.matcher(hostName)
-		                                                                  .matches();
+		return address instanceof InetSocketAddress && shouldProxy.test(address);
```
violetagg, line 163: Let's clarify what's returning `shouldProxy` and then we can thing about backwards compatibility.

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -197,7 +185,7 @@ public void addProxyHandler(Channel channel) {
 	public String toString() {
 		return "ProxyProvider {" +
 				"address=" + address.get() +
-				", nonProxyHosts=" + nonProxyHosts +
+				", nonProxyHosts=" + shouldProxy +
```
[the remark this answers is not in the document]
nhomble: I figured we could use the ```.toString()``` to make this useful. For the classes I added, I returned the regexs which was the existing behaviour.

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -245,7 +233,7 @@ private String getPasswordValue() {
 		String host;
 		int port;
 		Supplier<? extends InetSocketAddress> address;
-		String nonProxyHosts;
+		Predicate<SocketAddress> shouldProxy = a -> true;
```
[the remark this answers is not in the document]
nhomble: I found myself confused a couple of times handling the not notProxied double negative. Since we eventually call this in the ```shouldProxy()``` method I figured I could stick with that. But since this is related to specifically nonProxyHosts, I am happy to bring it back.

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -292,7 +280,12 @@ public final Builder address(Supplier<? extends InetSocketAddress> addressSuppli
 
 		@Override
 		public final Builder nonProxyHosts(String nonProxyHostsPattern) {
-			this.nonProxyHosts = nonProxyHostsPattern;
+			return nonProxyHostsPredicate(new RegexShouldProxyPredicate(nonProxyHostsPattern));
+		}
+
+		@Override
+		public final Builder nonProxyHostsPredicate(Predicate<SocketAddress> nonProxyHostsPredicate){
+			this.shouldProxy = nonProxyHostsPredicate;
```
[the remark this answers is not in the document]
nhomble: assertion makes sense and is there a codestyle xml I could import? You had another comment about tabs/spaces and I definitely want to be consistent

src/main/java/reactor/netty/transport/ProxyProvider.java
nhomble: true this was a bit of a hack to get the other boilerplate through, I'll follow up with something more legitimate 😄

src/main/java/reactor/netty/transport/ProxyProvider.java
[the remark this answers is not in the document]
violetagg: Did you check https://github.com/reactor/reactor-netty/blob/master/CONTRIBUTING.md#code-style

src/main/java/reactor/netty/transport/ProxyProvider.java
nhomble: Are there problems with using ```DefaultProxySelector``` from ```sun.net.spi```? It handles transforming the string to java pattern in ```disjunctToRegex```

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -324,6 +317,57 @@ public ProxyProvider build() {
...
+			return pattern;
+		}
+
+		@Override
+		public boolean test(SocketAddress socketAddress) {
+			if (!(socketAddress instanceof InetSocketAddress)) {
+				return false;
+			}
+			InetSocketAddress isa = (InetSocketAddress) socketAddress;
+			return isa.getHostName() == null || !getPattern().matcher(isa.getHostName()).matches();
```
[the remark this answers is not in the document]
nhomble: bad refactor by me - I didn't intend to change the original behavior
```
	public boolean shouldProxy(SocketAddress address) {
		return address instanceof InetSocketAddress && shouldProxy(((InetSocketAddress) address).getHostString());
	}
```

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -324,6 +340,65 @@ public ProxyProvider build() {
...
+
+		/**
+		 * The test returns true when the given {@link SocketAddress} should be contacted through the configured proxy.
+		 * When this method returns false, the client will reach out to the address directly and bypass the proxy
+		 *
+		 * @param socketAddress
+		 * @return true when we should go through proxy
+		 */
+		@Override
+		public boolean test(SocketAddress socketAddress) {
```
violetagg, line 388: We have this `Builder nonProxyHostsPredicate(Predicate<SocketAddress> nonProxyHostsPredicate);`

```
		 * A standard predicate expression for a configured list of hosts that should be reached directly, bypassing
 		 * the proxy.
```

So the predicate returns true when you want to bypass the proxy while here it is the opposite?

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -324,6 +340,65 @@ public ProxyProvider build() {
...
+		}
+
+		private RegexShouldProxyPredicate(String pattern) {
+			this.regex = pattern;
+		}
+
+		private Pattern getPattern() {
+			if (pattern == null) {
+				pattern = Pattern.compile(regex, Pattern.CASE_INSENSITIVE);
+			}
```
violetagg, line 376: Let's compile the pattern when the object is created. When there is a regex this mean that we will always check it so no need of lazy compilation of the regex.

src/main/java/reactor/netty/transport/ProxyProvider.java
violetagg: `sun.net.spi` I don't think we want to use the internal java packages

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -324,6 +340,87 @@ public ProxyProvider build() {
...
+		private final String regex;
+		private final Pattern pattern;
+
+		/**
+		 * Create a {@link RegexShouldProxyPredicate} based off the provided regular expression.
+		 *
+		 * @param regex The string regular expression
+		 * @return a predicate whether we should bypass the proxy
+		 */
+		public static RegexShouldProxyPredicate fromRegularExpression(String regex) {
```
violetagg, line 357: Do we need this, it is not used at the moment

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -240,11 +248,13 @@ private String getPasswordValue() {
 	}
 
 	static final class Build implements TypeSpec, AddressSpec, Builder {
+
 		String username;
 		Function<? super String, ? extends String> password;
 		String host;
 		int port;
 		Supplier<? extends InetSocketAddress> address;
+		Predicate<SocketAddress> nonProxyHostPredicate = RegexShouldProxyPredicate.ALWAYS_PROXY;
```
violetagg, line 257: From performance point of view which performs better `a -> false` or `pattern.matcher(isa.getHostString()).matches()`

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -293,6 +303,12 @@ public final Builder address(Supplier<? extends InetSocketAddress> addressSuppli
 		@Override
 		public final Builder nonProxyHosts(String nonProxyHostsPattern) {
 			this.nonProxyHosts = nonProxyHostsPattern;
+			return nonProxyHostsPattern == null ? this : nonProxyHostsPredicate(new RegexShouldProxyPredicate(nonProxyHostsPattern));
```
violetagg, line 306: Let's add here `StringUtil.isNullOrEmpty(nonProxyHostsPattern) ? this : nonProxyHostsPredicate(new RegexShouldProxyPredicate(nonProxyHostsPattern));`

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -324,6 +340,87 @@ public ProxyProvider build() {
...
+		 * @param socketAddress the address we are choosing to connect via proxy or not
+		 * @return true we should bypass the proxy
+		 */
+		@Override
+		public boolean test(SocketAddress socketAddress) {
+			if (!(socketAddress instanceof InetSocketAddress)) {
+				return false;
+			}
+			InetSocketAddress isa = (InetSocketAddress) socketAddress;
+			return isa.getHostString() != null && pattern.matcher(isa.getHostString()).matches();
```
violetagg, line 415: let's take this once and reuse it `isa.getHostString()`

src/test/java/reactor/netty/transport/ProxyProviderTest.java
```diff
@@ -16,16 +16,17 @@
...
 import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Predicate;
 
 import io.netty.handler.codec.http.HttpHeaders;
 import org.junit.Test;
 
+import static org.junit.Assert.*;
```
violetagg, line 28: Please do not use `*`

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -324,6 +340,87 @@ public ProxyProvider build() {
...
+			String[] parts = new String[]{"", "", ""};
+			if (in.startsWith("*")) {
+				parts[0] = ".*";
+				in = in.substring(1);
+			}
+			if (in.endsWith("*")) {
+				parts[2] = ".*";
+				in = in.substring(0, in.length() - 1);
+			}
+			parts[1] = in;
```
violetagg, line 393: you should use `Pattern.quote` so that metacharacters or escape sequences in the input sequence will be given no special meaning.
Let's say we have something like this

```
@Test
public void nonProxyHosts_wildcardInTheMiddle() {
	ProxyProvider.RegexShouldProxyPredicate pred = ProxyProvider.RegexShouldProxyPredicate.fromWildcardedPattern("some.*.com");
	assertFalse("Should proxy, nothing matching foo", pred.test(someAddress("some.other.com")));
	assertFalse("Should proxy, nothing matching foo", pred.test(someAddress("some.foo.com")));
}
```

nhomble: good catch, fixed!

src/main/java/reactor/netty/transport/ProxyProvider.java
nhomble: I was going to leave the lambda on ```Builder``` as a constant but then declaring something like
```
private static final Predicate<SocketAddress> ALWAYS_PROXY = a -> false
```
caused build warnings

```
warning: [UnnecessaryLambda] Returning a lambda from a helper method or saving it in a constant is unnecessary; prefer to implement the functional interface method directly and use a method reference instead.
```

instead of adding a static method on the builder or making it non-constant, I chose to create a trivial implementation ```AlwaysProxyPredicate``` but I am happy to move this.

src/main/java/reactor/netty/transport/ProxyProvider.java
nhomble: I did it for completeness since I made the constructor private and I figured I should support to existing way of defining a normal java pattern to represent nonProxyHosts. I am happy to remove though.

src/main/java/reactor/netty/transport/ProxyProvider.java
violetagg: Yes this warning is so that you think about whether you want really to use the lambda as a constant. It depends on your use case. 
You can suppress it here, like the other one in this class `NO_HTTP_HEADERS`

So let's do it like this

```
@SuppressWarnings("UnnecessaryLambda")
private static final Predicate<SocketAddress> ALWAYS_PROXY = a -> false
```

src/main/java/reactor/netty/transport/ProxyProvider.java
violetagg: yes please remove it

src/test/java/reactor/netty/transport/ProxyProviderTest.java
```diff
@@ -103,6 +107,83 @@ public void connectTimeoutWithDefault() {
...
+		assertFalse("Should proxy, nothing matching foo", pred.test(someAddress("some.other.com")));
+		assertTrue("Should not proxy, prefix in wildcard", pred.test(someAddress("some.foo.com")));
+	}
+
+	@Test
+	public void nonProxyHosts_wildcardFinish() {
+		ProxyProvider.RegexShouldProxyPredicate pred = ProxyProvider.RegexShouldProxyPredicate.fromWildcardedPattern("foo*");
+		assertFalse("Should proxy, nothing matching prefix", pred.test(someAddress("other.foo.com")));
+		assertTrue("Should not proxy, anything in wildcard", pred.test(someAddress("foo.other.com")));
+		assertTrue("Should proxy, nothing in wildcard", pred.test(someAddress("foo.")));
```
violetagg, line 126: `Should proxy` should be `Should not proxy`

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -155,24 +172,14 @@ public final ProxyHandler newProxyHandler() {
...
-	/**
-	 * Should proxy the given hostname
-	 *
-	 * @param hostName the hostname to test
-	 * @return true if candidate to proxy
-	 */
-	public boolean shouldProxy(@Nullable String hostName) {
-		return nonProxyHosts == null || hostName == null || !nonProxyHosts.matcher(hostName)
-		                                                                  .matches();
+		return address instanceof InetSocketAddress && !nonProxyHostPredicate.test(address);
```
violetagg, line 182: Can we keep the backwards compatibility with something like this?

```
 * @deprecated as of 0.9.10. Use {@link #shouldProxy(SocketAddress)}. This method will be removed in version 1.0.0.
 */
@Deprecated
public boolean shouldProxy(@Nullable String hostName) {
	return nonProxyHostPredicate == ALWAYS_PROXY || hostName == null ||
	        !((RegexShouldProxyPredicate) nonProxyHostPredicate).pattern.matcher(hostName)
	                                                                    .matches();
}
```

nhomble:
> I think we are almost there.

I really appreciate the time you have given me 😄

src/main/java/reactor/netty/transport/ProxyProvider.java
nhomble: makes sense!

violetagg:
@simonbasle PTAL
Please note that the target branch is 0.9.x, when I merge this change to master I'll remove the deprecated methods.

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -155,24 +174,29 @@ public final ProxyHandler newProxyHandler() {
...
-	 * @return true if candidate to proxy
+	 * @return true if should be proxied
+	 * @deprecated as of 0.9.10. use {@link #shouldProxy(SocketAddress)}. This method will be removed in version 1.0.0.
 	 */
+	@Deprecated
 	public boolean shouldProxy(@Nullable String hostName) {
-		return nonProxyHosts == null || hostName == null || !nonProxyHosts.matcher(hostName)
-		                                                                  .matches();
+		return nonProxyHostPredicate == ALWAYS_PROXY || hostName == null ||
+				!((RegexShouldProxyPredicate) nonProxyHostPredicate).pattern.matcher(hostName)
```
simonbasle, line 198: dangerous assumption that this can always be cast to `RegexShouldProxyPredicate`. Couldn't it directly use the `Predicate#test` method instead? (which is more generic)

src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -393,11 +499,21 @@ public ProxyProvider build() {
...
+		 *                             for a configured list of hosts that should be reached directly, bypassing the proxy.
 		 * @return {@code this}
 		 */
 		Builder nonProxyHosts(String nonProxyHostsPattern);
 
+		/**
+		 * A standard predicate expression for a configured list of hosts that should be reached directly, bypassing
+		 * the proxy.
+		 *
+		 * @param nonProxyHostsPredicate A Predicate {@link Predicate} on {@link SocketAddress} to determine if the
```
simonbasle, line 511: ideally the javadoc should be rephrased a bit: "to determine if address should not go through" is weird to read. maybe `that returns {@literal true} if the host should bypass the configured proxy`

src/main/java/reactor/netty/transport/ProxyProvider.java
simonbasle: if that is the case I'd definitely recommend to make it consistent

src/main/java/reactor/netty/transport/ProxyProvider.java
violetagg: This should not be used by anybody outside of Reactor Netty.
However you are right.
I would suggest here to check only for `instanceof` as before that change there wasn't a possibility for using a Predicate and also as there an explicit recommendation to use the new method.

So this one should work as it was before:

```
		return nonProxyHostPredicate == ALWAYS_PROXY || hostName == null ||
				((nonProxyHostPredicate instanceof RegexShouldProxyPredicate) && !((RegexShouldProxyPredicate) nonProxyHostPredicate).pattern.matcher(hostName))
```

nhomble: The concerns make sense to me - I'll push up the change.

src/main/java/reactor/netty/transport/ProxyProvider.java
violetagg: 👍

violetagg:
@nhomble Please fix the remaining javadoc recommendation. Thanks.

violetagg:
@nhomble I squashed the changed and committed them to 0.9.x branch (
f6f05ae), then forward merge to master.
Thanks for the contribution.

PR #1208, opened 2020-07-15
provides context caching

OlegDokuka:
This PR unignore `contextShouldBeTransferredFromDownStreamToUpStream` and ensures that context is cached per `ChannelOperation` class

### DIFF
--- a/src/main/java/reactor/netty/channel/MonoSendMany.java
+++ b/src/main/java/reactor/netty/channel/MonoSendMany.java
@@ -119,6 +119,7 @@ final class MonoSendMany<I, O> extends MonoSend<I, O> implements Scannable {
 		final EventLoop                    eventLoop;
 		final MonoSendMany<I, O>           parent;
 		final CoreSubscriber<? super Void> actual;
+		final Context                      actualContext;
 		final Runnable                     asyncFlush;
 
 
@@ -140,6 +141,7 @@ final class MonoSendMany<I, O> extends MonoSend<I, O> implements Scannable {
 		SendManyInner(MonoSendMany<I, O> parent, CoreSubscriber<? super Void> actual) {
 			this.parent = parent;
 			this.actual = actual;
+			this.actualContext = actual.currentContext();
 			this.requested = MAX_SIZE;
 			this.ctx = parent.ctx;
 			this.eventLoop = ctx.channel().eventLoop();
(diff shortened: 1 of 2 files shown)


violetagg:
@simonbasle PTAL

src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -140,6 +141,7 @@ public Object scanUnsafe(Attr key) {
 		SendManyInner(MonoSendMany<I, O> parent, CoreSubscriber<? super Void> actual) {
 			this.parent = parent;
 			this.actual = actual;
+			this.actualContext = actual.currentContext();
```
simonbasle, line 144: so now we have a field capturing the `Context` of downstream... I'm confused now, as you opposed doing that in the prveious PR when I suggested we store the context as  `this.context = actual.currentContext().put(KEY_ON_DISCARD, this)` in the constructor ?

OlegDokuka: yeah. it is confusing. The difference is that

`this.actualContext = actual.currentContext();`

is obtaining already created object whilst

`this.actualContext = actual.currentContext().put(...);`

create a new one along with the previous.


The idea here is not to produce a new object but store `actual.currentContext();` since underneath  it calls 

https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/channel/ChannelOperations.java#L503

which in turn end up calling 

https://github.com/reactor/reactor-netty/blob/41e73261a5935dac3b5773ab3668d0dd99d4cf42/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java#L364.

thus, in order to avoid that invocation all the time, we resolve context once at the construction

violetagg: note that `owner()` checks always a channel attribute which is an expensive operation
https://github.com/reactor/reactor-netty/blob/41e73261a5935dac3b5773ab3668d0dd99d4cf42/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java#L445-L462

violetagg:
@OlegDokuka I committed the fix to 0.9.x (3945a9a) and then forward merge to master.
Thanks for the PR

PR #1211, opened 2020-07-16
fixes class cast exception to no appear

OlegDokuka:
Fixes class cast exception appeared when non ByteBufs are discarded by the upstream


violetagg:
@simonbasle PTAL
reactor.netty.http.client.HttpClientTest#testIssue700AndIssue876 can be used for verification of the fix

violetagg:
@OlegDokuka I committed the fix to 0.9.x (1f406f2) and then forward merge to master.
Thanks for the PR

PR #1212, opened 2020-07-16
Proxy provider cleanup

violetagg:


src/main/java/reactor/netty/tcp/ProxyProvider.java
```diff
@@ -358,7 +352,6 @@ public final Builder address(Supplier<? extends InetSocketAddress> addressSuppli
 
 		@Override
 		public final Builder nonProxyHosts(String nonProxyHostsPattern) {
-			this.nonProxyHosts = nonProxyHostsPattern;
 			return StringUtil.isNullOrEmpty(nonProxyHostsPattern) ? this : nonProxyHostsPredicate(new RegexShouldProxyPredicate(nonProxyHostsPattern));
```
simonbasle, line 355: just something that I noticed reviewing this PR: should `nonProxyHosts(null)` call disable any `nonProxyHostPredicate` in place? it used to "disable" `nonProxyHostPattern`...

violetagg: yep agree

violetagg: fixed

PR #1216, opened 2020-07-18
Fix #1159 Expose requestTimeout setting on HttpClient/HttpClientRequest level

### LINKED ISSUE #1159: HTTP client request timeout
I'd like the http client to allow specifying a request timeout (i.e. the time it takes to receive a response after sending a request).  Something similar to the [java http client's request timeout](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpRequest.Builder.html#timeout(java.time.Duration)).

This is for use in WebFlux's WebClient, originally requested in https://github.com/spring-projects/spring-framework/issues/25115

## Motivation

To be able to specify an HTTP-level request timeout, that does not end up closing idle connections or take into account the time it takes to establish connections  (see Considered Alternatives below)

## Desired solution
The ability to specify:
* the default request timeout for an http client.
* a specific timeout for a specific request  (i.e. override the default)

I'd like to easily be able to use these features in WebFlux's WebClient.  This might or might not require changes to WebClient, but I'd like it to be considered as part of a "complete" solution.

## Considered alternatives

The use of netty's `ReadTimeoutHandler` / `WriteTimeoutHandler` is insufficient to use as an HTTP request timeout.  Specifically they apply at the TCP level, which leads to problems like this:
1. They apply during the SSL handshake, which might take longer than a typical HTTP response, due to the cryptography involved.  Therefore, they would need to be set higher than desired for HTTP responses.
2. They apply even when an HTTP request is not being processed.  For example, they could cause a connection sitting in the connection pool to be closed, even though it might be able to be used a split-second later by another request.

The `.timeout` operator on the reactive stream is insufficient to use as an HTTP request timeout as well.  Specifically, it operates on the reactive stream, which includes things like obtaining a connection from the connection pool and potentially creating a new connection, in addition to the time it takes the client to receive a response.  This leads to having to use a `.timeout` value this is greater than the connection timeout plus the time to obtain the connection from the connection pool.   So, the `.timeout` operator cannot be used as an HTTP request timeout.  I'm willing to "pay the price" of establishing connections occasionally (leading to waiting longer occasionally for the stream to emit), but I'd still like to set a lower HTTP request timeout, to ensure the stream emits values fast when a new connection is not established.

violetagg:
ReadTimeoutHandler is added when request sending is finished successfully.
The handler is removed before returning the connection to the pool.
The setting on request level overrides the one on client level.
Fixes #1159

### DIFF
--- a/src/main/java/reactor/netty/NettyPipeline.java
+++ b/src/main/java/reactor/netty/NettyPipeline.java
@@ -81,6 +81,7 @@ public interface NettyPipeline {
 	String WsCompressionHandler = LEFT + "wsCompressionHandler";
 	String ProxyProtocolDecoder = LEFT + "proxyProtocolDecoder";
 	String ProxyProtocolReader  = LEFT + "proxyProtocolReader";
+	String RequestTimeoutHandler = LEFT + "requestTimeoutHandler";
 
 	/**
 	 * Create a new {@link ChannelInboundHandler} that will invoke
--- a/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -17,6 +17,7 @@ package reactor.netty.http.client;
 
 import java.net.SocketAddress;
 import java.net.URI;
+import java.time.Duration;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.BiConsumer;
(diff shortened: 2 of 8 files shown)


philsttr:
This looks perfect @violetagg !  Thank you so much 🎉 

Now we just need a way to set the request-specific timeout via `WebClient`.   @rstoyanchev, how would you like to handle that?

rstoyanchev:
@philsttr I've re-opened https://github.com/spring-projects/spring-framework/issues/25115.

Issue #1155, opened 2020-06-14
Make PooledConnectionProviderMetrics configureable 


peterjurkovic:
Currently the `PooledConnectionProviderMetrics` exposes metrics per id, name, remote address and name. Due to the visibility and the way how it's been implemented it is not possible to override it. The issue is `remoteAddress` tag. If an application sending requests to N remote addresses It also generates a lot of data. We have a use case where an actuator metrics endpoints generates body size up to 2MB (due to the `remoteAddress` tag).

## Desired solution
Nice solution would be to do it similarly like HttpClient does it. expose `.metric(boolean, Registrator)` 

## Nice to have
Also would be nice to see a ratio - how many connections were reused from a pool

PR #1218, opened 2020-07-22
1155 pooled connection provider metrics configurable

aneveu:


src/main/java/reactor/netty/resources/DefaultPooledConnectionProviderMeterRegistrar.java
```diff
@@ -29,13 +29,19 @@
...
+ *
  * @author Violeta Georgieva
  * @since 0.9
  */
-final class PooledConnectionProviderMetrics {
+final class DefaultPooledConnectionProviderMeterRegistrar implements PooledConnectionProvider.MeterRegistrar {
 
-	static void registerMetrics(String poolName, String id, String remoteAddress,
-			InstrumentedPool.PoolMetrics metrics) {
+	 public void registerMetrics(String poolName, String id, String remoteAddress,
```
bsideup, line 43: `@Overrides` is missing :)

ericbottard:
Issue #1155 is not about abstracting micrometer (although that would be a nice to have benefit), but rather about letting users decide exactly which (and how) meters are registered (in particular wrt explosion of cardinality of the `tags`) isn't it?

As such, some of the decisions in this PR don't go as far as shielding the user from micrometer.

If we do want to abstract micrometer away, I think it would be better to do that in a separate PR, because it leaks in other places, wdyt?

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -91,8 +92,8 @@ protected PooledConnectionProvider(Builder builder) {
 
 				InstrumentedPool<T> newPool = createPool(config, poolFactory, remoteAddress, resolverGroup);
 
-				if (poolFactory.metricsEnabled || config.metricsRecorder() != null) {
-					PooledConnectionProviderMetrics.registerMetrics(name,
+				if (poolFactory.metricsEnabled) {
```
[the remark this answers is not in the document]
ericbottard: The fact that connection provider metrics are sometimes driven by HttpClient metrics is what puzzled us the most. Do we really want that? At the very least, it should be documented as a "feature" in either `HttpMetrics.metrics()` or `DefaultConnectionProvider.metrics()`, wdyt?

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -269,4 +270,16 @@ public int hashCode() {
...
 		}
 	}
+
+
+	/**
+	 * A strategy to register which {@link io.micrometer.core.instrument.Meter} are collected by the {@link PooledConnectionProvider}.
+	 *
+	 * Default implementation of this interface is {@link DefaultPooledConnectionProviderMeterRegistrar}
+	 */
+	public static interface MeterRegistrar {
```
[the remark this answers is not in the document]
ericbottard: We thought about that but only `PooledConnectionProvider` provides metrics concept. Indeed `InstrumentedPool.PoolMetrics` only makes sense for pooled connections.

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -269,4 +270,16 @@ public int hashCode() {
...
+
+		void registerMetrics(String poolName, String id, String remoteAddress, InstrumentedPool.PoolMetrics metrics);
```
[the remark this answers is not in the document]
ericbottard: > With other APIs we expose SocketAddress and not the string representation of the remote address.

:+1:

As for the shadowing part, yes that is what will need to be done. But it only shifts the problem one step further: if reactor-pool APIs change, this one will need to change as well...

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -269,4 +270,16 @@ public int hashCode() {
 			return Objects.hash(fqdn, holder, pipelineKey);
 		}
 	}
+
+
+	/**
+	 * A strategy to register which {@link io.micrometer.core.instrument.Meter} are collected by the {@link PooledConnectionProvider}.
```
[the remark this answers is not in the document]
ericbottard: See my comment at the top PR level: the initial intent of issue #1155 is not to abstract away from micrometer IMO

src/main/java/reactor/netty/resources/DefaultPooledConnectionProviderMeterRegistrar.java
```diff
@@ -29,13 +29,19 @@
...
+ *
+ * Registers gauges for every metric in {@link reactor.pool.InstrumentedPool.PoolMetrics}.
+ *
+ * Every gauge uses id, poolName and remoteAddress as tags.
+ *
  * @author Violeta Georgieva
  * @since 0.9
  */
-final class PooledConnectionProviderMetrics {
+final class DefaultPooledConnectionProviderMeterRegistrar implements PooledConnectionProvider.MeterRegistrar {
```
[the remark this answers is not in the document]
ericbottard: idem regarding the purpose of this PR. Happy to use Micrometer in the name, but even then, there may be a "Default" that registers some gauges in a certain way (in micrometer) and users may want a (custom) registrar that register gauges differently (but still in Micrometer). WDYT?

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -372,10 +373,33 @@ public final SPEC maxLifeTime(Duration maxLifeTime) {
 		 * </pre>
 		 * <p>By default this is not enabled.
 		 *
+		 * If enabled, this method is providing a {@link DefaultPooledConnectionProviderMeterRegistrar} to define which meters are going to be collected.
+		 *
 		 * @param metricsEnabled true enables metrics collection; false disables it
 		 * @return {@literal this}
 		 */
 		public final SPEC metrics(boolean metricsEnabled) {
+			return this.metrics(metricsEnabled, new DefaultPooledConnectionProviderMeterRegistrar());
```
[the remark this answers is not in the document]
ericbottard: We can indeed use a singleton here, but can you clarify how this would be called N times _"If you have different configurations per remote host"_?

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -372,10 +373,33 @@ public final SPEC maxLifeTime(Duration maxLifeTime) {
...
+		 * MeterFilter filter = ... ;
+		 * Metrics.globalRegistry.config().meterFilter(MeterFilter.maximumAllowableTags(CONNECTION_PROVIDER_PREFIX, 100, filter));
+		 * </pre>
+		 * <p>By default this is not enabled.
+		 *
+		 * @param metricsEnabled true enables metrics collection; false disables it
+		 * @param registrar the registrar used to define what meters are going to be collected.
+		 * @return {@literal this}
+		 */
+		public final SPEC metrics(boolean metricsEnabled, PooledConnectionProvider.MeterRegistrar registrar) {
```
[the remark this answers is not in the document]
ericbottard: > The idea of the new method is to remove the dependency on Micrometer.

see my comment above

> The user will provide an implementation of the second parameter (it is better to be a Supplier<> so that we will call it only when the metrics are enabled)

if the user calls this with `false`, he/she is more likely to use the overload with only one parameter anyway, so we don't understand what using a Supplier gives in terms of benefits. Registrars are meant to be lightweight anyway. It's actually calling the registration method that is costly, not creating the registrar, no?

src/main/java/reactor/netty/resources/ConnectionProvider.java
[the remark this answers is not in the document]
violetagg: Reactor Netty provides so called default configuration and specific per remote address

```
ConnectionProvider.builder("test")
                  .maxConnections(2)
                  .metrics(true)
                  .forRemoteHost(InetSocketAddress.createUnresolved("localhost", server.port()),
                                 spec -> spec.maxConnections(1).metrics(false))
                  .build();
```

This example above specifies that all connection pools for remote address different than `localhost:PORT` will have 2 connections and the metrics will be enabled while the connection pool for `localhost:PORT` will have 1 connection and the metrics will be disabled.

src/main/java/reactor/netty/resources/ConnectionProvider.java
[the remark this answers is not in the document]
violetagg: In some environments metrics can be activated (test/canary) but on the production disabled. The configuration can be controlled by an environment variable or something else (system property, configuration service). To avoid many if/else clauses we offer boolean parameter to enable/disable the whole feature and when it is disabled the instance will not be created. Similar features were requested by `Spring Cloud Gateway` for example `HttpClient.wiretap`, `HttpClient.keepAlive` etc.
With Supplier we the instance will be created only when it is needed.

```
boolean metricsDisabled = Boolean.parseBoolean(System.getProperty("metricsEnabled", "false"));

ConnectionProvider.builder("test")
                  .maxConnections(2)
                  .metrics(metricsDisabled, () -> ...)
                  .forRemoteHost(InetSocketAddress.createUnresolved("localhost", server.port()),
                                 spec -> spec.maxConnections(1).metrics(metricsDisabled, () -> ...))
                  .build();
```

src/test/java/reactor/netty/resources/DefaultPooledConnectionProviderTest.java
```diff
@@ -235,6 +241,84 @@ public void fixedPoolTwoAcquire()
 		}
 	}
 
+
+	@Test
+	public void testDefaultProviderMetrics() throws Exception {
```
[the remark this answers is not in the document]
aneveu: The `DefaultPooledConnectionProviderMeterRegistrarTest` has been deleted and the test moved to this class 😄

src/main/java/reactor/netty/resources/ConnectionProvider.java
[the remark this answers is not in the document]
ericbottard: Gotcha. Still, the registrar should be lightweight. Using a Supplier seems to go against the argument you made in https://github.com/reactor/reactor-netty/pull/1218#discussion_r459312327

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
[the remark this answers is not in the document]
violetagg: It is better to not expose at all Reactor Pool API and keep the current approach of loose coupling. This gives Reactor Netty freedom to change the implementation in the future when/if needed in a compatible way.

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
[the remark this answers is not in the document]
violetagg: The functionality below is related to connection pools only, as this interface is for the connection pools, it fits better as part of `ConnectionProvider` interface.
`reactor.netty.resources.ConnectionProvider.HostSpecificSpec`
`reactor.netty.resources.ConnectionProvider.ConnectionPoolSpec`

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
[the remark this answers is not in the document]
violetagg: It is documented here
https://projectreactor.io/docs/netty/release/reference/index.html#_metrics_5
https://projectreactor.io/docs/netty/release/reference/index.html#_metrics_3

and I agree that the javadoc is good to be extended also.

violetagg:
> Issue #1155 is not about abstracting micrometer (although that would be a nice to have benefit), but rather about letting users decide exactly which (and how) meters are registered (in particular wrt explosion of cardinality of the `tags`) isn't it?
> 
> As such, some of the decisions in this PR don't go as far as shielding the user from micrometer.
> 
> If we do want to abstract micrometer away, I think it would be better to do that in a separate PR, because it leaks in other places, wdyt?

Actually the #1155 is not so detailed and more details are available in this Gitter thread https://gitter.im/reactor/reactor-netty?at=5ee65f59013105125a3543fe where the user agreed to have his own implementation that will process the metrics.

The approach of exposing an API so that users can provide their own metrics `processor` is used for the rest of the metrics in Reactor Netty and it is documented for servers/clients.
https://projectreactor.io/docs/netty/release/reference/index.html#_metrics_5 (at the end of the topic)

src/main/java/reactor/netty/resources/ConnectionProvider.java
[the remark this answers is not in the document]
violetagg: Do you mean that the instance will be created even without calling the `Supplier.get()`?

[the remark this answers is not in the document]
violetagg: > Registrars are meant to be lightweight anyway. 

We do not control the user's implementation, we do not know whether it is lightweight or not.

src/main/java/reactor/netty/resources/DefaultPooledConnectionProviderMeterRegistrar.java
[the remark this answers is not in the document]
violetagg: > (custom) registrar that register gauges differently

This is a user's implementation how it is enforced that user's implementation uses Micrometer and not something else?

violetagg:
Please fix the failing tests

```
    java.lang.NullPointerException
        at reactor.netty.resources.PooledConnectionProvider.lambda$acquire$2(PooledConnectionProvider.java:155)
        at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1688)
        at reactor.netty.resources.PooledConnectionProvider.lambda$acquire$3(PooledConnectionProvider.java:146)
        at reactor.core.publisher.MonoCreate.subscribe(MonoCreate.java:57)
        at reactor.netty.http.client.HttpClientConnect$MonoHttpConnect.lambda$subscribe$0(HttpClientConnect.java:327)
        at reactor.core.publisher.MonoCreate.subscribe(MonoCreate.java:57)
        at reactor.core.publisher.FluxRetryWhen.subscribe(FluxRetryWhen.java:79)
        at reactor.core.publisher.MonoRetryWhen.subscribeOrReturn(MonoRetryWhen.java:46)
        at reactor.core.publisher.InternalMonoOperator.subscribe(InternalMonoOperator.java:57)
        at reactor.netty.http.client.HttpClientConnect$MonoHttpConnect.subscribe(HttpClientConnect.java:330)
        at reactor.core.publisher.InternalFluxOperator.subscribe(InternalFluxOperator.java:62)
        at reactor.netty.ByteBufFlux.subscribe(ByteBufFlux.java:327)
        at reactor.core.publisher.InternalMonoOperator.subscribe(InternalMonoOperator.java:64)
        at reactor.core.publisher.MonoDefer.subscribe(MonoDefer.java:52)
        at reactor.core.publisher.InternalMonoOperator.subscribe(InternalMonoOperator.java:64)
        at reactor.netty.ByteBufMono.subscribe(ByteBufMono.java:134)
        at reactor.core.publisher.Mono.subscribe(Mono.java:4213)
        at reactor.core.publisher.Mono.block(Mono.java:1703)
        at reactor.netty.channel.AddressResolverGroupMetricsTest.test(AddressResolverGroupMetricsTest.java:80)
```

src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
```diff
@@ -0,0 +1,103 @@
...
+     * resource is immediately handy (and the pool cannot allocate more resources).
+     * <p>
+     * A pool pending queue might be unbounded, in which case this method returns
+     * {@link Integer#MAX_VALUE}.
+     *
+     * @return the maximum number of pending acquire that can be enqueued by this pool
+     */
+    int getMaxPendingAcquireSize();
+
+    class ConnectionPoolWrapper implements ConnectionPoolMetrics {
```
violetagg, line 64: `ConnectionPoolWrapper` name is a bit misleading as this class is related only to metrics.

src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Get the maximum number of acquire this pool can queue in a pending state when no available
+     * resource is immediately handy (and the pool cannot allocate more resources).
+     * <p>
+     * A pool pending queue might be unbounded, in which case this method returns
+     * {@link Integer#MAX_VALUE}.
+     *
+     * @return the maximum number of pending acquire that can be enqueued by this pool
+     */
+    int getMaxPendingAcquireSize();
```
violetagg, line 62: `getMaxAllocatedSize()` and `getMaxPendingAcquireSize()` are settings that the user will configure when building the `ConnectionProvider`
I think there is no need to expose them with this interface. Wdyt?

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -575,6 +600,7 @@ public final SPEC metrics(boolean metricsEnabled) {
 				}
 			}
 			this.metricsEnabled = metricsEnabled;
+			this.registrar = Objects.requireNonNull(registrar.get());
```
violetagg, line 603: It is better to obtain the instance lazy when it is needed.

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -80,13 +80,15 @@
 	final String                          name;
 	final Map<SocketAddress, PoolFactory> poolFactoryPerRemoteHost = new HashMap<>();
 	final PoolFactory                     defaultPoolFactory;
+	MeterRegistrar 				  		  registrar;
```
violetagg, line 83: `registrar` is not a configuration on ConnectionProvider level, it is related to a particular connection pool.
Let's take this example

```
ConnectionProvider.builder("test")
                  .maxConnections(2)
                  .metrics(true, () -> new CustomImpl1())
                  .forRemoteHost(InetSocketAddress.createUnresolved("localhost", server.port()),
                                 spec -> spec.maxConnections(1).metrics(true, () -> new CustomImpl2()))
                  .build();
```

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -150,10 +152,16 @@ else if (origin instanceof InetSocketAddress &&
 						new PooledConnectionAllocator(bootstrap, poolFactory, opsFactory).pool;
 
 				if (poolFactory.metricsEnabled || BootstrapHandlers.findMetricsSupport(bootstrap) != null) {
-					PooledConnectionProviderMetrics.registerMetrics(name,
+					// in the case metrics are enable at the HttpClient level, the registrar has not been
+					// initialized through the ConnectionPoolSpec, we need to provide the default one
+					if (null == registrar) {
+						registrar = DefaultPooledConnectionProviderMeterRegistrar.INSTANCE;
+					}
```
violetagg, line 159: What do you think about this

```
MeterRegistrar registrar = poolFactory.registrar;
if (null == registrar) {
    // registrar is null when metrics are enabled on HttpClient level
    registrar = DefaultPooledConnectionProviderMeterRegistrar.INSTANCE;
}
```

violetagg:
Add copyright license to the new file

```
Execution failed for task ':licenseMain'.
1050
> License violations were found: /home/runner/work/reactor-netty/reactor-netty/src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java}
1051
```

src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
aneveu: `ConnectionPoolMetricsWrapper`? we struggled to find a proper name for this one indeed 😞

violetagg: `DelegatingConnectionPoolMetrics` ?

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
aneveu: way much better, thanks!

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -384,6 +408,7 @@ public final SPEC metrics(boolean metricsEnabled) {
 				}
 			}
 			this.metricsEnabled = metricsEnabled;
+			this.registrar = registrar;
```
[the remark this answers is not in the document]
violetagg: Better keep the input validation

src/main/java/reactor/netty/resources/DefaultPooledConnectionProviderMeterRegistrar.java
```diff
@@ -29,13 +29,24 @@
...
+ *
+ * Registers gauges for every metric in {@link ConnectionPoolMetrics}.
+ *
+ * Every gauge uses id, poolName and remoteAddress as tags.
+ *
  * @author Violeta Georgieva
  * @since 0.9
- */
-final class PooledConnectionProviderMetrics {
+q */
```
violetagg, line 40: Extra `q` char?

Unused import reactor.pool.InstrumentedPool

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -150,10 +151,14 @@ else if (origin instanceof InetSocketAddress &&
 						new PooledConnectionAllocator(bootstrap, poolFactory, opsFactory).pool;
 
 				if (poolFactory.metricsEnabled || BootstrapHandlers.findMetricsSupport(bootstrap) != null) {
-					PooledConnectionProviderMetrics.registerMetrics(name,
+					MeterRegistrar registrar = poolFactory.registrar != null ?
+							// registrar is null when metrics are enabled on HttpClient level
+							poolFactory.registrar.get() : DefaultPooledConnectionProviderMeterRegistrar.INSTANCE;;
```
violetagg, line 156: extra `;` at the end

violetagg:
> > Issue #1155 is not about abstracting micrometer (although that would be a nice to have benefit), but rather about letting users decide exactly which (and how) meters are registered (in particular wrt explosion of cardinality of the `tags`) isn't it?
> > As such, some of the decisions in this PR don't go as far as shielding the user from micrometer.
> > If we do want to abstract micrometer away, I think it would be better to do that in a separate PR, because it leaks in other places, wdyt?
> 
> Actually the #1155 is not so detailed and more details are available in this Gitter thread https://gitter.im/reactor/reactor-netty?at=5ee65f59013105125a3543fe where the user agreed to have his own implementation that will process the metrics.
> 
> The approach of exposing an API so that users can provide their own metrics `processor` is used for the rest of the metrics in Reactor Netty and it is documented for servers/clients.
> https://projectreactor.io/docs/netty/release/reference/index.html#_metrics_5 (at the end of the topic)

This comment also should be addressed in order to satisfy the requirement by the stakeholder.

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -563,10 +564,34 @@ public final SPEC maxLifeTime(Duration maxLifeTime) {
...
+		 * under the name {@link reactor.netty.Metrics#CONNECTION_PROVIDER_PREFIX}. Applications can
+		 * separately register their own
+		 * {@link io.micrometer.core.instrument.config.MeterFilter filters} associated with this name.
+		 * For example, to put an upper bound on the number of tags produced:
+		 * <pre class="code">
+		 * MeterFilter filter = ... ;
+		 * Metrics.globalRegistry.config().meterFilter(MeterFilter.maximumAllowableTags(CONNECTION_PROVIDER_PREFIX, 100, filter));
+		 * </pre>
+		 * <p>By default this is not enabled.
+		 *
```
violetagg, line 588: Please fix the javadoc

src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -575,6 +600,7 @@ public final SPEC metrics(boolean metricsEnabled) {
 				}
 			}
```
violetagg, line 601: This code is obsolete because stakeholders can use metering system different than Micrometer even if Micrometer is in the class path.

aneveu: This code is indeed obsolete if we accept that the `registrar` might be null in the `metrics(boolean metricsEnabled, Supplier<? extends ConnectionProvider.MeterRegistrar> registrar)` method, in which case it will end up with the default configuration being settled in the `acquire` method of the `ConnectionProvider`. 

I'm also not sure that the boolean `metricsEnabled` as a parameter is relevant: at the opposite of the `HttpClient.metrics` method, in case the boolean value is false, nothing has to be removed of the configuration. So we might simply give it `false` as a default value and pass it to true only when the method is called.

aneveu:
> The deleted test had two tests
> 
> ```
> 	@Test
> 	public void testClientMetricsEnabled() throws Exception {
> 		doTest(ConnectionProvider.create("test", 1), true);
> 	}
> 
> 	@Test
> 	public void testClientMetricsDisabled() throws Exception {
> 		doTest(ConnectionProvider.builder("test").maxConnections(1).metrics(true).lifo().build(),
> 		       false);
> 	}
> ```
> 
> The first was testing connection provider metrics enabled as part of the HttpClient metrics, the second one connection provider metrics enabled from the connection provider configuration.
> 
> This change removed the first test.
> 
> Also there is no test for the new functionality.

We removed it at first because we were puzzled with the fact that connection provider metrics are sometimes driven by HttpClient metrics. But the new functionality is actually covered by the remaining one which has been moved to the `PooledConnectionProviderTest` under the name `testDefaultProviderMetrics`.

However, I noticed that the method called by the test was `HttpClient.metrics(boolean metricsEnabled, @Nullable Function<String, String> uriTagValue)` which is deprecated, shouldn't we use `HttpClient.HttpClient metrics(boolean metricsEnabled, Supplier<? extends HttpClientMetricsRecorder> recorder)` instead?

aneveu:
> > > Issue #1155 is not about abstracting micrometer (although that would be a nice to have benefit), but rather about letting users decide exactly which (and how) meters are registered (in particular wrt explosion of cardinality of the `tags`) isn't it?
> > > As such, some of the decisions in this PR don't go as far as shielding the user from micrometer.
> > > If we do want to abstract micrometer away, I think it would be better to do that in a separate PR, because it leaks in other places, wdyt?
> > 
> > 
> > Actually the #1155 is not so detailed and more details are available in this Gitter thread https://gitter.im/reactor/reactor-netty?at=5ee65f59013105125a3543fe where the user agreed to have his own implementation that will process the metrics.
> > The approach of exposing an API so that users can provide their own metrics `processor` is used for the rest of the metrics in Reactor Netty and it is documented for servers/clients.
> > https://projectreactor.io/docs/netty/release/reference/index.html#_metrics_5 (at the end of the topic)
> 
> This comment also should be addressed in order to satisfy the requirement by the stakeholder.

As for me this is achievable only by keeping the `metrics(boolean metricsEnabled)` method to provide a default value based on the Micrometer implementation for the registrar (only the boolean might be unneeded as mentioned previously) and remove any mention to `DelegatingConnectionPoolMetrics` in the `ConnectionProvider`. Would it be ok for you?

src/main/java/reactor/netty/resources/ConnectionProvider.java
violetagg: > This code is indeed obsolete if we accept that the `registrar` might be null in the `metrics(boolean metricsEnabled, Supplier<? extends ConnectionProvider.MeterRegistrar> registrar)` method, in which case it will end up with the default configuration being settled in the `acquire` method of the `ConnectionProvider`.
> 

We can skip this check for `metrics(boolean metricsEnabled, Supplier<? extends ConnectionProvider.MeterRegistrar> registrar)`, while for `metrics(boolean metricsEnabled)` we definitely need it. This is similar to what we have for `HttpClient.metrics...`
https://github.com/reactor/reactor-netty/blob/d8f7acc1eb00775f7f2eb7681c8c8cb987296b62/src/main/java/reactor/netty/http/client/HttpClient.java#L1109-L1130
https://github.com/reactor/reactor-netty/blob/d8f7acc1eb00775f7f2eb7681c8c8cb987296b62/src/main/java/reactor/netty/http/client/HttpClient.java#L1159-L1163
I was thinking for something like this:

We keep the current `metrics(boolean metricsEnabled)` as it is at the moment, `registrar` is null and we will apply the default in `acquire`

```
public final SPEC metrics(boolean metricsEnabled) {
	if (metricsEnabled) {
		if (!Metrics.isInstrumentationAvailable()) {
			throw new UnsupportedOperationException(
					"To enable metrics, you must add the dependency `io.micrometer:micrometer-core`" +
							" to the class path first");
		}
	}
	this.metricsEnabled = metricsEnabled;
	return get();
}
```

And introduce a new method like this:

```
public final SPEC metrics(boolean metricsEnabled, Supplier<? extends ConnectionProvider.MeterRegistrar> registrar) {
	this.metricsEnabled = metricsEnabled;
	this.registrar = Objects.requireNonNull(registrar, "registrar");
	return get();
}
```

> I'm also not sure that the boolean `metricsEnabled` as a parameter is relevant: at the opposite of the `HttpClient.metrics` method, in case the boolean value is false, nothing has to be removed of the configuration. So we might simply give it `false` as a default value and pass it to true only when the method is called.

`metricsEnabled` is relevant and it is checked in `acquire` method.
`metricsEnabled ` is controlled by the user, see comment https://github.com/reactor/reactor-netty/pull/1218#discussion_r459555510

violetagg:
> > The deleted test had two tests
> > ```
> > 	@Test
> > 	public void testClientMetricsEnabled() throws Exception {
> > 		doTest(ConnectionProvider.create("test", 1), true);
> > 	}
> > 
> > 	@Test
> > 	public void testClientMetricsDisabled() throws Exception {
> > 		doTest(ConnectionProvider.builder("test").maxConnections(1).metrics(true).lifo().build(),
> > 		       false);
> > 	}
> > ```
> > 
> > 
> > The first was testing connection provider metrics enabled as part of the HttpClient metrics, the second one connection provider metrics enabled from the connection provider configuration.
> > This change removed the first test.
> > Also there is no test for the new functionality.
> 
> We removed it at first because we were puzzled with the fact that connection provider metrics are sometimes driven by HttpClient metrics, but this should be covered, indeed.
> 
> But the new functionality on its side is actually covered by the remaining one which has been moved to the `PooledConnectionProviderTest` under the name `testDefaultProviderMetrics`(but is indeed very similar, not sure there is a different proper way to test that precise part of the code)

`PooledConnectionProviderTest#testDefaultProviderMetrics` covers only: ConnectionProvider metrics are enabled, and HttpClient metrics are disabled

```
@Test
public void testClientMetricsDisabled() throws Exception {
	doTest(ConnectionProvider.builder("test").maxConnections(1).metrics(true).lifo().build(),
	       false);
}
```

While the one below: ConnectionProvider metrics are disabled, and HttpClient metrics are enabled

```
@Test
public void testClientMetricsEnabled() throws Exception {
	doTest(ConnectionProvider.create("test", 1), true);
}
```

violetagg:
> and remove any mention to `DelegatingConnectionPoolMetrics` in the `ConnectionProvider`.

I do not see where `DelegatingConnectionPoolMetrics` is mentioned in the `ConnectionProvider`.

aneveu:
> > and remove any mention to `DelegatingConnectionPoolMetrics` in the `ConnectionProvider`.
> 
> I do not see where `DelegatingConnectionPoolMetrics` is mentioned in the `ConnectionProvider`.

Line 161, when the registrar is calling its method `registerMetrics`. `DelegatingConnectionPoolMetrics` is used as a last parameter to retrieve default `ConnectionPoolMetrics`.

violetagg:
> > > and remove any mention to `DelegatingConnectionPoolMetrics` in the `ConnectionProvider`.
> > 
> > 
> > I do not see where `DelegatingConnectionPoolMetrics` is mentioned in the `ConnectionProvider`.
> 
> Line 161, when the registrar is calling its method `registerMetrics`. `DelegatingConnectionPoolMetrics` is used as a last parameter to retrieve default `ConnectionPoolMetrics`.

This is `PooledConnectionProvider`, not `ConnectionProvider` and it creates the object that we provide to the user, why should we remove that?

aneveu:
> `cs` is used as a last parameter to retrieve default `ConnectionPoolMetrics`.

Forget about it, I was confused 🙃

src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
```diff
@@ -0,0 +1,88 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty.resources;
+
+import reactor.pool.InstrumentedPool;
+
+public interface ConnectionPoolMetrics {
+
```
violetagg, line 21: Please switch to tabs instead of spaces
https://github.com/reactor/reactor-netty/blob/master/CONTRIBUTING.md#code-style

src/main/java/reactor/netty/resources/MicrometerPooledConnectionProviderMeterRegistrar.java
```diff
@@ -29,13 +28,24 @@
 import static reactor.netty.Metrics.TOTAL_CONNECTIONS;
 
 /**
+ * Default implementation of {@link reactor.netty.resources.PooledConnectionProvider.MeterRegistrar}.
```
violetagg, line 31: {@link ConnectionProvider.MeterRegistrar} instead of {@link reactor.netty.resources.PooledConnectionProvider.MeterRegistrar}

src/test/java/reactor/netty/resources/PooledConnectionProviderDefaultMetricsTest.java
```diff
@@ -0,0 +1,172 @@
...
+import static reactor.netty.Metrics.PENDING_CONNECTIONS;
+import static reactor.netty.Metrics.NAME;
+import static reactor.netty.Metrics.REMOTE_ADDRESS;
+import static reactor.netty.Metrics.TOTAL_CONNECTIONS;
+
+/**
+ * @author Violeta Georgieva
+ */
+public class PooledConnectionProviderDefaultMetricsTest {
+    private MeterRegistry registry;
```
violetagg, line 53: Please switch to tabs instead of spaces

src/test/java/reactor/netty/resources/PooledConnectionProviderCustomMetricsTest.java
```diff
@@ -0,0 +1,73 @@
...
+	@Test
+	public void customRegistrarIsUsed() {
+		AtomicBoolean used = new AtomicBoolean();
+		ConnectionProvider pool = ConnectionProvider.builder("test")
+				.metrics(true, () -> (a, b, c, d) -> used.set(true))
+				.build();
+
+		Bootstrap bootstrap = new Bootstrap().remoteAddress("localhost", 0)
+			.channelFactory(NioSocketChannel::new)
+			.group(new NioEventLoopGroup(2));
```
violetagg, line 39: Shut down NioEventLoopGroup at the end of the test.

src/test/java/reactor/netty/resources/PooledConnectionProviderCustomMetricsTest.java
```diff
@@ -0,0 +1,73 @@
...
+	@Test
+	public void customRegistrarSupplierNotInvokedWhenMetricsDisabled() {
+		AtomicBoolean used = new AtomicBoolean();
+		ConnectionProvider pool = ConnectionProvider.builder("test")
+				.metrics(false, () -> {used.set(true); return null;})
+				.build();
+
+		Bootstrap bootstrap = new Bootstrap().remoteAddress("localhost", 0)
+				.channelFactory(NioSocketChannel::new)
+				.group(new NioEventLoopGroup(2));
```
violetagg, line 61: Shut down NioEventLoopGroup at the end of the test.

src/test/java/reactor/netty/resources/PooledConnectionProviderCustomMetricsTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+	@After
+	public void tearDown() throws Exception {
+		bootstrap.config().group().shutdownGracefully().get(10L, TimeUnit.SECONDS);
+		pool.dispose();
+	}
+
+
+	@Test
+	public void customRegistrarIsUsed() throws Exception {
```
violetagg, line 54: `throws Exception` is not needed.

violetagg:
@aneveu @ericbottard Please rebase and squash the changes.

PR #1229, opened 2020-07-28
fix #1223 Update `how to use BOM with gradle` section in the reference documentation

### LINKED ISSUE #1223: Misleading statement about Gradle support for Maven BOMs
"Gradle has no core support for Maven BOMs"
https://projectreactor.io/docs/netty/release/reference/index.html#_gradle_installation

This hasn't been true since Gradle 5, which was released in Nov, 2018.

```
implementation(platform("io.projectreactor:reactor-bom:Dysprosium-SR10"))
```

violetagg:
Fixes #1223


violetagg:
> maybe take that as an opportunity to change the BOM in the example to `Dysprosium-RELEASE` and make a note that this is an example, not the latest one ?

Sure let me do that

PR #1236, opened 2020-07-31
fix #739 Separate Reactor Netty on modules

### LINKED ISSUE #739: Consider making netty-codec-http[2] dependencies optional
While working with `r2dbc-postgresql` and WebFlux i a Gradle project, I faced a very annoying issue where the [exclusions](https://github.com/r2dbc/r2dbc-postgresql/blob/efca10386046e70abec93a35154546ed4cca6802/pom.xml#L87-L99)  of `netty-codec-http` and `netty-codec-http2` in `r2dbc-postgresql` break completely WebFlux since it removes the `netty-codec-http` dependency from the project. We are going to remove for now these excludes from R2DBC but that just a workaround since it will now pull some dependencies it does not need.

Since `reactor-netty` usage is wider than just HTTP (R2DBC is a good example), my proposal is to make these 2 dependencies optionals and update consumer projects like WebFlux to add these explicitly.

violetagg:
Fixes #739


reactor-netty-core/src/main/java/reactor/netty/ReactorNetty.java
```diff
@@ -335,7 +334,7 @@ static void addHandlerAfterReactorCodecs(Connection context, String
 
 	static boolean mustChunkFileTransfer(Connection c, Path file) {
 		// if channel multiplexing a parent channel as an http2 stream
-		if (c.channel().parent() != null && c.channel().parent().pipeline().get(Http2ConnectionHandler.class) != null) {
+		if (c.channel().parent() != null && c.channel().parent().pipeline().get(NettyPipeline.H2MultiplexHandler) != null) {
```
simonbasle, line 337: isn't that an unrelated change?

reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
```diff
@@ -629,6 +631,30 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
...
 
 	static final Logger log = Loggers.getLogger(SslProvider.class);
+
+	/**
+	 * <a href="https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility">Mozilla Modern Cipher
+	 * Suites</a> minus the following cipher suites that are black listed by the
+	 * <a href="https://tools.ietf.org/html/rfc7540#appendix-A">HTTP/2 RFC</a>.
+	 * Copied from io.netty.handler.codec.http2.Http2SecurityUtil
+	 */
+	static final List<String> HTTP2_TLS12_CIPHERS =
```
simonbasle, line 641: it seems there's another unrelated change that creeped in the PR

simonbasle: ah got it, it was using something from netty's http2 artifact previously and we don't want that dependency anymore 👍

reactor-netty-core/src/main/java/reactor/netty/ReactorNetty.java
violetagg: `reactor-netty-core` does not have dependency to `netty-codec-http2`
this change removes the usage of `Http2ConnectionHandler` which is in `netty-codec-http2`

reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
violetagg: `reactor-netty-core` does not have dependency to `netty-codec-http2`
this change removes the usage of `Http2SecurityUtil` which is in `netty-codec-http2`

build.gradle
```diff
@@ -13,29 +13,30 @@
...
-	id "com.github.hierynomus.license" version "0.15.0"
+	id 'com.github.hierynomus.license' version '0.15.0'
 	id 'org.asciidoctor.convert' version '1.5.11'
-	id 'com.github.johnrengelman.shadow' version '5.2.0'
-	id "com.jfrog.artifactory" version "4.9.8" apply false
-	id "de.undercouch.download" version "4.1.1"
-	id 'me.champeau.gradle.japicmp' version '0.2.9'
+	id 'com.google.osdetector' version '1.6.2'
+	id 'com.github.johnrengelman.shadow' version '5.2.0' apply false
+	id 'com.jfrog.artifactory' version '4.9.8' apply false
```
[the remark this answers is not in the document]
violetagg: sure let me do that
in a separate change I intend to bump the gradle version

Issue #1224, opened 2020-07-27
TLSv1.3 not working with HTTP/2 for HttpServer using JDK as SSL Provider


sephiroth-j:
If an HttpServer is configured to use TLSv1.3 and HTTP/2, connections using HTTP/2 and TLSv1.3 can not be established. TLSv1.3 works only without HTTP/2. If TLSv1.2 and TLSv1.3 are configured, only TLSv1.2 can be used. If only TLSv1.3 is configured, no connection can be established at all.

## Expected Behavior
Connecting to an HttpServer using HTTP/2 and TLSv1.3 should be possible.

## Actual Behavior
HTTP/2 connections with TLSv1.3 can not be established and the following exceptions is thrown.

```
(reactor-http-nio-3) [id: 0xa7207b9e, L:/127.0.0.1:8080 ! R:/127.0.0.1:50370] onUncaughtException(SimpleConnection{channel=[id: 0xa7207b9e, L:/127.0.0.1:8080 ! R:/127.0.0.1:50370]}) - io.netty.handler.codec.DecoderException: javax.net.ssl.SSLHandshakeException: The client supported protocol versions [TLSv1.3] are not accepted by server preferences [TLS12]
io.netty.handler.codec.DecoderException: javax.net.ssl.SSLHandshakeException: The client supported protocol versions [TLSv1.3] are not accepted by server preferences [TLS12]
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:471)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:276)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
[...]

PR #1239, opened 2020-08-02
do not set cipher suites while configuring HTTP/2

sephiroth-j:
the cipher list from `Http2SecurityUtil.CIPHERS` does not contain ciphers for TLSv1.3 supported by JDK 11+, namely TLS_AES_128_GCM_SHA256 and TLS_AES_256_GCM_SHA384.
the list contains TLS_CHACHA20_POLY1305_SHA256 which is only supported by OpenSSL.
this results in TLSv1.3 being unavailable when HTTP/2 is enabled and JDK 11+ is used as SSL provider.

leave it up to Netty to set appropriate default ciphers (see `JdkSslContext`, `ReferenceCountedOpenSslContext` and `SslUtils`).

**bonus**:
- use of the unstable api `Http2SecurityUtil` removed
- users can define their own list of ciphers in `SslContextBuilder` and still use `DefaultConfigurationType.H2`

fixes gh-issue #1224

p.s.
A word about why I removed `JdkSslContext` from `SslProviderTests`. Since netty-tcnative-boringssl-static is part of the test runtime, all these checks on `OpenSslContext` or `JdkSslContext` always returned `OpenSslContext`. If one day this test no longer uses netty-tcnative-boringssl-static, then further changes are required anyway.

### DIFF
--- a/src/main/java/reactor/netty/http/client/HttpClientSecure.java
+++ b/src/main/java/reactor/netty/http/client/HttpClientSecure.java
@@ -21,13 +21,11 @@ import javax.annotation.Nullable;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLParameters;
 
-import io.netty.handler.codec.http2.Http2SecurityUtil;
 import io.netty.handler.ssl.ApplicationProtocolConfig;
 import io.netty.handler.ssl.ApplicationProtocolNames;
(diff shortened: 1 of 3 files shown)


violetagg:
Closing this PR as an updated list with ciphers will be available with Netty 4.1.52.Final

PR #1241, opened 2020-08-02
Move the documentation examples to the examples submodule

violetagg:
When there is a change in the API, the samples code will be adapted immediately and the documentation will be up-to-date.

### DIFF
--- a/docs/asciidoc/http-client.adoc
+++ b/docs/asciidoc/http-client.adoc
@@ -1,4 +1,5 @@
 :sourcedir: ./../../reactor-netty-http/src/main/java
+:examplesdir: ./../../reactor-netty-examples/src/main/java/reactor/netty/examples/documentation/http/client
 :javadoc: https://projectreactor.io/docs/netty/release/api
 
 [[http-client]]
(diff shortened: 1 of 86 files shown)


docs/asciidoc/http-client.adoc
```diff
@@ -16,22 +17,10 @@ To connect the `HTTP` client to a given `HTTP` endpoint, you must create and con
 The following example shows how to do so:
 
 ====
-[source,java]
+[source,java,indent=0]
+./../../reactor-netty-examples/src/main/java/reactor/netty/examples/documentation/http/client/connect/Application.java
```
simonbasle, line 21: this seems redundant with what's below. can't it use the `{examplesdir}` as well ?

violetagg: I wasn't able to find a way to do that. Do you know how it can be done?

violetagg: scratch that, found it

violetagg: done

PR #1243, opened 2020-08-03
Adapt to the changed API in reactor core

violetagg:


maxim-bandurko-lsvt:
@violetagg @simonbasle Is it possible to release this update to Spring Boot 2.4.0-SNAPSHOT? Thing is that recent reactor core updates completely broke RSocket and will be very thankful if this fix with `signal.getContextView()` will be updated before August 10.

Thank you.

maxim-bandurko-lsvt:
Actually I am wrong. Reactor Core at Spring Boot 2.4.0-SNAPSHOT still has old version with not implemented `getContextView` method

PR #1248, opened 2020-08-05
Adapt to Reactor Core changes

violetagg:
Adapt to the Processor changes
Adapt to the Context changes

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
@@ -119,8 +119,8 @@ class HttpClientConnect extends HttpClient {
 
 		if (config.doOnRequestError != null) {
 			mono = mono.onErrorResume(error ->
-					Mono.subscriberContext()
-					    .doOnNext(ctx -> config.doOnRequestError.accept(new FailedHttpClientRequest(ctx, config), error))
+					Mono.deferContextual(Mono::just)
+					    .doOnNext(ctx -> config.doOnRequestError.accept(new FailedHttpClientRequest(Context.of(ctx), config), error))
 					    .then(Mono.error(error)));
 		}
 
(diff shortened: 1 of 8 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -119,8 +119,8 @@ public String toString() {
 
 		if (config.doOnRequestError != null) {
 			mono = mono.onErrorResume(error ->
-					Mono.subscriberContext()
-					    .doOnNext(ctx -> config.doOnRequestError.accept(new FailedHttpClientRequest(ctx, config), error))
+					Mono.deferContextual(Mono::just)
+					    .doOnNext(ctx -> config.doOnRequestError.accept(new FailedHttpClientRequest(Context.of(ctx), config), error))
```
simonbasle, line 123: would it make sense for the `FailedHttpClientRequest` to accept `ContextView` instead of `Context`?

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientFormEncoder.java
```diff
@@ -49,7 +48,7 @@
 final class HttpClientFormEncoder extends HttpPostRequestEncoder
 		implements ChunkedInput<HttpContent>, Runnable, HttpClientForm {
 
-	final FluxIdentityProcessor<Long> progressFlux;
+	final Sinks.Many<Long> progressFlux;
```
simonbasle, line 51: could be renamed to `progressSink` ;)

violetagg: ok

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
violetagg: `FailedHttpClientRequest` provides API 

```
	@Override
	public Context currentContext() {
		return context;
	}
```

So we need a `Context`

Issue #1230, opened 2020-07-29
Disposing PooledConnections that are not actively used under LRU policy.


jun-lee:
## Motivation

My application is using **HttpClient** to access a massive number of endpoints to provide a **Link Preview** service. Over time, Netty-related resources are taking a lot of memory as shown below. Mostly by the PooledConnections, they have never been destroyed once they have been initialized for a unique endpoint with the InetSocketAddress as a key, and redundant data like HeaderEntity, DefaultAttribute in each config of the pool.

<img width="1233" alt="스크린샷 2020-07-29 오전 11 53 24" src="https://user-images.githubusercontent.com/12035676/88767519-f0ea9000-d1a3-11ea-84ce-d61cee6efef1.png">

## Desired solution

Could we have an LRU mechanism to release the old PooledConnections which are not actively reused?

violetagg:
@jun-lee Do you think that this feature request will serve your requirement?
https://github.com/reactor/reactor-pool/issues/86

For which version do you need this feature: 0.9.x or 1.0.0?

/cc @simonbasle

jun-lee:
@violetagg Thanks for the link. Yes, that's exactly what I'm expecting. And I'm using 1.0.0-M1 because of the Multi Content-Length issues.

violetagg:
@jun-lee Multi Content-Length issues?

jun-lee:
@violetagg this one https://github.com/netty/netty/pull/10349

violetagg:
@jun-lee all Reactor Netty releases depend on the latest released Netty version at that point of the release.
So 0.8.20 and 0.9.10 releases use 4.1.51.Final
https://github.com/reactor/reactor-netty/releases/tag/v0.9.10.RELEASE
https://github.com/reactor/reactor-netty/releases/tag/v0.8.20.RELEASE

jun-lee:
@violetagg Sorry, after checking my code again. This is not related to 1.0.0-M1, but somehow, I was using spring boot 2.4.0-M1, therefore, 1.0.0-M1 was used.

violetagg:
@jun-lee Ok so we are targeting 1.0.0? I need this in order to be able to fix the version where this feature request is needed.

jun-lee:
@violetagg Yes, we are.

violetagg:
This is depending on reactor/reactor-pool#86

PR #1259, opened 2020-08-11
#1230 Disposing ConnectionPool that are not actively used under LRU policy.

jun-lee:
Based on the issue(https://github.com/reactor/reactor-netty/issues/1230), I'm thinking that it can be on the `reactor-netty` application layer as well, instead of adding consumer logic into the `reactor-pool`.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -16,7 +16,6 @@
 package reactor.netty.resources;
 
 import io.netty.resolver.AddressResolverGroup;
-import io.netty.util.internal.PlatformDependent;
 import org.reactivestreams.Publisher;
 import reactor.core.CoreSubscriber;
 import reactor.core.Disposable;
@@ -41,11 +40,11 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.time.Duration;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.WeakHashMap;
-import java.util.concurrent.ConcurrentMap;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@jun-lee Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1259)!

[Click here](https://cla.pivotal.io/sync/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1259) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jun-lee Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1259)!

violetagg:
@jun-lee With issue #1230 do you want to remove the connections that reached their TTL within a connection pool or to remove a connection pool from the map with all connections pool?

The feature request in Reactor Pool is for removing the connections that reached their TTL within a connection pool. While with this PR you want to remove a connection pool from the map with all connection pools.

jun-lee:
@violetagg 
Because I don't see how the resources referenced by PoolKey can be garbage collected unless the entry is not removed from the map after it has established. Let's say a ConnectionPool has 10 connections that reached to TTL so that a process of `reactor-pool` evicted all the connections belong to the Connection Pool, so now we have empty ConnectionPool on the 1channelPools` Map. However, the `PoolKey` and cached resource in the Connection Pool that I shared the Screenshot on the issue would not be removed.

In my application, if I'm using enough numbers for the limit, removing entry by LRU policy is a kind of desired feature, rather than keeping them a long time or causing another heavy evicting operation that checks that all the connections in a pool have reached to TTL. I have tested that the simultaneous new request that establishing a new ConnectionPool to the destroying connection pool was not causing any problem either.

violetagg:
> @violetagg
> Because I don't see how the resources referenced by PoolKey can be garbage collected unless the entry is removed from the map after it has established. Let's say a ConnectionPool is holding 10 subsequent connections that reached to TTL so that a process of `reactor-pool` evicted all the connections belong to the Connection Pool,

We still do not have this background eviction from Reactor Pool, so currently the connections are checked for TTL only when they are released/acquired.

> so now we have empty ConnectionPool on the `channelPools` Map. However, the `PoolKey` and cached resource in the Connection Pool that I shared the Screenshot on the issue would not be removed.

Let say that the connection pool is empty then the cleanup from the map should not be done based on the eldest but based on `InstrumentedPool.metrics().allocatedSize() == 0`

> 
> In my application, if I'm using enough numbers for the limit, removing entry by LRU policy is a kind of the desired feature, rather than keeping them a long time or causing another heavy evicting operation that checks that all the connections in a pool have reached to TTL. I have tested that the simultaneous new request that establishing a new ConnectionPool to the destroying connection pool was not causing any problem either.

Let say that you destroy a connection pool with connections (active or idle). Then you are right this will not cause problems as the next request for a connection will create it again, BUT then you have to pay the price for establishing again a new connection to the remote host.

jun-lee:
Yeah this PR is not supposed to merged. We need to hit the map by calling get() on the acquiring Operation as well to put the most heavy one at the recently accessed position. I just want to point that the root cause of my application was the growing entities of the map holding the connection pools. If there is any way to clean up the connection pool and related PoolKey that holding InetAddress of the remote host, then it would work for my application.

violetagg:
@jun-lee So from what I understood you connect to many different remote hosts, is this correct?
We provide this method `reactor.netty.resources.ConnectionProvider#disposeWhen` will it help?

jun-lee:
Yes, that's what I have just found and I now can handle the LRU from my applicaion side. And it works well. Thanks for the help.

Issue #1257, opened 2020-08-11
Target host should not be resolved when proxy is configured


anilgursel:
When HTTP client is configured with proxy, it should not try to resolve the target host.  Resolving the target address should be proxy's responsibility:

```java
HttpClient.create()
    .baseUrl(""http://some-random-locally-not-resolvable-address-12345.com"")
    .metrics(metricsEnabled, () -> MicrometerHttpClientMetricsRecorder.INSTANCE)
    .proxy(proxyOptions);
```

## Expected Behavior

Target host should not be attempted to be resolved.

## Actual Behavior

- 0.9.x:
  - Metrics disabled: Works as expected and does not attempt to resolve the target host.
  - Metrics enabled: throws `UnknownHostException`.

- 1.0.0-M1: thrown `UnknownHostException`.

## Steps to Reproduce

I will be sharing a PR shortly.

PR #1260, opened 2020-08-11
#1257 Target host should not be resolved when proxy is configured

anilgursel:


pivotal-issuemaster:
@anilgursel Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1260)!

[Click here](https://cla.pivotal.io/sync/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1260) to manually synchronize the status of this Pull Request.

violetagg:
@anilgursel Please sign the CLA

pivotal-issuemaster:
@anilgursel Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1260)!

anilgursel:
@violetagg just signed.  Thanks!

src/test/java/reactor/netty/http/client/HttpClientProxyTest.java
```diff
@@ -165,13 +170,52 @@ public void testIssue804() {
 				    .verify(Duration.ofSeconds(30));
 	}
 
-	private Mono<Tuple2<String, HttpHeaders>> sendRequest(
+    @Test
+    public void shouldNotResolveTargetHostnameWhenMetricsEnabled() {
```
violetagg, line 174: nit: use tabs not spaces

violetagg:
@anilgursel I see one additional problem but I'll create another issue for it

src/test/java/reactor/netty/http/client/HttpClientProxyTest.java
anilgursel: @violetagg I was trying to be careful with the tabs overall, but apparently I missed it in a couple of places.  All should have been addressed with the recent commits.

src/main/java/reactor/netty/channel/BootstrapHandlers.java
```diff
@@ -487,6 +489,22 @@ public static Bootstrap removeMetricsSupport(Bootstrap b) {
 		return b;
 	}
 
+	public static Bootstrap updateResolverForProxySupport(Bootstrap b) {
```
simonbasle, line 492: this will need some javadoc

src/main/java/reactor/netty/tcp/TcpUtils.java
```diff
@@ -41,9 +41,7 @@ static Bootstrap updateProxySupport(Bootstrap b, ProxyProvider proxyOptions) {
 				NettyPipeline.ProxyHandler,
 				new ProxyProvider.DeferredProxySupport(proxyOptions));
 
-		if (b.config().resolver() == DefaultAddressResolverGroup.INSTANCE) {
-			return b.resolver(NoopAddressResolverGroup.INSTANCE);
-		}
+		BootstrapHandlers.updateResolverForProxySupport(b);
```
ericbottard, line 44: This does not use the return value of the method. It's better to either mutate the passed in parameter OR return a value to be used, but not both IMHO.

violetagg: @anilgursel Yes just return the result from `BootstrapHandlers.updateResolverForProxySupport(b)`.
@ericbottard Just to clarify that we do not clone the `b`.

anilgursel: The only reason I did this way is to follow the convention few lines above.  I am making the change now.

src/main/java/reactor/netty/channel/BootstrapHandlers.java
anilgursel: Added Javadoc.  Please let me know if you would like more details.

src/main/java/reactor/netty/channel/BootstrapHandlers.java
```diff
@@ -489,6 +489,10 @@ public static Bootstrap removeMetricsSupport(Bootstrap b) {
 		return b;
 	}
 
+	/**
+	 * Changes the resolver to {@link NoopAddressResolverGroup} if the current one is
+	 * {@link DefaultAddressResolverGroup}.
+	 */
 	public static Bootstrap updateResolverForProxySupport(Bootstrap b) {
```
simonbasle, line 496: I wonder if the name of the method shouldn't be changed to something more meaningful like `forceNoopResolver`, or something like that. Here the name makes me think it checks `Bootstrap` for proxy configuration. And also makes me think it sets up a meaningful `resolver`, not merely a `noop` one.

violetagg: It forces to `Noop...` only if there is no custom resolver. Also it takes care to update the resolver used when metrics are enabled. So I think `updateResolverForProxySupport` describes exactly what happens.

simonbasle: well, if resolver is already noop, then the end result is the same ;) but fair enough.

violetagg:
@anilgursel I squashed the change and applied it to 0.9.x (6afaa71)

violetagg:
@anilgursel Thanks for the PR

PR #1264, opened 2020-08-12
fix #1257 Target host should not be resolved when proxy is configured (1.0.0)

### LINKED ISSUE #1257: Target host should not be resolved when proxy is configured

violetagg:
A forward merge of the fix for #1257 (0.9.x) cannot be achieved because of
the changes in the implementation, so the fix for 1.0.0 is done separately.

### DIFF
--- a/reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientProxyTest.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientProxyTest.java
@@ -17,12 +17,13 @@ package reactor.netty.http.client;
 
 import io.netty.channel.ChannelHandler;
 import io.netty.handler.codec.http.HttpHeaders;
+import io.netty.resolver.AddressResolverGroup;
+import io.netty.resolver.NoopAddressResolverGroup;
 import io.specto.hoverfly.junit.core.HoverflyConfig;
 import io.specto.hoverfly.junit.rule.HoverflyRule;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.ClassRule;
-import org.junit.Ignore;
 import org.junit.Test;
 import reactor.core.publisher.Mono;
 import reactor.netty.DisposableServer;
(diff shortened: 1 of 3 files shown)


reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientProxyTest.java
```diff
@@ -203,6 +204,39 @@ public void shouldNotResolveTargetHostnameWhenMetricsDisabled() {
...
+		client.proxy(ops -> ops.type(ProxyProvider.Proxy.HTTP)
+		                       .host("localhost")
+		                       .port(hoverflyRule.getProxyPort()))
+		      .doOnConnect(config -> resolver2.set(config.resolver()))
+		      .get()
+		      .uri(LOCALLY_NOT_RESOLVABLE_ADDRESS)
+		      .responseSingle((response, body) -> Mono.zip(body.asString(),
+		                                                   Mono.just(response.responseHeaders())))
+		      .block(Duration.ofSeconds(30));
+
```
anilgursel, line 233: Should we add another http call here after the proxy just to verify that we are still able to make the calls without proxy and proxy configuration actually modified a copy?

```java
client.doOnConnect(config -> resolver1.set(config.resolver()))
		      .get()
		      .uri("http://example.com")
		      .responseSingle((response, body) -> Mono.zip(body.asString(),
		                                                   Mono.just(response.responseHeaders())))
		      .block(Duration.ofSeconds(30));
```

violetagg: yep let's extend the test

Issue #436, opened 2018-09-17
Provide more options for customising the access log


violetagg:

jkubrynski:
Is there any idea of this customization? Should it be on a code-level or just a string format parsed by the AccessLog ?

sonofelice:
I also have the same problem. I want to add some info from httpServeletRequest, like the access log of tomcat.How can I customize the log format?

sonofelice:
Is there any progress here?

jkubrynski:
I offered a PR, but there is no decision on how this customization should look like. Programmatic or string pattern

sonofelice:
I did not find your pr link. Would you like put the link here?

sonofelice:
> I offered a PR, but there is no decision on how this customization should look like. Programmatic or string pattern

I did not find your pr link. Would you like put the link here?

naturallight:
Is there any update?

ek-ex:
Could you also include the request headers to this enhancement ? Thank you

violetagg:
Fixed with #1357

Issue #233, opened 2017-12-04
How to debug http request & responses with body in a human readable way


thekalinga:
Currently if we try to log the request & response, the libray logs the request & response bodies in a machine readable way (I'm not sure what the format is). I am not sure which library can be used to view these logged requests & responses in a human readable way. Would be great if the request can be logged in a human readable way directly instead of using special tools.

### Expected behavior
Ability to view http request & response in a readable & developer friendly way

### Actual behavior
Some hex structure (not sure what this format is) is getting logged instead

```
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 47 45 54 20 2f 53 65 61 72 63 68 5f 47 43 2e 61 |GET /Search_GC.a|
|00000010| 73 70 78 20 48 54 54 50 2f 31 2e 31 0d 0a 75 73 |spx HTTP/1.1..us|
|00000020| 65 72 2d 61 67 65 6e 74 3a 20 52 65 61 63 74 6f |er-agent: Reacto|
|00000030| 72 4e 65 74 74 79 2f 30 2e 37 2e 32 2e 52 45 4c |rNetty/0.7.2.REL|
|00000040| 45 41 53 45 0d 0a 68 6f 73 74 3a 20 63 65 6f 6b |EASE..host: ceok|
|00000050| 61 72 6e 61 74 61 6b 61 2e 6b 61 72 2e 6e 69 63 |arnataka.kar.nic|
|00000060| 2e 69 6e 0d 0a 61 63 63 65 70 74 3a 20 2a 2f 2a |.in..accept: */*|
|00000070| 0d 0a 61 63 63 65 70 74 2d 65 6e 63 6f 64 69 6e |..accept-encodin|
|00000080| 67 3a 20 67 7a 69 70 0d 0a 63 6f 6e 74 65 6e 74 |g: gzip..content|
|00000090| 2d 6c 65 6e 67 74 68 3a 20 30 0d 0a 0d 0a       |-length: 0....  |
+--------+-------------------------------------------------+----------------+
```

### Steps to reproduce
Enable debugging for package reactor.ipc.netty

### Reactor Netty version
0.7.2.RELEASE

### JVM version (e.g. `java -version`)
JRE: 1.8.0_152-release-1024-b6 amd64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
[...]

PR #1266, opened 2020-08-12
Added possibility to use a custom logging handler. #436

magoeke:
It's a pretty straight forward and simple solution. I added an additional "wiretap" method to Transport that takes a LoggingHandler. 

Another way would be to pass a class that extends ChannelDuplexHandler instead of LoggingHandler.

Let me know what you think about it. We can also discuss different solutions.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/Transport.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/Transport.java
@@ -137,8 +137,8 @@ public abstract class Transport<T extends Transport<T, C>, C extends TransportCo
 		if (enable) {
 			if (!Metrics.isInstrumentationAvailable()) {
 				throw new UnsupportedOperationException(
-						"To enable metrics, you must add the dependency `io.micrometer:micrometer-core`" +
-								" to the class path first");
+					"To enable metrics, you must add the dependency `io.micrometer:micrometer-core`" +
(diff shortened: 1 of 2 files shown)


violetagg:
@magoeke #436 or #233 ?

violetagg:
for #233 I was thinking an API that enables this functionality and not asking people to provide `LoggingHandler`

Providing `LoggingHandler` you can do even with the current API using reactor.netty.transport.Transport#doOnChannelInit

magoeke:
Probably #233 fits better with what I was doing.

So your idea is to add something like a "format" method to the Transport class ( e.g. String format(ByteBuf msg) )?

violetagg:
@magoeke I haven't thought about how this can be implemented but just as a rough idea: can we do something around the new method with `ByteBufFormat`? Reactor Netty can expose its own enum with `ByteBufFormat` that contains the two from Netty and one additional. When we have `wiretap` with those from Netty we will use directly `LoggingHandler`, when it's the one that will be responsible for the non hex format then we can add our `CustomLoggingHandler`. Something around this?

magoeke:
@violetagg sure sounds like a good idea. I'll have a look at it later. I think I'll create a new PR and link the commits to  #233.

violetagg:
Superseded by #1270

PR #1273, opened 2020-08-17
Ensure HttpClient's default SslProvider takes in account the configured protocol

violetagg:


reactor-netty-http/src/test/java/reactor/netty/http/Http2Tests.java
```diff
@@ -414,4 +415,34 @@ public void testHttp2ForMemoryLeaks() throws Exception {
...
+
+	@Test
+	public void testHttpClientDefaultSslProvider() {
+		HttpClient client = HttpClient.create()
+		                              .wiretap(true);
+
+		doTestHttpClientDefaultSslProvider(client.protocol(HttpProtocol.H2));
+		doTestHttpClientDefaultSslProvider(client.protocol(HttpProtocol.H2)
+		                               .secure());
+		doTestHttpClientDefaultSslProvider(client.secure()
```
ericbottard, line 427: What's the difference with the line above?

reactor-netty-http/src/test/java/reactor/netty/http/Http2Tests.java
```diff
@@ -414,4 +415,34 @@ public void testHttp2ForMemoryLeaks() throws Exception {
...
+		                               .protocol(HttpProtocol.H2));
+		doTestHttpClientDefaultSslProvider(client.protocol(HttpProtocol.HTTP11)
+		                               .secure()
+		                               .protocol(HttpProtocol.H2));
```
ericbottard, line 431: Is this intended?

reactor-netty-http/src/test/java/reactor/netty/http/Http2Tests.java
violetagg: - When the configuration is `client.protocol(HttpProtocol.H2)` and the url starts with `https` then we add the default `SslProvider` in `HttpClientConnect` - https://github.com/reactor/reactor-netty/pull/1273/files#diff-23ed16da152b0327324bd951bd35198dR209
- When the configuration is `client.protocol(HttpProtocol.H2).secure()` or `client.secure().protocol(HttpProtocol.H2)`, we ensure that the corresponding methods take into account the current configuration. See `HttpClient#protocol` and `HttpClient#secure`.
Let's say we have `client.secure().protocol(HttpProtocol.H2)`, when `secure` is invoked we do not have info for the protocol so we add the default `SslProvider` for `HTTP/1.1`, later when the protocol is set we need to change the `SslProvider`

reactor-netty-http/src/test/java/reactor/netty/http/Http2Tests.java
violetagg: Yes checking that applying a protocol configuration multiple times sets the correct `SslProvider`.
Let's say we have this:

```
HttpClient client1 =
    HttpClient....
           .protocol(HttpProtocol.HTTP11)
	   .secure();
HttpClient client2 = client1.protocol(HttpProtocol.H2)
```

`client1` should run with the default `SslProvider` for `HTTP/1.1`
`client2` for this for `HTTP/2`

PR #1278, opened 2020-08-18
fix #573 Add server SNI support

### LINKED ISSUE #573: add server SNI support
We want to use `spring-cloud-gateway` to fully replace `nginx`, so `SNI` support is needed. currently `HttpServer` can have only one SslContext, although I can add a `SniHandler` myself, but if your framework can support this, it would be better.

violetagg:
Fixes #573


violetagg:
@smaldini I added the SNI feature as part of the SslProvider. PTAL

rstoyanchev (approved):

PR #1297, opened 2020-09-01
Add support to use URI mapper with custom metrics recorder

fkjellberg:
Add support to use URI mapper function together with custom metrics recorder. It's more efficient to let HttpClientMetricsHandler and HttpServerMetricsHandler call the mapper function before calling the recorder methods compared to do the mapping in the custom metrics recorder since the handlers will reuse an already mapped value when calling different methods in the recorder.

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -1132,6 +1132,24 @@ public abstract class HttpClient extends ClientTransport<HttpClient, HttpClientC
 		return super.metrics(enable, recorder);
 	}
 
+	public final HttpClient metrics(boolean enable, Supplier<? extends ChannelMetricsRecorder> recorder, Function<String, String> uriTagValue) {
+		if (enable) {
+			HttpClient dup = duplicate();
+			dup.configuration().metricsRecorder(recorder);
+			dup.configuration().uriTagValue = uriTagValue;
+			return dup;
+		}
+		else if (configuration().metricsRecorder() != null) {
+			HttpClient dup = duplicate();
+			dup.configuration().metricsRecorder(null);
+			dup.configuration().uriTagValue = null;
+			return dup;
+		}
+		else {
+			return this;
+		}
+	}
+
 	/**
 	 * Removes any previously applied SSL configuration customization
 	 *
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -1132,6 +1132,21 @@ public abstract class HttpClient extends ClientTransport<HttpClient, HttpClientC
 		return super.metrics(enable, recorder);
 	}
 
+	/**
+	 * Specifies whether the metrics are enabled on the {@link HttpClient}.
+	 * All generated metrics are provided to the specified recorder which is only
+	 * instantiated if metrics are being enabled.
+	 * <p>{@code uriTagValue} function receives the actual uri and returns the uri tag value
+	 * that will be used for the metrics with {@link reactor.netty.Metrics#URI} tag.
+	 * For example instead of using the actual uri {@code "/users/1"} as uri tag value, templated uri
+	 * {@code "/users/{id}"} can be used.
+	 *
+	 * @param enable true enables metrics collection; false disables it
+	 * @param recorder a supplier for the metrics recorder that receives the collected metrics
+	 * @param uriTagValue a function that receives the actual uri and returns the uri tag value
+	 * that will be used for the metrics with {@link reactor.netty.Metrics#URI} tag
+	 * @return a new {@link HttpClient}
+	 */
 	public final HttpClient metrics(boolean enable, Supplier<? extends ChannelMetricsRecorder> recorder, Function<String, String> uriTagValue) {
 		if (enable) {
 			HttpClient dup = duplicate();
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
@@ -342,6 +342,24 @@ public abstract class HttpServer extends ServerTransport<HttpServer, HttpServerC
 		return super.metrics(enable, recorder);
 	}
 
+	public final HttpServer metrics(boolean enable, Supplier<? extends ChannelMetricsRecorder> recorder, Function<String, String> uriTagValue) {
+		if (enable) {
+			HttpServer dup = duplicate();
+			dup.configuration().metricsRecorder(recorder);
+			dup.configuration().uriTagValue = uriTagValue;
+			return dup;
+		}
+		else if (configuration().metricsRecorder() != null) {
+			HttpServer dup = duplicate();
+			dup.configuration().metricsRecorder(null);
+			dup.configuration().uriTagValue = null;
+			return dup;
+		}
+		else {
+			return this;
+		}
+	}
+
 	/**
 	 * Removes any previously applied SSL configuration customization
 	 *
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
@@ -342,6 +342,21 @@ public abstract class HttpServer extends ServerTransport<HttpServer, HttpServerC
 		return super.metrics(enable, recorder);
 	}
 
+	/**
+	 * Specifies whether the metrics are enabled on the {@link HttpServer}.
+	 * All generated metrics are provided to the specified recorder which is only
+	 * instantiated if metrics are being enabled.
+	 * <p>{@code uriTagValue} function receives the actual uri and returns the uri tag value
+	 * that will be used for the metrics with {@link reactor.netty.Metrics#URI} tag.
+	 * For example instead of using the actual uri {@code "/users/1"} as uri tag value, templated uri
+	 * {@code "/users/{id}"} can be used.
+	 *
+	 * @param enable true enables metrics collection; false disables it
+	 * @param recorder a supplier for the metrics recorder that receives the collected metrics
+	 * @param uriTagValue a function that receives the actual uri and returns the uri tag value
+	 * that will be used for the metrics with {@link reactor.netty.Metrics#URI} tag
+	 * @return a new {@link HttpServer}
+	 */
 	public final HttpServer metrics(boolean enable, Supplier<? extends ChannelMetricsRecorder> recorder, Function<String, String> uriTagValue) {
 		if (enable) {
 			HttpServer dup = duplicate();
--- a/reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
@@ -32,7 +32,6 @@ import java.nio.file.Paths;
 import java.security.cert.CertificateException;
 import java.time.Duration;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -52,6 +51,9 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
 import javax.net.ssl.SSLException;
 
 import io.netty.buffer.ByteBuf;
@@ -2501,24 +2503,24 @@ public class HttpClientTest {
 							.wiretap(true)
 							.bindNow();
 
-		final List<String> collectedUris = Collections.synchronizedList(new ArrayList<>());
+		Sinks.Many<String> collectedUris = Sinks.many().replay().all();
 
 		HttpClient.create()
 				.metrics(true,
 						() -> new HttpClientMetricsRecorder() {
 							@Override
 							public void recordDataReceived(SocketAddress remoteAddress, String uri, long bytes) {
-								collectedUris.add(uri);
+								collectedUris.emitNext(uri);
 							}
 
 							@Override
 							public void recordDataSent(SocketAddress remoteAddress, String uri, long bytes) {
-								collectedUris.add(uri);
+								collectedUris.emitNext(uri);
 							}
 
 							@Override
 							public void incrementErrorsCount(SocketAddress remoteAddress, String uri) {
-								collectedUris.add(uri);
+								collectedUris.emitNext(uri);
 							}
 
 							@Override
@@ -2547,17 +2549,17 @@ public class HttpClientTest {
 
 							@Override
 							public void recordDataReceivedTime(SocketAddress remoteAddress, String uri, String method, String status, Duration time) {
-								collectedUris.add(uri);
+								collectedUris.emitNext(uri);
 							}
 
 							@Override
 							public void recordDataSentTime(SocketAddress remoteAddress, String uri, String method, Duration time) {
-								collectedUris.add(uri);
+								collectedUris.emitNext(uri);
 							}
 
 							@Override
 							public void recordResponseTime(SocketAddress remoteAddress, String uri, String method, String status, Duration time) {
-								collectedUris.add(uri);
+								collectedUris.emitNext(uri);
 							}
 						},
 						s -> s.startsWith("/stream/") ? "/stream/{n}" : s
@@ -2568,11 +2570,12 @@ public class HttpClientTest {
 				.aggregate()
 				.block(Duration.ofSeconds(30));
 
-		synchronized (collectedUris) {
-			assertThat(collectedUris).isNotEmpty();
-			for(String uri : collectedUris) {
-				assertThat(uri).isEqualTo("/stream/{n}");
-			}
-		}
+		List<String> expected = IntStream.range(0, 5)
+				.mapToObj(i -> "/stream/{n}")
+				.collect(Collectors.toList());
+		StepVerifier.create(collectedUris.asFlux().take(5))
+				.expectNextSequence(expected)
+				.expectComplete()
+				.verify(Duration.ofSeconds(30));
 	}
 }
--- a/reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
@@ -19,6 +19,7 @@ package reactor.netty.http.client;
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.ByteBuffer;
@@ -31,6 +32,7 @@ import java.nio.file.Paths;
 import java.security.cert.CertificateException;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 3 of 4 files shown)


pivotal-issuemaster:
@fkjellberg Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1297)!

[Click here](https://cla.pivotal.io/sync/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1297) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@fkjellberg Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1297)!

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -315,14 +315,23 @@ public final HttpServer httpRequestDecoder(Function<HttpRequestDecoderSpec, Http
...
+		return metrics(enable, recorder, configuration().uriTagValue());
+	}
+
+	public final HttpServer metrics(boolean enable, Supplier<? extends ChannelMetricsRecorder> recorder, Function<String, String> uriTagValue) {
 		if (enable) {
 			if (!Metrics.isInstrumentationAvailable()) {
 				throw new UnsupportedOperationException(
 						"To enable metrics, you must add the dependency `io.micrometer:micrometer-core`" +
 								" to the class path first");
 			}
```
violetagg, line 332: Do we need to check for Micrometer as this is a custom recorder that might not need it?
We need it only in case of the default implementation.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1105,14 +1105,23 @@ public final HttpClient mapConnect(
...
+		return metrics(enable, recorder, configuration().uriTagValue());
+	}
+
+	public final HttpClient metrics(boolean enable, Supplier<? extends ChannelMetricsRecorder> recorder, Function<String, String> uriTagValue) {
 		if (enable) {
 			if (!Metrics.isInstrumentationAvailable()) {
 				throw new UnsupportedOperationException(
 						"To enable metrics, you must add the dependency `io.micrometer:micrometer-core`" +
 								" to the class path first");
 			}
```
violetagg, line 1122: Do we need to check for Micrometer as this is a custom recorder that might not need it?
We need it only in case of the default implementation.

violetagg (changes requested):
Can you add a test for the new functionality?

Do you need this feature only for 1.0.0 (still in development) or 0.9.x (the current version with releases)?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
fkjellberg: No, you are right and I now understand why the code looked like it did before my change. I'll fix.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
fkjellberg: I'll fix here as well

fkjellberg:
@violetagg I fixed your requested changes and added two tests, one for HttpClient and one for HttpServer.

I need this functionality in the Spring Cloud Gateway so it would be great if this can be backported to the 0.9.x branch. Do you want me to create a separate PR for that?

violetagg:
> @violetagg I fixed your requested changes and added two tests, one for HttpClient and one for HttpServer.
> 
> I need this functionality in the Spring Cloud Gateway so it would be great if this can be backported to the 0.9.x branch. Do you want me to create a separate PR for that?

OK I'll mark this PR for 1.0.0.RC1. Once we agree on the change, if you are able to create a PR for 0.9.x it will be great.

fkjellberg:
I noticed that the test cases failed in some of the automatic builds. After struggling a bit they now pass after my latest push.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1132,6 +1132,24 @@ public final HttpClient metrics(boolean enable, Supplier<? extends ChannelMetric
 		return super.metrics(enable, recorder);
 	}
 
+	public final HttpClient metrics(boolean enable, Supplier<? extends ChannelMetricsRecorder> recorder, Function<String, String> uriTagValue) {
```
violetagg, line 1135: Add javadoc

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -342,6 +342,24 @@ public final HttpServer metrics(boolean enable, Supplier<? extends ChannelMetric
 		return super.metrics(enable, recorder);
 	}
 
+	public final HttpServer metrics(boolean enable, Supplier<? extends ChannelMetricsRecorder> recorder, Function<String, String> uriTagValue) {
```
violetagg, line 345: Add javadoc

violetagg (changes requested):

violetagg (approved):

fkjellberg:
@violetagg PR updated with javadoc but could you please give me some advice regarding the unit tests as one of them again fails on one of the build bots? I took inspiration from other similar test cases but apparently I'm missing something?

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -2489,4 +2491,88 @@ public void testLoopAndResolver() {
...
+	@Test
+	public void testCustomMetricsWithUriMapper() {
+		disposableServer =
+				HttpServer.create()
+							.port(0)
+							.handle((req, resp) -> resp.sendString(Mono.just("OK")))
+							.wiretap(true)
+							.bindNow();
+
+		final List<String> collectedUris = Collections.synchronizedList(new ArrayList<>());
```
violetagg, line 2504: Instead of this can we use `Sinks`
`Sinks.Many<String> collectedUris = Sinks.many().replay().all();`

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -2489,4 +2491,88 @@ public void testLoopAndResolver() {
...
+
+		HttpClient.create()
+				.metrics(true,
+						() -> new HttpClientMetricsRecorder() {
+							@Override
+							public void recordDataReceived(SocketAddress remoteAddress, String uri, long bytes) {
+								collectedUris.add(uri);
```
violetagg, line 2511: `collectedUris.emitNext(uri);`

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -2489,4 +2491,88 @@ public void testLoopAndResolver() {
...
+							}
+
+							@Override
+							public void recordDataSent(SocketAddress remoteAddress, String uri, long bytes) {
+								collectedUris.add(uri);
+							}
+
+							@Override
+							public void incrementErrorsCount(SocketAddress remoteAddress, String uri) {
+								collectedUris.add(uri);
```
violetagg, line 2521: `collectedUris.emitNext(uri);`

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -2489,4 +2491,88 @@ public void testLoopAndResolver() {
...
+				.metrics(true,
+						() -> new HttpClientMetricsRecorder() {
+							@Override
+							public void recordDataReceived(SocketAddress remoteAddress, String uri, long bytes) {
+								collectedUris.add(uri);
+							}
+
+							@Override
+							public void recordDataSent(SocketAddress remoteAddress, String uri, long bytes) {
+								collectedUris.add(uri);
```
violetagg, line 2516: `collectedUris.emitNext(uri);`

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -2489,4 +2491,88 @@ public void testLoopAndResolver() {
...
+							public void recordConnectTime(SocketAddress remoteAddress, Duration time, String status) {
+							}
+
+							@Override
+							public void recordResolveAddressTime(SocketAddress remoteAddress, Duration time, String status) {
+							}
+
+							@Override
+							public void recordDataReceivedTime(SocketAddress remoteAddress, String uri, String method, String status, Duration time) {
+								collectedUris.add(uri);
```
violetagg, line 2550: `collectedUris.emitNext(uri);`

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -2489,4 +2491,88 @@ public void testLoopAndResolver() {
...
+							}
+
+							@Override
+							public void recordDataSentTime(SocketAddress remoteAddress, String uri, String method, Duration time) {
+								collectedUris.add(uri);
```
violetagg, line 2555: `collectedUris.emitNext(uri);`

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -2489,4 +2491,88 @@ public void testLoopAndResolver() {
...
+							}
+
+							@Override
+							public void recordResponseTime(SocketAddress remoteAddress, String uri, String method, String status, Duration time) {
+								collectedUris.add(uri);
```
violetagg, line 2560: `collectedUris.emitNext(uri);`

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -2489,4 +2491,88 @@ public void testLoopAndResolver() {
...
+				.responseContent()
+				.aggregate()
+				.block(Duration.ofSeconds(30));
+
+		synchronized (collectedUris) {
+			assertThat(collectedUris).isNotEmpty();
+			for(String uri : collectedUris) {
+				assertThat(uri).isEqualTo("/stream/{n}");
+			}
+		}
```
violetagg, line 2576:
```
		List<String> expected = IntStream.range(0, 5)
				.mapToObj(i -> "/stream/{n}")
				.collect(Collectors.toList());
		StepVerifier.create(collectedUris.asFlux().take(5))
				.expectNextSequence(expected)
				.expectComplete()
				.verify();
```

violetagg (changes requested):
Can we use `Sinks` for the test?

violetagg:
> @violetagg PR updated with javadoc but could you please give me some advice regarding the unit tests as one of them again fails on one of the build bots? I took inspiration from other similar test cases but apparently I'm missing something?

We can use `Sinks`. As it is a new API in Reactor Core, when you create the PR for 0.9.x, the corresponding type is `ReplayProcessor`

fkjellberg:
@violetagg Thanks for the help! PR updated with your proposed changes.

violetagg (approved):

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -2489,4 +2493,89 @@ public void testLoopAndResolver() {
...
+	@Test
+	public void testCustomMetricsWithUriMapper() {
+		disposableServer =
+				HttpServer.create()
+							.port(0)
+							.handle((req, resp) -> resp.sendString(Mono.just("OK")))
+							.wiretap(true)
+							.bindNow();
+
+		Sinks.Many<String> collectedUris = Sinks.many().replay().all();
```
simonbasle, line 2506: I see little sense in using a Sink/Flux here, a `List` (make it a `CopyOnWriteArrayList` if you're concerned about concurrency) would be perfectly fine here.

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -2489,4 +2493,89 @@ public void testLoopAndResolver() {
...
+								collectedUris.emitNext(uri);
+							}
+
+							@Override
+							public void incrementErrorsCount(SocketAddress remoteAddress, String uri) {
+								collectedUris.emitNext(uri);
+							}
+
+							@Override
+							public void recordDataReceived(SocketAddress remoteAddress, long bytes) {
```
simonbasle, line 2527: @violetagg do we have an abstract/no-op base class (either in prod or test sources) that could remove the need to implement all these empty methods?

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2109,4 +2111,88 @@ else if (msg instanceof ByteBuf) {
 		            .expectComplete()
 		            .verify(Duration.ofSeconds(30));
 	}
+
+	@Test
+	public void testCustomMetricsWithUriMapper() {
+		Sinks.Many<String> collectedUris = Sinks.many().replay().all();
```
simonbasle, line 2117: same as above

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2109,4 +2111,88 @@ else if (msg instanceof ByteBuf) {
...
+				HttpServer.create()
+							.port(0)
+							.handle((req, resp) -> resp.sendString(Mono.just("OK")))
+							.wiretap(true)
+							.metrics(true,
+									() -> new HttpServerMetricsRecorder() {
+										@Override
+										public void recordDataReceived(SocketAddress remoteAddress, String uri, long bytes) {
+											collectedUris.emitNext(uri);
+										}
```
simonbasle, line 2129: same as above

simonbasle (changes requested):
nitpicks on the tests. also @fkjellberg I'm not sure I understand your comment about efficiency, since I don't see any change in the way the mapper is applied in this PR, can you clarify?

simonbasle (changes requested):
nitpicks on the tests. also @fkjellberg I'm not sure I understand your comment about efficiency, since I don't see any change in the way the mapper is applied in this PR, can you clarify?

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
violetagg: @simonbasle The PR used List, but with List it is not guaranteed that exactly after receiving the response  all values will be in the least (see previous commits and the failed CI). So we either use `Sinks` or other mechanism (CountDownLatch) in order to be sure that when we start checking the values all data is there.

We can use CoundDownLatch if this is your recommendation.

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
violetagg: no there isn't

violetagg:
> nitpicks on the tests. also @fkjellberg I'm not sure I understand your comment about efficiency, since I don't see any change in the way the mapper is applied in this PR, can you clarify?

When there is `uriTagValue`, Reactor Netty calculates the path once (when the request is received)
https://github.com/reactor/reactor-netty/blob/3528badfa032195e8325f35ba351e993cf186a9a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientMetricsHandler.java#L70

without `uriTagValue` for the custom recorders, they (the recorders) has to do this on every invocation.

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
simonbasle: I see. I would have expected the metrics to be recorded by the time the client`.get()` is finished (since it is blocking). I found the pattern of using `Sinks` a bit surprising and harder to read than a list-based assertion in that particular case.

Using `take(5)` is a potential subtle issue, too. It wouldn't catch a bug where a 6th uri is recorded.

So my preference would be to switch back to a `List`+`assertJ` approach. reactor-netty tests don't depend on `Awaitility` from what I can see, but it could be a better alternative than `CountdownLatch` here, wdyt @violetagg ?

Plus the assertion would be simpler and be less brittle:

```java
assertThat(collectedUris).hasSize(5).containsOnly("/stream/{n}");
```

Either way is fine in the end, I'll leave that decision to you @violetagg. I would just add a clarifying comment to the use of `Sinks` if we keep this solution.

violetagg: the number `5` is because we have 5 types of metrics and not 5 requests, because of that we cannot expect on `get` to have all of them.
For one request that we make we have `recordDataReceived`, `recordDataSent`, `recordDataReceivedTime`, `recordDataSentTime`, `recordResponseTime`

simonbasle: yeah, I understood that. I meant, if a bug is introduced in the `MetricsRecorder` (or the internal code that calls it) and ends up calling eg. `recordDataReceived` twice, this test wouldn't see it (due to the `take(5)`) whereas a list assertion like I suggested would.

simonbasle:
> When there is `uriTagValue`, Reactor Netty calculates the path once (when the request is about to be sent)

Thanks for clarifying that for me. So in essence the "plumbing" was there but the API didn't guide you into setting the pattern along activation of the metrics. 👍

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
violetagg: @fkjellberg Please switch to `CountDownLatch` or `Awaitility` instead of `Sinks`

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2109,4 +2111,87 @@ else if (msg instanceof ByteBuf) {
...
+										@Override
+										public void recordResponseTime(String uri, String method, String status, Duration time) {
+											collectedUris.add(uri);
+										}
+									},
+									s -> s.startsWith("/stream/") ? "/stream/{n}" : s)
+							.bindNow();
+
+		HttpClient.create()
+				.doOnRequest((req, conn) -> conn.onTerminate()
```
violetagg, line 2185: It is better to count down when every metric is recorded, thus `latch.await` will wait for all to appear.
Wdyt?

simonbasle: like initializing as `CountDownLatch(5)` and counting down next to each `collectedUris.add()` call?

why not, especially if you think counting down on the connection termination could lead to making the assertion too early. it has the same drawback as my original concern with `take(5)`, but that is more minor.

I like how the test reads much more with the list 👍

violetagg: yes because terminating is competing with recorder
https://github.com/reactor/reactor-netty/blob/8afc89e173733e97ce67d8bcbf4e3083e8a40f1f/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpTrafficHandler.java#L280
https://github.com/reactor/reactor-netty/blob/8afc89e173733e97ce67d8bcbf4e3083e8a40f1f/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerMetricsHandler.java#L77

fkjellberg:
@simonbasle @violetagg Thanks for the feedback! Please review the updated PR where the test cases are now using CountDownLatch. One of the build bots failed but it seems unrelated to my changes as it was in the DefaultPooledConnectionProviderTest.

violetagg:
> @simonbasle @violetagg Thanks for the feedback! Please review the updated PR where the test cases are now using CountDownLatch. One of the build bots failed but it seems unrelated to my changes as it was in the DefaultPooledConnectionProviderTest.

I do not see the difference between the last two commits?

simonbasle (approved):

violetagg:
@fkjellberg I squashed the changes and committed them with 6a79577. Also I backported the change to 0.9.x with 96da5d0.
Thanks a lot for the PR.
0.9.12.RELEASE and 1.0.0-RC1 are scheduled for today.

fkjellberg:
Thank you, @violetagg ! Looking forward to get this into the next Spring Cloud Gateway release.

PR #1305, opened 2020-09-10
fix #1261 restore default resolver on proxy remove

### LINKED ISSUE #1261: When removing the proxy configuration, return the correct resolver
When adding proxy configuration the resolver is changed as follows:
- 0.9.x - https://github.com/reactor/reactor-netty/blob/72528d28731345d1d96cc746d2a4ab27a750b0b9/src/main/java/reactor/netty/channel/BootstrapHandlers.java#L492-L510

- 1.0.0 - https://github.com/reactor/reactor-netty/blob/5be51c788c4a393a468ea23da84b212da1406ea5/reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java#L181-L198

When removing the proxy configuration we need to restore the default resolver.

ericbottard:


ericbottard:
> Can we test this?

I'm sure we can. Wanted to make sure the approach is the correct one. I couldn't find a test for the proxy setup, let me know if you know there is one.

violetagg:
> > Can we test this?
> 
> I'm sure we can. Wanted to make sure the approach is the correct one. I couldn't find a test for the proxy setup, let me know if you know there is one.

May be this one reactor.netty.http.client.HttpClientProxyTest

violetagg:
@ericbottard I'm going to merge that as I want this in the release, the test will be provided later.

violetagg (approved):

violetagg:
Test case https://github.com/reactor/reactor-netty/commit/55386a708873fda21a65b58f3884ea7863b58ddb

PR #1307, opened 2020-09-11
Add Brave instrumentation for HttpServer

violetagg:
Related to #539

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/Http2StreamBridgeServerHandler.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/Http2StreamBridgeServerHandler.java
@@ -18,12 +18,14 @@ package reactor.netty.http.server;
 import java.net.SocketAddress;
 import java.util.Optional;
 import java.util.function.BiFunction;
+import java.util.function.Function;
 
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.cookie.ServerCookieDecoder;
 import io.netty.handler.codec.http.cookie.ServerCookieEncoder;
 import io.netty.handler.ssl.SslHandler;
+import reactor.core.publisher.Mono;
 import reactor.netty.Connection;
 import reactor.netty.ConnectionObserver;
 import reactor.netty.http.Http2StreamBridgeHandler;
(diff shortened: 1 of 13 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -290,6 +290,21 @@ public final HttpServer httpRequestDecoder(Function<HttpRequestDecoderSpec, Http
...
+	 * See {@link #handle(BiFunction)}.
+	 *
+	 * @param mapHandle A function to decorate the configured I/O handler
+	 * @return a new {@link HttpServer}
+	 */
+	public final HttpServer mapHandle(Function<? super Mono<Void>, ? extends Mono<Void>> mapHandle) {
+		Objects.requireNonNull(mapHandle, "mapHandle");
+		HttpServer dup = duplicate();
+		Function<? super Mono<Void>, ? extends Mono<Void>> current = configuration().mapHandle;
+		dup.configuration().mapHandle = current == null ? mapHandle : current.andThen(mapHandle);
```
simonbasle, line 304: do you foresee a case where one would want to remove mapHandle `Function` (or part(s) of it)? after it has been set?

reactor-netty-http/src/main/java/reactor/netty/http/server/Http2StreamBridgeServerHandler.java
```diff
@@ -40,17 +42,20 @@
 	final ServerCookieEncoder                                     cookieEncoder;
 	final ConnectionObserver                                      listener;
 	final BiFunction<ConnectionInfo, HttpRequest, ConnectionInfo> forwardedHeaderHandler;
+	final Function<? super Mono<Void>, ? extends Mono<Void>> mapHandle;
```
simonbasle, line 45: nitpick: breaking the nice alignment here :p

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -553,22 +560,25 @@ public void handlerAdded(ChannelHandlerContext ctx) {
 		final ConnectionObserver                                      listener;
 		final ServerCookieEncoder                                     cookieEncoder;
 		final ServerCookieDecoder                                     cookieDecoder;
+		final Function<? super Mono<Void>, ? extends Mono<Void>> mapHandle;
```
simonbasle, line 563: alignment here too

simonbasle (approved):
only small formatting nitpick, plus a separate comment on "removing" `mapHandle`. From my perspective, looks good.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
violetagg: fixed

reactor-netty-http/src/main/java/reactor/netty/http/server/Http2StreamBridgeServerHandler.java
violetagg: fixed

simonbasle (approved):

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
violetagg: For the moment no, but if it is needed we can add an API.
I will update the javadoc for this that is adds this function to the existing functions (if any)
wdyt?

simonbasle: ah yeah that's probably worth mentioning in the javadoc

marcingrzejszczak (approved):
At this point LGTM. I'll be able to tell more once I take it out for a spin in Sleuth.

simonbasle (approved):

Issue #1058, opened 2020-04-08
Add fluent mechanism to construct encoded uri parameters using QueryStringEncoder


thekalinga:
## Motivation
Currently, when making a HttpClient request, the fluent api expects the URI to be single string. This leads to ugly url string concatanation & also the API does not give a hint whether the underlying library would escape URI appropriately/it expects the user to do the escaping

## Desired solution
Add an overload of uri that allows users to consurtct uri in a fluent manner. One potential implementation detail would be to use `QueryStringEncoder` from netty internally and expose a fluent API to the user, the same way we currently do for `headers`

With fluent API user can be certain that the encoding is done for sure & also he has a saner way of constructing the uri parameters

violetagg:
I'm closing this issue as `uri(URI)` provides the requested functionality. Reactor Netty users can choose which library to use in order to construct the needed URI (for example #1323 where `QueryStringEncoder` is used). 
If there is a broader interest for something like `uri(UriBuilder)` we can reopen the issue and discuss the further steps.

PR #1323, opened 2020-10-03
Add fluent interface to append query parameters

l0s:
This adds several fluent methods for adding query parameters to an `HttpClient`. The new methods ensure that the parameters are encoded first. In addition, any existing parameters retain their position.

Resolves: #1058

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -17,7 +17,10 @@ package reactor.netty.http.client;
 
 import java.net.SocketAddress;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.time.Duration;
+import java.util.Collections;
+import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.BiConsumer;
@@ -26,6 +29,7 @@ import java.util.function.BiPredicate;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.regex.Pattern;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.handler.codec.http.HttpHeaderNames;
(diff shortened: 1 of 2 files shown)


violetagg:
@l0s Thanks for looking into this.

I think the API should be more like what we have for the `headers`, later the query parameters can be added as we add the headers. See here
https://github.com/reactor/reactor-netty/blob/abaa3e417afd824aa3346144156ff4ec54aa0937/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L430-L447

Also do we want these query parameters to be applied to all requests or we want them per request?

/cc @thekalinga

l0s:
@violetagg thanks for the feedback!

I can refactor it to store the query parameters in the `HttpClientConfig`. Is that what you had in mind? Also, would you mind clarifying the differences between the fields: `baseUrl: String`, `uri: URI`, and `uriStr: String` in that class? I want to make sure I apply the query parameters in all the places I need to.

As for your questions, I'll defer to you and @thekalinga as to what would work best.

rstoyanchev:
> Or may be we need an uri method with some URI builder?

Indeed I don't think this stops at adding query parameters. Once that's available, it is reasonable to expect requests for more complete, full modification capabilities for any part of the URI but I'm wondering if it's worth building this into `HttpClient` vs leaving it as an external exercise via `QueryStringEncoder` or similar depending on what's on your classpath.

violetagg:
> > Or may be we need an uri method with some URI builder?
> 
> Indeed I don't think this stops at adding query parameters. Once that's available, it is reasonable to expect requests for more complete, full modification capabilities for any part of the URI but I'm wondering if it's worth building this into `HttpClient` vs leaving it as an external exercise via `QueryStringEncoder` or similar depending on what's on your classpath.

Exactly I think that the method `uri(java.net.URI)` can be used and the user can decide what library to use in order to build that URI.

l0s:
@violetagg, @rstoyanchev are you saying the onus on encoding and appending query parameters should be on the user? Or do you still want a fluent mechanism for adding query parameters?

rstoyanchev:
@l0s from a framework perspective what looks like a straight forward enhancement can often turn into a much larger feature over time. Here we are looking at query parameters but what about other parts of the URI, shouldn't there be some sort of a URI builder as @violetagg already asked? Should Reactor Netty really be solving this problem, when using `QueryStringEncoder` or similar externally is quite straight forward.

l0s:
Got it. So to confirm, you propose a new class or interface called `UriBuilder` and we should overload the `uri` methods on `UriConfiguration` with `uri( UriBuilder ): S`. Do I understand correctly?

rstoyanchev:
No, not proposing that but rather questioning whether anything should be done at all.

violetagg:
@l0s I'm proposing to close the issue #1058 because `uri(URI)` provides the requested functionality. Reactor Netty users can choose which library to use in order to construct the needed URI (for example this PR where `QueryStringEncoder` is used).
If there is a broader interest for something like `uri(UriBuilder)` we can reopen the issue/PR and discuss the further steps.

PR #1330, opened 2020-10-09
Revert to use Context instead of ContextView

violetagg:


simonbasle (approved):
approved but do we still really need this, from the discussions we had around users migrating to reactor-netty 1.0 ?

violetagg:
@simonbasle 
> approved but do we still really need this, from the discussions we had around users migrating to reactor-netty 1.0 ?

OK let me close that. If we need it we can reopen it.

Issue #1327, opened 2020-10-08
ConnectionObserver#currentContext visible in doOnChannelInit should have the context of the subscription that drove it.


crankydillo:
We feel that `doOnChannelInit` should have access to the client 'request' subscription that drove it firing.  The only place we saw to look was in `ConnectionObserver#currentContext`.  Seems like it could go there; however, ultimately, we just need it somewhere:)

I added the following to `TcpClientTests` to show what we are thinking:

```
       @Test
	public void testSubscriberContextAndDoOnChannelInit() throws Exception {
		DisposableServer server = null;
		try {
			server = TcpServer.create().port(0).bindNow();

			final String contextKey = "marcels-key";
			final AtomicReference<String> contextualData = new AtomicReference<>();
			final CountDownLatch channelInitialized = new CountDownLatch(1);
			final String contextValue = "marcels-context";

			TcpClient.create()
					.doOnChannelInit((connectionObserver, channel, remoteAddress) -> {
						if (connectionObserver.currentContext().hasKey(contextKey)) {
							contextualData.set(connectionObserver.currentContext().get(contextKey));
						}
						channelInitialized.countDown();
					})
					.remoteAddress(server::address)
					.wiretap(true)
					.connect()
					.contextWrite(Context.of(contextKey, contextValue))
					.subscribe();

			assertTrue(channelInitialized.await(30, TimeUnit.SECONDS));
			assertNotNull(contextualData.get());
			assertEquals(contextValue, contextualData.get());
		} finally {
			if (server != null)
				server.disposeNow();
		}
	}
```

anilgursel:
I think the propagation of context is broken in two places:

1. In `SimpleDequePool`, `primary.subscribe(alreadyPropagated -> { }, alreadyPropagatedOrLogged -> { });`
2. Context not being passed to `PendingConnectionObserver`.

Sending a PR to tackle the second one for your review.

violetagg:
@crankydillo @anilgursel Do you want to have the PR and continue using `Hooks.onLastOperator` or should we wait for Reactor Pool to start propagating the Context?

/cc @simonbasle

anilgursel:
We already use `Hooks.onLastOperator` for tracing in general.  So, if the PR can be merged, it should help with our immediate issue.  We can re-visit later when Reactor Pool starts propagating the Context. (I need to see why the PR build failed though).

violetagg:
[...]

PR #1335, opened 2020-10-13
#1327 Propagate context to ConnectionObserver in doOnChannelInit

anilgursel:


reactor-netty-core/src/test/java/reactor/netty/ConnectionObserverContextTest.java
```diff
@@ -0,0 +1,166 @@
+package reactor.netty;
```
violetagg, line 1: Add copyright header, this is the cause for the failure

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
```diff
@@ -471,7 +484,7 @@ ConnectionObserver owner() {
 				obs = channel.attr(OWNER)
 				             .get();
 				if (obs == null) {
-					obs = new PendingConnectionObserver();
+					obs = new PendingConnectionObserver(context);
```
violetagg, line 487: I prefer this context to be available only when establishing a connection.
Basically we do not have OWNER (or PendingConnectionObserver can appear) in two cases:
- when establishing a connection
- when returning a connection to the pool it is removed and once the connection is acquired it is set again

Can we add something like the code below after this row
https://github.com/reactor/reactor-netty/blob/631f7a67269ac450195d2336714edb9098e6736f/reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java#L533

```
channel.attr(OWNER).compareAndSet(null, obs)
```

anilgursel:
Only `build (ubuntu-18.04, native)` failed and I do not think it is related to my changes:

```
reactor.netty.http.server.HttpsSendFileTests > sendFileAsync1024 FAILED
    org.junit.ComparisonFailure: 
```

```
ERROR: Test: Test testIssue1071MaxContentLengthSpecified(reactor.netty.http.Http2Tests) produced resource leak: 20:13:58.170 [reactor-http-epoll-3] ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
```

reactor-netty-core/src/test/java/reactor/netty/ConnectionObserverContextTest.java
```diff
@@ -0,0 +1,182 @@
...
+ * {@link Context} is propagated to {@link ConnectionObserver}, not really
+ * interested in verifying that {@link Context} is propagated from outer stream
+ * to the inner stream.  That's why it uses {@link Hooks} to propagate
+ * {@link Context} from outer stream to the inner stream.
+ */
+public class ConnectionObserverContextTest {
+
+	private static final String CONTEXT_KEY = "marcels-key";
+	private static final String CONTEXT_VALUE = "marcels-context";
+	private static final ThreadLocal<String> helloWorld = new ThreadLocal();
```
violetagg, line 60: nit: `new ThreadLocal<>()`

reactor-netty-core/src/test/java/reactor/netty/ConnectionObserverContextTest.java
```diff
@@ -0,0 +1,182 @@
...
+					}
+					channelInitialized.countDown();
+				})
+				.remoteAddress(server::address)
+				.wiretap(true)
+				.connect()
+				.contextWrite(Context.of(CONTEXT_KEY, CONTEXT_VALUE))
+				.subscribe();
+
+		Thread.sleep(1000);
```
violetagg, line 100: What is the purpose of `Thread.sleep(1000)`?

reactor-netty-core/src/test/java/reactor/netty/ConnectionObserverContextTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+		/**
+		 * This intended to be used with reactor functions such as {@link Hooks#onLastOperator(Function)}.
+		 *
+		 * @param contextModifier a function that modifies the {@link Context} given to it.
+		 */
+		public static <T> Function<? super Publisher<T>, ? extends Publisher<T>> asOperator(
+				final UnaryOperator<Context> contextModifier
+		) {
+			return Operators.liftPublisher((Publisher publisher, CoreSubscriber<? super T> sub) -> {
```
violetagg, line 171:
```
/.../reactor-netty/reactor-netty-core/src/test/java/reactor/netty/ConnectionObserverContextTest.java:171: warning: [rawtypes] found raw type: Publisher
			return Operators.liftPublisher((Publisher publisher, CoreSubscriber<? super T> sub) -> {
			                                ^
  missing type arguments for generic class Publisher<T>
  where T is a type-variable:
    T extends Object declared in interface Publisher
1 warning
```

May be we can suppress that?

violetagg (changes requested):
I triggered the tests again and now they are OK

reactor-netty-core/src/test/java/reactor/netty/ConnectionObserverContextTest.java
```diff
@@ -0,0 +1,182 @@
...
+				.doOnChannelInit((connectionObserver, channel, remoteAddress) -> {
+					if (connectionObserver.currentContext().hasKey(CONTEXT_KEY)) {
+						contextualData.set(connectionObserver.currentContext().get(CONTEXT_KEY));
+					}
+					channelInitialized.countDown();
+				})
+				.remoteAddress(server::address)
+				.wiretap(true)
+				.connect()
+				.contextWrite(Context.of(CONTEXT_KEY, CONTEXT_VALUE))
```
simonbasle, line 97: I don't understand the purpose of having both `contextWrite` and the threadLocal-to-`Context` operator decorator / hook. isn't the purpose of the change to validate that the `Context` from the `connect()` `Mono` is visible within `ConnextionObserver`? Having the value set by both an explicit operator and an operator hook makes it hard to reason about this test. If we expect one to take precedence over the other, then at a minimum the values put in the context should be different ?

reactor-netty-core/src/test/java/reactor/netty/ConnectionObserverContextTest.java
anilgursel: It was there in the example in the issue.  It does not look like needed, so removing.

reactor-netty-core/src/test/java/reactor/netty/ConnectionObserverContextTest.java
anilgursel: I agree with you and will remove this line.  However, the original goal was to add a context and the outer stream and the context should be available in `doOnChannelInit`.  The reason Hooks are added is because of the first issue mentioned in https://github.com/reactor/reactor-netty/issues/1327#issuecomment-707849473.  Once https://github.com/reactor/reactor-pool/issues/103 is addressed, we should hopefully not need the `Hooks`.  

For now, I will change the test to just rely on `Hooks` and will add comment why that's the case.

violetagg (approved):

violetagg:
@anilgursel I squashed and committed the changes 4cef559

Thanks for the PR!

Issue #476, opened 2018-10-25
Log Remote IP address in the access log


hota911:
If a server is running behind a load balancer, the remote client IP address is not logged in access log.  I'd like to have an option to log the IP address stored in `X-Forwarded-For` header if it exists, otherwise the client IP address.

Maybe this could be addressed in #436, but I'm filing this issue for the specific need.

LyStormrage:
I had the same problem，it's always nginx server ip address. And I can't customize the format

jkubrynski:
Is there any progress here? If no, are you open for pull requests?

violetagg:
@jkubrynski Sure we are OK with pull requests. If you have an idea for a pull request go for it.

violetagg:
Fixed with #1357

PR #1357, opened 2020-10-29
Provide customized access log

limaoning:


violetagg:
Related issues #436 and #476

violetagg:
@limaoning @simonbasle What do you think if the access log functionality is moved to an own package (i.e. `logging`)?

@limaoning Can you add javadoc to the public API? I like the idea, once we agree on the package and we have a javadoc, I'll make a detailed review.

simonbasle:
> @limaoning @simonbasle What do you think if the access log functionality is moved to its own package (e.g. `logging`)?

Yeah that makes sense to me if that possible 👍

limaoning:
@violetagg The code has been submitted, please take a look.

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogArgProviderH2.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * @author limaoning
+ */
+final class AccessLogArgProviderH2 extends AbstractAccessLogArgProvider {
+
+	static final String H2_PROTOCOL_NAME = "HTTP/2.0";
+
+	Http2HeadersFrame requestHeaders;
+	Http2HeadersFrame responseHeaders;
```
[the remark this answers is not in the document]
limaoning: Now the response is only for status, do you mean to replace responseHeaders(Http2HeadersFrame responseHeaders) with status(CharSequence status)?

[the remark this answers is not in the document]
limaoning: > For the moment do we need the response for something else than status?

Now the response is only for status, do you mean to replace responseHeaders(Http2HeadersFrame responseHeaders) with status(CharSequence status)?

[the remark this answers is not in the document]
violetagg: The class is internal so if we need the response (for example for response headers) we can always change it, so let's now just set the status, otherwise we need NULL check when accessing the response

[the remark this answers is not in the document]
limaoning: Ok. And I am thinking initialize the accessLogArgProvider only on the first time the request is received, clear the fields in accessLogArgProvider (e.g. requestHeaders) after log,
`accessLog .apply(accessLogArgProvider).log();
 accessLogArgProvider.clear();`
so can reuse the accessLogArgProvider and help GC. Do you think is that ok?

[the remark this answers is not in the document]
violetagg: this should be ok

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogHandler.java
```diff
@@ -14,39 +14,45 @@
...
+import java.util.function.Function;
 
 /**
+ * {@link ChannelHandler} for access log of HTTP/1.1.
+ *
  * @author Violeta Georgieva
+ * @author limaoning
  */
-final class AccessLogHandler extends ChannelDuplexHandler {
+public final class AccessLogHandler extends BaseAccessLogHandler {
```
simonbasle, line 39: to maintain consistency, I'd rename this class to `AccessLogHandlerH1`.

@violetagg @limaoning I'm also wondering about the `public` status of this and the H2 implementations. maybe moving _everything_ in `logging` package wasn't such a good idea 😖 

I wonder if it wouldn't be better to put the concrete classes back into the base `http.server` package, as package-private, and only keep the interfaces and user-facing classes here in `http.server.logging`. ie classes that make sense for a user to implement/customize/inherit from

simonbasle (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogArgProvider.java
```diff
@@ -0,0 +1,106 @@
...
+	 */
+	@Nullable
+	String zonedDateTime();
+
+	/**
+	 * Returns the address of the remote peer or {@code null} in case of Unix Domain Sockets.
+	 *
+	 * @return the peer's address
+	 */
+	SocketAddress remoteAddress();
```
violetagg, line 42: `@Nullable`

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogArgProviderH1.java
```diff
@@ -0,0 +1,77 @@
...
+	@Override
+	public CharSequence uri() {
+		return request == null ? null : request.uri();
+	}
+
+	@Override
+	public String protocol() {
+		return request == null ? null : request.protocolVersion().text();
+	}
+
```
violetagg, line 54: Can we move those in the abstract class and then in the children to override the `onRequest` method where we will initialise `method`, `uri` and `protocol`

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogArgProviderH2.java
```diff
@@ -0,0 +1,74 @@
...
+
+	@Override
+	public CharSequence uri() {
+		return requestHeaders == null ? null : requestHeaders.headers().path();
+	}
+
+	@Override
+	public String protocol() {
+		return H2_PROTOCOL_NAME;
+	}
```
violetagg, line 55: Can we move those in the abstract class and then in the children to override the onRequest method where we will initialise method, uri and protocol

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/AccessLogArgProviderH1Tests.java
```diff
@@ -0,0 +1,113 @@
...
+import io.netty.handler.codec.http.HttpRequest;
+import io.netty.handler.codec.http.HttpVersion;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.net.InetSocketAddress;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNullPointerException;
+import static reactor.netty.http.server.logging.LoggingTests.*;
```
violetagg, line 31: Do not use `*`

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/AccessLogArgProviderH2Tests.java
```diff
@@ -0,0 +1,104 @@
...
+import io.netty.handler.codec.http2.Http2Headers;
+import io.netty.handler.codec.http2.Http2HeadersFrame;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.net.InetSocketAddress;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNullPointerException;
+import static reactor.netty.http.server.logging.LoggingTests.*;
```
violetagg, line 30: Do not use `*`

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/AccessLogArgProviderTests.java
```diff
@@ -0,0 +1,135 @@
...
+		assertThat(accessLogArgProvider.startTime).isNotNull();
+		accessLogArgProvider.clear();
+		assertThat(accessLogArgProvider.zonedDateTime()).isNull();
+		assertThat(accessLogArgProvider.status()).isNull();
+		assertThat(accessLogArgProvider.chunked).isFalse();
+		assertThat(accessLogArgProvider.contentLength).isZero();
+		assertThat(accessLogArgProvider.startTime).isZero();
+	}
+
+	static class MyAccessLogArgProvider extends AbstractAccessLogArgProvider<MyAccessLogArgProvider> {
```
violetagg, line 104: `TestAccessLogArgProvider` instead of `MyAccessLogArgProvider`

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/AccessLogHandlerH2Tests.java
```diff
@@ -0,0 +1,76 @@
...
+import io.netty.handler.codec.http2.DefaultHttp2DataFrame;
+import io.netty.handler.codec.http2.DefaultHttp2Headers;
+import io.netty.handler.codec.http2.DefaultHttp2HeadersFrame;
+import io.netty.handler.codec.http2.Http2Headers;
+import org.junit.jupiter.api.Test;
+
+import java.net.SocketAddress;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static reactor.netty.http.server.logging.LoggingTests.*;
```
violetagg, line 32: Do not use `*`

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/AccessLogHandlerTests.java
```diff
@@ -0,0 +1,111 @@
...
+import io.netty.buffer.DefaultByteBufHolder;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.handler.codec.http.*;
+import org.junit.jupiter.api.Test;
+
+import java.net.SocketAddress;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static reactor.netty.http.server.logging.LoggingTests.*;
```
violetagg, line 29: Do not use `*`

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/AccessLogHandlerTests.java
```diff
@@ -0,0 +1,111 @@
...
+
+		channel.writeInbound(newHttpRequest());
+
+		channel.writeOutbound(newHttpResponse(true));
+
+		ByteBuf byteBuf = Unpooled.buffer(RESPONSE_CONTENT.length);
+		byteBuf.writeBytes(RESPONSE_CONTENT);
+		channel.writeOutbound(byteBuf);
+
+		ByteBufHolder byteBufHolder = new DefaultByteBufHolder(byteBuf);
```
violetagg, line 71: May be using `DefaultHttpContent` is better?

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
```diff
@@ -0,0 +1,51 @@
...
+		this.logFormat = logFormat;
+		this.args = args;
+	}
+
+	public static AccessLog create(String logFormat, Object... args) {
+		return new AccessLog(logFormat, args);
+	}
+
+	void log() {
+		if (log.isInfoEnabled()) {
```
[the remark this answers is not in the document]
violetagg: I would keep such change out of this PR.
Typically access log  logger is isolated in a separate file.

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogHandler.java
violetagg: > to maintain consistency, I'd rename this class to AccessLogHandlerH1.

Agree with that

And let's move to the `http.server` package ONLY the handlers `BaseAccessLogHandler`, `AccessLogHandlerH1`, `AccessLogHandlerH2`

limaoning: But `AccessLogArgProviderH1`, `AccessLogArgProviderH2` will be `public`.

violetagg: if we move them to `http.server` they can be package private - that's the idea

limaoning: You are saying move handlers to `http.server`, but `AccessLogArgProviderH1`, `AccessLogArgProviderH2` are not, and they will be `public`. Otherwise, move `AccessLogArgProviderH1`, `AccessLogArgProviderH2` to `http.server` too?

violetagg: > And let's move to the http.server package ONLY the handlers BaseAccessLogHandler, AccessLogHandlerH1, AccessLogHandlerH2

yes

violetagg: @simonbasle Here I think we either move everything on keep the handlers public  as @limaoning noted in the comment above

violetagg: Let me think a bit here

violetagg: some rough idea

```
public class AccessLogHandlerFactory {

	ChannelHandler newAccessLogHandler(@Nullable Function<AccessLogArgProvider, AccessLog> accessLog, boolean isHttp2) {
		return isHttp2 ? new AccessLogHandlerH2(accessLog) : new AccessLogHandlerH1(accessLog);
	}
}
```

wdyt?

violetagg: or may be even something like this https://github.com/reactor/reactor-netty/blob/master/reactor-netty-core/src/main/java/reactor/netty/transport/logging/AdvancedByteBufFormat.java

limaoning: Should be OK.

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AbstractAccessLogArgProvider.java
```diff
@@ -0,0 +1,144 @@
...
+
+	final SocketAddress remoteAddress;
+	final String user = MISSING;
+	String zonedDateTime;
+	CharSequence method;
+	CharSequence uri;
+	String protocol;
+	CharSequence status;
+	boolean chunked;
+	long contentLength;
```
violetagg, line 45: There are use cases when the I/O handler may send `Transfer-Encoding: chunked` to a HEAD request. As the HEAD request contains only headers, the access log for such request will be with content length `0` which is not quite correct.

`reactor.netty.http.server.AccessLog - /0:0:0:0:0:0:0:1:49307 - - [03/Nov/2020:19:58:38 +0200] "GET / HTTP/1.1" 200 0 10 ms`

This is also valid for HTTP/2
`INBOUND HEADERS: streamId=1 headers=DefaultHttp2Headers[:method: HEAD, :path: /, :scheme: https, :authority: localhost:8080, user-agent: XXX, accept: */*] padding=0 endStream=true`
`OUTBOUND HEADERS: streamId=1 headers=DefaultHttp2Headers[:status: 200] padding=0 endStream=true`

May be we can initialise/clean it `contentLength=-1` so that later it is replaced with `MISSING`?
But please keep in mind that this is relevant only when there is no body, when there is we need to guarantee that we start counting from `0`.

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AbstractAccessLogArgProvider.java
limaoning: Maybe I found a bug, when I defined GET /hello, but client call HEAD /hello, it returns 404, it should returns the GET /hello headers.

violetagg: Do you use `route` if yes then the default implementation does not support declaration of GET but call with HEAD, you need a declaration with HEAD.
For testing purposes you can use `handle`, or custom predicates with `route`.
Check these tests
https://github.com/reactor/reactor-netty/blob/fdd7b1b967fe84f0fa39ae3e57ed0eb41db7b16f/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java#L570
https://github.com/reactor/reactor-netty/blob/fdd7b1b967fe84f0fa39ae3e57ed0eb41db7b16f/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java#L666

limaoning: Should I just ignore the content length like this
`
SELF increaseContentLength(long contentLength) {
	if (chunked && contentLength >= 0 && !HttpMethod.HEAD.asciiName().contentEqualsIgnoreCase(method)) {
		if (this.contentLength == -1) {
			this.contentLength = 0;
		}
		this.contentLength += contentLength;
	}
	return get();
}
`

limaoning: Should I just ignore the content length like this
```
	SELF increaseContentLength(long contentLength) {
		if (chunked && contentLength >= 0 && !HttpMethod.HEAD.asciiName().contentEqualsIgnoreCase(method)) {
			if (this.contentLength == -1) {
				this.contentLength = 0;
			}
			this.contentLength += contentLength;
		}
		return get();
	}
```

violetagg: yes I think this will work

limaoning:
```
.route(r -> r.head("/hello",
		(req, res) -> res.header(CONTENT_TYPE, TEXT_PLAIN)
				.sendString(Mono.just("Hello World!"))));
```
But HTTP/2 still returns the content.

violetagg: I'll create an issue for this and we will track it separately. I can reproduce it.

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/BaseAccessLogHandler.java
```diff
@@ -0,0 +1,53 @@
...
+		this.accessLog = accessLog == null ? DEFAULT_ACCESS_LOG : accessLog;
+	}
+
+	static String applyAddress(@Nullable SocketAddress socketAddress) {
+		if (socketAddress == null) {
+			return MISSING;
+		}
+		InetSocketAddress inetSocketAddress = (InetSocketAddress) socketAddress;
+		return inetSocketAddress.getHostString() + ":" + inetSocketAddress.getPort();
+	}
```
violetagg, line 51: Let's change it to the snippet below. Wdyt?

```
static String applyAddress(@Nullable SocketAddress socketAddress) {
	if (socketAddress instanceof InetSocketAddress) {
		InetSocketAddress inetSocketAddress = (InetSocketAddress) socketAddress;
		return inetSocketAddress.getHostString() + ":" + inetSocketAddress.getPort();
	}
	else {
		return MISSING;
	}
}
```

limaoning: Sorry, I will change it.

violetagg (approved):

violetagg:
@simonbasle PTAL

simonbasle (approved):

violetagg:
@limaoning I squashed and committed the changes 9244b6d. Thanks for that contribution!

IvanSmurygin:
Hello! Do you have some readme regarding changes for access log?

violetagg:
@IvanSmurygin Currently there is no reference documentation for this particular change. Are you interested in providing a PR with such?

nikban95:
There is no way to log response headers in access log even after the above change. Are we planning to add it in future??

violetagg:
@nikban95 By default Reactor Netty provides Common Log Format. If you need another format you can easily provide it by yourself. Here is a documentation:
https://projectreactor.io/docs/netty/release/reference/index.html#_http_access_log

Please ask you questions on [Gitter](https://gitter.im/reactor/reactor-netty) or [stackoverflow.com](https://stackoverflow.com/questions/tagged/reactor-netty)

limaoning:
@violetagg What he means is that there is no way to get response header, only the method to get request header.

nikban95:
@violetagg  As @limaoning has pointed out. Currently we can only log request headers only but the response headers can't be logged. So was asking about the same.

violetagg:
@nikban95 @limaoning mmmm correct. Can you create an issue to expose the response headers via `AccessLogArgProvider`. Are you interested in providing a PR?

nikban95:
Sure @violetagg. CReated an issue: https://github.com/reactor/reactor-netty/issues/1505. Will check if i can provide the pr for the same.

violetagg:
> Sure @violetagg. Created an issue: #1505. Will check if i can provide the pr for the same.

great!

Issue #1360, opened 2020-11-03
Add hooks for address resolver to propagate trace context


anilgursel:
We would like to be able to trace address resolution while establishing a connection.  Accordingly, proposing to add new hooks to be called around `resolve`.

PR will be submitted shortly.

PR #1361, opened 2020-11-03
#1360 Add hooks for address resolver to propagate trace context

anilgursel:


reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
```diff
@@ -140,6 +142,48 @@ public T doOnDisconnected(Consumer<? super Connection> doOnDisconnected) {
 		return dup;
 	}
 
+	/**
+	 * Set or add a callback called before {@link SocketAddress} is resolved.
+	 *
+	 * @param doOnResolve a consumer observing resolve events
+	 * @return a new {@link ClientTransport} reference
+	 */
+	public T doOnResolve(Consumer<? super Channel> doOnResolve) {
```
violetagg, line 151: In the `do*` API we work with `Connection` and not `Channel`.
Is it possible to make the solution based on `ConnectionObserver` events?

anilgursel: @violetagg I have few questions about this request:

1. Inside `doResolveAndConnect` I do not have access to `Connection`.  Do you have a suggestion on how to achieve it?

2. I would like to be able to pass information from `doOnResolve` to `doAfterResponse`.  The use case is that we would like to time the resolution time and add to the trace.  Unless I put a mutable object inside the Reactor `Context`, I am not sure how I can do it by just relying on `ConnectionObserver`.

3. I see the following logic in `doResolveAndConnect`:

```java
			if (!resolver.isSupported(remoteAddress) || resolver.isResolved(remoteAddress)) {
				MonoChannelPromise monoChannelPromise = new MonoChannelPromise(channel);
				doConnect(remoteAddress, bindAddress, monoChannelPromise);
				return monoChannelPromise;
			}
```

Does this mean that it would call the `resolve` method only once?  The subsequent resolutions would be from cache?  If so, is there an invalidation time for the cache?

violetagg: > @violetagg I have few questions about this request:
> 
> 1. Inside `doResolveAndConnect` I do not have access to `Connection`.  Do you have a suggestion on how to achieve it?

`reactor.netty.Connection#from(io.netty.channel.Channel)`

> 2. I would like to be able to pass information from `doOnResolve` to `doAfterResponse`.  The use case is that we would like to time the resolution time and add to the trace.  Unless I put a mutable object inside the Reactor `Context`, I am not sure how I can do it by just relying on `ConnectionObserver`.

You can either use Reactor Context or Netty Channel Attributes
With channel attributes it is something like `channel.attr(...).get()`

> 3. I see the following logic in `doResolveAndConnect`:
> 
> ```java
> 			if (!resolver.isSupported(remoteAddress) || resolver.isResolved(remoteAddress)) {
> 				MonoChannelPromise monoChannelPromise = new MonoChannelPromise(channel);
> 				doConnect(remoteAddress, bindAddress, monoChannelPromise);
> 				return monoChannelPromise;
> 			}
> ```
> 
> Does this mean that it would call the `resolve` method only once? The subsequent resolutions would be from cache? If so, is there an invalidation time for the cache?

Did you check the new functionality around DNS resolver?
`reactor.netty.transport.NameResolverProvider`
`reactor.netty.transport.ClientTransport#resolver(java.util.function.Consumer<reactor.netty.transport.NameResolverProvider.NameResolverSpec>)`

anilgursel: 1. Thanks @violetagg!  I just pushed a commit with updates to use `Connection` instead.  I was not aware of the `Connection#from(io.netty.channel.Channel)` API before.

2. I previously misunderstood and thought I would pass `ConnectionObserver` in the lamdbda, that' why couldn't find a way to access to `Channel`.  Now allowing access to `Connection` object, I can use Netty Channel Attributes.

3. Thanks for the pointer.

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
```diff
@@ -140,6 +142,48 @@ public T doOnDisconnected(Consumer<? super Connection> doOnDisconnected) {
...
+	/**
+	 * Set or add a callback called after {@link SocketAddress} is resolved.
+	 *
+	 * @param doAfterResolve a consumer observing resolved events
+	 * @return a new {@link ClientTransport} reference
+	 */
+	public T doAfterResolve(BiConsumer<? super Connection, ? super SocketAddress> doAfterResolve) {
+		Objects.requireNonNull(doAfterResolve, "doAfterResolve");
+		T dup = duplicate();
+		BiConsumer<Connection, SocketAddress> current = ((ClientTransportConfig)configuration()).doAfterResolve;
```
violetagg, line 168: Wdyt?

```
@SuppressWarnings("unchecked")
BiConsumer<Connection, SocketAddress> current = (BiConsumer<Connection, SocketAddress>) configuration().doAfterResolve;
```

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
```diff
@@ -140,6 +142,48 @@ public T doOnDisconnected(Consumer<? super Connection> doOnDisconnected) {
...
+	/**
+	 * Set or add a callback called if an exception happens whlie resolving to a {@link SocketAddress}.
+	 *
+	 * @param doOnResolveError a consumer observing resolve error events
+	 * @return a new {@link ClientTransport} reference
+	 */
+	public T doOnResolveError(BiConsumer<? super Connection, ? super Throwable> doOnResolveError) {
+		Objects.requireNonNull(doOnResolveError, "doOnResolveError");
+		T dup = duplicate();
+		BiConsumer<Connection, Throwable> current = ((ClientTransportConfig)configuration()).doOnResolveError;
```
violetagg, line 182: Wdyt?

```
@SuppressWarnings("unchecked")
BiConsumer<Connection, Throwable> current = (BiConsumer<Connection, Throwable>) configuration().doOnResolveError;
```

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
```diff
@@ -18,9 +18,11 @@
 import java.net.SocketAddress;
 import java.time.Duration;
 import java.util.Objects;
+import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
+import io.netty.channel.Channel;
```
violetagg, line 25: nit: unnecessary import

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
```diff
@@ -140,6 +142,48 @@ public T doOnDisconnected(Consumer<? super Connection> doOnDisconnected) {
...
+	public T doOnResolve(Consumer<? super Connection> doOnResolve) {
+		Objects.requireNonNull(doOnResolve, "doOnResolve");
+		T dup = duplicate();
+		Consumer<Connection> current = ((ClientTransportConfig)configuration()).doOnResolve;
+		dup.configuration().doOnResolve = current == null ? doOnResolve : current.andThen(doOnResolve);
+		return dup;
+	}
+
+	/**
+	 * Set or add a callback called after {@link SocketAddress} is resolved.
```
violetagg, line 160: nit: `is resolved successfully.` instead of `is resolved.`

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
```diff
@@ -140,6 +142,48 @@ public T doOnDisconnected(Consumer<? super Connection> doOnDisconnected) {
...
+	public T doAfterResolve(BiConsumer<? super Connection, ? super SocketAddress> doAfterResolve) {
+		Objects.requireNonNull(doAfterResolve, "doAfterResolve");
+		T dup = duplicate();
+		BiConsumer<Connection, SocketAddress> current = ((ClientTransportConfig)configuration()).doAfterResolve;
+		dup.configuration().doAfterResolve = current == null ? doAfterResolve : current.andThen(doAfterResolve);
+		return dup;
+	}
+
+	/**
+	 * Set or add a callback called if an exception happens whlie resolving to a {@link SocketAddress}.
```
violetagg, line 174: nit: typo `whlie`

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConnector.java
```diff
@@ -219,7 +220,31 @@ static void doConnect(SocketAddress remoteAddress, @Nullable Supplier<? extends
...
+						}
+					}
+					else {
+						if (clientTransportConfig.doAfterResolve != null) {
+							clientTransportConfig.doAfterResolve.accept(Connection.from(channel), future.getNow());
+						}
+					}
+				});
+			}
+
```
violetagg, line 247: I was thinking that here we can use `config.connectionObserver()` and send events instead of casting to a specific config but with the current `ConnectionObserver` we cannot propagate `Connection` + `SocketAddress`. (this was why I mentioned `ConnectionObserver` in my previous comment)

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConnector.java
```diff
@@ -219,7 +220,31 @@ static void doConnect(SocketAddress remoteAddress, @Nullable Supplier<? extends
...
+			if (config instanceof ClientTransportConfig) {
+				final ClientTransportConfig clientTransportConfig = (ClientTransportConfig)config;
+				if (clientTransportConfig.doOnResolve != null) {
+					clientTransportConfig.doOnResolve.accept(Connection.from(channel));
+				}
+			}
+
 			Future<SocketAddress> resolveFuture = resolver.resolve(remoteAddress);
+
+			if (config instanceof ClientTransportConfig) {
```
violetagg, line 232: What about checking that we have `doOnResolveError`/`doAfterResolve` and then adding a listener.
May be in the common cases (where we do not need to propagate something) these callbacks will not exist.

violetagg (changes requested):

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConnector.java
```diff
@@ -219,7 +220,31 @@ static void doConnect(SocketAddress remoteAddress, @Nullable Supplier<? extends
 				return monoChannelPromise;
 			}
 
+			if (config instanceof ClientTransportConfig) {
+				final ClientTransportConfig clientTransportConfig = (ClientTransportConfig)config;
```
violetagg, line 224: `final ClientTransportConfig<?> clientTransportConfig = (ClientTransportConfig<?>) config;`

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConnector.java
```diff
@@ -219,7 +220,31 @@ static void doConnect(SocketAddress remoteAddress, @Nullable Supplier<? extends
...
+				if (clientTransportConfig.doOnResolve != null) {
+					clientTransportConfig.doOnResolve.accept(Connection.from(channel));
+				}
+			}
+
 			Future<SocketAddress> resolveFuture = resolver.resolve(remoteAddress);
+
+			if (config instanceof ClientTransportConfig) {
+				final ClientTransportConfig clientTransportConfig = (ClientTransportConfig)config;
```
violetagg, line 233: `final ClientTransportConfig<?> clientTransportConfig = (ClientTransportConfig<?>) config;`

violetagg (approved):

violetagg:
@simonbasle PTAL

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
simonbasle: @anilgursel please explain how this conversation is resolved, nothing seems to have changed so at a minimum you should answer the comment rather than sneaky closing the conversation...

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConnector.java
```diff
@@ -219,7 +220,35 @@ static void doConnect(SocketAddress remoteAddress, @Nullable Supplier<? extends
...
+					clientTransportConfig.doOnResolve.accept(Connection.from(channel));
+				}
+			}
+
 			Future<SocketAddress> resolveFuture = resolver.resolve(remoteAddress);
+
+			if (config instanceof ClientTransportConfig) {
+				final ClientTransportConfig<?> clientTransportConfig = (ClientTransportConfig<?>) config;
+
+				if (clientTransportConfig.doOnResolveError != null) {
```
simonbasle, line 235: I'd combine these two blocks (test+addListener) into a single listener, but I guess the cost of adding an extra listener is not that high

reactor-netty-core/src/test/java/reactor/netty/transport/ClientTransportResolverHooksTest.java
```diff
@@ -0,0 +1,97 @@
...
+import reactor.netty.tcp.TcpClient;
+import reactor.netty.tcp.TcpServer;
+
+import java.net.UnknownHostException;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ClientTransportResolverHooksTest {
```
simonbasle, line 31: JUnit5 tests need not have `public` modifier on class and methods. let's have newly introduced test classes follow that best practice.

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
violetagg: @anilgursel @simonbasle I think javadoc update is added by mistake on the `doOnResolve` method above instead of `doAfterResolve` method

reactor-netty-core/src/test/java/reactor/netty/transport/ClientTransportResolverHooksTest.java
```diff
@@ -0,0 +1,97 @@
...
+				})
+				.host("localhost")
+				.port(server.port())
+				.doOnResolve(conn -> doOnResolve.set(conn.channel().attr(TRACE_ID_KEY).get()))
+				.doAfterResolve((conn, socketAddress) -> doAfterResolve.set(conn.channel().attr(TRACE_ID_KEY).get()))
+				.doOnResolveError((conn, th) -> doOnResolveError.set(conn.channel().attr(TRACE_ID_KEY).get()))
+				.connect()
+				.block();
+
+		assertThat(doOnResolve.get()).isEqualTo(TRACE_ID_VALUE);
```
simonbasle, line 57: `assertThat(doOnResolve).hasValue(TRACE_ID_VALUE)` is more idiomatic in AssertJ

simonbasle: (also applies to other assertions below)

reactor-netty-core/src/test/java/reactor/netty/transport/ClientTransportResolverHooksTest.java
```diff
@@ -0,0 +1,97 @@
...
+	public void shouldCallHooksInErrorScenario() {
+
+		final AtomicLong doOnResolve = new AtomicLong(0);
+		final AtomicLong doAfterResolve = new AtomicLong(0);
+		final AtomicLong doOnResolveError = new AtomicLong(0);
+		final AtomicReference<Throwable> throwable = new AtomicReference<>();
+
+		final DisposableServer server = TcpServer.create().port(0).bindNow();
+
+		try {
```
simonbasle, line 72: maybe convert to either A) a `StepVerifier` or B) an `assertThatExceptionOfType(SpecificException.class).isThrownBy(() -> { ... }) block ?

simonbasle (commented):
some minor comments on tests, assertions and the way the 2 listeners are added

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConnector.java
anilgursel: This was originally a single listener.  However, @violetagg had a good suggestion that we do not want to register the listener unnecessarily when the hooks are not there.  I can do a single listener and achieve both requests; however, that would cause the code to look a little uglier (making some checks multiple times).  Here is how it would look like:

```java
if (config instanceof ClientTransportConfig) {
	final ClientTransportConfig<?> clientTransportConfig = (ClientTransportConfig<?>) config;

	if (clientTransportConfig.doOnResolveError != null || clientTransportConfig.doAfterResolve != null) {
		resolveFuture.addListener((FutureListener<SocketAddress>) future -> {
			if (future.cause() != null) {
				if (clientTransportConfig.doOnResolveError != null) {
					clientTransportConfig.doOnResolveError.accept(Connection.from(channel), future.cause());
				}
			} else {
				if (clientTransportConfig.doAfterResolve != null) {
					clientTransportConfig.doAfterResolve.accept(Connection.from(channel), future.getNow());
				}
			}
		});
	}
}
```

I could potentially make `doOnResolveError` and `doAfterResolve` use `Optional` to eliminate some of the `if` checks, but that would cause a deviation from other hooks.  In addition, I am not sure if the code would be any nicer.

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
anilgursel: @simonbasle Apparently, I added the comment on the wrong one.  Sorry for the mistake, and thank you for catching it!

reactor-netty-core/src/test/java/reactor/netty/transport/ClientTransportResolverHooksTest.java
anilgursel: Removed `public` from class and test methods.

reactor-netty-core/src/test/java/reactor/netty/transport/ClientTransportResolverHooksTest.java
anilgursel: Made the suggested code changes.

reactor-netty-core/src/test/java/reactor/netty/transport/ClientTransportResolverHooksTest.java
anilgursel: Changed the code to `.as(StepVerifier::create).verifyError(UnknownHostException.class);`.

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConnector.java
simonbasle: yeah that's what I was getting at. seeing a snippet of code helps, indeed there's no huge gain in readability so let's keep as is 👍

simonbasle (approved):

violetagg:
@anilgursel I squashed and committed the changes bb1326b. Thanks for that contribution!

PR #1364, opened 2020-11-04
[tests] Switch from deprecated Processors to Sinks

violetagg:


violetagg:
@simonbasle PTAL

simonbasle (commented):
looks good, although technically all these sinks could use the `Sinks.unsafe()` rootSpec. and in tests, it would also be ok to use `tryEmitNext(t).orThrow()`.

violetagg:
> looks good, although technically all these sinks could use the `Sinks.unsafe()` rootSpec. and in tests, it would also be ok to use `tryEmitNext(t).orThrow()`.

For the rest of tests I agree but for `MonoSendManyTest` we cannot do that because we try to race between `cancel` and `emit`?

```
RaceTestUtils.race(testSubscriber::cancel, () -> {
	for (ByteBuf buf : buffersToSend) {
		source.tryEmitNext(buf).orThrow();
	}
});
```

```
reactor.core.publisher.Sinks$EmissionException: Sink emission failed with FAIL_CANCELLED
        at reactor.core.publisher.Sinks$EmitResult.orThrow(Sinks.java:174)
        at reactor.netty.channel.MonoSendManyTest.lambda$shouldNotLeakIfFusedOnRacingCancelAndOnNext$21(MonoSendManyTest.java:234)
```

violetagg:
> looks good, although technically all these sinks could use the `Sinks.unsafe()` rootSpec. and in tests, it would also be ok to use `tryEmitNext(t).orThrow()`.

PTAL

simonbasle (approved):

PR #1365, opened 2020-11-04
Switch from deprecated Processors to Sinks

violetagg:


violetagg:
@simonbasle PTAL

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
```diff
@@ -133,11 +132,10 @@ protected ChannelOperations(ChannelOperations<INBOUND, OUTBOUND> replaced) {
 	 * @param connection the new {@link Connection} connection
 	 * @param listener the events callback
 	 */
-	@SuppressWarnings("deprecation")
 	public ChannelOperations(Connection connection, ConnectionObserver listener) {
 		this.connection = requireNonNull(connection, "connection");
 		this.listener = requireNonNull(listener, "listener");
-		this.onTerminate = MonoProcessor.create();
+		this.onTerminate = Sinks.empty();
```
simonbasle, line 138: I think this could be `Sinks.unsafe().empty()`, unless there is a need to add overhead of multi-threaded access detection ?

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
```diff
@@ -377,16 +377,14 @@ public Context currentContext() {
...
+		final Sinks.Empty<Void> onTerminate;
 		final InstrumentedPool<PooledConnection> pool;
 
 		PooledRef<PooledConnection> pooledRef;
 
-		@SuppressWarnings("deprecation")
 		PooledConnection(Channel channel, InstrumentedPool<PooledConnection> pool) {
 			this.channel = channel;
-			this.onTerminate = MonoProcessor.create();
+			this.onTerminate = Sinks.empty();
```
simonbasle, line 387: same here, `Sinks.unsafe().empty()`

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientFormEncoder.java
```diff
@@ -80,7 +78,7 @@
 		this.newCharset = charset;
 		this.request = request;
 		this.cleanOnTerminate = true;
-		this.progressFlux = DirectProcessor.create();
+		this.progressSink = Sinks.many().multicast().directBestEffort();
```
simonbasle, line 81: just as a reminder: if there are multiple subscribers with different demand, the slowest subscribers will not see pushed values (when they're at request 0) but they won't be terminated either, contrary to `DirectProcessor`. if NONE of the subscribers have enough request to process the value, however, the sink terminates with an overflow exception

simonbasle (changes requested):
globally, if you were using the processors directly (via `onNext`) it should be safe to use `Sinks.unsafe()` and avoid the extra overhead of the serialized versions (which check for multi-threaded access). this is more of a concern in production code than in tests, so be sure to review usage carefully. from what I can see though, unsafe can be used everywhere in this PR

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
```diff
@@ -453,7 +451,7 @@ protected final void terminate() {
 			// when there is no response state
 			onInboundComplete();
 			afterInboundComplete();
-			onTerminate.onComplete();
+			onTerminate.emitEmpty(Sinks.EmitFailureHandler.FAIL_FAST);
```
rstoyanchev, line 454: I think `tryEmit()` could be sufficient here. If there are competing attempts to terminate, I'm guessing it doesn't matter which one wins? If that is the case you can also initialize via `Sinks.unsafe().empty()` and avoid unnecessary protection.

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
```diff
@@ -454,7 +452,7 @@ public void onStateChange(Connection connection, State newState) {
 				                                 pool.metrics().acquiredSize(),
 				                                 pool.metrics().idleSize());
 				                     }
-				                     onTerminate.onComplete();
+				                     onTerminate.emitEmpty(Sinks.EmitFailureHandler.FAIL_FAST);
```
rstoyanchev, line 455: I think this can also be `Sinks.unsafe().empty()` with `tryEmit()` since the signals are serialized already. See [this mention](https://github.com/reactor/reactor-core/blob/332b18ba2cf25ff9d93d01456d6ec79f30a61ccc/reactor-core/src/main/java/reactor/core/publisher/Sinks.java#L99-L101) for example.

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -225,7 +223,8 @@ public NettyOutbound send(Publisher<? extends ByteBuf> dataStream) {
 			return FutureMono.deferFuture(() -> {
 				if (CLOSE_SENT.getAndSet(this, 1) == 0) {
 					discard();
-					onCloseState.onNext(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()));
+					onCloseState.emitValue(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()),
+							Sinks.EmitFailureHandler.FAIL_FAST);
```
rstoyanchev, line 227: This should probably be handled differently. Either ignore a competing attempt (which could be underlying connection closed?) or use `tryEmit` and handle return value in order to at least release the frame. Otherwise `emitValue` could raise `EmissionException` which would bubble up.

reactor-netty-http/src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
```diff
@@ -219,7 +218,8 @@ protected void onInboundCancel() {
 			return FutureMono.deferFuture(() -> {
 				if (CLOSE_SENT.getAndSet(this, 1) == 0) {
 					discard();
-					onCloseState.onNext(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()));
+					onCloseState.emitValue(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()),
+							Sinks.EmitFailureHandler.FAIL_FAST);
```
rstoyanchev, line 222: Similar comment as for the client side.

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
violetagg: It is guaranteed that this will be invoked in an event loop. I'll change it.

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
violetagg: It is guaranteed that this will be invoked in an event loop. I'll change it to unsafe.

rstoyanchev (commented):
Some comments before mostly for the 0 and 0..1 cases. 

I don't know enough about `Sinks.many().multicast().directBestEffort()` so I'll let someone else comment. One thing I wonder about it is whether the new logic can raise an exception under some circumstances and whether that should be handled in some way or allowed to propagate?

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientFormEncoder.java
violetagg: > just as a reminder: if there are multiple subscribers with different demand, the slowest subscribers will not see pushed values (when they're at request 0) but they won't be terminated either, contrary to `DirectProcessor`.

Reactor Netty is the subscriber here, no other subscribers.

> if NONE of the subscribers have enough request to process the value, however, the sink terminates with an overflow exception

Reactor Netty may stop requesting if the channel is not writable.

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
violetagg: It is guaranteed that will be invoked in an event loop, so I think `Sinks.unsafe().empty()` should be OK.
Your suggestion is to change from `emitEmpty(Sinks.EmitFailureHandler.FAIL_FAST)` to `tryEmit`?

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
violetagg: > a competing attempt (which could be underlying connection closed?)

Isn't this guarded by `if (CLOSE_SENT.getAndSet(this, 1) == 0) {`
We should try to emit just once

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientFormEncoder.java
simonbasle: if there is a single subscriber, wouldn't `unicast()` fit better? note there is two flavors: onBackpressureBuffer and onBackpressureError

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
rstoyanchev: True this is only the attempt to emit. In that case `.unsafe()` with `tryEmit()` should suffice.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientFormEncoder.java
violetagg: ok I'll change that to `unicast()`

about back pressure strategy: till now we have `progressSink.asFlux().onBackpressureLatest()`?

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
violetagg: done

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
violetagg: done

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientFormEncoder.java
violetagg: please check

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
violetagg: please check

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
violetagg: please check

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
violetagg: please check

reactor-netty-http/src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
violetagg: please check

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -223,8 +223,7 @@ public NettyOutbound send(Publisher<? extends ByteBuf> dataStream) {
 			return FutureMono.deferFuture(() -> {
 				if (CLOSE_SENT.getAndSet(this, 1) == 0) {
 					discard();
-					onCloseState.emitValue(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()),
-							Sinks.EmitFailureHandler.FAIL_FAST);
+					onCloseState.tryEmitValue(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()));
```
simonbasle, line 226: I don't think that was what @rstoyanchev was advocating for. unlike with `tryEmitEmpty`, it's riskier to ignore the `EmitResult` of `tryEmitValue`. Have you gone through the possible failure modes and made sure all of them can either be ignored or are guaranteed not to occur?

I suspect the `CLOSE_SENT` guard protects against most (all) possible failures, but in that case be sure to again put a comment that `EmitResult` is ignored on purpose. maybe even go as far as detailing the ignored failure modes (so that if one is added later, it forces us to reason about this piece of code again)

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -245,13 +244,11 @@ void sendCloseNow(@Nullable CloseWebSocketFrame frame) {
 		}
 		if (CLOSE_SENT.getAndSet(this, 1) == 0) {
 			if (frame != null) {
-				onCloseState.emitValue(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()),
-						Sinks.EmitFailureHandler.FAIL_FAST);
+				onCloseState.tryEmitValue(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()));
```
simonbasle, line 247: same as the other `tryEmitValue`

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -245,13 +244,11 @@ void sendCloseNow(@Nullable CloseWebSocketFrame frame) {
...
 				channel().writeAndFlush(frame)
 				         .addListener(ChannelFutureListener.CLOSE);
 			} else {
-				onCloseState.emitValue(new WebSocketCloseStatus(-1, ""),
-						Sinks.EmitFailureHandler.FAIL_FAST);
+				onCloseState.tryEmitValue(new WebSocketCloseStatus(-1, ""));
```
simonbasle, line 251: same as the other `tryEmitValue`

reactor-netty-http/src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
```diff
@@ -218,8 +218,7 @@ protected void onInboundCancel() {
 			return FutureMono.deferFuture(() -> {
 				if (CLOSE_SENT.getAndSet(this, 1) == 0) {
 					discard();
-					onCloseState.emitValue(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()),
-							Sinks.EmitFailureHandler.FAIL_FAST);
+					onCloseState.tryEmitValue(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()));
```
simonbasle, line 221: same as the other `tryEmitValue`

reactor-netty-http/src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
```diff
@@ -240,13 +239,11 @@ void sendCloseNow(@Nullable CloseWebSocketFrame frame, ChannelFutureListener lis
 		}
 		if (CLOSE_SENT.getAndSet(this, 1) == 0) {
 			if (frame != null) {
-				onCloseState.emitValue(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()),
-						Sinks.EmitFailureHandler.FAIL_FAST);
+				onCloseState.tryEmitValue(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()));
```
simonbasle, line 242: same as the other `tryEmitValue`

reactor-netty-http/src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
```diff
@@ -240,13 +239,11 @@ void sendCloseNow(@Nullable CloseWebSocketFrame frame, ChannelFutureListener lis
...
 				channel().writeAndFlush(frame)
 				         .addListener(listener);
 			} else {
-				onCloseState.emitValue(new WebSocketCloseStatus(-1, ""),
-						Sinks.EmitFailureHandler.FAIL_FAST);
+				onCloseState.tryEmitValue(new WebSocketCloseStatus(-1, ""));
```
simonbasle, line 246: same as the other `tryEmitValue`

simonbasle (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
rstoyanchev: This looks good to me. It's already synchronized so it can't be non-serialized, it can't overflow, and it doesn't matter if it's cancelled.

violetagg: `FAIL_TERMINATED` - `CLOSE_SENT` guarantees only one emission will happen
`FAIL_OVERFLOW`, `FAIL_NON_SERIALIZED`, `FAIL_ZERO_SUBSCRIBER` - should not be a problem here?
`FAIL_CANCELLED` -  what to do here?

rstoyanchev: Whoever is listening for CloseStatus, cancelled, which means they no longer care, so I'd say nothing to do. FWIW we have a [similar case](https://github.com/spring-projects/spring-framework/blob/e1f51b4bf80c7284f7c6ad918385a24b680b6549/spring-webflux/src/main/java/org/springframework/web/reactive/socket/adapter/AbstractListenerWebSocketSession.java#L219) in the Spring Framework.

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
```diff
@@ -451,7 +451,7 @@ protected final void terminate() {
 			// when there is no response state
 			onInboundComplete();
 			afterInboundComplete();
-			onTerminate.emitEmpty(Sinks.EmitFailureHandler.FAIL_FAST);
+			onTerminate.tryEmitEmpty();
```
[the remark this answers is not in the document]
violetagg: done

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
```diff
@@ -442,7 +442,7 @@ public void onStateChange(Connection connection, State newState) {
 				                                 pool.metrics().idleSize(),
 				                                 t);
 				                     }
-				                     onTerminate.emitEmpty(Sinks.EmitFailureHandler.FAIL_FAST);
+				                     onTerminate.tryEmitEmpty();
```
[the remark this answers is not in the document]
violetagg: done

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
```diff
@@ -452,7 +452,7 @@ public void onStateChange(Connection connection, State newState) {
 				                                 pool.metrics().acquiredSize(),
 				                                 pool.metrics().idleSize());
 				                     }
-				                     onTerminate.emitEmpty(Sinks.EmitFailureHandler.FAIL_FAST);
+				                     onTerminate.tryEmitEmpty();
```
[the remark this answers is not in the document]
violetagg: done

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
violetagg: done

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
violetagg: done

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
violetagg: done

reactor-netty-http/src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
violetagg: done

reactor-netty-http/src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
violetagg: done

reactor-netty-http/src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
violetagg: done

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
simonbasle: great 👍

simonbasle (approved):

Issue #1370, opened 2020-11-07
Performance regression in Spring Cloud Gateway


violetagg:
Related to https://github.com/spring-cloud/spring-cloud-gateway/issues/1956

johndudo:
when are you going to release 0.9.14.RELEASE, before  0.9.14.RELEASE come out,  use 0.9.7 instead?  Is 0.9.7 stable?

violetagg:
The issue is a regression caused by #1185 that is released with 0.9.10.RELEASE

@johndudo For the moment 0.9.14.RELEASE is scheduled for 07.12
The regression is introduced in 0.9.10.RELEASE so you can either use 0.9.7.RELEASE or 0.9.9.RELEASE

violetagg:
`0.9.14.RELEASE` is available with the fix from the PR #1371

PR #1371, opened 2020-11-07
Guarantee there is no race between subscribe and other operations invoked in an event loop

violetagg:
This is a quick fix for the regression introduced with #1185

Related to #1370 and #1350


violetagg:
@OlegDokuka @simonbasle PTAL

PR #1384, opened 2020-11-16
Fix NoClassDefFoundError: LaQute/bnd/build/Project

violetagg:
Bnd plugin is not available from Spring's plugins repository


violetagg:
This has been fixed with 6774bbec8252bd3057f9c150d6448932fd24ab67

PR #1403, opened 2020-12-01
fix #1394 Ensure the correct protocol (H2C/H2) is applied in case SslHandler is applied based on the http/https scheme

### LINKED ISSUE #1394: HttpClient cannot be configured with both HTTP11 and H2
## Expected Behavior

HttpClient should be able to automatically select HTTP2 for https if supported or HTTP1.1 for both http and https.

## Actual Behavior

When I configure HttpClient with .protocol(HTTP11, H2) it fails when connecting to http URLs with IllegalArgumentException("Configured H2 protocol without TLS. Use H2 Clear-Text protocol via HttpClient#protocol or configure TLS via HttpClient#secure") 

## Steps to Reproduce

1. Configure HTTP client with SSL context and .protocol(HTTP11, H2)
2. Try to send request to http URLs 

## Possible Solution

Here is the code fragment from HttpClientConnect.MonoHttpConnect with my comments
```java
                               if (handler.toURI.isSecure()) {
					if (_config.sslProvider == null) {
						_config = new HttpClientConfig(config);
						_config.sslProvider = HttpClientSecure.defaultSslProvider(_config);
					}
				}
				else {
					if (_config.sslProvider != null) {
						_config = new HttpClientConfig(config);
						_config.sslProvider = null;//We clear sslProvider for http URLs
					}
				}

				if (_config.sslProvider != null) {
					if (_config.sslProvider.getDefaultConfigurationType() == null) {
						if ((_config._protocols & HttpClientConfig.h2) == HttpClientConfig.h2) {
							_config.sslProvider = SslProvider.updateDefaultConfiguration(_config.sslProvider,
									SslProvider.DefaultConfigurationType.H2);
						}
						else {
							_config.sslProvider = SslProvider.updateDefaultConfiguration(_config.sslProvider,
									SslProvider.DefaultConfigurationType.TCP);
						}
					}

					if ((_config._protocols & HttpClientConfig.h2c) == HttpClientConfig.h2c) {
						sink.error(new IllegalArgumentException(
								"Configured H2 Clear-Text protocol with TLS. " +
										"Use the non Clear-Text H2 protocol via HttpClient#protocol or disable TLS " +
										"via HttpClient#noSSL()"));
						return;
					}
				}
				else {
                                        //This check should only be done for HTTPS URLs. We can configure multiple supported protocols besides H2
					if ((_config._protocols & HttpClientConfig.h2) == HttpClientConfig.h2) {
						sink.error(new IllegalArgumentException(
								"Configured H2 protocol without TLS. Use H2 Clear-Text " +
										"protocol via HttpClient#protocol or configure TLS via HttpClient#secure"));
						return;
					}
				}
```

## Your Environment

* Reactor version(s) used: 
Maven:io.projectreactor.netty:reactor-netty-core:1.0.1 
Maven: io.projectreactor.netty:reactor-netty-http:1.0.1 
Maven: io.projectreactor:reactor-core:3.4.0

* Other relevant libraries versions (eg. `netty`, ...):netty 4.1.54.Final
* JVM version (`java -version`):Oracle 11.0.4
* OS and version (eg. `uname -a`):Windows 10

violetagg:
Fixes #1394 

In case of HTTP/1.1
- if the scheme is HTTPS and there IS NO security provider specified -> the default one is used
- if the scheme is HTTP and there IS security provider specified -> the security provider is removed

In case of HTTP/2
- if the scheme is HTTPS and the specified protocol is H2C -> the protocol will be replaced with H2.
The security provider will be configured AFTER the protocol so that Reactor Netty can choose the correct
default security provider based on the protocol.
- if the scheme is HTTP and the specified protocol is H2 -> the protocol will be replaced with H2C.
The specified security provider is removed.

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
@@ -18,8 +18,10 @@ package reactor.netty.http.client;
 
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.time.Duration;
 import java.util.Objects;
@@ -51,6 +53,7 @@ import reactor.netty.ConnectionObserver;
 import reactor.netty.NettyOutbound;
 import reactor.netty.channel.AbortedException;
 import reactor.netty.http.HttpOperations;
+import reactor.netty.http.HttpProtocol;
 import reactor.netty.resources.ConnectionProvider;
 import reactor.netty.tcp.TcpClientConfig;
 import reactor.netty.transport.AddressUtils;
@@ -208,19 +211,22 @@ class HttpClientConnect extends HttpClient {
 				if (handler.toURI.isSecure()) {
 					if (_config.sslProvider == null) {
 						_config = new HttpClientConfig(config);
+						if (checkProtocol(_config, HttpClientConfig.h2c) && _config.protocols.length > 1) {
+							removeIncompatibleProtocol(_config, HttpProtocol.H2C);
+						}
 						_config.sslProvider = HttpClientSecure.defaultSslProvider(_config);
 					}
-				}
-				else {
-					if (_config.sslProvider != null) {
-						_config = new HttpClientConfig(config);
-						_config.sslProvider = null;
-					}
-				}
 
-				if (_config.sslProvider != null) {
+					if (checkProtocol(_config, HttpClientConfig.h2c)) {
+						sink.error(new IllegalArgumentException(
+								"Configured H2 Clear-Text protocol with TLS. " +
+										"Use the non Clear-Text H2 protocol via HttpClient#protocol or disable TLS " +
+										"via HttpClient#noSSL()"));
+						return;
+					}
+
 					if (_config.sslProvider.getDefaultConfigurationType() == null) {
-						if ((_config._protocols & HttpClientConfig.h2) == HttpClientConfig.h2) {
+						if (checkProtocol(_config, HttpClientConfig.h2)) {
 							_config.sslProvider = SslProvider.updateDefaultConfiguration(_config.sslProvider,
 									SslProvider.DefaultConfigurationType.H2);
 						}
@@ -229,17 +235,17 @@ class HttpClientConnect extends HttpClient {
 									SslProvider.DefaultConfigurationType.TCP);
 						}
 					}
-
-					if ((_config._protocols & HttpClientConfig.h2c) == HttpClientConfig.h2c) {
-						sink.error(new IllegalArgumentException(
-								"Configured H2 Clear-Text protocol with TLS. " +
-										"Use the non Clear-Text H2 protocol via HttpClient#protocol or disable TLS " +
-										"via HttpClient#noSSL()"));
-						return;
-					}
 				}
 				else {
-					if ((_config._protocols & HttpClientConfig.h2) == HttpClientConfig.h2) {
+					if (_config.sslProvider != null) {
+						_config = new HttpClientConfig(config);
+						if (checkProtocol(_config, HttpClientConfig.h2) && _config.protocols.length > 1) {
+							removeIncompatibleProtocol(_config, HttpProtocol.H2);
+						}
+						_config.sslProvider = null;
+					}
+
+					if (checkProtocol(_config, HttpClientConfig.h2)) {
 						sink.error(new IllegalArgumentException(
 								"Configured H2 protocol without TLS. Use H2 Clear-Text " +
 										"protocol via HttpClient#protocol or configure TLS via HttpClient#secure"));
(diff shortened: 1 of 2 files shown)


jGleitz:
From my point of view, this solution implements behaviour that I, as a user, would not expect: It uses protocols I have not configured. I think `reactor-netty` should only ever use a protocol that was configured on the client. It should pick the protocol with the highest version number that supports the target’s URI scheme. Hence, I’d expect the behaviour to be as follows:

Configured Client Protocols | Server URI scheme | Used Protocol
--- | --- | ---
{`H2`} | `http` | _error_
{`H2`} | `https` | `H2` (pre-known)
{`H2C`} | `http` | `H2C` (pre-known)
{`H2C`} | `https` | _error_
{`HTTP11`} | `http` | `HTTP11`
{`HTTP11`} | `https` | `HTTP11`
{`H2`, `H2C`} | `http` | `H2C` (pre-known)
{`H2`, `H2C`} | `https` | `H2` (pre-known)
{`H2`, `HTTP11`} | `http` | `HTTP11`
{`H2`, `HTTP11`} | `https` | `H2` (offered via ALPN, fallback to `HTTP11`)
{`H2C`, `HTTP11`} | `http` | `H2C` (offered via `Upgrade` header, fallback to `HTTP11`)
{`H2C`, `HTTP11`} | `https` | `HTTP11` 
{`H2`, `H2C`, `HTTP11`} | `http` | `H2C` (offered via `Upgrade` header, fallback to `HTTP11`)
{`H2`, `H2C`, `HTTP11`} | `https` | `H2` (offered via ALPN, fallback to `HTTP11`)

violetagg:
@jGleitz Then isn't it the current behaviour the expected one?

jGleitz:
@violetagg unfortunately, no. As reported in #1394, if you configure `HTTP11` and `H2` and make a request against an `http` URL, the client will throw an error, while it should be using `HTTP11`.

simonbasle:
my sentiment while reviewing this is a bit similar to what @jGleitz said: it feels like the code makes a sneaky change to the configuration, one that isn't even documented anywhere. Ultimately I'll defer to you @violetagg, but regarding the table provided by @jGleitz, from an external perspective it makes sense to me. I'd also prefer the approach of an error if reactor-netty cannot reconcile the scheme with the protocols configured by the user.

violetagg:
@jGleitz @simonbasle PTAL

reactor-netty-http/src/test/java/reactor/netty/http/Http2Tests.java
```diff
@@ -491,29 +506,57 @@ void testIssue1394_HttpEndpointH2Protocol() {
...
+
+	private void doTestIssue1394_ProtocolSchemeNotCompatible(HttpProtocol protocol, String scheme) {
+		HttpClient.create()
+		          .protocol(protocol)
+		          .wiretap(true)
+		          .get()
+		          .uri(scheme + "://example.com")
+		          .responseSingle((res, bytes) -> Mono.just(res.responseHeaders().get("x-http2-stream-id")))
+		          .as(StepVerifier::create)
+		          .expectError(IllegalArgumentException.class)
```
simonbasle, line 559: having a meaningful message would be beneficial, and testing it here would be a form of minimal documentation

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -209,24 +211,32 @@ public void subscribe(CoreSubscriber<? super Connection> actual) {
 				if (handler.toURI.isSecure()) {
 					if (_config.sslProvider == null) {
 						_config = new HttpClientConfig(config);
-						for (int i = 0; i < _config.protocols.length; i++) {
-							if (_config.protocols[i] == HttpProtocol.H2C) {
-								_config.protocols[i] = HttpProtocol.H2;
+						if ((_config._protocols & HttpClientConfig.h2c) == HttpClientConfig.h2c &&
```
simonbasle, line 214: consider extracting that choice algorithm into a separate method, which can be tested separately

reactor-netty-http/src/test/java/reactor/netty/http/Http2Tests.java
```diff
@@ -479,7 +480,21 @@ private void doTestMonoRequestBodySentAsFullRequest(Publisher<? extends ByteBuf>
 	}
 
 	@Test
-	void testIssue1394_HttpEndpointH2Protocol() {
+	void testIssue1394_SchemeHttp() {
+		// When only H2C is specified, "prior-knowledge" is used and stream id is 3
+		doTestIssue1394_SchemeHttp("3", HttpProtocol.H2C);
+		doTestIssue1394_SchemeHttp("3", HttpProtocol.H2, HttpProtocol.H2C);
+
+		// When only HTTP11 is specified, "Upgrade" header is used and stream id is 1
```
simonbasle, line 488: the comment should read "when both HTTP11 and H2C are specified"

jGleitz:
From reading the tests, I think that this PR now implements the behaviour I would `reactor-netty` expect to have! :+1:

reactor-netty-http/src/test/java/reactor/netty/http/Http2Tests.java
simonbasle: isn't this a case where H2C was expected? or does the streamid 1 indicate H2C was selected ? (sorry I'm not familiar with the protocol and concepts of prior knowledge vs upgrade header). My feeling is that the implementation and the tests are hard to correlate with the matrix from the discussion, and would benefit from an isolated utility method that plainly returns the selected protocol, given a scheme and an array of protocols as input. wdyt?

simonbasle (changes requested):

reactor-netty-http/src/test/java/reactor/netty/http/Http2Tests.java
violetagg: done

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
violetagg: done

reactor-netty-http/src/test/java/reactor/netty/http/Http2Tests.java
violetagg: comments are updated

rstoyanchev (approved):

simonbasle (approved):

simonbasle (approved):
I would still add a few comments detailing what happens when one falls through the ifs (as an alternative to explicit `else` branches), but otherwise looks ok

PR #1416, opened 2020-12-09
fix #1406 When connection is closed abnormally, emit WebSocketCloseStatus.ABNORMAL_CLOSURE

### LINKED ISSUE #1406: onCloseState in WebsocketServerOperations#sendCloseNow should emit 1006 if the session is closed without a close frame
## Expected Behavior

onCloseState in reactor.netty.http.server.WebsocketServerOperations#sendCloseNow should emit 1006 (NO_CLOSE_FRAME) instead of -1 if no close frame.
```
	void sendCloseNow(@Nullable CloseWebSocketFrame frame, ChannelFutureListener listener) {
		if (frame != null && !frame.isFinalFragment()) {
			channel().writeAndFlush(frame);
			return;
		}
		if (CLOSE_SENT.getAndSet(this, 1) == 0) {
			if (frame != null) {
				onCloseState.tryEmitValue(new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()));
				channel().writeAndFlush(frame)
				         .addListener(listener);
			} else {
				onCloseState.tryEmitValue(new WebSocketCloseStatus(-1, "")); // should emit 1006 instead of -1
				channel().writeAndFlush(new CloseWebSocketFrame())
				         .addListener(listener);
			}
		}
		else if (frame != null) {
			frame.release();
		}
	}
```

## Actual Behavior

org.springframework.web.reactive.socket.CloseStatus throw "java.lang.IllegalArgumentException: Invalid status code" because reactor.netty.http.server.WebsocketServerOperations#sendCloseNow tries to create an WebSocketCloseStatus with the status code -1.

## Steps to Reproduce

The bug happens when a WebSocket session is closed without a close frame.

## Possible Solution

onCloseState in reactor.netty.http.server.WebsocketServerOperations#sendCloseNow should emit 1006 (NO_CLOSE_FRAME) instead of -1 if no close frame.

## Your Environment

reactor-netty-http-1.0.1
spring-webflux-5.3.1

violetagg:
According to the specification (below) the emitted WebSocketCloseStatus will be
WebSocketCloseStatus.ABNORMAL_CLOSURE, but the CloseWebSocketFrame
will be still created with status code "-1" as 1006 cannot be set as a status code for
CloseWebSocketFrame.

"1006 is a reserved value and MUST NOT be set as a status code in a
Close control frame by an endpoint.  It is designated for use in
applications expecting a status code to indicate that the
connection was closed abnormally, e.g., without sending or
receiving a Close control frame."

Fixes #1406

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
@@ -147,10 +147,8 @@ final class WebsocketClientOperations extends HttpClientOperations
 			if (log.isDebugEnabled()) {
 				log.debug(format(channel(), "CloseWebSocketFrame detected. Closing Websocket"));
 			}
-			CloseWebSocketFrame close = (CloseWebSocketFrame) msg;
-			sendCloseNow(new CloseWebSocketFrame(true,
-					close.rsv(),
-					close.content()));
+			sendCloseNow(new CloseWebSocketFrame(true, ((CloseWebSocketFrame) msg).rsv(),
+					((CloseWebSocketFrame) msg).content()));
 			onInboundComplete();
 		}
 		else if (msg != LastHttpContent.EMPTY_LAST_CONTENT) {
@@ -163,7 +161,7 @@ final class WebsocketClientOperations extends HttpClientOperations
 		if (log.isDebugEnabled()) {
 			log.debug(format(channel(), "Cancelling Websocket inbound. Closing Websocket"));
 		}
-		sendCloseNow(null);
+		sendCloseNow(new CloseWebSocketFrame(), WebSocketCloseStatus.ABNORMAL_CLOSURE);
 	}
 
 	@Override
@@ -181,7 +179,7 @@ final class WebsocketClientOperations extends HttpClientOperations
 			if (log.isDebugEnabled()) {
 				log.debug(format(channel(), "Outbound error happened"), err);
 			}
-			sendCloseNow(new CloseWebSocketFrame(1002, "Client internal error"));
+			sendCloseNow(new CloseWebSocketFrame(WebSocketCloseStatus.PROTOCOL_ERROR));
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


simonbasle (approved):

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -240,32 +240,22 @@ public NettyOutbound send(Publisher<? extends ByteBuf> dataStream) {
 	}
 
 	@SuppressWarnings("FutureReturnValueIgnored")
-	void sendCloseNow(@Nullable CloseWebSocketFrame frame) {
-		if (frame != null && !frame.isFinalFragment()) {
+	void sendCloseNow(CloseWebSocketFrame frame, WebSocketCloseStatus closeStatus) {
+		if (!frame.isFinalFragment()) {
```
rstoyanchev, line 244: Really minor but `WebSocketCloseStatus` could be `@Nullable`, and if not provided, derived as previously so that way for the normal case `CloseWebSocketFrame` could contain the status but for the 1006 case it is provided separately.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
```diff
@@ -739,7 +741,7 @@ public void operationComplete(ChannelFuture future)  {
 		public void onComplete() {
 			if (ops.channel()
 			       .isActive()) {
-				ops.sendCloseNow(null, this);
+				ops.sendCloseNow(new CloseWebSocketFrame(), new WebSocketCloseStatus(-1, ""), this);
```
rstoyanchev, line 744: My understanding was that emitting a close status of -1 leads to ISE in WebFlux. Couldn't this be a status of 1001 indicating normal completion?

rstoyanchev (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
violetagg: done

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
violetagg: done

violetagg:
@rstoyanchev PTAL

rstoyanchev (approved):

JamesChenX (approved):

PR #1421, opened 2020-12-11
fix #1417 Add helpers for filtering the access log

### LINKED ISSUE #1417: Expose more convenient way to selectively perform access logs
## Motivation
Before open this issue,i read issue about access log #1357 and #436,but what i want is not log some api,for example, access log  not log this API。
~~~
@GetMapping(value = "/v5/gateway/liveness")
public Mono<Void> liveness() {
	return Mono.empty();
}
~~~
## Desired solution
If use tomcat,i can do like this:
~~~java
public class SelfTomcatAccessLog extends AccessLogValve {

    private final TomcatLogger tomcatLogger = new TomcatLogger();

    public SelfTomcatAccessLog() {
        super();
    }

    @Override
    public void log(java.io.CharArrayWriter message) {
        String messageContent = message.toString();
        if (!messageContent.contains("liveness")) {
            tomcatLogger.info(messageContent);
        }
    }
}
~~~
can reactor-netty support method like this or other methods?

aneveu:
Provides two factories methods into `AccessLog` to allow filtering logs on predicate matching, with or without custom String format.

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
@@ -228,7 +228,7 @@ public abstract class HttpServer extends ServerTransport<HttpServer, HttpServerC
 	 *           .route(r -> r.get("/hello",
 	 *                   (req, res) -> res.header(CONTENT_TYPE, TEXT_PLAIN)
 	 *                                    .sendString(Mono.just("Hello World!"))))
-	 *           .accessLog(AccessLogFactory.create(
+	 *           .accessLog(true, AccessLogFactory.createFilter(
 	 *                   args -> String.valueOf(args.uri()).startsWith("/health"),
 	 *                   args -> AccessLog.create("user-agent={}", args.requestHeader("user-agent"))
 	 *            )
@@ -239,7 +239,7 @@ public abstract class HttpServer extends ServerTransport<HttpServer, HttpServerC
 	 * </pre>
 	 * <p>
 	 * The {@link AccessLogFactory} class offers several helper methods to generate such a function,
-	 * notably if one wants to {@link AccessLogFactory#create(Predicate) filter} some requests out of the access log.
+	 * notably if one wants to {@link AccessLogFactory#createFilter(Predicate) filter} some requests out of the access log.
 	 *
 	 * Note that this method takes precedence over the {@value reactor.netty.ReactorNetty#ACCESS_LOG_ENABLED} system property.
 	 *
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
@@ -23,6 +23,7 @@ import java.util.function.BiFunction;
 import java.util.function.BiPredicate;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 
 import io.netty.channel.group.ChannelGroup;
@@ -227,15 +228,18 @@ public abstract class HttpServer extends ServerTransport<HttpServer, HttpServerC
 	 *           .route(r -> r.get("/hello",
 	 *                   (req, res) -> res.header(CONTENT_TYPE, TEXT_PLAIN)
 	 *                                    .sendString(Mono.just("Hello World!"))))
-	 *           .accessLog(true, AccessLogFactory.create(
+	 *           .accessLog(AccessLogFactory.create(
+	 *                   args -> String.valueOf(args.uri()).startsWith("/health"),
 	 *                   args -> AccessLog.create("user-agent={}", args.requestHeader("user-agent"))
-	 *           )
+	 *            )
 	 *           .bindNow()
 	 *           .onDispose()
 	 *           .block();
 	 * }
 	 * </pre>
 	 * <p>
+	 * The {@link AccessLogFactory} class offers several helper methods to generate such a function,
+	 * notably if one wants to {@link AccessLogFactory#create(Predicate) filter} some requests out of the access log.
 	 *
 	 * Note that this method takes precedence over the {@value reactor.netty.ReactorNetty#ACCESS_LOG_ENABLED} system property.
 	 *
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
@@ -22,6 +22,9 @@ import java.util.Objects;
 
 /**
  * Log the http access information into a Logger named {@code reactor.netty.http.server.AccessLog} at INFO level.
+ * <p>
+ * See {@link AccessLogFactory} for convenience methods to create an access log factory to be passed to
+ * {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory)} during server configuration.
  *
  * @author limaoning
  * @since 1.0.1
@@ -42,6 +45,7 @@ public final class AccessLog {
 	public static AccessLog create(String logFormat, Object... args) {
 		return new AccessLog(logFormat, args);
 	}
+
 
 	void log() {
 		if (LOG.isInfoEnabled()) {
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
@@ -46,7 +46,6 @@ public final class AccessLog {
 		return new AccessLog(logFormat, args);
 	}
 
-
 	void log() {
 		if (LOG.isInfoEnabled()) {
 			LOG.info(logFormat, args);
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
@@ -16,18 +16,53 @@
 package reactor.netty.http.server.logging;
 
 import java.util.function.Function;
+import java.util.function.Predicate;
 
-/**
- * An interface to declare more concisely a {@link Function} that apply an {@link AccessLog} by an
- * {@link AccessLogArgProvider}.
- * <p>
- * Can be used in {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory) accessLog} method
- * for example.
- *
+ /**
  * @author Simon Baslé
  * @author Audrey Neveu
  * @since 1.0.3
  */
 public interface AccessLogFactory extends Function<AccessLogArgProvider, AccessLog> {
+
+	/**
+	 * Helper method to create an access log factory that selectively enables access logs.
+	 * <p>
+	 * Any request (represented as an {@link AccessLogArgProvider}) that doesn't match the
+	 * provided {@link Predicate} is excluded from the access log. Other requests are logged
+	 * using the default format.
+	 *
+	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
+	 * @return an access log factory {@link Function} to be used in
+	 * {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory)}
+	 * @since 1.0.3
+	 */
+	static AccessLogFactory create(Predicate<AccessLogArgProvider> predicate) {
+		return input -> predicate.test(input) ? BaseAccessLogHandler.DEFAULT_ACCESS_LOG.apply(input) : null;
+	}
+
+	/**
+	 * Helper method to create an access log factory that selectively enables access logs and customizes
+	 * the format to apply.
+	 * <p>
+	 * Any request (represented as an {@link AccessLogArgProvider}) that doesn't match the
+	 * provided {@link Predicate} is excluded from the access log. Other requests are logged
+	 * using the provided formatting {@link Function}.
+	 * Said function is expected to {@link AccessLog#create(String, Object...) create} an {@link AccessLog} instance,
+	 * defining both the String format and a vararg of the relevant arguments, extracted from the
+	 * {@link AccessLogArgProvider}.
+	 * <p>
+	 *
+	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
+	 * @param formatFunction the {@link Function} that creates {@link AccessLog} instances, encapsulating the format
+	 * and the extraction of relevant arguments
+	 * @return an access log factory {@link Function} to be used in
+	 * {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory)}
+	 * @since 1.0.3
+	 */
+	static AccessLogFactory create(Predicate<AccessLogArgProvider> predicate,
+			Function<AccessLogArgProvider, AccessLog> formatFunction) {
+		return input -> predicate.test(input) ? formatFunction.apply(input) : null;
+	}
 
 }
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
@@ -57,7 +57,7 @@ public interface AccessLogFactory extends Function<AccessLogArgProvider, AccessL
 	 * defining both the String format and a vararg of the relevant arguments, extracted from the
 	 * {@link AccessLogArgProvider}.
 	 * <p>
- 	 *
+	 *
 	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
 	 * @param formatFunction the {@link AccessLogFactory} that creates {@link AccessLog} instances, encapsulating the
 	 * format
@@ -65,7 +65,7 @@ public interface AccessLogFactory extends Function<AccessLogArgProvider, AccessL
 	 * @return an {@link AccessLogFactory} to be used in
 	 * {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory)}
 	 * @since 1.0.3
- 	 */
+	 */
 	static AccessLogFactory createFilter(Predicate<AccessLogArgProvider> predicate, AccessLogFactory formatFunction) {
 		return input -> predicate.test(input) ? formatFunction.apply(input) : null;
 	}
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
@@ -18,7 +18,7 @@ package reactor.netty.http.server.logging;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
- /**
+/**
  * An interface to declare more concisely a {@link Function} that apply an {@link AccessLog} by an
  * {@link AccessLogArgProvider}.
  * <p>
@@ -38,7 +38,7 @@ public interface AccessLogFactory extends Function<AccessLogArgProvider, AccessL
 	 * using the default format.
 	 *
 	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
-	 * @return an access log factory {@link Function} to be used in
+	 * @return an {@link AccessLogFactory} to be used in
 	 * {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory)}
 	 * @since 1.0.3
 	 */
@@ -57,14 +57,15 @@ public interface AccessLogFactory extends Function<AccessLogArgProvider, AccessL
 	 * defining both the String format and a vararg of the relevant arguments, extracted from the
 	 * {@link AccessLogArgProvider}.
 	 * <p>
-	 *
+ 	 *
 	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
-	 * @param formatFunction the {@link Function} that creates {@link AccessLog} instances, encapsulating the format
+	 * @param formatFunction the {@link AccessLogFactory} that creates {@link AccessLog} instances, encapsulating the
+	 * format
 	 * and the extraction of relevant arguments
-	 * @return an access log factory {@link Function} to be used in
+	 * @return an {@link AccessLogFactory} to be used in
 	 * {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory)}
 	 * @since 1.0.3
-	 */
+ 	 */
 	static AccessLogFactory createFilter(Predicate<AccessLogArgProvider> predicate, AccessLogFactory formatFunction) {
 		return input -> predicate.test(input) ? formatFunction.apply(input) : null;
 	}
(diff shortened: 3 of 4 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
```diff
@@ -19,15 +19,22 @@
...
 import java.util.Objects;
+import java.util.function.Function;
+import java.util.function.Predicate;
 
 /**
  * Log the http access information.
+ * All logs are written to a Logger named {@code reactor.netty.http.server.AccessLog} at INFO level.
+ * <p>
+ * This class also exposes convenience methods to create an access log factory to be passed to {@link reactor.netty.http.server.HttpServer#accessLog(Function)}
+ * during server configuration. Note that access logging must be globally enabled first for that configuration to
```
violetagg, line 30: This class has a single purpose to log with the provided logFormat and arguments using a specific logger.
I propose to keep this single purpose and create a new factory class containing the two additional methods.
With this change we will have single and simple purpose for each class.
Wdyt?

violetagg (changes requested):

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -91,25 +91,20 @@
...
 import reactor.netty.http.client.HttpClientRequest;
 import reactor.netty.http.client.PrematureCloseException;
+import reactor.netty.http.server.logging.AccessLog;
 import reactor.netty.resources.ConnectionProvider;
 import reactor.netty.resources.LoopResources;
 import reactor.netty.tcp.TcpClient;
 import reactor.netty.tcp.TcpServer;
 import reactor.netty.transport.TransportConfig;
 import reactor.test.StepVerifier;
+import reactor.test.util.TestLogger;
```
violetagg, line 107: I think this class can be removed from the change?

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
simonbasle: mmh yeah I see what you mean. `AccessLog` seemed the most logical existing place a user would look for such convenience methods, since the configuration at `HttpServer` level requires a `Function` that produces an `AccessLog`.

We also thought that encapsulating the filtering in the generated `Function` was better than exposing an API that tries too hard to stick to single responsibility principle by directly returning an `AccessLog`, eg.:

```java
//option 1
static public Function<AccessLogArgsProvider, AccessLog> filtering(Predicate<AccessLogArgsProvider> predicate);

//option 2
static public AccessLog filtering(AccessLogArgsProvider args, Predicate<AccessLogArgsProvider> predicate);

HttpServer.create()
    .accessLog(AccessLog.filtering(somePredicate))
    .accessLog(args -> AccessLog.filtering(args, somePredicate))
```

The second option seems redundant, and the goal is to facilitate the writing of the `Function` itself after all...

Also note that while users could arguably write the `somePredicate.test(args)` part and `return null` part themselves, when the predicate matches they'd have no way to mimic the default `AccessLog` pattern:

```java
HttpServer.create()
    .accessLog(args -> somePredicate.test(args) ? new AccessLog(????uh oh????) : null)
```

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
simonbasle: indeed

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
violetagg: what about to deprecate the current method and to expose 
`.accessLog(*Factory)`, `*Factory` will implement Function and also expose this convenient methods?

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
simonbasle: fixed

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
simonbasle: ok we tried that in the latest commit, but frankly I'm not convinced there is high value in adding an interface (and consequently a configuration method) just for that...

the `AccessLog` class has little to no public responsibility (the only public API is the `create` factory method), so I'd tend to say having it expose factory methods for factory functions is not that much of a stretch (these functions do ultimately represent "a way to create an `AccessLog`", after all).

having a look at the relevant commit, do you still strongly feel there is value in separating these concerns @violetagg?

violetagg: from API point of view it is more obvious to use `accessLog(AccessLogFormat)` in comparison with
`accessLog(Function)` and then checking that actually `AccsessLog` provides also some kind of factory methods that can be used with `accessLog(Function)`

wdyt?

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -0,0 +1,64 @@
...
+	 * Helper method to create an access log factory that selectively enables access logs.
+	 * <p>
+	 * Any request (represented as an {@link AccessLogArgProvider}) that doesn't match the
+	 * provided {@link Predicate} is excluded from the access log. Other requests are logged
+	 * using the default format.
+	 *
+	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
+	 * @return an access log factory {@link Function} to be used in {@link reactor.netty.http.server.HttpServer#accessLog(Function)}
+	 */
+	static AccessLogFactory createFilter(Predicate<AccessLogArgProvider> predicate) {
```
violetagg, line 37: Do we really need to name the methods depending on the arguments? Can't we use just `create`?

For example in `ConnectionProvider` we do not name the `create` methods depending on the arguments.
`reactor.netty.resources.ConnectionProvider#create(java.lang.String)`
`reactor.netty.resources.ConnectionProvider#create(java.lang.String, int)`

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -91,14 +91,7 @@
...
 import reactor.core.publisher.SignalType;
-import reactor.netty.ByteBufFlux;
-import reactor.netty.ChannelBindException;
-import reactor.netty.Connection;
-import reactor.netty.ConnectionObserver;
-import reactor.netty.DisposableServer;
-import reactor.netty.FutureMono;
-import reactor.netty.NettyOutbound;
-import reactor.netty.NettyPipeline;
+import reactor.netty.*;
```
violetagg, line 94: we can remove this class from the change.

simonbasle: yes, incremental changes prevented me from realizing the class was supposed to be back to what's on `master` 👍

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
simonbasle: we do not really _need_ to, it's just a practice that allows extensibility. It is one of the benefits of factory methods over constructors: you can add more factory methods later on, taking the same _type_ of parameters but with different _semantics_ that are made clearer through the factory method's name.

that said, in the current case I don't expect a high chance of adding variants later on, except maybe the `defaultFactory()` one.

violetagg: In Reactor Netty we use `create` for the most used combinations of configurations, everything should be a `Builder` if we expect to have a lot of configurations.

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
simonbasle: yeah, it's marginally better at naturally redirecting users to the correct place for factory methods. ok, let's keep it in `AccessLogFactory` then 👍

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
simonbasle: ok, let's stick with the common pattern 👍

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
```diff
@@ -22,12 +22,18 @@
 
 /**
  * Log the http access information.
+ * All logs are written to a Logger named {@code reactor.netty.http.server.AccessLog} at INFO level.
+ * <p>
+ * See {@link AccessLogFactory} for convenience methods to create an access log factory to be passed to
+ * {@link reactor.netty.http.server.HttpServer#accessLog(AccessLogFactory)} during server configuration.
+ * Note that access logging must be globally enabled first for that configuration to
+ * be taken into account, see the {@link reactor.netty.ReactorNetty#ACCESS_LOG_ENABLED} system property.
```
simonbasle, line 30: the `@since 1.0.1` can be added there as well @aneveu (class existed before that but was package private)

simonbasle:
converting to draft since we'd like to address #1428 first, then rework this one on top

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -16,18 +16,53 @@
...
+import java.util.function.Predicate;
 
-/**
- * An interface to declare more concisely a {@link Function} that apply an {@link AccessLog} by an
- * {@link AccessLogArgProvider}.
- * <p>
- * Can be used in {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory) accessLog} method
- * for example.
- *
+ /**
```
simonbasle, line 21: The comment about factories can be put back

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -16,18 +16,53 @@
...
+	 * <p>
+	 *
+	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
+	 * @param formatFunction the {@link Function} that creates {@link AccessLog} instances, encapsulating the format
+	 * and the extraction of relevant arguments
+	 * @return an access log factory {@link Function} to be used in
+	 * {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory)}
+	 * @since 1.0.3
+	 */
+	static AccessLogFactory create(Predicate<AccessLogArgProvider> predicate,
```
simonbasle, line 63: maybe use an `AccessLogFactory` as the second parameter here too?

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -16,18 +16,53 @@
...
+	 * Any request (represented as an {@link AccessLogArgProvider}) that doesn't match the
+	 * provided {@link Predicate} is excluded from the access log. Other requests are logged
+	 * using the default format.
+	 *
+	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
+	 * @return an access log factory {@link Function} to be used in
+	 * {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory)}
+	 * @since 1.0.3
+	 */
+	static AccessLogFactory create(Predicate<AccessLogArgProvider> predicate) {
```
simonbasle, line 40: I still get a strong feeling the `create` keyword is too generic. An `AccessLogFactory` is not inherently filtering... wdyt @violetagg?

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/AccessLogTest.java
```diff
@@ -135,29 +123,83 @@ void secondCallToAccessLogOverridesPreviousOne() {
...
+		disposableServer = HttpServer.create()
+				.port(8080)
+				.handle((req, resp) -> {
+					resp.withConnection(conn -> {
+						ChannelHandler handler = conn.channel().pipeline().get(NettyPipeline.AccessLogHandler);
+						resp.header(ACCESS_LOG_HANDLER, handler != null ? FOUND : NOT_FOUND);
+					});
+					return resp.send();
+				})
+				.accessLog(true, AccessLogFactory.create(p -> !String.valueOf(p.uri()).startsWith("/foo/"),
```
simonbasle, line 165: note: we now try to avoid `foo` as a best practice, `example` is better

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/AccessLogTest.java
```diff
@@ -135,29 +123,83 @@ void secondCallToAccessLogOverridesPreviousOne() {
...
+		disposableServer = HttpServer.create()
+				.port(8080)
+				.handle((req, resp) -> {
+					resp.withConnection(conn -> {
+						ChannelHandler handler = conn.channel().pipeline().get(NettyPipeline.AccessLogHandler);
+						resp.header(ACCESS_LOG_HANDLER, handler != null ? FOUND : NOT_FOUND);
+					});
+					return resp.send();
+				})
+				.accessLog(true, AccessLogFactory.create(p -> !String.valueOf(p.uri()).startsWith("/foo/")))
```
simonbasle, line 142: note: we now try to avoid `foo` as a best practice, `example` is better

simonbasle (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -169,15 +170,18 @@ public final HttpServer accessLog(boolean enable) {
 	 *           .route(r -> r.get("/hello",
 	 *                   (req, res) -> res.header(CONTENT_TYPE, TEXT_PLAIN)
 	 *                                    .sendString(Mono.just("Hello World!"))))
-	 *           .accessLog(true, AccessLogFactory.create(
+	 *           .accessLog(AccessLogFactory.create(
```
violetagg, line 173: The example should stay with `boolean`

`.accessLog(true, AccessLogFactory.create(`

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
violetagg: @simonbasle @aneveu I will leave the decision to you

violetagg:
@aneveu you can switch to regular PR as it is still marked as draft

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
aneveu: nice catch, thanks

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
simonbasle: nit: extraneous leading space

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLog.java
```diff
@@ -43,6 +46,7 @@ public static AccessLog create(String logFormat, Object... args) {
 		return new AccessLog(logFormat, args);
 	}
 
+
```
simonbasle, line 49: nit: extraneous blank line

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -16,18 +16,57 @@
...
+	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
+	 * @return an access log factory {@link Function} to be used in
+	 * {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory)}
+	 * @since 1.0.3
+	 */
+	static AccessLogFactory createFilter(Predicate<AccessLogArgProvider> predicate) {
+		return input -> predicate.test(input) ? BaseAccessLogHandler.DEFAULT_ACCESS_LOG.apply(input) : null;
+	}
+
+	/**
```
simonbasle, line 49: there seems to be something wrong with how your IDE is configured: there's an extra space before each comment line in this block

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -16,18 +16,57 @@
...
  *
  * @author Simon Baslé
  * @author Audrey Neveu
  * @since 1.0.3
  */
 public interface AccessLogFactory extends Function<AccessLogArgProvider, AccessLog> {
 
-}
\ No newline at end of file
+	/**
```
simonbasle, line 33: there seems to be something wrong with how your IDE is configured: there's an extra space before each comment line in this block

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -16,18 +16,57 @@
...
+	 * Any request (represented as an {@link AccessLogArgProvider}) that doesn't match the
+	 * provided {@link Predicate} is excluded from the access log. Other requests are logged
+	 * using the provided formatting {@link Function}.
+	 * Said function is expected to {@link AccessLog#create(String, Object...) create} an {@link AccessLog} instance,
+	 * defining both the String format and a vararg of the relevant arguments, extracted from the
+	 * {@link AccessLogArgProvider}.
+	 * <p>
+	 *
+	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
+	 * @param formatFunction the {@link Function} that creates {@link AccessLog} instances, encapsulating the format
```
simonbasle, line 62: not a `Function` anymore, change the link to `AccessLogFactory`

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -16,18 +16,57 @@
...
+	 * and the extraction of relevant arguments
+	 * @return an access log factory {@link Function} to be used in
```
simonbasle, line 64: not a `Function` anymore, change the link to `AccessLogFactory`

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -16,18 +16,57 @@
...
\ No newline at end of file
+	/**
+	 * Helper method to create an access log factory that selectively enables access logs.
+	 * <p>
+	 * Any request (represented as an {@link AccessLogArgProvider}) that doesn't match the
+	 * provided {@link Predicate} is excluded from the access log. Other requests are logged
+	 * using the default format.
+	 *
+	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
+	 * @return an access log factory {@link Function} to be used in
```
simonbasle, line 41: not a `Function` anymore, change the link to `AccessLogFactory`

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
simonbasle: for reference: https://github.com/reactor/reactor-netty/pull/1421#discussion_r547262245

I voted in favor of using a more specific suffix rather than just `create`.

simonbasle (changes requested):
some last minute comments and nitpicks

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
aneveu: I don't really know why is that as I don't see anything special in my configuration 😞

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
simonbasle: ah, nevermind, only the first line was problematic. the other lines are fine, but the rendering in the ui (probably with a plugin I use) displays a small artifact that looks like an extra space:
image

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -16,18 +16,58 @@
...
+	 * the format to apply.
+	 * <p>
+	 * Any request (represented as an {@link AccessLogArgProvider}) that doesn't match the
+	 * provided {@link Predicate} is excluded from the access log. Other requests are logged
+	 * using the provided formatting {@link Function}.
+	 * Said function is expected to {@link AccessLog#create(String, Object...) create} an {@link AccessLog} instance,
+	 * defining both the String format and a vararg of the relevant arguments, extracted from the
+	 * {@link AccessLogArgProvider}.
+	 * <p>
+ 	 *
```
simonbasle, line 60: newly introduced leading space here

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -16,18 +16,58 @@
...
+	 * @param predicate the filter that returns {@code true} if the request should be logged, {@code false} otherwise
+	 * @param formatFunction the {@link AccessLogFactory} that creates {@link AccessLog} instances, encapsulating the
+	 * format
+	 * and the extraction of relevant arguments
+	 * @return an {@link AccessLogFactory} to be used in
+	 * {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory)}
+	 * @since 1.0.3
+ 	 */
```
simonbasle, line 68: newly introduced leading space here

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
simonbasle: see above, this block was actually fine.

there is however now an extra leading space in some of the pseudo-blank lines below
image

simonbasle (changes requested):

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/AccessLogTest.java
```diff
@@ -135,29 +123,83 @@ void secondCallToAccessLogOverridesPreviousOne() {
...
+		HttpClientResponse response = getHttpClientResponse("/example/test");
+
+		HttpClient.create()
 				.port(disposableServer.port())
 				.wiretap(true)
 				.get()
-				.uri("/test")
+				.uri("/filtered/test")
 				.response()
 				.block();
```
violetagg, line 178: Can we replace this code with `getHttpClientResponse("/filtered/test");`?

violetagg (approved):

simonbasle (approved):

Issue #1425, opened 2020-12-12
Concurrent HTTP client creation leads to multiple LoopResources creation


anilgursel:
When multiple HTTP clients are create concurrently and there is no Reactor Netty server instance created prior to clients,  there multiple loop resources are created (all but one is disposed though).  When doing a JFR analysis, one would expect the number of reactor-http-nio-* thread count to be equal to `LoopResources.DEFAULT_IO_WORKER_COUNT`; however, with the concurrency problem it can go `num clients * LoopResources.DEFAULT_IO_WORKER_COUNT`.  Here is a reproducer of the issue:

```java
	@Test
	public void testDefaultLoopResources() {

		final String[] host = new String[] {"example.com", "google.com", "amazon.com"};

		IntStream.range(0, 2)
				.parallel()
				.forEach(i ->
						HttpClient
								.create(ConnectionProvider.builder("pool-" + i).build())
								.host(host[i])
								.port(80)
								.get()
								.response()
								.subscribe());

		final Set<Thread> threadSet = Thread.getAllStackTraces().keySet();
		threadSet.forEach(t -> System.out.println(t.getName()));
		assertThat(threadSet.stream().filter(t -> t.getName().startsWith("reactor-http-nio")))
				.hasSize(LoopResources.DEFAULT_IO_WORKER_COUNT);
	}
```

## Expected Behavior

No matter how many clients are created concurrently, `reactor-http-nio-*` threads should not go beyond `LoopResources.DEFAULT_IO_WORKER_COUNT`.

## Actual Behavior

Many resources are created and disposed which causes quite bit of confusion.

## Possible Solution

I will send a PR with a possible solution which we can discuss.

violetagg:
We will address this issue with #1455

Issue #1023, opened 2020-03-09
How to properly warmup connection pool


maczikasz:
Hey team. I am creating an application that needs to call another microservice, and add the minimum possible amount of latency doing so. So for that I wanted to give netty and the Spring webclient a try. 

On the first try i saw that calling the first time the netty version was much slower than the simple RestTemplate based ones, after some debugging and trace logging I realized that the reason this initial call was so slow is because the connection pool is only created on the first call. 

I tried adding a warmup call, which helped but I find it a not so elegant solution:
Without warmup
```
2020-03-09 22:28:10.696  INFO 4483 --- [nio-8080-exec-1] com.example.demo.ControllerConfigs       : Total took 317
2020-03-09 22:28:14.647  INFO 4483 --- [nio-8080-exec-2] com.example.demo.ControllerConfigs       : Total took 16
2020-03-09 22:28:15.746  INFO 4483 --- [nio-8080-exec-4] com.example.demo.ControllerConfigs       : Total took 14
2020-03-09 22:28:17.923  INFO 4483 --- [nio-8080-exec-6] com.example.demo.ControllerConfigs       : Total took 14
```

With warmup:
```
2020-03-09 22:29:05.307  INFO 4489 --- [nio-8080-exec-1] com.example.demo.ControllerConfigs       : Total took 32
2020-03-09 22:29:09.161  INFO 4489 --- [nio-8080-exec-2] com.example.demo.ControllerConfigs       : Total took 14
2020-03-09 22:29:09.558  INFO 4489 --- [nio-8080-exec-3] com.example.demo.ControllerConfigs       : Total took 13
2020-03-09 22:29:09.817  INFO 4489 --- [nio-8080-exec-4] com.example.demo.ControllerConfigs       : Total took 14
2020-03-09 22:29:09.992  INFO 4489 --- [nio-8080-exec-6] com.example.demo.ControllerConfigs       : Total took 13
```
Is there anything else I can do to optimize the first call performance? In my production I am running the app with very low CPU settings, so the difference is between 200ms for a normal call and 6-8 seconds for the initial one.

```kotlin
package com.example.demo

import kotlinx.coroutines.Dispatchers
import kotlinx.coroutines.reactive.awaitSingle
import kotlinx.coroutines.runBlocking
import kotlinx.coroutines.withContext
import org.slf4j.Logger
import org.slf4j.LoggerFactory
import org.springframework.beans.factory.annotation.Value
import org.springframework.context.annotation.Bean
[...]

Issue #560, opened 2018-12-28
Slow first request of Spring's WebClient caused by Netty Reactor Http Client


codependent:
The first request made with Spring WebClient takes around 6 seconds, whereas the next ones are way faster (30ms). The following Spring JIRA ticket points at Netty's initialization as the reason for this delay: https://jira.spring.io/browse/SPR-17200

I have uploaded a sample project with instructions to reproduce the problem in its readme: https://github.com/codependent/slow-webclient-sample

Reactor netty version: 0.8.3.RELEASE

JVM: Java(TM) SE Runtime Environment (build 1.8.0_161-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.161-b12, mixed mode)

OS Version: MacOS Darwin user.local 18.2.0 Darwin Kernel Version 18.2.0: Fri Oct  5 19:41:49 PDT 2018; root:xnu-4903.221.2~2/RELEASE_X86_64 x86_64

phantomedc:
i am facing a similar problem, webclient will hang on after being idle for several minutes
i reproduced this problem in my dev envrionment,I did 4 requests at ```2019-09-06 00:10:14.739```(t1)、```2019-09-06 00:19:51.274```(t2)、```2019-09-06 00:30:37.992```(t3)、```2019-09-06 00:33:13.314```(t4)，
the request1(at t1), request2(at t2)  webclient didn't idle too long time, the requests were processed fine, but request3(at t3), webclient was idle 10minutes, webclient got **Read Timeout**, then i did request4 , it was processed fine as well.
and i noticed that request1、request2、request3，they used same local port: 59838
but request4 used local port:60928

so i guess that: the connection(or channel in netty) will be cached in the connection pool, reactor netty will manage and reuse it, but when the connection idle too long, the connection may be closed or cut down by remote side(F5 loadbalance or Nginx?) or some other reason, it became disconnected on the link layer, but the connection pool did not realize what happened actually, the connection pool still thinks this connection is connected and provide it to webclient, and boom: read time out.
```java
        HttpClient httpClient = HttpClient.create()
                .tcpConfiguration(tcpClient -> {
                    tcpClient = tcpClient.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, 5000);
                    tcpClient = tcpClient.doOnConnected(conn -> conn
                            .addHandlerLast(new ReadTimeoutHandler(15000, TimeUnit.MILLISECONDS)));
                    return tcpClient;
                });
```
[...]

PR #1426, opened 2020-12-12
#1425 Do not create multiple loop resources if multiple clients are c…

anilgursel:
…reated concurrently.

I would also add the following test case; however, it requires JVM isolation to for this test to be valuable. If you know of a way to isolate the specific test to a separate JVM, please share (all I can find is isolating every test, which would be very costly).

```java
	@Test
	public void testDefaultLoopResources() {

		final String[] host = new String[] {"example.com", "google.com", "amazon.com"};
		
		IntStream.range(0, 2)
				.parallel()
				.forEach(i ->
						HttpClient
								.create(ConnectionProvider.builder("pool-" + i).build())
								.host(host[i])
								.port(80)
								.get()
								.response()
								.subscribe());

		final Set<Thread> threadSet = Thread.getAllStackTraces().keySet();
		threadSet.forEach(t -> System.out.println(t.getName()));
		assertThat(threadSet.stream().filter(t -> t.getName().startsWith("reactor-http-nio")))
				.hasSize(LoopResources.DEFAULT_IO_WORKER_COUNT);
	}
```

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/DefaultLoopResources.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/DefaultLoopResources.java
@@ -49,6 +49,7 @@ final class DefaultLoopResources extends AtomicLong implements LoopResources {
 	final AtomicReference<EventLoopGroup> cacheNativeServerLoops;
 	final AtomicReference<EventLoopGroup> cacheNativeSelectLoops;
 	final AtomicBoolean                   running;
+	final AtomicBoolean                   eventLoopGroupCreationInProgress;
 
 	DefaultLoopResources(String prefix, int workerCount, boolean daemon) {
 		this(prefix, -1, workerCount, daemon);
@@ -56,6 +57,7 @@ final class DefaultLoopResources extends AtomicLong implements LoopResources {
 
 	DefaultLoopResources(String prefix, int selectCount, int workerCount, boolean daemon) {
 		this.running = new AtomicBoolean(true);
+		this.eventLoopGroupCreationInProgress = new AtomicBoolean(false);
 		this.daemon = daemon;
 		this.workerCount = workerCount;
 		this.prefix = prefix;
@@ -201,17 +203,28 @@ final class DefaultLoopResources extends AtomicLong implements LoopResources {
 
 	@SuppressWarnings("FutureReturnValueIgnored")
 	EventLoopGroup cacheNioServerLoops() {
-		EventLoopGroup eventLoopGroup = serverLoops.get();
-		if (null == eventLoopGroup) {
+		final EventLoopGroup eventLoopGroup = serverLoops.get();
+
+		if (null != eventLoopGroup) {
+			return eventLoopGroup;
+		}
+
+		if (eventLoopGroupCreationInProgress.compareAndSet(false, true)) {
+			if (null == serverLoops.get()) {
 			EventLoopGroup newEventLoopGroup = new NioEventLoopGroup(workerCount,
 					threadFactory(this, "nio"));
-			if (!serverLoops.compareAndSet(null, newEventLoopGroup)) {
+				if (serverLoops.compareAndSet(null, newEventLoopGroup)) {
+					eventLoopGroupCreationInProgress.set(false);
+				} else {
 				//"FutureReturnValueIgnored" this is deliberate
 				newEventLoopGroup.shutdownGracefully();
 			}
-			eventLoopGroup = cacheNioServerLoops();
 		}
-		return eventLoopGroup;
+		} else {
+			while (null == serverLoops.get()) {}
+		}
+
+		return cacheNioServerLoops();
 	}
 
 	EventLoopGroup cacheNativeClientLoops() {


violetagg:
@anilgursel Interesting. If you delay a bit the checks you will see that the extra event loops were successfully destroyed. This is because by default `io.netty.util.concurrent.EventExecutorGroup#shutdownGracefully()` has `15s` timeout.
So is the requirement not to create/destroy event loop many times? Because I do not think that once the extra event loops are destroyed there will be more than `LoopResources.DEFAULT_IO_WORKER_COUNT`.
Also I think that one alternative solution would be to invoke `io.netty.util.concurrent.EventExecutorGroup#shutdownGracefully(long quietPeriod, long timeout, TimeUnit unit)` so that we destroy immediately the event loops as they we not used and we do not need neither `quietPeriod ` nor `timeout `.

```java
@Test
public void testDefaultLoopResources() {

	final String[] host = new String[] {"example.com", "google.com", "amazon.com"};
		
	IntStream.range(0, 2)
			.parallel()
			.forEach(i ->
					HttpClient
							.create(ConnectionProvider.builder("pool-" + i).build())
							.host(host[i])
							.port(80)
							.get()
							.response()
							.subscribe());

	Thread.sleep(20000); <--------Delay the checks
	final Set<Thread> threadSet = Thread.getAllStackTraces().keySet();
	threadSet.forEach(t -> System.out.println(t.getName()));
	assertThat(threadSet.stream().filter(t -> t.getName().startsWith("reactor-http-nio")))
			.hasSize(LoopResources.DEFAULT_IO_WORKER_COUNT);
}
```

anilgursel:
@violetagg Yes, I also mentioned in https://github.com/reactor/reactor-netty/issues/1425 that all but one of the loops are destroyed.  However, 

- seeing `reactor-http-nio-*` that has a number more the worker count causes confusion during JFR analysis.  When a Reactor Netty server is not in place (which was loaded prior to clients) and if an app has more than on client, this scenario is almost guaranteed to occur.  
- The resources are created with the first request (not during startup time).  The first request is slow for which there is an issue already: https://github.com/reactor/reactor-netty/issues/560.  I believe creating/destroying resources may add to this slowness.

anilgursel:
@violetagg Do you have major concerns with the proposed change?

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultLoopResources.java
```diff
@@ -201,17 +203,28 @@ EventLoopGroup cacheNioSelectLoops() {
...
+					eventLoopGroupCreationInProgress.set(false);
+				} else {
 				//"FutureReturnValueIgnored" this is deliberate
 				newEventLoopGroup.shutdownGracefully();
 			}
-			eventLoopGroup = cacheNioServerLoops();
 		}
-		return eventLoopGroup;
+		} else {
+			while (null == serverLoops.get()) {}
```
simonbasle, line 224: rather than using recursion as the last step, you could reuse `eventLoopGroup` (making it not final) and continuously try to reload it here, then return `eventLoopGroup` at the end? (`newEventLoopGroup` can also be replaced by `eventLoopGroup` in that case)

I also wonder if anything can go wrong in the `NioEventloopGroup` ctor, which would mean the other thread would get stuck in this infinite loop (ie can this be proven to be guaranteed to make progress?)

violetagg:
@anilgursel @simonbasle Basically if we add this change to NIO transport, we need the same for the native transport.

I'm thinking: We are going to add `warmup` to the server and client because in some use cases we want to preload the natives used for the transport/dns resolution/TLS. If we add this warmup, do we still need to add this change?

anilgursel:
@violetagg Is `warmup` going to be a new API which would allow creating the resources during client creation (instead of with the first request)?  If so, that should  help.  When is `warmup` planned to be implemented?

(I was actually considering a suggestion to add an option to `LoopResources.create` to aggressively create resources instead of being lazy until the first request.)

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultLoopResources.java
anilgursel: I try to avoid mutability as much as possible, that's the reason.  But, I am not objecting your suggestion.  Also, yes, there is risk.  Depending on how the `warmup` api discussion goes with @violetagg, we may close this PR.  If not, I will make the suggestions you made and also will add ways to prevent infinite loop.

violetagg:
@anilgursel I wanted to work on this in the next days, it is related to these issues #1023 and https://github.com/reactor/reactor-netty/issues/560#issuecomment-730561732

anilgursel:
Thanks @violetagg!  Looking forward to seeing it!  Between, that would cover non-native as well, right?

violetagg:
@anilgursel Yes in that warmup we can include early creation of the event loop group instead of a lazy creation (regardless whether it is NIO or native transport)

violetagg:
This PR is superseded by #1453 and #1455

PR #1429, opened 2020-12-14
Fix #1422 Remove appender from logger after each test in ReactorNettyLoggingHandlerTest

### LINKED ISSUE #1422: Mocked appender are never removed from Logger in ReactorNettyLoggingHandlerTest
Mocked appender declared into `setup` method, which is called with `@beforeEach`, are never removed from the `root` `Logger`into `ReactorNettyLoggingHandlerTest`.

## Possible Solution

Remove mocked appenders created in each method from logger into a teardown method, annotated with`@afterEach`.

aneveu:


reactor-netty-core/src/test/java/reactor/netty/transport/logging/ReactorNettyLoggingHandlerTest.java
```diff
@@ -44,23 +45,28 @@
 
 	private Appender<ILoggingEvent> mockedAppender;
 	private ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor;
+	private Logger root = (Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
```
violetagg, line 48: nit: `static final`

violetagg (changes requested):

violetagg (approved):

PR #1432, opened 2020-12-15
fix #1418 let configure WebSocket version of the client

### LINKED ISSUE #1418: How to switch WebSocketClientHandshaker version
I now use spring-boot-starter-webflux:2.4.0->reactor-netty:1.0.1, the WebSocketClient handshake fails, and the reactor-netty uses WebSocketVersion.V13, can this be dynamic?

aneveu:
@violetagg could you please tell me if the way to manage `null` and `UNKNOWN` version is fine for you?

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientSpec.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientSpec.java
@@ -15,7 +15,10 @@
  */
 package reactor.netty.http.client;
 
+import io.netty.handler.codec.http.websocketx.WebSocketVersion;
 import reactor.netty.http.websocket.WebsocketSpec;
+
+import java.util.Objects;
 
 /**
  * Websocket client configuration
@@ -26,8 +29,16 @@ import reactor.netty.http.websocket.WebsocketSpec;
 public interface WebsocketClientSpec extends WebsocketSpec {
 
 	/**
+	 * Returns the configured WebSocket version
+	 *
+	 * @return returns the configured WebSocket version
+	 * @since 1.0.3
+	 */
+	WebSocketVersion version();
+
+	/**
 	 * Create builder with default properties:<br>
-	 * version = 13
+	 * version = {@link io.netty.handler.codec.http.websocketx.WebSocketVersion#V13}
 	 * <br>
 	 * protocols = null
 	 * <br>
@@ -44,8 +55,28 @@ public interface WebsocketClientSpec extends WebsocketSpec {
 	}
 
 	final class Builder extends WebsocketSpec.Builder<Builder> {
+		WebSocketVersion version = WebSocketVersion.V13;
 
 		private Builder() {
+		}
+
+		/**
+		 * Sets websocket version to use.
+		 * Set to {@link io.netty.handler.codec.http.websocketx.WebSocketVersion#V13} by default
+		 *
+		 * @param version WebSocket version to be used
+		 * @return {@literal this}
+		 * @throws NullPointerException if version is null
+		 * @throws IllegalArgumentException if the version is unknown
+		 * @since 1.0.3
+		 */
+		public final Builder version(WebSocketVersion version) {
+			Objects.requireNonNull(version, "version");
+			if (WebSocketVersion.UNKNOWN.equals(version)) {
+				throw new IllegalArgumentException("Unknown web socket version:" + version);
+			}
+			this.version = version;
+			return this;
 		}
 
 		/**
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientSpec.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientSpec.java
@@ -55,6 +55,7 @@ public interface WebsocketClientSpec extends WebsocketSpec {
 	}
 
 	final class Builder extends WebsocketSpec.Builder<Builder> {
+
 		WebSocketVersion version = WebSocketVersion.V13;
 
 		private Builder() {
@@ -73,7 +74,7 @@ public interface WebsocketClientSpec extends WebsocketSpec {
 		public final Builder version(WebSocketVersion version) {
 			Objects.requireNonNull(version, "version");
 			if (WebSocketVersion.UNKNOWN.equals(version)) {
-				throw new IllegalArgumentException("Unknown web socket version:" + version);
+				throw new IllegalArgumentException("WebSocketVersion.UNKNOWN represents an invalid version, please provide a proper version");
 			}
 			this.version = version;
 			return this;
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientSpec.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientSpec.java
@@ -27,6 +27,8 @@ public interface WebsocketClientSpec extends WebsocketSpec {
 
 	/**
 	 * Create builder with default properties:<br>
+	 * version = 13
+	 * <br>
 	 * protocols = null
 	 * <br>
 	 * maxFramePayloadLength = 65536
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientSpecImpl.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientSpecImpl.java
@@ -15,6 +15,7 @@
  */
 package reactor.netty.http.client;
 
+import io.netty.handler.codec.http.websocketx.WebSocketVersion;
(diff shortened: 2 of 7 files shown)


violetagg:
@aneveu As discussed let's fail when `NULL` and `UNKNOWN` as we expect that when a user wants to provide a version he/she will provide a concrete version different than the default one.

violetagg (changes requested):

reactor-netty-http/src/test/java/reactor/netty/http/client/WebsocketClientOperationsTest.java
```diff
@@ -67,7 +68,7 @@ private void failOnClientServerError(
 				          .port(httpServer.port())
 				          .wiretap(true)
 				          .headersWhen(h -> login(httpServer.port()).map(token -> h.set("Authorization", token)))
-				          .websocket(WebsocketClientSpec.builder().protocols(clientSubprotocol).build())
+				          .websocket(WebsocketClientSpec.builder().version(WebSocketVersion.V08).protocols(clientSubprotocol).build())
```
violetagg, line 71: Instead of adding here a configuration for the version, I suggest to have a dedicated test for this feature.
Wdyt?

aneveu: Better idea indeed, I'll be able to test failing cases this way

reactor-netty-http/src/test/java/reactor/netty/http/client/WebsocketClientOperationsTest.java
```diff
@@ -88,4 +94,34 @@ private void failOnClientServerError(
 		                 .uri("/login")
 		                 .responseSingle((res, buf) -> Mono.just(res.status().code() + ""));
 	}
+
+	@Test
+	public void testConfigureWebSocketVersion(){
+		DisposableServer httpServer = HttpServer.create()
+				.port(0)
+				.handle((in, out) -> out.sendWebsocket((i, o) -> o.sendString(Mono.just("test"))))
+				.bindNow();
```
violetagg, line 103: I propose to check on the server that the version that was sent by the client is really `8`.
You can do that with checking this header `sec-websocket-version: 8`

violetagg:
@simonbasle PTAL

reactor-netty-http/src/test/java/reactor/netty/http/client/WebsocketClientOperationsTest.java
aneveu: great, thanks for the tip!

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientSpecImpl.java
```diff
@@ -15,15 +15,23 @@
...
 
+import io.netty.handler.codec.http.websocketx.WebSocketVersion;
 import reactor.netty.http.websocket.WebsocketSpecImpl;
 
 /**
  * @author Violeta Georgieva
  * @since 0.9.7
  */
 final class WebsocketClientSpecImpl extends WebsocketSpecImpl implements WebsocketClientSpec {
+	private final WebSocketVersion version;
```
simonbasle, line 26: nitpick: add a blank line here

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientSpec.java
```diff
@@ -42,10 +55,30 @@ static Builder builder() {
 	}
 
 	final class Builder extends WebsocketSpec.Builder<Builder> {
+		WebSocketVersion version = WebSocketVersion.V13;
```
simonbasle, line 58: nitpick: blank line before

reactor-netty-http/src/test/java/reactor/netty/http/client/WebsocketClientOperationsTest.java
simonbasle: good idea! please do capture the header in `handle` above in order to verify it

simonbasle (changes requested):

violetagg (approved):

violetagg:
@aneveu You somehow reverted the test for the header?

reactor-netty-http/src/test/java/reactor/netty/http/client/WebsocketClientOperationsTest.java
```diff
@@ -15,9 +15,13 @@
  */
 package reactor.netty.http.client;
 
+import com.sun.org.apache.xpath.internal.Arg;
```
simonbasle, line 18: bad import I think

reactor-netty-http/src/test/java/reactor/netty/http/client/WebsocketClientOperationsTest.java
```diff
@@ -54,73 +58,77 @@ private void failOnClientServerError(
...
-				                        int token = Integer.parseInt(req.requestHeaders().get("Authorization"));
-				                        if (token >= 400) {
-				                            return res.status(token).send();
-				                        }
-				                        return res.sendWebsocket((i, o) -> o.sendString(Mono.just("test")),
-				                                WebsocketServerSpec.builder().protocols(serverSubprotocol).build());
-				                    }))
-				          .wiretap(true)
-				          .bindNow();
+						.port(0)
```
simonbasle, line 61: please configure your IDE to not run formatting on the whole file (only changed line, ideally)

reactor-netty-http/src/test/java/reactor/netty/http/client/WebsocketClientOperationsTest.java
```diff
@@ -54,73 +58,77 @@ private void failOnClientServerError(
...
 	@Test
-	public void testConfigureWebSocketVersion(){
+	public void testConfigureWebSocketVersion() {
+		ArgumentCaptor<HttpHeaders> argumentCaptor = ArgumentCaptor.forClass(HttpHeaders.class);
+
 		DisposableServer httpServer = HttpServer.create()
 				.port(0)
-				.handle((in, out) -> out.sendWebsocket((i, o) -> o.sendString(Mono.just("test"))))
+				.handle((in, out) -> {
+					assertThat(in.requestHeaders().get("sec-websocket-version")).isEqualTo("8");
```
simonbasle, line 109: does the `AssertionError` escape the `handle` and actually fails the test?
I had a doubt and verified locally, it seems to log the assertion error but then hangs...

another way would be for the `sendString` below to return the content of the header, and then we can assert it in the `HttpClient...block()` call, as the body of the response.

simonbasle (changes requested):

reactor-netty-http/src/test/java/reactor/netty/http/client/WebsocketClientOperationsTest.java
simonbasle: just to be clear: the test passes in its current form _because the assertion succeeds_. If you change it to `isEqualTo("foo")` an `AssertionError` is logged but the test then hangs.

simonbasle: also, it seems you don't need Mockito at all anymore

reactor-netty-http/src/test/java/reactor/netty/http/client/WebsocketClientOperationsTest.java
```diff
@@ -48,44 +58,78 @@ private void failOnClientServerError(
...
+				.bindNow();
+
+		 HttpClient.create()
+				.port(httpServer.port())
+				.wiretap(true)
+				.websocket(WebsocketClientSpec.builder().version(WebSocketVersion.V08).build())
+				.uri("/test")
+				.handle((in, out) -> in.receive().aggregate().asString())
+				.collectList()
+				.block();
```
violetagg, line 122: I suggest to specify timeout for the `block` invocation

violetagg (changes requested):

reactor-netty-http/src/test/java/reactor/netty/http/client/WebsocketClientOperationsTest.java
simonbasle: but then if the test fails the error message would be one of timeout... unless you're suggesting to do that in addition to extracting the assertion to perform it on the result of the block?

violetagg: in addition to what you suggested
We always use timeout for `block`, `verify` etc in order to prevent tests hang forever in something happens

simonbasle (approved):

violetagg (approved):

PR #1438, opened 2020-12-21
fix #1428 allow access log to be enabled programatically

### LINKED ISSUE #1428: Provide a way to configure `AccessLog` programatically
Should we allow the users to configure `AccesLog` programmatically?

## Motivation
Actually in order to enable `AccessLog`, one needs to activate it through the `ACCESS_LOG_ENABLED` system property declared in server configuration whereas the `accessLog` method just provides configuration option, which might be confusing for the user.

## Desired solution
The user could used different combination to enable or not AccessLog, and specify or not a custom configuration, for example:

```java
HttpServer.create()
    .port(8080)
    .route(r -> r.get("/hello",
    .(req, res) -> res.header(CONTENT_TYPE, TEXT_PLAIN)
    .sendString(Mono.just("Hello World!"))))

    .accessLog(false, null) // don't enable AccessLog
    .accessLog(true, DEFAULT) // enable AccessLog and use default configuration
    .accessLog(true, argProvider -> AccessLog.create("user-agent={}", argProvider.requestHeader("user-agent"))) //enable AccessLog and use custom configuration

    .bindNow()
    .onDispose()
    .block();
```

aneveu:


violetagg (approved):

violetagg:
@simonbasle PTAL

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogFactory.java
```diff
@@ -0,0 +1,33 @@
...
+ * {@link AccessLogArgProvider}.
+ * <p>
+ * Can be used in {@link reactor.netty.http.server.HttpServer#accessLog(boolean, AccessLogFactory) accessLog} method
+ * for example.
+ *
+ * @author Simon Baslé
+ * @author Audrey Neveu
+ * @since 1.0.3
+ */
+public interface AccessLogFactory extends Function<AccessLogArgProvider, AccessLog> {
```
simonbasle, line 31: the plan is still to expose the filtering factories here in the follow-up PR, correct ?

aneveu: Yes, it is.

PR #1454, opened 2021-01-07
Add system properties for enabling SamplingAllocationStrategy

violetagg:
`reactor.netty.pool.getPermitsSamplingRate` and
`reactor.netty.pool.returnPermitsSamplingRate` are exposed
in order to be able to enable Reactor Pool's `SamplingAllocationStrategy`

### DIFF
--- a/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -55,6 +55,7 @@ import reactor.core.publisher.Operators;
 import reactor.netty.Connection;
 import reactor.netty.ConnectionObserver;
 import reactor.netty.FutureMono;
+import reactor.netty.ReactorNetty;
 import reactor.netty.channel.BootstrapHandlers;
 import reactor.netty.channel.ChannelOperations;
 import reactor.pool.InstrumentedPool;
@@ -62,6 +63,7 @@ import reactor.pool.PoolBuilder;
 import reactor.pool.PoolConfig;
 import reactor.pool.PooledRef;
 import reactor.pool.PooledRefMetadata;
+import reactor.pool.introspection.SamplingAllocationStrategy;
 import reactor.util.Logger;
 import reactor.util.Loggers;
 import reactor.util.annotation.NonNull;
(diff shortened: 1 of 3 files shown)


src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -741,12 +748,25 @@ public int hashCode() {
...
+
+			if (DEFAULT_POOL_GET_PERMITS_SAMPLING_RATE > 0d && DEFAULT_POOL_GET_PERMITS_SAMPLING_RATE < 1d
+					&& DEFAULT_POOL_RETURN_PERMITS_SAMPLING_RATE > 0d && DEFAULT_POOL_RETURN_PERMITS_SAMPLING_RATE < 1d) {
+				poolBuilder = poolBuilder.allocationStrategy(SamplingAllocationStrategy.sizeBetweenWithSampling(
+						0,
+						maxConnections,
+						DEFAULT_POOL_GET_PERMITS_SAMPLING_RATE,
+						DEFAULT_POOL_RETURN_PERMITS_SAMPLING_RATE));
+			}
+			else {
```
simonbasle, line 761: I would consider at least logging a warning if (some) of the values are set but out of 0-1 bounds (eg. somebody sets sampling rate at 50 thinking it's 50 percent). Negative can probably be silently ignored, and 0 (defaults) totally can be silently ignored, but positives above `1.0` should definitely trigger some warning. That said, maybe that check should be done in a `static final` initializer block upon loading the system property ?

violetagg: fixed

src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -750,8 +773,8 @@ public int hashCode() {
 					           .maxPendingAcquire(pendingAcquireMaxCount)
 					           .evictInBackground(evictionInterval);
 
-			if (DEFAULT_POOL_GET_PERMITS_SAMPLING_RATE > 0d && DEFAULT_POOL_GET_PERMITS_SAMPLING_RATE < 1d
-					&& DEFAULT_POOL_RETURN_PERMITS_SAMPLING_RATE > 0d && DEFAULT_POOL_RETURN_PERMITS_SAMPLING_RATE < 1d) {
+			if (DEFAULT_POOL_GET_PERMITS_SAMPLING_RATE > 0d && DEFAULT_POOL_GET_PERMITS_SAMPLING_RATE <= 1d
```
simonbasle, line 776: good catch !

simonbasle (approved):

PR #1459, opened 2021-01-11
API for disabling force retry support for HTTP client

lovingcodesk:
# Motivation
although reactor-netty's HttpClient has a retry-once on connection-reset behavior and It is effective to prevent  problems in some scenarios ,for example connection prematurely closed BEFORE response, but it can not prevent this problem in all scenarios, as you said in this pr :https://github.com/reactor/reactor-netty/pull/1183#issuecomment-739917771.
In most cases, netty serves as a high-performance API gateway server,so it is mainly responsible for forwarding requests and request is stateless, In some extreme high concurrency scenarios, the probability of connection closed BEFORE response is very high, which seriously affects the user experience. It is necessary for reactor-netty to provide an api to ensure that it will try again when there is a problem.
Our system requests more than 200 million times a day, After adding the mandatory request to try again, this problem is completely solved, and the effect is very obvious
# solution
Only if retryDisable and forceRetryDisable are set to false, the forced retry function can be enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,8 @@ With `Gradle` from [repo.spring.io](https://repo.spring.io) or `Maven Central` r
     }
 
     dependencies {
-      //compile "io.projectreactor.netty:reactor-netty:0.9.17.BUILD-SNAPSHOT"
-      compile "io.projectreactor.netty:reactor-netty:0.9.16.RELEASE"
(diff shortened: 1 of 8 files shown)


violetagg:
Superseded by #1461

Issue #1472, opened 2021-01-22
PooledConnectionProvider is creating multiple channel pools for same http client and address


kushagraThapar:
After upgrading from reactor-netty 0.9.15.RELEASE to 1.x.x (tested 1.0.3) as well, we are seeing `PooledConnectionProvider` is creating multiple channel pools from same http client and for same remote addresses. 

## Expected Behavior
For the same remote address and from the same client, existing channel pools should be re-used. 

## Actual Behavior
In latest version of reactor-netty, `PooledConnectionProvider` is creating multiple channel pools from same http client and for same remote addresses. 

This change in behavior is causing multiple channel pools being created, and exhausting the network resources, resulting in `ConnectionTimeout` Exceptions. 

## Steps to Reproduce
Here is the repro code: 
https://github.com/kushagraThapar/reactor-netty-test

Run the main application, and you will see that for 1000 requests, 1000 new channel pools are created. 
Interesting thing to note is, if you change the pom.xml and point to 0.9.15.RELEASE of reactor-netty, you will only see 5 channel pools being created (1 for each type of remote address). 

P.S: To change the pom.xml to reactor-netty version 0.9.15.RELEASE, you will have to change slight code in `ReactorNettyClient.java` in the repro, because of non-backward compatibility in the major version. 

## Possible Solution
Afters some digging, I found out that this is because of the recent change in the way of creating `PooledConnectionProvider.PoolKey` object. When calculating the `pipelineKey`, `config.channelHash()` is called, which has added new fields for creating hash value for `TransportConfig`, specially `ConnectionObserver`. 

I believe this is causing the issue for our Azure Cosmos DB Java SDK internal http client, because we use `observe` API for each http request, which is important for us to calculate some internal metrics. You can see it in the sample code here : https://github.com/kushagraThapar/reactor-netty-test/blob/master/src/main/java/com/example/ReactorNettyClient.java#L93 

## Your Environment

* Reactor version(s) used: 1.0.1, 1.0.2, 1.0.3
* Other relevant libraries versions (eg. `netty`, ...): reactor-core - latest versions (3.4.x), slf4j, log4j
* JVM version (`java -version`): 1.8
* OS and version (eg. `uname -a`): macOS 11.1

violetagg:
@kushagraThapar You are setting the observer every time when you send a request. For us this is a different observer, we do not check that configuration in 0.9.x, it is checked only in 1.0.x as we consider this incompatible change.
[...]

PR #1473, opened 2021-01-22
[tests] Test various HttpClient configurations that may affect the creation of the connection pool

violetagg:
Related to #1472

### DIFF
--- a/reactor-netty-http/src/test/java/reactor/netty/http/client/ConnectionPoolTests.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/client/ConnectionPoolTests.java
@@ -1,0 +1,353 @@
+/*
+ * Copyright (c) 2011-Present VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


reactor-netty-http/src/test/java/reactor/netty/http/client/ConnectionPoolTests.java
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	@Test
+	void testClientWithPort() {
+		HttpClient localClient = client.port(server1.port());
+		checkExpectations(
+				"server1-ConnectionPoolTests",
+				"server2-ConnectionPoolTests",
+				localClient,
+				localClient.port(server2.port()));
```
simonbasle, line 118: just a readability comment: why not put the initialisation of second `HttpClient` in its own line as `localClient2` next to `localClient`, rename the other one to `localClient1` and pass both variables to the `checkExpectations` method? this would simplify the call and allow to quickly get a sense of the two http calls made by each test.

simonbasle (commented):
just a readability comment (which applies to all these tests)

reactor-netty-http/src/test/java/reactor/netty/http/client/ConnectionPoolTests.java
violetagg: I fixed that PTAL

simonbasle (approved):

PR #1487, opened 2021-02-01
fix #1485 When CloseWebSocketFrame with -1 status code (undefined/empty), emit WebSocketCloseStatus.EMPTY

### LINKED ISSUE #1485: WebsocketServerOperations emits a close frame with an invalid status code -1 when a ws client closes without a status code
## Expected Behavior

WebsocketServerOperations should create a WebSocketCloseStatus with a status code 1005 (meaning `NO_STATUS_CODE`) when receives an empty close frame (because it's possible for a WebSocket client to close without a status code)

## Actual Behavior

WebsocketServerOperations create a WebSocketCloseStatus with a status code -1 when receives an empty close frame (because a WebSocket client can close without a status code)
And, org.springframework.web.reactive.socket.CloseStatus will thows "IllegalArgumentException: Invalid status code" because WebsocketServerOperations emit an illegal status code -1

## Steps to Reproduce

1. Setup a WebSocket server based on reactor-netty
2. Setup a WebSocket client no matter which implementations, the client connects to the server
3. The client closes the WebSocket session without a status code (meaning an empty CloseWebSocketFrame)
4. On the server side, the WebsocketServerOperations#onInboundNext will receive the empty CloseWebSocketFrame (see the following image)
image
5. And the status will be converted to -1 by io.netty.handler.codec.http.websocketx.CloseWebSocketFrame#statusCode (see the following codes) on WebsocketServerOperations#sendCloseNow
```java
	void sendCloseNow(CloseWebSocketFrame frame, ChannelFutureListener listener) {
		sendCloseNow(frame, new WebSocketCloseStatus(frame.statusCode(), frame.reasonText()), listener);
	}
```
```java
    public int statusCode() {
        ByteBuf binaryData = content();
        if (binaryData == null || binaryData.capacity() == 0) {
            return -1; // Notice here
        }

        binaryData.readerIndex(0);
        return binaryData.getShort(0);
    }
```
6. Spring-Webflux will try to create a org.springframework.web.reactive.socket.CloseStatus#CloseStatus instance based on the WebSocketCloseStatus but it passes a invalid status code (-1)
7. As a result, the throwable occurs.

## Possible Solution
Convert the -1 passed by io.netty.handler.codec.http.websocketx.CloseWebSocketFrame#statusCode to 1005 `NO_STATUS_CODE`

## Your Environment
* Reactor version(s) used: 1.0.3

violetagg:
According to the specification the endpoint may echo the received Close frame,
and Reactor Netty echos the received Close frame:

https://tools.ietf.org/html/rfc6455#section-5.5.1
```
If an endpoint receives a Close frame and did not previously send a
Close frame, the endpoint MUST send a Close frame in response.  (When
sending a Close frame in response, the endpoint typically echos the
status code it received.)
```

However for the emitted Close status (the one that is used by the application),
if a status code is not defined or empty, the Close status should be considered 1005,
so Reactor Netty will emit 1005 instead of -1 with `WebsocketInbound#receiveCloseStatus`.

https://tools.ietf.org/html/rfc6455#section-7.1.5
```
If this Close control frame contains no status code, _The WebSocket
Connection Close Code_ is considered to be 1005.
```

Fixes #1485

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
@@ -147,8 +147,14 @@ final class WebsocketClientOperations extends HttpClientOperations
 			if (log.isDebugEnabled()) {
 				log.debug(format(channel(), "CloseWebSocketFrame detected. Closing Websocket"));
 			}
-			sendCloseNow(new CloseWebSocketFrame(true, ((CloseWebSocketFrame) msg).rsv(),
-					((CloseWebSocketFrame) msg).content()));
+			CloseWebSocketFrame closeFrame = new CloseWebSocketFrame(true, ((CloseWebSocketFrame) msg).rsv(),
+					((CloseWebSocketFrame) msg).content());
+			if (closeFrame.statusCode() != -1) {
+				sendCloseNow(closeFrame);
+			}
+			else {
+				sendCloseNow(closeFrame, WebSocketCloseStatus.EMPTY);
+			}
 			onInboundComplete();
 		}
 		else if (msg != LastHttpContent.EMPTY_LAST_CONTENT) {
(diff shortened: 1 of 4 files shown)


OlegDokuka (approved):
LGTM

violetagg:
@OlegDokuka thanks for the review

Issue #360, opened 2018-05-17
DisposableServer.isDisposed() doesn't work


smaldini:
Seems like server selector channel never changes active state after close:
```
//some server
server.isDisposed();; //false
server.channel().isActive(); //true
server.channel().isOpen(); //true

//stopping
server.dispose();

//later
server.isDisposed();; //false
server.channel().isActive(); //true
server.channel().isOpen(); //false
```

smaldini:
cc @nebhale

SyedMuhammadSufyian:
Is this bug still relevant ?

violetagg:
I cannot reproduce this with the current `1.0.4-SNAPSHOT` version. Test case will be added with PR #1501 in order to ensure that `DisposableServer.isDisposed()` works as expected.

PR #1501, opened 2021-02-08
[test] Test that the server selector channel is closed when the server is disposed.

violetagg:
Related to #360


simonbasle (approved):
approved. Side note: especially when dealing with booleans and other primitive types, I recommend consistent use of `.as("description")` so that in case of assertion failure it becomes clear which assertion fails (even if eg. the stacktrace gets truncated)

violetagg:
> approved. Side note: especially when dealing with booleans and other primitive types, I recommend consistent use of `.as("description")` so that in case of assertion failure it becomes clear which assertion fails (even if eg. the stacktrace gets truncated)

I added descriptions

violetagg:

PR #1503, opened 2021-02-09
fix #1471 Test existence of an AttributeKey before creation

### LINKED ISSUE #1471: '$PERSISTENT_CHANNEL'  Exception when ReactorNetty is used in multi classloader project
Hi,

I'm currently using reactor-netty (0.9.11.RELEASE) through SpringData R2DBC (v 1.1.0-RELEASE).
I'm facing an issue in my projet with the ReactorNetty class.

Here is some Context. My application use a plugin system, each plugin has this own classloader that inherit from the "main classloader". The main classloader uses netty in order to expose HTTP Endpoint. I need to implement two different SpringData/R2DBC plugins that may run simultanously according to the application deployment.

When the two plugins run simultanously, I have the following exception:
```
Caused by: java.lang.IllegalArgumentException: '$PERSISTENT_CHANNEL' is already in use
	at io.netty.util.ConstantPool.createOrThrow(ConstantPool.java:109)
	at io.netty.util.ConstantPool.newInstance(ConstantPool.java:91)
	at io.netty.util.AttributeKey.newInstance(AttributeKey.java:55)
	at reactor.netty.ReactorNetty.<clinit>(ReactorNetty.java:829)
	... 136 common frames omitted
```
https://github.com/reactor/reactor-netty/blob/8a87ab297166d03cd1ab492842f11b421589f106/reactor-netty-core/src/main/java/reactor/netty/ReactorNetty.java#L943

This is due to the fact that the ContantPool is hold by the main classload through the static variable of AttributeKey named `pool` that is also used by the Http Connection pool.  When the first R2DBC want to initialize the AttributeKey  the ConstantPool create the instance but when the second one try to do it, the exception occurs because '$PERSISTENT_CHANNEL' has been initialized by the first plugin.

I don't know if some one already faced to this issue and if there are a clean workaround.

## Expected Behavior
Both plugin start and initialize ReactorNetty without exception.

## Actual Behavior
When second plugin start, the `java.lang.IllegalArgumentException: '$PERSISTENT_CHANNEL' is already in use` occurs 

## Possible Solution

I'm not comfortable with the ReactoNetty source code, but may be providing the ReactorNetty class as a Singleton taking as 'getInstance' parameters the AttributeKey names may solve this issue. Thanks to this, it will be possible to define custom names at plugin bootstrap and avoid the name collision.

## Your Environment

* Reactor version(s) used: 0.9.11
* Other relevant libraries versions (eg. `netty`, ...): Netty 4.1.49 / SpringData R2DBC (v 1.1.0-RELEASE).
* JVM version (`java -version`): 11.0.7

leleueri:
When ReactorNetty is used in a multi-classloaders context, PERSISTENT_CHANNEL and CONNECTION attributeKey instantitation may throw an IllegalArgumentException. This fix test the existence of the AttributeKeys before creating them.

### DIFF
--- a/src/main/java/reactor/netty/ReactorNetty.java
+++ b/src/main/java/reactor/netty/ReactorNetty.java
@@ -837,9 +837,9 @@ public final class ReactorNetty {
 	static final ConnectionObserver NOOP_LISTENER = (connection, newState) -> {};
 
 	static final Logger log                               = Loggers.getLogger(ReactorNetty.class);
-	static final AttributeKey<Boolean> PERSISTENT_CHANNEL = AttributeKey.newInstance("$PERSISTENT_CHANNEL");
+	static final AttributeKey<Boolean> PERSISTENT_CHANNEL = AttributeKey.valueOf("$PERSISTENT_CHANNEL");
 
-	static final AttributeKey<Connection> CONNECTION = AttributeKey.newInstance("$CONNECTION");
+	static final AttributeKey<Connection> CONNECTION = AttributeKey.valueOf("$CONNECTION");
 
 	static final Consumer<? super FileChannel> fileCloser = fc -> {
 		try {


src/main/java/reactor/netty/ReactorNetty.java
```diff
@@ -837,9 +837,30 @@ public synchronized Throwable fillInStackTrace() {
...
 	static final Logger log                               = Loggers.getLogger(ReactorNetty.class);
-	static final AttributeKey<Boolean> PERSISTENT_CHANNEL = AttributeKey.newInstance("$PERSISTENT_CHANNEL");
 
-	static final AttributeKey<Connection> CONNECTION = AttributeKey.newInstance("$CONNECTION");
+	static final AttributeKey<Boolean> PERSISTENT_CHANNEL = getOrCreateAttribute("$PERSISTENT_CHANNEL", Boolean.class);
+
+	static final AttributeKey<Connection> CONNECTION = getOrCreateAttribute("$CONNECTION", Connection.class);
+
+	static <T> AttributeKey<T> getOrCreateAttribute(String name, Class<T> attributeClass) {
+		if (AttributeKey.exists(name)) {
```
violetagg, line 846: I think that we can safely invoke just `AttributeKey.valueOf(name)`
https://github.com/netty/netty/blob/83895f0f7a784225333c99a194009e7f9b7c7406/common/src/main/java/io/netty/util/AttributeKey.java#L34-L40
https://github.com/netty/netty/blob/83895f0f7a784225333c99a194009e7f9b7c7406/common/src/main/java/io/netty/util/ConstantPool.java#L46-L57

Wdyt?

violetagg (changes requested):

src/main/java/reactor/netty/ReactorNetty.java
leleueri: Indeed your are right. I made the change and override my last commit.

violetagg (approved):

simonbasle (approved):
👍 But please indicate in the commit message why there is no unit/integration test, and that the fix was validated in the original multi-classloader environment.

violetagg:
> 👍 But please indicate in the commit message why there is no unit/integration test, and that the fix was validated in the original multi-classloader environment.

I updated the commit message.
Thanks for the review!

violetagg:
@leleueri Thanks for the PR!
I merged it into 0.9.x and integrated it also into 1.0.x.

PR #1518, opened 2021-02-23
Add more debug info for inbound double subscription use case

violetagg:
I addition to the error propagation add also a debug info with channel info
so that the logs can be correlated.


rstoyanchev (approved):

violetagg:
@rstoyanchev Thanks for the review

PR #1520, opened 2021-02-24
Add memory leak hint when the inbound message is provided to the receiver

violetagg:
- In case of memory leaks there will be a record whether
the buffer was provided to the receiver or not.
- Add a special logger so that when debugging memory leaks only
this logger can be enabled in order to have memory leak hints in place.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
@@ -266,11 +266,11 @@ final class FluxReceive extends Flux<Object> implements Subscription, Disposable
 				try {
 					if (logLeakDetection.isDebugEnabled()) {
 						if (v instanceof ByteBuf) {
-							((ByteBuf) v).touch(format(channel, "Bounded receiver " + a.getClass().getName() +
+							((ByteBuf) v).touch(format(channel, "Receiver " + a.getClass().getName() +
 									" will handle the message from this point"));
 						}
 						else if (v instanceof ByteBufHolder) {
-							((ByteBufHolder) v).touch(format(channel, "Bounded receiver " + a.getClass().getName() +
+							((ByteBufHolder) v).touch(format(channel, "Receiver " + a.getClass().getName() +
 									" will handle the message from this point"));
 						}
 					}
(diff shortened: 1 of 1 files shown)


reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -264,6 +264,16 @@ final void drainReceiver() {
...
 
 				try {
+					if (logLeakDetection.isDebugEnabled()) {
+						if (v instanceof ByteBuf) {
+							((ByteBuf) v).touch(format(channel, "Bounded receiver " + a.getClass().getName() +
+									" will handle the message from this point"));
+						}
+						else if (v instanceof ByteBufHolder) {
+							((ByteBufHolder) v).touch(format(channel, "Bounded receiver " + a.getClass().getName() +
+									" will handle the message from this point"));
```
rstoyanchev, line 274: "Bound receiver" perhaps you mean, or maybe just "Receiver".

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -339,14 +349,14 @@ final void onInboundNext(Object msg) {
 
 		if (receiverFastpath && receiver != null) {
 			try {
-				if (log.isDebugEnabled()) {
+				if (logLeakDetection.isDebugEnabled()) {
 					if (msg instanceof ByteBuf) {
-						((ByteBuf) msg).touch(format(channel, "Unbounded receiver, bypass inbound " +
-								"buffer queue"));
+						((ByteBuf) msg).touch(format(channel, "Unbounded receiver " + receiver.getClass().getName() +
+								" will handle the message from this point"));
```
rstoyanchev, line 355: Also here, "Unbound receiver" or just "Receiver"?

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -485,4 +494,6 @@ public String toString() {
 	};
 
 	static final Logger log = Loggers.getLogger(FluxReceive.class);
+
+	static final Logger logLeakDetection = Loggers.getLogger("reactor.netty.channel.LeakDetection");
```
rstoyanchev, line 498: Maybe this could be a log category that remains off by default when "reactor.netty" is on. In the Spring Framework we use an ["_" prefix](https://github.com/spring-projects/spring-framework/blob/31d9f31806fec88b268e1ad97cfa2db428d3f734/spring-core/src/main/java/org/springframework/core/log/LogDelegateFactory.java#L73), one possible idea.

rstoyanchev (approved):

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
violetagg: fixed

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
violetagg: fixed

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
violetagg: fixed

violetagg:
@rstoyanchev Thanks for the review

Issue #1511, opened 2021-02-16
Add support for redirecting HTTP to HTTPS


JamesChenX:
## Motivation

This should be a very common use case that a user sends an HTTP request to a Web server that uses TLS and the server better redirect HTTP to HTTPS rather than responding with an error message which is confusing for users.

## Desired solution

reactor-netty should use OptionalSslHandler instead of SslHandler so it's possible to redirect HTTP to HTTPS,
And the reference code to redirect HTTP to HTTPS is (Note that the code cannot work with SslHandler because SslHandler starts handshake processing at io.netty.handler.ssl.SslHandler#channelActive while we cannot detect whether the following request is HTTP or HTTPS):

```java
public class RedirectHttpsHandler extends ChannelInboundHandlerAdapter {

    private HttpRequestDecoder requestDecoder;
    private HttpResponseEncoder responseEncoder;
    private DelegatingChannelHandlerContext handlerContext;

    @Override
    public void channelRegistered(ChannelHandlerContext ctx) throws Exception {
        handlerContext = new DelegatingChannelHandlerContext(ctx) {
            @Override
            public ChannelHandlerContext fireChannelRead(Object msg) {
                if (msg instanceof HttpRequest) {
                    HttpMessage message = (HttpMessage) msg;
                    String address = message.headers().get(HOST);
                    if (address == null) {
                        ctx.close();
                        return ctx;
                    }
                    HttpResponse response = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.PERMANENT_REDIRECT);
                    response.headers().set(LOCATION, "https://" + address);
                    try {
                        responseEncoder.write(ctx, response, new DefaultChannelPromise(ctx.channel()) {
                        });
                    } catch (Exception e) {
                        log.error(e);
                    }
                    ctx.channel().writeAndFlush(response)
                            .addListener(ChannelFutureListener.CLOSE);
                }
                return ctx;
            }
        };
        super.channelRegistered(ctx);
    }

    @Override
    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
        boolean encrypted = SslHandler.isEncrypted((ByteBuf) msg);
        if (!encrypted) {
            if (requestDecoder == null) {
                requestDecoder = new HttpRequestDecoder();
                responseEncoder = new HttpResponseEncoder();
            }
            // Redirect to another pipeline
            requestDecoder.channelRead(handlerContext, msg);
        } else {
[...]

PR #1521, opened 2021-02-25
Add support for redirecting HTTP to HTTPS (#1511)

JamesChenX:


pivotal-issuemaster:
@JamesChenX Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1521)!

[Click here](https://cla.pivotal.io/sync/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1521) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@JamesChenX Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-netty&pullRequestId=1521)!

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
```diff
@@ -0,0 +1,130 @@
...
+
+import javax.annotation.Nullable;
+import java.net.SocketAddress;
+
+import static io.netty.handler.codec.ByteToMessageDecoder.MERGE_CUMULATOR;
+
+/**
+ * @author James Chen
+ */
+final class NonSslRedirectDetector extends ChannelInboundHandlerAdapter {
```
violetagg, line 37: Can we extend `io.netty.handler.ssl.OptionalSslHandler`?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -652,6 +652,17 @@ public final HttpServer proxyProtocol(ProxyProtocolSupportType proxyProtocolSupp
 		return dup;
 	}
 
+	/**
+	 * Specifies whether HTTP should be redirected to HTTPS if SSL is enabled.
+	 *
+	 * @param redirectHttp2Https true enables redirecting HTTP to HTTPS if SSL is enabled.
+	 */
+	public final HttpServer redirectHttp2Https(boolean redirectHttp2Https) {
```
violetagg, line 660: I'm wondering whether this has to be a configuration on `HttpServer` level or a configuration on `SslProvider` level. Wdyt?

JamesChenX: I was thinking the same thing.

**Another solution**
It also reasonable for `redirectHttp2Https` to be a parameter of `secure(SslProvider sslProvider, boolean redirectHttp2Https=false)` because redirectHttp2Https can only work when ssl is enabled.

**The reason it can also be an independent config**
I added `redirectHttp2Https` as an independent config just for fluent API especially the `sslProvider` in `secure(SslProvider sslProvider, boolean redirectHttp2Https=false)`, is usually used like `secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder))` and it's a little long and isn't that fluent to add `redirectHttp2Https` as the second parameter of `secure` method.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
JamesChenX: No. According to the implementation of OptionalSslHandler, it can only replace itself with another handler for non-ssl data or ssl data. But `sslProvider.addSslHandler(ctx.channel(), remoteAddress, sslDebug);` will add multiple handlers in `NonSslRedirectDetector#handlePotentialHeader` (and then NonSslRedirectHandler will just remove itself).

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
violetagg: > And it seems not good to add redirectHttp2Https as a config of SslProvider because SslProvider is designed at TCP level.

Yes agree

> It's also reasonable for redirectHttp2Https to be a parameter of secure(SslProvider sslProvider, boolean redirectHttp2Https=false) because redirectHttp2Https can only work when ssl is enabled.
 

```
public final HttpServer secure(Consumer<? super SslProvider.SslContextSpec> sslProviderBuilder, boolean redirectHttp)
public final HttpServer secure(SslProvider sslProvider, boolean redirectHttp)
```

Even though that it is a little longer, it is clear what's the intend.
Wdyt?

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
violetagg: Yes true.
Why don't we use the approach that `OptionalSslHandler` uses and extend `ByteToMessageDecoder`? Thus it will handle for us the cumulation work.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
```diff
@@ -0,0 +1,130 @@
...
+		pipeline.remove(this);
+		if (SslHandler.isEncrypted(buf)) {
+			// Lazy init for better performance
+			sslProvider.addSslHandler(ctx.channel(), remoteAddress, sslDebug);
+			ChannelHandlerContext context = pipeline.context(NettyPipeline.SslHandler);
+			SslHandler sslHandler = (SslHandler) context.handler();
+			sslHandler.channelRead(context, buf);
+		}
+		else {
+			redirectHandler.enableRedirecting();
```
violetagg, line 118: Instead of adding `NonSslRedirectHandler` always and then setting a flag here, is it possible to add `NonSslRedirectHandler` here, so the handler will be available in the pipeline only if it is needed.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
```diff
@@ -0,0 +1,80 @@
...
+		if (msg instanceof HttpRequest) {
+			ctx.pipeline().remove(this);
+			if (redirectHttp2Https) {
+				HttpRequest request = (HttpRequest) msg;
+				String url = getRequestedUrlInHttps(request);
+				if (url == null) {
+					ctx.close();
+				}
+				else {
+					HttpResponse response = new DefaultHttpResponse(request.protocolVersion(), HttpResponseStatus.PERMANENT_REDIRECT);
```
violetagg, line 55: What about returning here a full response? `DefaultFullHttpResponse`?
After writing the response, should we continue with the pipeline?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -819,7 +832,24 @@ public void onChannelInit(ConnectionObserver observer, Channel channel, @Nullabl
...
-				sslProvider.addSslHandler(channel, remoteAddress, SSL_DEBUG);
+				NonSslRedirectHandler nonSslRedirectHandler = null;
+				ChannelPipeline pipeline = channel.pipeline();
+				if (redirectHttp2Https) {
+					nonSslRedirectHandler = new NonSslRedirectHandler();
+					NonSslRedirectDetector nonSslRedirectDetector = new NonSslRedirectDetector(nonSslRedirectHandler,
+							sslProvider,
+							remoteAddress,
+							SSL_DEBUG);
+					if (pipeline.get(NettyPipeline.ProxyHandler) != null) {
```
violetagg, line 843: Check `NettyPipeline` javadoc. `ProxyHandler` is applied only when we are on the client, so we do not need to check for this handler on the server.

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
JamesChenX: 1. `DefaultFullHttpResponse` extends `DefaultHttpResponse` with an additional a http body, and we don't need the body currently.
2. I am sure that we need to continue with the pipeline so that the handler of HttpServer can accept the request and handle it. Otherwise, the handler isn't aware of the request.
https://github.com/reactor/reactor-netty/blob/31664c33b2e78d26b54b6be8e7979b5bec25762d/reactor-netty-http/src/test/java/reactor/netty/http/client/HttpRedirectTest.java#L528-L534

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
JamesChenX: No. If we add `NonSslRedirectHandler` here (earlier), there is no HTTP request aggregator to aggregate and so we cannot get the host from the HTTP request in NonSslRedirectHandler. And that's why I added two handlers.

BTW: NonSslRedirectDetector is added before all HTTP handlers because we need to check whether it's an SSL header or a potential HTTP request.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
JamesChenX: Thanks. I will remove the check.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
violetagg: The idea of `DefaultFullHttpResponse` is that it combines `DefaultHttpResponse` and `LastHttpContent`, otherwise you have to send two messages `DefaultHttpResponse` and `LastHttpContent` in order to consider the response complete. By default the content is `Unpooled.buffer(0)` so we are OK.

violetagg: What is the use case where the handler should know about the redirect message?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
JamesChenX: OK, it's reasonable and I will update.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
violetagg: Interesting
I was thinking about something like this

```
		else {
			pipeline.addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.NonSslRedirectHandler, new NonSslRedirectHandler());
			ctx.fireChannelRead(buf);
		}
```

Wdyt?
At that point we will have everything that is needed from HTTP protocol point of view.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
JamesChenX: Just for code clarity, it's eaiser to read `NonSslRedirectDetector` + `ChannelInboundHandlerAdapter` instead of `NonSslRedirectDetector` + `ByteToMessageDecoder` + `ChannelInboundHandlerAdapter`

e.g. The code in channelRead() of `NonSslRedirectDetector` is far easier to read than the one in ByteToMessageDecoder
https://github.com/reactor/reactor-netty/blob/31664c33b2e78d26b54b6be8e7979b5bec25762d/reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java#L57-L68

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
JamesChenX: 1. OK, I will use replace `DefaultHttpResponse` with `DefaultFullHttpResponse`.

2. Continue the pipeline for:
* Other handlers can handle it (especially for metrics-related handlers)
* It's reasonable to pass the HTTP request just because that's what `reactor.netty.http.server.HttpServer#handle` means, and developers can choose whether to just ignore it flexibly.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
violetagg: Agree, on the other hand, having `NonSslRedirectDetector` as an extension of `ByteToMessageDecoder` gives us the ability to focus only on the functionality that this handler needs to do and leave `ByteToMessageDecoder` to handle various use cases when the cumulation has to be released.
Wdyt?

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
violetagg: > 2. Continue the pipeline for:
> 
> * Other handlers can handle it (especially for metrics-related handlers)

If we add the handler lazy it will be added before `ReactiveBridge` so no handlers between these two handlers.

> * It's reasonable to pass the HTTP request just because that's what `reactor.netty.http.server.HttpServer#handle` means, and developers can choose whether to just ignore it flexibly.

I cannot understand that.
In the redirect handler, you send a complete response, if you continue with the pipeline then
- the I/O handler will need to distinguish between these two types of requests instead of focusing on serving the necessary data as there was never a redirect.
- the I/O handler cannot do anything to undo the already sent response

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
JamesChenX: OK, I will replace with `pipeline.addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.NonSslRedirectHandler, new NonSslRedirectHandler())` and it makes code cleaner.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
JamesChenX: You can try to implement the same functions by extending `ByteToMessageDecoder` locally and you may understand what I mean:

> it's far eaiser to read `NonSslRedirectDetector` + `ChannelInboundHandlerAdapter` instead of `NonSslRedirectDetector` + `ByteToMessageDecoder` + `ChannelInboundHandlerAdapter`

The code extends from `ByteToMessageDecoder` is definitely shorter, but definitely harder to debug and read (bad maintainability).

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
JamesChenX: > we add the handler lazy it will be added before ReactiveBridge so no handlers between these two handlers.

No. Even if we add the handler lazily, there are still some useful handlers between these two handlers (e.g. `reactor.netty.http.server.HttpTrafficHandler` will run before `NonSslRedirectHandler`)

> the I/O handler will need to distinguish between these two types of requests instead of focusing on serving the necessary data as there was never a redirect.

No. the I/O handler doesn't need to distinguish any type of requests and can just handle it like what they usually do. And if they try to send a response, they will just get the exception `AbortedException` and this's reasonable because the handler has redirected HTTP to HTTPS.
https://github.com/reactor/reactor-netty/blob/31664c33b2e78d26b54b6be8e7979b5bec25762d/reactor-netty-http/src/test/java/reactor/netty/http/client/HttpRedirectTest.java#L528-L534

> the I/O handler cannot do anything to undo the already sent response

It's fine for the developer's custom handler cannot undo the already sent response because that's what `redirectHttp2Https` should do (redirect HTTP to HTTPS if SSL enabled). And I don't want to make it complicated (add support for undoing).
If developers really need complicated logic, they should consider writing netty handlers for their cases.

JamesChenX:
@violetagg I can finish TODOs that I have promised, but for extensions (or more advanced features. e.g. undo redirect respond), I am afraid that I don't have so many efforts for them.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
violetagg: > No. Even if we add the handler lazily, there are still some useful handlers between these two handlers (e.g. reactor.netty.http.server.HttpTrafficHandler will run before NonSslRedirectHandler)

We have this pipeline

`HttpTrafficHandler` ...  `NonSslRedirectHandler` (this is only Inbound handler)  `ChannelOperationsHandler` (this is only Inbound handler)

When we read the incoming data the pipeline is the following

`HttpTrafficHandler` -> ... -> `NonSslRedirectHandler` (this is only Inbound handler) -> `ChannelOperationsHandler` (this is only Inbound handler)

so `HttpTrafficHandler` processes the message first and then `NonSslRedirectHandler` is invoked and the last one is `ChannelOperationsHandler`. 

When we write the outgoing data the pipeline is the following
`HttpTrafficHandler` <- ... some outbound handlers (`NonSslRedirectHandler` and `ChannelOperationsHandler` are not invoked when writing)

So there is no handler between `NonSslRedirectHandler` and `ChannelOperationsHandler` that will miss the incoming message.

violetagg: > No. the I/O handler doesn't need to distinguish any type of requests and can just handle it like what they usually do. And if they try to send a response, they will just get the exception `AbortedException` and this's reasonable because the handler has redirected HTTP to HTTPS.
> 
> https://github.com/reactor/reactor-netty/blob/31664c33b2e78d26b54b6be8e7979b5bec25762d/reactor-netty-http/src/test/java/reactor/netty/http/client/HttpRedirectTest.java#L528-L534
> 
> > the I/O handler cannot do anything to undo the already sent response
> 
> It's fine for the developer's custom handler cannot undo the already sent response because that's what `redirectHttp2Https` should do (redirect HTTP to HTTPS if SSL enabled). And I don't want to make it complicated (add support for undoing).
> If developers really need complicated logic, they should consider writing netty handlers for their cases.

I suggest as a first implementation to skip notifying the I/O handler for the redirect response and later if there is a use case to add that functionality and also think about undo.
Wdyt?

violetagg:
> @violetagg I can finish TODOs that I have promised, but for extensions (or more advanced features. e.g. undo redirect respond), I am afraid that I don't have so many efforts for them.

yep thanks.
we can always extend the functionality if there is a use case.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
violetagg: I was looking at `OptionalSslHandler`. Do we need some more functionality that what `OptionalSslHandler` does?

JamesChenX: Yes, we need more functionality than what `OptionalSslHandler` can provide.
For the obvious one, we need to call `sslProvider.addSslHandler(ctx.channel(), remoteAddress, sslDebug);` that will add multiple SSL-related handlers in `NonSslRedirectDetector`, rather than just `context.pipeline().replace(this, newSslHandlerName(), sslHandler);` in `OptionalSslHandler`

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
JamesChenX: OK, let's make things easier currently and just skip notifying the I/O handler for the redirect response.
And I will update the PR in 2 hours.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
violetagg: I'm sorry I didn't explain correctly. I was thinking for something like this:

```
    @Override
    protected void decode(ChannelHandlerContext context, ByteBuf in, List<Object> out) throws Exception {
        if (in.readableBytes() < SslUtils.SSL_RECORD_HEADER_LENGTH) {
            return;
        }
		if (SslHandler.isEncrypted(buf)) {
			// Lazy init for better performance
			sslProvider.addSslHandler(ctx.channel(), remoteAddress, sslDebug);
			ChannelHandlerContext context = pipeline.context(NettyPipeline.SslHandler);
			SslHandler sslHandler = (SslHandler) context.handler();
			sslHandler.channelRead(context, buf);
		}
		else {
			pipeline.addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.NonSslRedirectHandler, new NonSslRedirectHandler());
			ctx.fireChannelRead(buf);
		}
    }
```

Wdyt?

JamesChenX: Our new tests will fail because of `Caused by: io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1` with the code. Full code:

```
final class NonSslRedirectDetector extends ByteToMessageDecoder {

	private static final int SSL_RECORD_HEADER_LENGTH = 5;

	private final SslProvider sslProvider;
	private final SocketAddress remoteAddress;
	private final boolean sslDebug;

	public NonSslRedirectDetector(SslProvider sslProvider, @Nullable SocketAddress remoteAddress, boolean sslDebug) {
		this.sslProvider = sslProvider;
		this.remoteAddress = remoteAddress;
		this.sslDebug = sslDebug;
	}

	@Override
	protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
		if (in.readableBytes() < SSL_RECORD_HEADER_LENGTH) {
			return;
		}

		ChannelPipeline pipeline = ctx.pipeline();
		pipeline.remove(this);
		if (SslHandler.isEncrypted(in)) {
			// Lazy init for better performance
			sslProvider.addSslHandler(ctx.channel(), remoteAddress, sslDebug);
			ChannelHandlerContext context = pipeline.context(NettyPipeline.SslHandler);
			SslHandler sslHandler = (SslHandler) context.handler();
			sslHandler.channelRead(context, in);
		}
		else {
			pipeline.addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.NonSslRedirectHandler, new NonSslRedirectHandler());
			ctx.fireChannelRead(in);
		}
	}

}
```

Caused by: io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
	at io.netty.util.internal.ReferenceCountUpdater.toLiveRealRefCnt(ReferenceCountUpdater.java:74)
	at io.netty.util.internal.ReferenceCountUpdater.release(ReferenceCountUpdater.java:138)
	at io.netty.buffer.AbstractReferenceCountedByteBuf.release(AbstractReferenceCountedByteBuf.java:100)
	at io.netty.buffer.WrappedByteBuf.release(WrappedByteBuf.java:1037)
	at io.netty.buffer.SimpleLeakAwareByteBuf.release(SimpleLeakAwareByteBuf.java:102)
	at io.netty.buffer.AdvancedLeakAwareByteBuf.release(AdvancedLeakAwareByteBuf.java:941)
	at io.netty.handler.codec.ByteToMessageDecoder.handlerRemoved(ByteToMessageDecoder.java:256)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:515)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:447)

violetagg: what I have to replace in the PR in order to see this exception? just the detector?

JamesChenX: Yes. Just need to replace `NonSslRedirectDetector` with the code. And the exception should occur when running `HttpRedirectTests`

JamesChenX:
@violetagg I just updated the code that we have made a consensus, ptal.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
violetagg: Ok try with

```
	@Override
	protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
		if (in.readableBytes() < SSL_RECORD_HEADER_LENGTH) {
			return;
		}

		ChannelPipeline pipeline = ctx.pipeline();
		if (SslHandler.isEncrypted(in)) {
			sslProvider.addSslHandler(ctx.channel(), remoteAddress, sslDebug);
		}
		else {
			pipeline.addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.NonSslRedirectHandler, new NonSslRedirectHandler());
		}
		pipeline.remove(this);
	}
```

and in the SslProvider#addSslHandler

```
		ChannelPipeline pipeline = channel.pipeline();
		if (pipeline.get(NettyPipeline.ProxyHandler) != null) {
			pipeline.addAfter(NettyPipeline.ProxyHandler, NettyPipeline.SslHandler, sslHandler);
		}
+		else if (pipeline.get(NettyPipeline.NonSslRedirectDetector) != null) {
+			pipeline.addAfter(NettyPipeline.NonSslRedirectDetector, NettyPipeline.SslHandler, sslHandler);
+		}
		else {
			pipeline.addFirst(NettyPipeline.SslHandler, sslHandler);
		}
```

and in the SniProvider#addSniHandler

```
		if (pipeline.get(NettyPipeline.NonSslRedirectDetector) != null) {
			pipeline.addAfter(NettyPipeline.NonSslRedirectDetector, NettyPipeline.SslHandler, newSniHandler());
		}
		else {
			pipeline.addFirst(NettyPipeline.SslHandler, newSniHandler());
		}
```

JamesChenX: updated. ptal.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -231,6 +240,7 @@ public SslProvider sslProvider() {
 	HttpProtocol[]                                          protocols;
 	int                                                     _protocols;
 	ProxyProtocolSupportType                                proxyProtocolSupportType;
+	boolean                                                 redirectHttp2Https;
```
violetagg, line 243: Can we name redirectHttp2Https either redirectHttp or redirectHttpToHttps so that it is not misinterpreted as HTTP/2 protocol.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -823,7 +836,16 @@ public void onChannelInit(ConnectionObserver observer, Channel channel, @Nullabl
 			boolean needRead = false;
 
 			if (sslProvider != null) {
-				sslProvider.addSslHandler(channel, remoteAddress, SSL_DEBUG);
+				ChannelPipeline pipeline = channel.pipeline();
+				if (redirectHttp2Https) {
```
violetagg, line 840: Let's also check that the protocol is not HTTP/2

```suggestion
				if (redirectHttp2Https && ((protocols & h2) != h2)) {
```

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
```diff
@@ -0,0 +1,62 @@
...
+import io.netty.handler.ssl.SslHandler;
+import reactor.netty.NettyPipeline;
+import reactor.netty.tcp.SslProvider;
+
+import javax.annotation.Nullable;
+import java.net.SocketAddress;
+import java.util.List;
+
+/**
+ * @author James Chen
```
violetagg, line 31: Add `@since 1.0.5`

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -713,9 +739,33 @@ public final HttpServer secure(Consumer<? super SslProvider.SslContextSpec> sslP
...
+	 *     SslContextBuilder sslContextBuilder =
+	 *             SslContextBuilder.forServer(cert.certificate(), cert.privateKey());
+	 *     secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder));
+	 * }
+	 * </pre>
+	 *
+	 * @param sslProvider The provider to set when configuring SSL
+	 * @param redirectHttp2Https true enables redirecting HTTP to HTTPS
+	 *
+	 * @return a new {@link HttpServer}
```
violetagg, line 762: Add `@since 1.0.5`

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -686,11 +686,37 @@ public final HttpServer route(Consumer<? super HttpServerRoutes> routesBuilder)
...
+	 *     SelfSignedCertificate cert = new SelfSignedCertificate();
+	 *     SslContextBuilder sslContextBuilder =
+	 *             SslContextBuilder.forServer(cert.certificate(), cert.privateKey());
+	 *     secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder));
+	 * }
+	 * </pre>
+	 *
+	 * @param sslProviderBuilder builder callback for further customization of SslContext.
+	 * @param redirectHttp2Https true enables redirecting HTTP to HTTPS
+	 * @return a new {@link HttpServer}
```
violetagg, line 711: Add `@since 1.0.5`

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
```diff
@@ -0,0 +1,75 @@
...
+import io.netty.handler.codec.http.HttpResponse;
+import io.netty.handler.codec.http.HttpResponseStatus;
+
+import javax.annotation.Nullable;
+
+import static io.netty.handler.codec.http.HttpHeaderNames.HOST;
+import static io.netty.handler.codec.http.HttpHeaderNames.LOCATION;
+
+/**
+ * @author James Chen
```
violetagg, line 32: Add `@since 1.0.5`

Can you add 1-2 sentences that describe the handler. Add also that this handler will not propagate the incoming message when a response is completed.

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+final class NonSslRedirectHandler extends ChannelInboundHandlerAdapter {
+
+	private static final String HTTP_PROTOCOL = "http://";
+	private static final String HTTPS_PROTOCOL = "https://";
+
+	@Override
+	public void channelRead(ChannelHandlerContext ctx, Object msg) {
```
violetagg, line 40:
```
> Task :reactor-netty-http:compileJava
.../reactor-netty/reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java:46: warning: [FutureReturnValueIgnored] Return value of methods returning Future must be checked. Ignoring returned Futures suppresses exceptions thrown from the code that completes the Future.
                                ctx.close();
                                         ^
    (see https://errorprone.info/bugpattern/FutureReturnValueIgnored)
  Did you mean to remove this line?
1 warning
```

Let's fix this warning:

```suggestion
	@SuppressWarnings("FutureReturnValueIgnored")
	public void channelRead(ChannelHandlerContext ctx, Object msg) {
```

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
```diff
@@ -0,0 +1,75 @@
...
+		if (msg instanceof HttpRequest) {
+			ctx.pipeline().remove(this);
+			HttpRequest request = (HttpRequest) msg;
+			String url = getRequestedUrlInHttps(request);
+			if (url == null) {
+				ctx.close();
```
violetagg, line 46:
```suggestion
				//"FutureReturnValueIgnored" this is deliberate
				ctx.close();
```

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -686,11 +686,37 @@ public final HttpServer route(Consumer<? super HttpServerRoutes> routesBuilder)
...
+	 * {@code
+	 *     SelfSignedCertificate cert = new SelfSignedCertificate();
+	 *     SslContextBuilder sslContextBuilder =
+	 *             SslContextBuilder.forServer(cert.certificate(), cert.privateKey());
+	 *     secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder));
+	 * }
+	 * </pre>
+	 *
+	 * @param sslProviderBuilder builder callback for further customization of SslContext.
+	 * @param redirectHttp2Https true enables redirecting HTTP to HTTPS
```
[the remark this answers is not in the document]
JamesChenX: updated to `redirectHttpToHttps`

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -713,9 +739,33 @@ public final HttpServer secure(Consumer<? super SslProvider.SslContextSpec> sslP
...
+	 * {@code
+	 *     SelfSignedCertificate cert = new SelfSignedCertificate();
+	 *     SslContextBuilder sslContextBuilder =
+	 *             SslContextBuilder.forServer(cert.certificate(), cert.privateKey());
+	 *     secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder));
+	 * }
+	 * </pre>
+	 *
+	 * @param sslProvider The provider to set when configuring SSL
+	 * @param redirectHttp2Https true enables redirecting HTTP to HTTPS
```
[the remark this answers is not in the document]
JamesChenX: updated to 'redirectHttpToHttps'

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
JamesChenX: updated to 'redirectHttpToHttps'

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
JamesChenX: Done.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
JamesChenX: Done.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
JamesChenX: Done.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
JamesChenX: Done.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
JamesChenX: Done.

> The handler sends an HTTP response with a status code of 308 and a
> location header consisting of the host header of the request and
> HTTPS protocol to the incoming HTTP requests.
> <p>
> The handler is applicable only for HTTP/1.x and will not propagate
> for incoming HTTP requests

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
JamesChenX: Done.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectHandler.java
JamesChenX: Done.

violetagg (approved):

violetagg:
@OlegDokuka @rstoyanchev PTAL

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
rstoyanchev: Please, fill out the Javadoc.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -686,11 +686,39 @@ public final HttpServer route(Consumer<? super HttpServerRoutes> routesBuilder)
...
+	 *     SelfSignedCertificate cert = new SelfSignedCertificate();
+	 *     SslContextBuilder sslContextBuilder =
+	 *             SslContextBuilder.forServer(cert.certificate(), cert.privateKey());
+	 *     secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder), true);
+	 * }
+	 * </pre>
+	 *
+	 * @param sslProviderBuilder  builder callback for further customization of SslContext.
+	 * @param redirectHttpToHttps true enables redirecting HTTP to HTTPS on the same port.
+	 *                            This configuration is applicable only for HTTP/1.x.
```
rstoyanchev, line 711:
```suggestion
	 * @param redirectHttpToHttps true enables redirecting HTTP to HTTPS by changing the
	 *                            scheme only but otherwise leaving the port the same.
	 *                            This configuration is applicable only for HTTP 1.x.
```

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -713,9 +741,34 @@ public final HttpServer secure(Consumer<? super SslProvider.SslContextSpec> sslP
...
+	 *     SelfSignedCertificate cert = new SelfSignedCertificate();
+	 *     SslContextBuilder sslContextBuilder =
+	 *             SslContextBuilder.forServer(cert.certificate(), cert.privateKey());
+	 *     secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder), true);
+	 * }
+	 * </pre>
+	 *
+	 * @param sslProvider         The provider to set when configuring SSL
+	 * @param redirectHttpToHttps true enables redirecting HTTP to HTTPS on the same port.
+	 *                            This configuration is applicable only for HTTP/1.x.
```
rstoyanchev, line 763:
```suggestion
	 * @param redirectHttpToHttps true enables redirecting HTTP to HTTPS by changing the
	 *                            scheme only but otherwise leaving the port the same.
	 *                            This configuration is applicable only for HTTP 1.x.
```

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -713,9 +741,34 @@ public final HttpServer secure(Consumer<? super SslProvider.SslContextSpec> sslP
...
+	 * @return a new {@link HttpServer}
+	 * @since 1.0.5
+	 */
+	public final HttpServer secure(SslProvider sslProvider, boolean redirectHttpToHttps) {
```
rstoyanchev, line 767: What about instead of the boolean args in two places, having a single method like in `HttpServerConfig`, something like `secureRedirectHttpToHttps(boolean)`? Boolean args (and overloading) can become challenging in the future if more need to be added. I'm thinking such a method could be expected to be called after `secure` and be rejected if secure wasn't called already. Just a thought.

rstoyanchev (changes requested):
LGTM. A couple of minor comments below.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
violetagg: @rstoyanchev Till now we do not have a requirement for ordering of the configuration, because of that we handle this when the respective configuration is made.
For example when we configure the HTTP/1.1 and HTTP/2 protocols and the default SslProvider when on the client side, we need to choose the correct default SslProvider depending on the protocol:

https://github.com/reactor/reactor-netty/blob/f8c75b9fea8d378cf93f7696ead8a2a40c750f2f/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java#L1268

https://github.com/reactor/reactor-netty/blob/f8c75b9fea8d378cf93f7696ead8a2a40c750f2f/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java#L1340

So the code will become a little bit more complicated.

OlegDokuka (approved):
LGTM

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
JamesChenX: updated.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
JamesChenX: updated.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
JamesChenX: @rstoyanchev Both using an independent method `redirectHttpToHttps` or using `redirectHttpToHttps` as a param in `secure` are reasonable. And I have discussed this with @violetagg. Personally, I am okay with both for the following reasons:

* using an independent method `redirectHttpToHttps` for a more fluent API, especially `secure(SslProvider sslProvider, boolean redirectHttp2Https=false)`, is usually used like `secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder), true)` and it's a little long and isn't that fluent compared to add redirectHttp2Https as the second parameter of secure method.

* It's also reasonable for redirectHttp2Https to be a parameter of secure(SslProvider sslProvider, boolean redirectHttp2Https=false) because redirectHttp2Https can only work when ssl is enabled.

@violetagg wdyt?

violetagg: @rstoyanchev As long term solution I prefer to think about `SslProvider` configurations per different protocol.
@JamesChenX Do you need this feature for both methods `secure(SslProvider sslProvider, true)` and `secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder), true)` or we can limit the configuration to one of them? @rstoyanchev wdyt?

rstoyanchev: Feel free to decide how to best proceed. Overloading only one of the methods is an option as well.

reactor-netty-http/src/main/java/reactor/netty/http/server/NonSslRedirectDetector.java
violetagg: @JamesChenX Can you add javadoc as per @rstoyanchev review comment

JamesChenX: @violetagg updated. ptal

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
violetagg: @JamesChenX shall we go with a reduced scope with only one method or to leave it as it is?

JamesChenX: @violetagg I prefer `secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder), true)`, and it's just a choice between "Fluent API" (one for `secure` and another for `redirectHttpToHttps`) and "Code clarity" (uses one method for the these two).

The current style `secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder), true)` is less confusing because the method indicates that `redirectHttpToHttps` can only work when SSL is enabled.

violetagg: @JamesChenX Agree
Ok let's reduce the scope and expose only `secure(sslContextSpec -> sslContextSpec.sslContext(sslContextBuilder), true)`
Can you add this change to the PR and I think after that we are ready to merge the change.

JamesChenX: @violetagg I misinterpreted your words, and we can just keep the current commit with both the existing method `secure(SslProvider sslProvider)` and the new method `secure(SslProvider sslProvider, boolean redirectHttpToHttps)` without making an unnecessary breaking change.

violetagg:
@JamesChenX Thanks a lot for the PR! 👍 
@OlegDokuka @rstoyanchev Thanks for the review!

PR #1526, opened 2021-03-02
Update TcpResources/UdpResources javadoc with preferred disposal methods

violetagg:


simonbasle (commented):
I would quickly explain why the alternative methods are preferred. Also, maybe use `Prefer using` wording which is a bit less prescriptive, wdyt?

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -175,7 +175,12 @@ static ConnectionProvider create(String name, int maxConnections) {
 			@Nullable AddressResolverGroup<?> resolverGroup);
 
 
-	default void disposeWhen(SocketAddress address) {
+	/**
+	 * Dispose all connection pools for the specified remote address.
```
simonbasle, line 179: maybe mention the default implementation is NO-OP but should be implemented in user-provided implementations the same way `disposeLater` should be implemented?

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
```diff
@@ -165,25 +165,54 @@ public boolean daemon() {
 	}
 
 	/**
-	 * Use {@link #disposeLoopsAndConnections()}
+	 * {@link TcpResources} provides the global TCP resources, as such {@link TcpResources} lifecycle is
```
simonbasle, line 168: phrasing looks great 👍

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
```diff
@@ -165,25 +165,54 @@ public boolean daemon() {
...
+	 * This method has {@code NOOP} implementation.
+	 */
+	@Override
+	public Mono<Void> disposeLater(Duration quietPeriod, Duration timeout) {
+		//noop on global by default
+		return Mono.empty();
+	}
+
+	/**
+	 * Dispose all connection pools for the specified remote address.
```
simonbasle, line 209: maybe specify that unlike eg. `disposeLater`, this method _does_ delegate to the underlying pool and has a global effect (comparable to some extent to `disposeLoopsAndConnectionsLater()`).

simonbasle (changes requested):
almost there I think 👍

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
violetagg: done

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
violetagg: done

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
```diff
@@ -207,6 +207,10 @@ public void dispose() {
 
 	/**
 	 * Dispose all connection pools for the specified remote address.
+	 * <p>
+	 * As an opposite to {@link #dispose()}, this method delegates to the underlying connection provider.
+	 * It has a global effect and removes all connection pools for this remote address from the
+	 * global TCP resources similar to {@link #disposeLoopsAndConnections()}.
```
simonbasle, line 213:
```suggestion
	 * As opposed to {@link #dispose()}, this method delegates to the underlying connection provider.
	 * It has a global effect and removes all connection pools for this remote address from the
	 * global TCP resources (making it closer to {@link #disposeLoopsAndConnections()} than
	 * to {@link #dispose()}).
```

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
simonbasle: suggested a phrasing tweak

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
violetagg: committed the proposal to the change

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
```diff
@@ -165,25 +165,58 @@ public boolean daemon() {
 	}
 
 	/**
-	 * Use {@link #disposeLoopsAndConnections()}
+	 * {@link TcpResources} provides the global TCP resources, as such {@link TcpResources} lifecycle is
+	 * longer than regular {@link LoopResources} and {@link ConnectionProvider} implementations lifecycle.
+	 * In order to prevent an unintended disposal of the global TCP resources, {@link #disposeLoopsAndConnections()}
+	 * is exposed and should be used.
+	 * <p>
+	 * This method has {@code NOOP} implementation.
```
rstoyanchev, line 173: The order in which this is explained is slightly confusing to me in the sense that if the method is a no-op why would it cause any issues. Perhaps something along these lines:

"This has is a no-op by default in order to prevent unintended disposal of the global TCP resources which has a longer lifecycle than regular LoopResources and ConnectionProvider. If you do mean to shut global resources, please use `disposeLoopsAndConnections` instead."

rstoyanchev (commented):
Looks very helpful overall. Just one small comment.

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
violetagg: Please take a look now

simonbasle (approved):

violetagg:
@simonbasle @rstoyanchev thanks for the review

PR #1529, opened 2021-03-04
Ensure client transports use one and the same name resolver when no custom configuration

### LINKED ISSUE #1515: UDP File Description Leak at DNS check in ConnectionProvider.newConnection() mode
## Expected Behavior

The UDP socket should be closed after DNS check.
## Actual Behavior

The UDP socket not close after DNS check and cause `Too many open files` err. This bug might occurred in the new DNS module after version 1.0.0. I only test on the newest version 1.0.4

Keep the socket open might work on the default HttpResources but not work in the `ConnectionProvider.newConnection()` mode.
## Steps to Reproduce

```java
@Test
public void repoCase() {
 Mono.defer(() -> HttpClient.create(ConnectionProvider.newConnection())
                .host("xx")
                .port(80)
                .baseUrl("/")
                .get()
                .response())
                .repeat(Integer.MAX_VALUE)
                .blockLast();
}
```

## Possible Solution

The UDP socket should be closed.

## Your Environment

* Reactor version(s) used:1.0.4
* Other relevant libraries versions (eg. `netty`, ...):netty 4.1.59.final
 java version "1.8.0_281"
Java(TM) SE Runtime Environment (build 1.8.0_281-b09)
Java HotSpot(TM) 64-Bit Server VM (build 25.281-b09, mixed mode)
Linux ubuntu 4.4.0-154-generic #181-Ubuntu SMP Tue Jun 25 05:29:03 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
and 
Darwin A-MBP 18.7.0 Darwin Kernel Version 18.7.0: Tue Nov 10 00:07:31 PST 2020; root:xnu-4903.278.51~1/RELEASE_X86_64 x86_64

violetagg:
When there isn't a custom name resolver and the global LoopResources
are in use then all client transports will share one and the same name resolver.

Fixes #1515

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
@@ -31,6 +31,7 @@ import reactor.netty.Connection;
 import reactor.netty.ConnectionObserver;
 import reactor.netty.resources.ConnectionProvider;
 import reactor.netty.resources.LoopResources;
+import reactor.netty.transport.NameResolverProvider;
 import reactor.netty.transport.TransportConfig;
 import reactor.util.Logger;
 import reactor.util.Loggers;
@@ -141,12 +142,14 @@ public class TcpResources implements ConnectionProvider, LoopResources {
 		return getOrCreate(tcpResources, loops, null, ON_TCP_NEW, "tcp");
 	}
 
-	final ConnectionProvider defaultProvider;
-	final LoopResources      defaultLoops;
+	final LoopResources                            defaultLoops;
+	final ConnectionProvider                       defaultProvider;
+	final AtomicReference<AddressResolverGroup<?>> defaultResolver;
 
 	protected TcpResources(LoopResources defaultLoops, ConnectionProvider defaultProvider) {
 		this.defaultLoops = defaultLoops;
 		this.defaultProvider = defaultProvider;
+		this.defaultResolver = new AtomicReference<>();
 	}
 
 	@Override
@@ -266,6 +269,7 @@ public class TcpResources implements ConnectionProvider, LoopResources {
 	 */
 	protected void _dispose() {
 		defaultProvider.dispose();
+		_disposeResolver();
 		defaultLoops.dispose();
 	}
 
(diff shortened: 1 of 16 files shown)


reactor-netty-core/src/main/java/reactor/netty/udp/UdpResources.java
```diff
@@ -220,7 +225,34 @@ protected void _dispose() {
...
-		return defaultLoops.disposeLater(quietPeriod, timeout);
+		return Mono.when(_disposeResolverLater(), defaultLoops.disposeLater(quietPeriod, timeout));
+	}
+
+	AddressResolverGroup<?> getOrCreateDefaultResolver() {
+		AddressResolverGroup<?> resolverGroup = defaultResolver.get();
+		if (resolverGroup == null) {
+			AddressResolverGroup<?> newResolverGroup =
+					DEFAULT_NAME_RESOLVER_PROVIDER.newNameResolverGroup(defaultLoops, LoopResources.DEFAULT_NATIVE);
+			defaultResolver.compareAndSet(null, newResolverGroup);
```
simonbasle, line 236: if the compareAndSet fails, is there any clean up to be done on the `newResolverGroup` variable?

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
```diff
@@ -281,7 +285,37 @@ protected void _dispose() {
...
 	 */
 	protected Mono<Void> _disposeLater(Duration quietPeriod, Duration timeout) {
-		return Mono.when(defaultLoops.disposeLater(quietPeriod, timeout), defaultProvider.disposeLater());
+		return Mono.when(
+				_disposeResolverLater(),
+				defaultLoops.disposeLater(quietPeriod, timeout),
+				defaultProvider.disposeLater());
+	}
+
+	protected AddressResolverGroup<?> getOrCreateDefaultResolver() {
```
simonbasle, line 294: would probably benefit from some javadoc

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransportConfig.java
```diff
@@ -177,14 +174,15 @@ protected ClientTransportConfig(ClientTransportConfig<CONF> parent) {
...
 		this.resolver = parent.resolver;
-		this.defaultResolver = parent.defaultResolver;
 	}
 
 	@Override
 	protected Class<? extends Channel> channelType(boolean isDomainSocket) {
 		return isDomainSocket ? DomainSocketChannel.class : SocketChannel.class;
 	}
 
+	protected abstract AddressResolverGroup<?> defaultAddressResolverGroup();
```
simonbasle, line 184: would probably benefit from some javadoc, and more importantly also document the concrete cases (eg. `HttpConfig` would mention that the default comes from `HttpResources` and is thus shared between all such configs)

simonbasle (changes requested):
looks good overall. I have one question around the `getOrCreateDefaultResolver` method(s) and suggestions around documenting at least the `defaultAddressResolverGroup` methods

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java
violetagg: done

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransportConfig.java
violetagg: done

reactor-netty-core/src/main/java/reactor/netty/udp/UdpResources.java
violetagg: Will invoke `newResolverGroup.close()`

simonbasle (approved):

PR #1530, opened 2021-03-04
Add API for obtaining request ID

violetagg:
Request ID is a combination of the id of the underlying connection
and the serial number of the request received on that connection.


simonbasle (approved):
seems to touch a little bit more than the new API (ReactorNettyLoggingHandler), so I'd mention that in the commit body, but otherwise looks good.

violetagg:
> seems to touch a little bit more than the new API (ReactorNettyLoggingHandler), so I'd mention that in the commit body, but otherwise looks good.

yes true

I updated the commit. Thanks for the review.

PR #1536, opened 2021-03-08
Ensure relative redirection is resolved correctly following to the specification

### LINKED ISSUE #1533: Handle Redirect Starting with "./"
The Reactor Netty HTTP client appears to have trouble handling redirects where the Location starts with "./".

(related question on SO: https://stackoverflow.com/questions/66510670/netty-httpclient-follow-relative-redirects)

## Expected Behavior
The expected behavior should be that the redirect is handled; the "./" is construed to be relative to the current request's host, port and path.

## Actual Behavior
`java.net.UnknownHostException: failed to resolve '.' after 3 queries`

I see in the code (UriEndpointFactory) that the pattern matcher picks up "." as the hostname, which isn't right.

## Steps to Reproduce
Make a request to a URL that returns a HTTP 30x redirect with Location starting with "./".  I happened to use Spring WebClient to do this, which leverages Netty under the hood.

## Possible Solution
Seems like this condition could be handled in the `reactor.netty.http.client.UriEndpointFactory` class's `createUriEndpoint` method.

## Your Environment
* Reactor version(s) used: 3.4.3
* Other relevant libraries versions (eg. `netty`, ...): netty 1.0.4
* OS and version (eg. `uname -a`): Mac OS X 10.15.7 Catalina

violetagg:
https://tools.ietf.org/html/rfc7231#section-7.1.2
https://tools.ietf.org/html/rfc3986#section-4.2
https://tools.ietf.org/html/rfc3986#appendix-A

Fixes #1533

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
@@ -18,6 +18,8 @@ package reactor.netty.http.client;
 
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


violetagg:
@OlegDokuka @rstoyanchev PTAL

OlegDokuka (approved):
LGTM

rstoyanchev (approved):

violetagg:
@rstoyanchev @OlegDokuka Thanks for the review!

Issue #1543, opened 2021-03-10
Configuration of ciphers is ignored for TCP


Johannes-Rost:
## Expected Behavior
Configuring a list of ciphers should result in a running server with this ciphers configured for Spring Boot 2.4.x with reactor-netty 1.0.x 

## Actual Behavior
reactor.netty.tcp.SslProvider is overriding the list of ciphers with `null` since commit https://github.com/reactor/reactor-netty/commit/41e73261a5935dac3b5773ab3668d0dd99d4cf42 for `reactor.netty.tcp.SslProvider.DefaultConfigurationType#TCP`. This results in a running server which falls back to default ciphers

## Steps to Reproduce
Start a Spring Boot WebFlux Application with SSL support on localhost port 8443 and the following properties set 
```
server.ssl.enabled-protocols=TLSv1.2,TLSv1.3
server.ssl.ciphers=TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
```

Open SSL should not be able to complete a SSL handshake for `ECDHE-RSA-AES256-SHA`, but it is:
```
my-machine:johannesrost$ openssl s_client -cipher ECDHE-RSA-AES256-SHA -connect localhost:8443

[...]
SSL handshake has read 2940 bytes and written 225 bytes
---
New, TLSv1/SSLv3, Cipher is ECDHE-RSA-AES256-SHA
Server public key is 2048 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : ECDHE-RSA-AES256-SHA
[...]

```

Spring Boot 2.3.x with reactor-netty 0.9.x is working as expected (handshake is not successfull for the same configuration)

## Possible Solution
`case TCP` in `reactor.netty.tcp.SslProvider#updateDefaultConfiguration()` should copy the configured ciphers instead of just using `null` 

## Your Environment

Spring Boot 2.4.3 with reactor-netty-core 1.0.3 on JDK11

Johannes-Rost:
[...]

PR #1549, opened 2021-03-12
GH-1543 Do not override externally configured ciphers

Johannes-Rost:
A fix for GH-1543.
Side effect: If no ciphers are configured for http/2, the intersection of the netty's DEFAULT_CIPHERS and the `reactor.netty.tcp.SslProvider#HTTP2_CIPHERS` is used. This is 12 instead of 13.


Johannes-Rost:
Since the `SslContext` has already been build, a default set of ciphers was selected, if no ciphers had been configured on the `SslContextBuilder`. It would be better to use the state from `SslContextBuilder`, but this would require a change to `io.netty:netty-handler` because the current implementation does not allow to query the ciphers from the builder.

violetagg:
@Johannes-Rost Reactor Netty will expose a new API that will be used by Spring Boot

PR #1552, opened 2021-03-16
Ensure "aggregateJavadoc" task configuration is the same as the "javadoc" task configuration

violetagg:
Since 1.0.0, the javadoc produced doesn't follow the same styling as reactor-core and 0.9.x javadocs due to the way the aggregation of individual modules javadocs is configured.
Notably, the CSS is different and links to javadocs of external classes (like JDK classes or `Flux`/`Mono`) don't appear.
This change fixes that.

### DIFF
--- a/gradle/javadoc.gradle
+++ b/gradle/javadoc.gradle
@@ -17,7 +17,7 @@ if (project.name == 'reactor-netty-examples') {
 	return
 }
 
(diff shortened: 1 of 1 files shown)


simonbasle (approved):

violetagg:

PR #1562, opened 2021-03-24
Ensure uriTagValue metrics function is not shared amongst the clients

### LINKED ISSUE #1559: uriTagValue is shared across different HttpClient instances
If I configure multiple HttpClient Instances as such
```java
HttpClient.create().metrics(enable, uri -> "/books/{id}")
```
```java
HttpClient.create().metrics(enable, uri -> "/authors/{id}")
```

## Expected Behavior
They're shared (Probably the latest overrides but I can't say)

## Actual Behavior
uriTagValue function belongs to individual HttpClient instance not global.

violetagg:
Fixes #1559

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java
@@ -106,7 +106,7 @@ public final class HttpClientConfig extends ClientTransportConfig<HttpClientConf
 
 	@Override
 	public int channelHash() {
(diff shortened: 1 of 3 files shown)


simonbasle (approved):
reading the second test, I'd consider a separate improvement of renaming that `checkExpectations` method. especially given that the interesting assertion seems to be the fact that the channel id returned by each client is different.
otherwise, I think LGTM

violetagg:
> reading the second test, I'd consider a separate improvement of renaming that `checkExpectations` method. especially given that the interesting assertion seems to be the fact that the channel id returned by each client is different.
> otherwise, I think LGTM

I changed the method name.

Thanks for the review!

simonbasle (approved):

PR #1572, opened 2021-03-31
Order the routes HttpPredicates

### LINKED ISSUE #594: Router routes macthing order
# Issue : conflict between "/yes/{param}" and "/yes/value" routes

# Actual behavior

The `HttpServerRoutes.route(*)` are impacted by the order. For example :

```java
httpServerRoutes.get("/yes/{param}", handler1);
httpServerRoutes.get("/yes/value", handler2);
```

In this case the route "/yes/value" will be occulted by "/yes/{param}".

# Proposed solution

Order the routes httpPredicates to place the more specific ones before?

This will allow the `HttpServerRoutes` routes matching to be not impacted by the methods calling order (since the backed list will be automatically sorted).

Thanks for your help

ctlove0523:
Fixes #594

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
@@ -19,9 +19,7 @@ package reactor.netty.http.server;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.Iterator;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CopyOnWriteArrayList;
@@ -81,8 +79,14 @@ final class DefaultHttpServerRoutes implements HttpServerRoutes {
 			handlers.add(new HttpRouteHandler(condition,
 					handler,
 					(HttpPredicate) condition));
+			handlers.sort((o1, o2) -> {
+				HttpPredicate predicate1 = (HttpPredicate) o1.condition;
+				HttpPredicate predicate2 = (HttpPredicate) o2.condition;
+				return predicate1.compareTo(predicate2);
+			});
 		}
 		else {
+			// not HttpPredicate,hos not method to sort handlers keep original order
 			handlers.add(new HttpRouteHandler(condition, handler, null));
 		}
 		return this;
@@ -93,12 +97,11 @@ final class DefaultHttpServerRoutes implements HttpServerRoutes {
 		final Iterator<HttpRouteHandler> iterator = handlers.iterator();
 		HttpRouteHandler cursor;
 
-		List<HttpRouteHandler> matchHandlers = new ArrayList<>();
 		try {
 			while (iterator.hasNext()) {
 				cursor = iterator.next();
 				if (cursor.test(request)) {
-					matchHandlers.add(cursor);
+					cursor.apply(request, response);
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
```diff
@@ -104,7 +107,30 @@ public HttpServerRoutes route(Predicate<? super HttpServerRequest> condition,
...
+		}
+
+		for (HttpRouteHandler handler : matchHandlers) {
+			// if condition is not HttpPredicate,have no method to compare
+			if (!(handler.condition instanceof HttpPredicate)) {
+				return matchHandlers.get(0).apply(request, response);
+			}
+		}
+
+		matchHandlers.sort((o1, o2) -> {
```
simonbasle, line 126: I wonder if it would make sense to offer customization here.
A) an option for the user to set the `Comparator` to use
B) an option for the user to opt out of sorting (eg. by providing `null` as the `Comparator`)

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
```diff
@@ -104,7 +107,30 @@ public HttpServerRoutes route(Predicate<? super HttpServerRequest> condition,
...
+		if (matchHandlers.isEmpty()) {
+			return response.sendNotFound();
+		}
+
+		// only one handler match request,just apply for speed
+		if (matchHandlers.size() == 1) {
+			return matchHandlers.get(0).apply(request, response);
+		}
+
+		for (HttpRouteHandler handler : matchHandlers) {
```
simonbasle, line 119: if the `HttpRouteHandler` itself becomes `Comparable`, maybe we can sort the `this.handlers` collection once set up is done, rather than on each application of the routes?

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
simonbasle: Also maybe we could make the `HttpRouteHandler` itself `Comparable`?
it could delegate to `HttpPredicate` in case both sides have such a predicate, or globally prioritize barebone `Predicate`s over `HttpPredicate`s ?

@violetagg in which case can we have a simple `Predicate` as the condition, btw?

simonbasle (commented):
leaving some ideas out there around comparing `HttpRouteHandler` rather than just the `HttpPredicate`

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
ctlove0523: it will be better if `HttpRouteHandler` itself becomes `Comparable` like `HandlerMapping` in spring-web-reactive project.Under this condition `this handlers` just need sort when add new handler.Now,we can also do this.

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
ctlove0523: > I wonder if it would make sense to offer customization here.
> A) an option for the user to set the `Comparator` to use
> B) an option for the user to opt out of sorting (eg. by providing `null` as the `Comparator`)

I agree with you.

violetagg: @ctlove0523 Can we introduce a configuration so that by default we keep the current behaviour and then if someone wants to configure ordering then this feature to be enabled

ctlove0523: @violetagg ok i will try

ctlove0523:
@violetagg @simonbasle 
thanks your code review,i will close this PR and create a new issue to discuss how to support choose request handler.

violetagg:
This is superseded by #1580

PR #1575, opened 2021-04-01
Check for backwards compatibility against the previous release

violetagg:
The check for backwards compatibility should be against the previous release
and not GA. Some APIs are added after GA so we need to guarantee backwards
compatibility for these APIs also.


violetagg:

Issue #1579, opened 2021-04-03
support routing sorting


ctlove0523:
## Motivation

issue about this #594

In http server we can add more routes,when request come in choose a route to handle the request. In current version，default choose the first route which match the request(may be there many routes match request),look at the following example:

~~~
httpServerRoutes.get("/yes/{value}",handler1); // route1
httpServerRoutes.get("/yes/value",handler2);   // route2
~~~

make a http call like:

curl -i http://host:port/yes/value

Which route will handle this request? This determined which route first add to server.

This feature will support：

* route add order insensitive（add route code can change）
* when many route match request support route ordering and choose the best match route to handle request

## Desired solution

1、`HttpServerRoute`s add one method to configure comparator 

~~~
HttpServerRoutes comparator(Supplier<Comparator<HttpRouteHandler>> supplier);
~~~

2、make `HttpRouteHandler` become public class and add field path

3、if configure comparator `HttpRouteHandler`s will be sorted

4、if not configure comparator choose first match route in route add order to handler request

**Any suggestions are welcome**

PR #1580, opened 2021-04-03
support routing sorting

ctlove0523:
Related issues #1579 #594


reactor-netty-http/src/main/java/reactor/netty/http/server/HttpRouteHandler.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.function.Predicate;
+
+import org.reactivestreams.Publisher;
+import reactor.util.annotation.Nullable;
+
+/**
+ * @author Stephane Maldini
+ * @author chentong
+ */
+public class HttpRouteHandler
```
violetagg, line 31: Basically from `HttpRouteHandler` we need only the `path`. Can we expose an interface that will have access only to `path`?

ctlove0523: expose an interface `RouteHandlerMetadata `
~~~
public interface RouteHandlerMetadata {
	
	String getRoutePath();
}
~~~

use `RouteHandlerMetadata`  to access `HttpRouteHandler's` metada like `path`?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
```diff
@@ -262,6 +264,15 @@ default HttpServerRoutes put(String path,
 	HttpServerRoutes route(Predicate<? super HttpServerRequest> condition,
 			BiFunction<? super HttpServerRequest, ? super HttpServerResponse, ? extends Publisher<Void>> handler);
 
+	/**
+	 * A comparator used to sort HttpRouteHandlers.If this method called more times,
```
simonbasle, line 268: also mention that currently it sorts the existing routes in place upon calling this method, then sorts them again each time a new route is added.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpRouteHandler.java
simonbasle: good idea @violetagg, yeah an interface for metadata sounds good @ctlove0523. This will make it extendable. Of course, the `HttpServerRoute#comparator` API will need to be based on said interface.

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
```diff
@@ -78,11 +80,23 @@ public HttpServerRoutes route(Predicate<? super HttpServerRequest> condition,
...
+
+		if (this.comparator != null) {
+			handlers.sort(this.comparator);
+		}
+
+		return this;
+	}
+
+	@Override
+	public HttpServerRoutes comparator(Supplier<Comparator<HttpRouteHandler>> supplier) {
```
simonbasle, line 97: is there a huge benefit in requiring a `Supplier`?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
simonbasle: we should also think about how on can go back to the default if a `Comparator` has previously been set. the most straightforward way is to make the `Comparator` nullable and pass `null` to remove any custom comparator. wdyt @violetagg ?

(A null-free way would be to add "insertion index" to the routeHandler's metadata, so that one can provide a `Comparator<HttpRouteHandlerMetadata>` restoring insertion order, but that's adding a field for a corner case.)

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpRouteHandler.java
simonbasle: the interface would be a good spot to offer two basic comparators: lexical and lexical-with-template

simonbasle (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
violetagg: we can either allow `null` in the API so that this will restore the default behaviour or add API similar to what we have for the security `HttpServer#secure(...)` vs. `HttpServer#noSSL`

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpRouteHandler.java
violetagg: For this one you mean something similar to what we have with `reactor.netty.http.server.logging.AccessLogFactory` right?

simonbasle: (by `lexical-with-template` I mean one that prioritize joker characters differently, like in the original issue)

ctlove0523: @simonbasle @violetagg let ou'rnbasle @violetagg let our users to define the order of route(we need to define new api)？
If one user add routes in the below order：
Route1(order = 1)
Route2(order = DEFAULT_ORDER),
Route3(order = 4)
Route4(order =DEFAULT_ORDER）
Route5(order = 2)

when get a request use `Route1 -> Route5 -> Route3 -> Route2 -> Route4` order to match request.

violetagg: @ctlove0523 The last proposal seems similar to the use case where we have the current implementation and you order the routes by yourself while configuring them.

simonbasle: I don't think asking the users to provide an index / priority is good ux, at least not terribly better than just asking them to invoke the routes in the desired order.

We _might_ want to consider internally adding the index information (as in "in which order did the `route` method did get invoked", eg. by incrementing an `AtomicInteger` and passing that to the constructor) so that users can restore "insertion order" so to speak, through a third built-in `Comparator`.

The index information doesn't even need to be public if the `Comparator` is implemented as package-private inside `HttpRouteHandler` and exposed as a static factory method in the `RouteHandlerMetadata` interface.

wdyt @violetagg @ctlove0523 ?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
simonbasle: I personally like the second approach if it works for you (avoiding `null`-based approaches where possible)

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpRouteHandler.java
ctlove0523: @violetagg use call method order is unstable,user explicit define route order is better. I will make new PR refer to your and simonbasle‘s review.

simonbasle: (please continue in this PR if possible, to limit leaving a breadcrumb trail of discussions in several closed PRs)

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
ctlove0523: no,I have change method to `comparator(Comparator<HttpRouteHandlerMetadata> comparator)`

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
ctlove0523: add a new method `noCompare()`?

violetagg: May be `noComparator()` or `disableComparator()`

ctlove0523: change to `noComparator()` like `HttpServer#noSSL`

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2474,6 +2475,48 @@ void testConnectionClosePropagatedAsError_TE() throws Exception {
 				"POST http://%s/ HTTP/1.1\r\nHost: %s\r\nTransfer-Encoding: chunked\r\n\r\n10\r\nhello");
 	}
 
+	@Test
+	void testOrderRoutes() {
```
violetagg, line 2479: we need to test that the comparator can be overridden and also removed so:

```
@Test
void testOrderRoutes() {
	HttpServerRoutes routes =
			HttpServerRoutes.newRoutes()
					.get("/yes/{value}", (request, response) -> response.status(200).send())
					.get("/yes/value", (request, response) -> response.status(201).send())
					.comparator(comparator);

	DisposableServer server = null;

	// 1. Case one .comparator(comparator)
	try {
		server = createServer().handle(routes).bindNow();

		StepVerifier.create(createClient(server.port()).get().uri("/yes/value")
				.response().map(httpClientResponse -> httpClientResponse.status().code()))
				.expectNext(201)
				.expectComplete()
	}
	finally {
		if (server != null) {
			server.disposeNow();
		}
	}

	// 2. Case two .comparator(comparator.reversed())
	try {
		server = createServer().handle(routes.comparator(comparator.reversed())).bindNow();

		StepVerifier.create(createClient(server.port()).get().uri("/yes/value")
				.response().map(httpClientResponse -> httpClientResponse.status().code()))
				.expectNext(200)
				.expectComplete()
	}
	finally {
		if (server != null) {
			server.disposeNow();
		}
	}

	// 3. Case three .noComparator()
	// TODO
}
```

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
```diff
@@ -76,13 +80,38 @@ public HttpServerRoutes route(Predicate<? super HttpServerRequest> condition,
...
 		else {
-			handlers.add(new HttpRouteHandler(condition, handler, null));
+			HttpRouteHandler httpRouteHandler = new HttpRouteHandler(condition, handler, null, null);
+			handlers.add(httpRouteHandler);
+			initialOrderHandlers.add(httpRouteHandler);
 		}
+
+		if (this.comparator != null) {
+			handlers.sort(this.comparator);
+		}
```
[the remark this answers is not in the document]
ctlove0523: do you mean,when `uri == nll`,no need to sort handlers？

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
```diff
@@ -76,13 +80,38 @@ public HttpServerRoutes route(Predicate<? super HttpServerRequest> condition,
...
+
+		return this;
+	}
+
+	@Override
+	public HttpServerRoutes comparator(Comparator<HttpRouteHandlerMetadata> comparator) {
+		this.comparator = comparator;
```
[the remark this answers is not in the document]
ctlove0523: already add，thanks

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
ctlove0523: I add four test case：1、no comparator;2、set comparator；3、comparator can override(has one comparator)；4、clear comparator(set comparator and then clear)

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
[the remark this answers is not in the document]
violetagg: Having an indication for nullable will solve this issue, so we can keep it like this. Thanks.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
```diff
@@ -262,6 +263,24 @@ default HttpServerRoutes put(String path,
...
 			BiFunction<? super HttpServerRequest, ? super HttpServerResponse, ? extends Publisher<Void>> handler);
 
+	/**
+	 * A comparator used to sort routes.When call this method or configure a route,already configured routes will be
+	 * sorted use this comparator.If you don't want to sort routes you can use noComparator() method to recover and
+	 * keep routes in they are configured order.By default there is no sorting and the routes will be interpreted in
+	 * the way they are configured order.
+	 *
+	 * @param comparator a HttpRouteHandlerMetadata comparator.
+	 * @return this {@link HttpServerRoutes}
```
violetagg, line 273: Add `@since 1.0.7`

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
```diff
@@ -262,6 +263,24 @@ default HttpServerRoutes put(String path,
...
+	 */
+	HttpServerRoutes comparator(Comparator<HttpRouteHandlerMetadata> comparator);
+
+	/**
+	 * This method make sure routes will be interpreted in the way they are configured order.
+	 *
+	 * @return this {@link HttpServerRoutes}
```
violetagg, line 280: Add `@since 1.0.7`

violetagg (changes requested):
I think we a good to go

@simonbasle PTAL

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
```diff
@@ -264,18 +264,18 @@ HttpServerRoutes route(Predicate<? super HttpServerRequest> condition,
...
-	 * @param comparator a HttpRouteHandler comparator supplier
+	 * @param comparator a HttpRouteHandlerMetadata comparator.
 	 * @return this {@link HttpServerRoutes}
 	 */
 	HttpServerRoutes comparator(Comparator<HttpRouteHandlerMetadata> comparator);
 
 	/**
-	 * If already call comparator() method routes will be ordered use this comparator,
-	 * this method can eliminate the side effect and keep routes in added order。
+	 * This method make sure routes will be interpreted in the way they are configured order.
```
simonbasle, line 278:
```suggestion
	 * Do not sort routes but use the order in which the routes were declared.
```

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
```diff
@@ -264,18 +264,18 @@ HttpServerRoutes route(Predicate<? super HttpServerRequest> condition,
 			BiFunction<? super HttpServerRequest, ? super HttpServerResponse, ? extends Publisher<Void>> handler);
 
 	/**
-	 * A comparator used to sort HttpRouteHandlers.When call this method,added routes will be sorted use
-	 * this comparator.When add route if already set comparator,routes will be sorted. If you don't want
-	 * to sort routes you can use noCompare() method to recover and keep routes in add order.
+	 * A comparator used to sort routes.When call this method or configure a route,already configured routes will be
+	 * sorted use this comparator.If you don't want to sort routes you can use noComparator() method to recover and
+	 * keep routes in they are configured order.By default there is no sorting and the routes will be interpreted in
+	 * the way they are configured order.
```
simonbasle, line 270:
```suggestion
	 * Use the provided {@link java.util.Comparator} to sort routes, rather than using
	 * declaration order. Routes that were already declared are also impacted by this
	 * change and will be sorted according to the comparator. You can revert to using the
	 * declaration order by calling the {@link #noComparator()} method (which is the default).
```

simonbasle (changes requested):
polishing the phrasing of the javadoc + one remark on tests. otherwise lgtm

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2477,37 +2477,73 @@ void testConnectionClosePropagatedAsError_TE() throws Exception {
...
-					return response.sendNotFound();
-				}).comparator(comparator.reversed())).bindNow();
-		StepVerifier.create(createClient(server.port()).get().uri("/yes/value")
-				.response().map(httpClientResponse -> httpClientResponse.status().code()))
-				.expectNext(200)
-				.expectComplete();
-	}
-
-	private final Comparator<HttpRouteHandlerMetadata> comparator = (o1, o2) -> {
+		HttpServerRoutes serverRoutes = HttpServerRoutes.newRoutes()
```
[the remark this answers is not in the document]
ctlove0523: agree, @violetagg what's your suggestion?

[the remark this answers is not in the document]
violetagg: yep agree

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2474,6 +2475,100 @@ void testConnectionClosePropagatedAsError_TE() throws Exception {
 				"POST http://%s/ HTTP/1.1\r\nHost: %s\r\nTransfer-Encoding: chunked\r\n\r\n10\r\nhello");
 	}
 
+	@Test
+	void testMatchRouteInConfiguredOrder() {
+		HttpServerRoutes serverRoutes = HttpServerRoutes.newRoutes()
+				.get("/yes/{value}", (request, response) -> response.sendString(Mono.just("/yes/{value}")))
+				.get("/yes/value", (request, response) -> response.sendString(Mono.just("/yes/value")));
+
+		DisposableServer server = HttpServer.create().handle(serverRoutes).bindNow();
```
violetagg, line 2484: Please use `disposableServer` and remove `server.disposeNow()`, as `disposableServer` will be disposed with `reactor.netty.BaseHttpTest#disposeServer`

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2474,6 +2475,100 @@ void testConnectionClosePropagatedAsError_TE() throws Exception {
...
+		server.disposeNow();
+	}
+
+	@Test
+	void testUseComparatorOrderRoues() {
+		HttpServerRoutes serverRoutes = HttpServerRoutes.newRoutes()
+				.get("/yes/{value}", (request, response) -> response.sendString(Mono.just("/yes/{value}")))
+				.get("/yes/value", (request, response) -> response.sendString(Mono.just("/yes/value")));
+
+		DisposableServer server = HttpServer.create().handle(serverRoutes.comparator(comparator)).bindNow();
```
violetagg, line 2500: Please use `disposableServer` and remove `server.disposeNow()`, as `disposableServer` will be disposed with `reactor.netty.BaseHttpTest#disposeServer`

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2474,6 +2475,100 @@ void testConnectionClosePropagatedAsError_TE() throws Exception {
...
+		server.disposeNow();
+	}
+
+	@Test
+	void testOverrideRoueOrder() {
+		HttpServerRoutes serverRoutes = HttpServerRoutes.newRoutes()
+				.get("/yes/{value}", (request, response) -> response.sendString(Mono.just("/yes/{value}")))
+				.get("/yes/value", (request, response) -> response.sendString(Mono.just("/yes/value")));
+
+		DisposableServer server = HttpServer.create().handle(serverRoutes.comparator(comparator)).bindNow();
```
violetagg, line 2516: Please use `try/finally` when invoking `server.disposeNow()`

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2474,6 +2475,100 @@ void testConnectionClosePropagatedAsError_TE() throws Exception {
...
+		server.disposeNow();
+	}
+
+	@Test
+	void testUseRoutesConfiguredOrder() {
+		HttpServerRoutes serverRoutes = HttpServerRoutes.newRoutes()
+				.get("/yes/{value}", (request, response) -> response.sendString(Mono.just("/yes/{value}")))
+				.get("/yes/value", (request, response) -> response.sendString(Mono.just("/yes/value")));
+
+		DisposableServer server = HttpServer.create().handle(serverRoutes.comparator(comparator)).bindNow();
```
violetagg, line 2542: Please use `try/finally` when invoking `server.disposeNow()`

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2474,6 +2475,100 @@ void testConnectionClosePropagatedAsError_TE() throws Exception {
...
+		StepVerifier.create(createClient(server.port()).get().uri("/yes/value")
+				.responseSingle((response, byteBufMono) -> byteBufMono.asString()))
+				.expectNext("/yes/{value}")
+				.verifyComplete();
+
+		server.disposeNow();
+	}
+
+	@Test
+	void testUseComparatorOrderRoues() {
```
violetagg, line 2495: nitpick: `Roues` may be should be `Routes`

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2474,6 +2475,100 @@ void testConnectionClosePropagatedAsError_TE() throws Exception {
...
+		StepVerifier.create(createClient(server.port()).get().uri("/yes/value")
+				.responseSingle((response, byteBufMono) -> byteBufMono.asString()))
+				.expectNext("/yes/value")
+				.verifyComplete();
+
+		server.disposeNow();
+	}
+
+	@Test
+	void testOverrideRoueOrder() {
```
violetagg, line 2511: nitpick: `Roue` may be should be `Route`

violetagg (changes requested):

violetagg (approved):
@simonbasle PTAL

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
```diff
@@ -262,6 +263,25 @@ default HttpServerRoutes put(String path,
...
+	 * to using the declaration order by calling the {@link #noComparator()} method (which is the default).
+	 *
+	 * @param comparator a HttpRouteHandlerMetadata comparator.
+	 * @return this {@link HttpServerRoutes}
+	 * @since 1.0.7
+	 */
+	HttpServerRoutes comparator(Comparator<HttpRouteHandlerMetadata> comparator);
+
+	/**
+	 * Do not sort routes but use the order which the routes were configured.
```
simonbasle, line 278: missing `in` in `user the order IN which`. Also, we talk about the `declaration` order above, so for consistency `configured` should be `declared`.

simonbasle (changes requested):
still a javadoc issue on `noComparator()`. after that, this PR is good to go 👍

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
ctlove0523: Removes any previously applied `{@link java.util.Comparator}` customization using `{@link HttpServerRoutes#comparator(Comparator)}`, and use the order in which the routes were configured.

violetagg: @ctlove0523 Yep 👍

ctlove0523: @simonbasle PTAL

simonbasle (approved):

violetagg:

violetagg:
@ctlove0523 Thanks for the PR!

PR #1581, opened 2021-04-05
Expose a new API AccessLogArgProvider#cookies for obtaining cookies information

### LINKED ISSUE #1577: Netty HTTP_Access_Log : Add requestCookie property to AccessLogArgProvider
Hello, it would be interesting to add a way to get cookie values directly by using their names inside the AccessLogArgProvider in order to feed access_log.

Thank'you

## Motivation
By this way we can get easily cookie values in order to make our own netty access_log format.

## Desired solution
add a "requestCookie" property to the class AccessLogArgProvider

## Considered alternatives
I'am trying unfortunately to use requestHeader() to obtain "Set-Cookie" headers value.

violetagg:
Fixes #1577

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AbstractAccessLogArgProvider.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AbstractAccessLogArgProvider.java
@@ -16,12 +16,15 @@
 package reactor.netty.http.server.logging;
 
 import io.netty.handler.codec.http.HttpMethod;
+import io.netty.handler.codec.http.cookie.Cookie;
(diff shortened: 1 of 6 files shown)


simonbasle (commented):
this is opt-in so I think this is fine, but should we document that there is a risk logs now contain sensitive data (cookies) when activating this feature?

otherwise, LGTM

violetagg:
> this is opt-in so I think this is fine, but should we document that there is a risk logs now contain sensitive data (cookies) when activating this feature?

Agree
I expended the javadoc

violetagg:
@simonbasle Thanks a lot for the review

PR #1589, opened 2021-04-07
Ensure Forwarded header is parsed according to the specification

violetagg:
```
Forwarded: by=<identifier>;for=<identifier>;host=<host>;proto=<http|https>

<identifier>
an IP address (v4 or v6, optionally with a port, and ipv6 quoted and enclosed in square brackets)
```

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/AddressUtils.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/AddressUtils.java
@@ -88,30 +88,46 @@ public final class AddressUtils {
 	}
 
 	/**
-	 * Parse unresolved InetSocketAddress. Numeric IP addresses will be detected and
-	 * resolved.
+	 * Parse unresolved InetSocketAddress. Numeric IP addresses will be detected and resolved.
 	 *
 	 * @param address ip-address or hostname
 	 * @param defaultPort the default port
-	 * @return InetSocketAddress for given parameters
+	 * @return {@link InetSocketAddress} for given parameters, only numeric IP addresses will be resolved
 	 */
 	public static InetSocketAddress parseAddress(String address, int defaultPort) {
+		return parseAddress(address, defaultPort, false);
+	}
+
+	/**
+	 * Parse unresolved InetSocketAddress. Numeric IP addresses will be detected and resolved.
+	 *
+	 * @param address ip-address or hostname
+	 * @param defaultPort the default port
+	 * @param strict if true throws an exception when the address cannot be parsed
+	 * @return {@link InetSocketAddress} for given parameters, only numeric IP addresses will be resolved
+	 */
+	public static InetSocketAddress parseAddress(String address, int defaultPort, boolean strict) {
 		requireNonNull(address, "address");
+		String host = address;
+		int port = defaultPort;
 		int separatorIdx = address.lastIndexOf(':');
 		int ipV6HostSeparatorIdx = address.lastIndexOf(']');
 		if (separatorIdx > ipV6HostSeparatorIdx) {
 			if (separatorIdx == address.indexOf(':') || ipV6HostSeparatorIdx > -1) {
-				String port = address.substring(separatorIdx + 1);
-				if (port.chars().allMatch(Character::isDigit)) {
-					return AddressUtils.createUnresolved(address.substring(0, separatorIdx),
-							Integer.parseInt(port));
+				host = address.substring(0, separatorIdx);
+				String portStr = address.substring(separatorIdx + 1);
+				if (!portStr.isEmpty() && portStr.chars().allMatch(Character::isDigit)) {
+					port = Integer.parseInt(portStr);
 				}
-				else {
-					return AddressUtils.createUnresolved(address.substring(0, separatorIdx), defaultPort);
+				else if (strict) {
+					throw new IllegalArgumentException("Failed to parse a port from " + address);
 				}
 			}
+			else if (strict) {
+				throw new IllegalArgumentException("Invalid IPv4 address " + address);
+			}
 		}
-		return AddressUtils.createUnresolved(address, defaultPort);
+		return AddressUtils.createUnresolved(host, port);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


reactor-netty-core/src/main/java/reactor/netty/transport/AddressUtils.java
```diff
@@ -88,30 +88,46 @@ public static InetSocketAddress createUnresolved(String hostname, int port) {
...
 	public static InetSocketAddress parseAddress(String address, int defaultPort) {
+		return parseAddress(address, defaultPort, false);
+	}
+
+	/**
+	 * Parse unresolved InetSocketAddress. Numeric IP addresses will be detected and resolved.
+	 *
+	 * @param address ip-address or hostname
+	 * @param defaultPort the default port
+	 * @param strict if true throws an exception when the address cannot be parsed
```
simonbasle, line 106: maybe mention clearly that if `false`, an _unresolved_ `INetSocketAddress` is instead returned, and that it can include the case of the host having been parsed but not the port (replaced by `defaultPort`)

reactor-netty-core/src/main/java/reactor/netty/transport/AddressUtils.java
```diff
@@ -88,30 +88,46 @@ public static InetSocketAddress createUnresolved(String hostname, int port) {
...
 	 */
 	public static InetSocketAddress parseAddress(String address, int defaultPort) {
+		return parseAddress(address, defaultPort, false);
+	}
+
+	/**
+	 * Parse unresolved InetSocketAddress. Numeric IP addresses will be detected and resolved.
+	 *
+	 * @param address ip-address or hostname
+	 * @param defaultPort the default port
```
simonbasle, line 105: mention that `defaultPort` is used when the host is parseable but not the port

reactor-netty-core/src/test/java/reactor/netty/transport/AddressUtilsTest.java
```diff
@@ -146,6 +147,13 @@ void shouldParseAddressForIPv6WithoutBrackets() {
 		assertThat(socketAddress.getHostString()).isEqualTo("1abc:2abc:3abc:0:0:0:5abc:6abc");
 	}
 
+	@Test
+	void shouldParseAddressForIPv6WithoutBrackets_Strict() {
```
simonbasle, line 151: nit: shouldn't these tests that assert an IllegalArgumentException be prefixed "shouldNotParseXxx"?

simonbasle (changes requested):

reactor-netty-core/src/main/java/reactor/netty/transport/AddressUtils.java
violetagg: fixed

reactor-netty-core/src/main/java/reactor/netty/transport/AddressUtils.java
violetagg: fixed

reactor-netty-core/src/test/java/reactor/netty/transport/AddressUtilsTest.java
violetagg: fixed

simonbasle (approved):

reactor-netty-core/src/main/java/reactor/netty/transport/AddressUtils.java
```diff
@@ -88,30 +88,48 @@ public static InetSocketAddress createUnresolved(String hostname, int port) {
...
+	 * Parse unresolved InetSocketAddress. Numeric IP addresses will be detected and resolved.
+	 *
+	 * @param address ip-address or hostname
+	 * @param defaultPort is used when the host is parseable but not the port
+	 * @param strict if true throws an exception when the address cannot be parsed,
+	 * otherwise an unresolved {@link InetSocketAddress} is returned. It can include the case of the host
+	 * having been parsed but not the port (replaced by {@code defaultPort})
+	 * @return {@link InetSocketAddress} for given parameters, only numeric IP addresses will be resolved
+	 */
+	public static InetSocketAddress parseAddress(String address, int defaultPort, boolean strict) {
```
rstoyanchev, line 111: I think the strict flag could probably be removed, even if the non-standard x-forwarded headers aren't clearly defined, an IPv6 address should have square brackets as per https://tools.ietf.org/html/rfc3986#section-3.2.2 so it can be used to construct URLs.

violetagg: This is a public API and it is not OK to change the behaviour.
Let me change the functionality for `X-Forwarded` in the `reactor-netty-http` module

rstoyanchev: It could also be seen as tightening the validation, i.e. the behavior that should be, that is if you want to avoid adding a new API option that might not be necessary (or could be added later if it turned out to be essential).

reactor-netty-core/src/main/java/reactor/netty/transport/AddressUtils.java
```diff
@@ -88,30 +88,48 @@ public static InetSocketAddress createUnresolved(String hostname, int port) {
...
 	 */
 	public static InetSocketAddress parseAddress(String address, int defaultPort) {
+		return parseAddress(address, defaultPort, false);
+	}
+
+	/**
+	 * Parse unresolved InetSocketAddress. Numeric IP addresses will be detected and resolved.
+	 *
+	 * @param address ip-address or hostname
+	 * @param defaultPort is used when the host is parseable but not the port
```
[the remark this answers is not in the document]
violetagg: fixed

reactor-netty-core/src/main/java/reactor/netty/transport/AddressUtils.java
violetagg: I prefer to expose a new method as this method for parsing the address is not directly connected to `Forwarded`/`X-Forwarded` and somebody may rely on the behaviour.

PR #1597, opened 2021-04-09
netty-handler-proxy runtime dependency is needed for examples with proxy functionality

violetagg:


simonbasle (approved):

violetagg:

PR #1605, opened 2021-04-15
Add workaround for "ConnectTimeoutException: connection timed out" on GitHub Actions

violetagg:
Sporadic failures are observed on GitHub Actions when running on Ubuntu 20.04/18.04
Related to https://github.com/actions/virtual-environments/issues/3185

### DIFF
--- a/.github/workflows/check_transport.yml
+++ b/.github/workflows/check_transport.yml
@@ -19,6 +19,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Update /etc/hosts file
+        if: matrix.os == 'ubuntu-20.04'
+        # /etc/hosts file needs to be updated as a workaround for
+        # https://github.com/actions/virtual-environments/issues/3185
+        run: echo -e "$(ip addr show eth0 | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)\t$(hostname -f) $(hostname -s)" | sudo tee -a /etc/hosts
       - uses: gradle/wrapper-validation-action@v1
       - name: Set up JDK 1.8
         uses: actions/setup-java@v1
--- a/.github/workflows/check_transport.yml
+++ b/.github/workflows/check_transport.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Update /etc/hosts file
-        if: matrix.os == 'ubuntu-20.04'
+        if: startsWith(matrix.os, 'ubuntu-')
         # /etc/hosts file needs to be updated as a workaround for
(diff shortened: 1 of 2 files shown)


simonbasle:
@lhotari 👋 since you noticed this issue 😄
while you're here, looks like we can't comment on your issue in github/virtual-environments. the way you phrased it made us initially think that **only** 20.04 was affected, but visibly it is also affecting 18.04. It would be interesting to mention that in your issue 🙇

.github/workflows/check_transport.yml
```diff
@@ -19,6 +19,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Update /etc/hosts file
+        if: matrix.os == 'ubuntu-20.04'
```
simonbasle, line 23: since we've determined that 18.04 is also affected, might as well use `startsWith(matrix.os, 'ubuntu-') wdyt?
source for syntax: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#startswith

simonbasle (changes requested):

violetagg:
Here for completeness the same failure observed on Ubuntu 18.04
https://github.com/reactor/reactor-netty/runs/2344428745?check_suite_focus=true#step:1:7

lhotari:
> @lhotari  since you noticed this issue 
> while you're here, looks like we can't comment on your issue in github/virtual-environments. the way you phrased it made us initially think that **only** 20.04 was affected, but visibly it is also affecting 18.04. It would be interesting to mention that in your issue

Hi @simonbasle and @violetagg , nice to meet again. :) 

The problem is that I'm not able to comment on the issue or edit the description. I guess the only way to communicate with the maintainers is to create new issues.

I added more information to [the GitHub community discussion where the problem was reported](https://github.community/t/github-action-and-oserror-errno-99-cannot-assign-requested-address). I mentioned that ubuntu-18.04 is also impacted (and possibly other VM types). Feel free to add more comments to the [GitHub community discussion](https://github.community/t/github-action-and-oserror-errno-99-cannot-assign-requested-address).

lhotari:
btw. It turns out that I'm able to edit the title of the issue (https://github.com/actions/virtual-environments/issues/3185), but not the description or add comments. @simonbasle  I'll remove "ubuntu-latest" from the title to clarify that it applies to other VM types as well.

simonbasle:
> The problem is that I'm not able to comment on the issue or edit the description

Ah bummer! Thanks for trying, and for reporting that in the community discussion @lhotari 👍

.github/workflows/check_transport.yml
violetagg: done

simonbasle (approved):

PR #1619, opened 2021-05-05
Add a new API for disposing in background the inactive connection pools

### LINKED ISSUE #1556: Memory leak in PooledConnectionProvider
## Expected Behavior
[channelPools](https://github.com/reactor/reactor-netty/blob/55f0f93b49aad10390b2f0de6f86ec166f09e846/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java#L103)  map should be cleaned-up of stale targets
## Actual Behavior
I observe heap memory growth over a week in Spring Cloud Gateway (3.0.1) and reactor-netty-1.0.3. After analyzing the heap memory dump I found that [channelPools](https://github.com/reactor/reactor-netty/blob/55f0f93b49aad10390b2f0de6f86ec166f09e846/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java#L103) accumulates stale targets and keeps on growing.  At the time it contained close to 2k entries, most of them pointing to stale, non-empty SimpleDequePools. This happens because we use service discovery and our platform is rather dynamic, it scales up and down often. Services change their IPs/ports and this results in constantly growing `channelPools` map. In our case memory growth rate is around 20MB per day

Setting non-zero `evictionInterval` and `maxIdleTime` improves the situation by emptying pooled connections, but it still does not clean up `channelPools`.

## Possible Solution
Running background connection eviction by setting `evictionInterval` could also evict stale hosts from `channelPools`, if the host points to an empty pool, or it was not accessed for longer than `maxIdleTime`

## Your Environment

* Reactor-netty 1.0.3
* Spring Cloud Gateway 3.0.1

violetagg:
Fixes #1556

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
@@ -243,7 +243,11 @@ public interface ConnectionProvider extends Disposable {
 	 */
 	final class Builder extends ConnectionPoolSpec<Builder> {
 
+		static final Duration DISPOSE_INACTIVE_POOLS_IN_BACKGROUND_DISABLED = Duration.ZERO;
+
 		String name;
+		Duration disposeInterval = DISPOSE_INACTIVE_POOLS_IN_BACKGROUND_DISABLED;
+		Duration poolInactivity;
 		final Map<SocketAddress, ConnectionPoolSpec<?>> confPerRemoteHost = new HashMap<>();
 
 		/**
@@ -266,6 +270,23 @@ public interface ConnectionProvider extends Disposable {
 		public final Builder name(String name) {
 			this.name = Objects.requireNonNull(name, "name");
 			return this;
+		}
+
+		/**
+		 * Set the options to use for configuring {@link ConnectionProvider} background disposal for inactive connection pools.
+		 * When this option is enabled, the connection pools are regularly checked whether they are empty and inactive
+		 * for a specified time, thus applicable for disposal.
+		 * Default to {@link #DISPOSE_INACTIVE_POOLS_IN_BACKGROUND_DISABLED} - the background disposal is disabled.
+		 * Providing an {@code interval} of {@link Duration#ZERO zero} means the background disposal is disabled.
+		 *
+		 * @param disposeInterval specifies the interval to be used for checking the connection pool inactivity, (resolution: ns)
+		 * @param poolInactivity specifies the time the pool was empty and inactive, (resolution: seconds)
+		 * @return {@literal this}
+		 */
+		public final Builder disposeInactivePoolsInBackground(Duration disposeInterval, Duration poolInactivity) {
+			this.disposeInterval = Objects.requireNonNull(disposeInterval, "disposeInterval");
+			this.poolInactivity = Objects.requireNonNull(poolInactivity, "poolInactivity");
+			return get();
 		}
 
 		/**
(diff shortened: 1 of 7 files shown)


reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -268,6 +272,23 @@ public final Builder name(String name) {
...
 
+		/**
+		 * Set the options to use for configuring {@link ConnectionProvider} background disposal for inactive connection pools.
+		 * When this option is enabled, the connection pools are regularly checked whether they are empty and inactive
+		 * for a specified time, thus applicable for disposal.
+		 * Default to {@link #DISPOSE_INACTIVE_POOLS_IN_BACKGROUND_DISABLED} - the background disposal is disabled.
+		 * Providing an {@code interval} of {@link Duration#ZERO zero} means the background disposal is disabled.
+		 *
+		 * @param disposeInterval specifies the interval to be used for checking the connection pool inactivity, (resolution: ns)
+		 * @param poolInactivity specifies the time the pool was empty and inactive, (resolution: seconds)
```
simonbasle, line 283:
```suggestion
		 * @param poolInactivity specifies the duration after which an empty pool with
		 * no recorded interactions is considered inactive (resolution: seconds)
```

reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -220,6 +233,34 @@ protected static void logPoolState(Channel channel, InstrumentedPool<? extends C
 				t == null ? "" : t);
 	}
 
+	final void scheduleInactivePoolsDisposal() {
+		if (!disposeInterval.isZero()) {
+			Schedulers.parallel()
+			          .schedule(this::disposeInactivePoolsInBackground, disposeInterval.toNanos(), TimeUnit.NANOSECONDS);
```
simonbasle, line 239: I wonder about the ns resolution for the interval. I'm pretty sure no JDK Executor can handle scheduling that precise, but I guess for tests with VTS it could be useful...

But milliseconds would be good enough, wouldn't they? Plus since the inactivity info is in seconds, it doesn't make much sense to check at nanosecond intervals, does it? (we can lag behind a few milliseconds, if the pool has been inactive for eg. 5 seconds it won't make a huge difference).

simonbasle (changes requested):
small question on the nanoseconds resolution as well as suggestion for javadoc wording, otherwise looks good

reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
violetagg: ok I'll change it

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
violetagg: fixed

simonbasle (approved):

violetagg:

eugene-sadovsky:
Thanks a lot for resolving this

violetagg:
@eugene-sadovsky if you are able to test it, it will be great. The new release is planed for Tuesday next week.

eugene-sadovsky:
@violetagg will definitely do and report back once I gather some stats

eugene-sadovsky:
@violetagg I can confirm the pool clean-up works properly in 1.0.7 :+1:

violetagg:
@eugene-sadovsky Thanks for testing it!

PR #1620, opened 2021-05-05
Rephrase `responseTimeout` Javadoc

hisener:
The `#responseTimeout` Javadoc and Netty's [`ReadTimeoutHandler` Javadoc](https://netty.io/4.0/api/io/netty/handler/timeout/ReadTimeoutHandler.html) don't seem to match.

> Raises a `ReadTimeoutException` when no data was read within a certain period of time.

This PR tries to clarify the response timeout configuration.

Also, does it make sense to introduce `#readTimeout` and deprecate these methods as their name is misleading?

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -1310,7 +1310,7 @@ public abstract class HttpClient extends ClientTransport<HttpClient, HttpClientC
 
 	/**
 	 * Specifies the response timeout duration in milliseconds.
-	 * This is time that takes to receive a response after sending a request.
+	 * This is the timeout for the read data operation, i.e., the maximum inactive time.
 	 * If the {@code timeout} is {@code null}, any previous setting will be removed and no response timeout
 	 * will be applied.
 	 * If the {@code timeout} is less than {@code 1ms}, then {@code 1ms} will be the response timeout.
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -1309,12 +1309,15 @@ public abstract class HttpClient extends ClientTransport<HttpClient, HttpClientC
 	}
 
 	/**
-	 * Specifies the response timeout duration.
-	 * If the {@code timeout} is {@code null}, any previous setting will be removed and no response timeout
-	 * will be applied.
-	 * If the {@code timeout} is less than {@code 1ms}, then {@code 1ms} will be the response timeout.
-	 * The response timeout setting on {@link HttpClientRequest} level overrides any response timeout
-	 * setting on {@link HttpClient} level.
+	 * Specifies the maximum interval allowed between read operations (resolution: ms).
+	 * {@link io.netty.handler.timeout.ReadTimeoutHandler} is added to the channel pipeline after sending the request
+	 * and is removed when the response is fully received.
+	 * If the {@code maxReadOperationInterval} is {@code null}, any previous setting will be removed and no
+	 * {@code maxReadOperationInterval} will be applied.
+	 * If the {@code maxReadOperationInterval} is less than {@code 1ms}, then {@code 1ms} will be the
+	 * {@code maxReadOperationInterval}.
+	 * The {@code maxReadOperationInterval} setting on {@link HttpClientRequest} level overrides any
+	 * {@code maxReadOperationInterval} setting on {@link HttpClient} level.
 	 *
 	 * @param maxReadOperationInterval the maximum interval allowed between read operations (resolution: ms).
 	 * @return a new {@link HttpClient}
@@ -1322,7 +1325,7 @@ public abstract class HttpClient extends ClientTransport<HttpClient, HttpClientC
 	 * @see io.netty.handler.timeout.ReadTimeoutHandler
 	 */
 	public final HttpClient responseTimeout(Duration maxReadOperationInterval) {
-		Objects.requireNonNull(maxReadOperationInterval, "timeout");
+		Objects.requireNonNull(maxReadOperationInterval, "maxReadOperationInterval");
 		if (Objects.equals(maxReadOperationInterval, configuration().responseTimeout)) {
 			return this;
 		}
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -1309,25 +1309,25 @@ public abstract class HttpClient extends ClientTransport<HttpClient, HttpClientC
 	}
 
 	/**
-	 * Specifies the response timeout duration in milliseconds.
-	 * This is the timeout for the read data operation, i.e., the maximum inactive time.
+	 * Specifies the response timeout duration.
 	 * If the {@code timeout} is {@code null}, any previous setting will be removed and no response timeout
 	 * will be applied.
 	 * If the {@code timeout} is less than {@code 1ms}, then {@code 1ms} will be the response timeout.
 	 * The response timeout setting on {@link HttpClientRequest} level overrides any response timeout
 	 * setting on {@link HttpClient} level.
 	 *
-	 * @param timeout the response timeout duration (resolution: ms)
+	 * @param maxReadOperationInterval the maximum interval allowed between read operations (resolution: ms).
 	 * @return a new {@link HttpClient}
 	 * @since 0.9.11
+	 * @see io.netty.handler.timeout.ReadTimeoutHandler
 	 */
-	public final HttpClient responseTimeout(Duration timeout) {
-		Objects.requireNonNull(timeout, "timeout");
-		if (Objects.equals(timeout, configuration().responseTimeout)) {
+	public final HttpClient responseTimeout(Duration maxReadOperationInterval) {
+		Objects.requireNonNull(maxReadOperationInterval, "timeout");
+		if (Objects.equals(maxReadOperationInterval, configuration().responseTimeout)) {
 			return this;
 		}
 		HttpClient dup = duplicate();
-		dup.configuration().responseTimeout = timeout;
+		dup.configuration().responseTimeout = maxReadOperationInterval;
 		return dup;
 	}
 
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -1309,9 +1309,9 @@ public abstract class HttpClient extends ClientTransport<HttpClient, HttpClientC
 	}
 
 	/**
-	 * Specifies the maximum interval allowed between read operations (resolution: ms).
-	 * {@link io.netty.handler.timeout.ReadTimeoutHandler} is added to the channel pipeline after sending the request
-	 * and is removed when the response is fully received.
+	 * Specifies the maximum duration allowed between each network-level read operation while reading a given response
+	 * (resolution: ms). In other words, {@link io.netty.handler.timeout.ReadTimeoutHandler} is added to the channel
+	 * pipeline after sending the request and is removed when the response is fully received.
 	 * If the {@code maxReadOperationInterval} is {@code null}, any previous setting will be removed and no
 	 * {@code maxReadOperationInterval} will be applied.
 	 * If the {@code maxReadOperationInterval} is less than {@code 1ms}, then {@code 1ms} will be the
@@ -1319,7 +1319,8 @@ public abstract class HttpClient extends ClientTransport<HttpClient, HttpClientC
 	 * The {@code maxReadOperationInterval} setting on {@link HttpClientRequest} level overrides any
 	 * {@code maxReadOperationInterval} setting on {@link HttpClient} level.
 	 *
-	 * @param maxReadOperationInterval the maximum interval allowed between read operations (resolution: ms).
+	 * @param maxReadOperationInterval the maximum duration allowed between each network-level read operations
+	 *                                 (resolution: ms).
 	 * @return a new {@link HttpClient}
 	 * @since 0.9.11
 	 * @see io.netty.handler.timeout.ReadTimeoutHandler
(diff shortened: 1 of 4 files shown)


Stephan202 (approved):
Agree that `#readTimeout` would be a clearer name, as `#responseTimeout` hints at "time to first byte" semantics.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1310,7 +1310,7 @@ public RequestSender request(HttpMethod method) {
 
 	/**
 	 * Specifies the response timeout duration in milliseconds.
-	 * This is time that takes to receive a response after sending a request.
+	 * This is the timeout for the read data operation, i.e., the maximum inactive time.
```
violetagg, line 1313: Can you clarify `the maximum inactive time`?
The response timeout setting is applied once the request is sent and removed once the response is received. It doesn't apply to the whole request/response.
See https://github.com/reactor/reactor-netty/blob/00201ebb2e696e26b5d2f2df6289a3de86e9cff5/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java#L557

violetagg:
> Agree that `#readTimeout` would be a clearer name, as `#responseTimeout` hints at "time to first byte" semantics.

Can you clarify what you mean with `readTimeout`?
The response timeout setting is applied once the request is sent and removed once the response is received. It doesn't apply to the whole request/response.

violetagg:
https://projectreactor.io/docs/netty/release/reference/index.html#_timeout_configuration

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
hisener: Maybe I misunderstood completely. But here is my interpretation: `ReadTimeoutHandler` that is added in the line you sent is an `IdleStateHandler` and it has a scheduled task for timeouts. The one we use is `readerIdleTimeout` and the task (`ReaderIdleTimeoutTask`) is rescheduled if there is a read operation. So that's why I thought it's read timeout rather than response timeout.
 https://github.com/netty/netty/blob/7b9a97a81f10ea574a8e6f2596075615d2eeb185/handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java#L482-L513

violetagg: so what's the question?

hisener: So if my understanding is correct then it should be read timeout instead of response timeout, right? Similar to https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URLConnection.html#setReadTimeout(int).

The response timeout name sounds like the timeout before getting the full response. However, getting the full response may take longer than the given duration in the case of a large payload, for instance.

violetagg: yeah, but we can just extend the javadoc so that it is clear, it is late for this release to change the method name ...

hisener: Extending the Javadoc sounds good to me. I was just wondering if another method name would be clearer. But I don't have a better suggestion.

Spring's `HttpComponentsClientHttpRequestFactory`, which is for a specific request, also calls it `readTimeout`.
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/client/HttpComponentsClientHttpRequestFactory.html#setReadTimeout-int-

violetagg: > Spring's HttpComponentsClientHttpRequestFactory, which is for a specific request, also calls it readTimeout.
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/client/HttpComponentsClientHttpRequestFactory.html#setReadTimeout-int-

isn't this again socket read timeout?

violetagg: Just some rough idea: What do you think about `readOperationInterval ` and then we can describe what `readOperationInterval` means?

```
public final HttpClient responseTimeout(Duration readOperationInterval) {
```

Stephan202: Yes, renaming the parameter could work. W.r.t. the suggestion: how about `maxReadOperationInterval`?

violetagg: > Yes, renaming the parameter could work. W.r.t. the suggestion: how about `maxReadOperationInterval`?

sounds OK

hisener: Proposed something, PTAL.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1309,25 +1309,25 @@ public RequestSender request(HttpMethod method) {
...
+	 * @param maxReadOperationInterval the maximum interval allowed between read operations (resolution: ms).
 	 * @return a new {@link HttpClient}
 	 * @since 0.9.11
+	 * @see io.netty.handler.timeout.ReadTimeoutHandler
 	 */
-	public final HttpClient responseTimeout(Duration timeout) {
-		Objects.requireNonNull(timeout, "timeout");
-		if (Objects.equals(timeout, configuration().responseTimeout)) {
+	public final HttpClient responseTimeout(Duration maxReadOperationInterval) {
+		Objects.requireNonNull(maxReadOperationInterval, "timeout");
```
violetagg, line 1325: nitpick: 
```suggestion
		Objects.requireNonNull(maxReadOperationInterval, "maxReadOperationInterval");
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1309,25 +1309,25 @@ public RequestSender request(HttpMethod method) {
 	}
 
 	/**
-	 * Specifies the response timeout duration in milliseconds.
-	 * This is time that takes to receive a response after sending a request.
+	 * Specifies the response timeout duration.
```
violetagg, line 1312: What do you think about this?

```suggestion
	 * Specifies the maximum interval allowed between read operations (resolution: ms). {@link io.netty.handler.timeout.ReadTimeoutHandler} is added to the channel pipeline after sending the request and is removed when the response is fully received. 
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientRequest.java
```diff
@@ -75,17 +75,17 @@
 	boolean isFollowRedirect();
 
 	/**
-	 * Specifies the response timeout duration in milliseconds.
-	 * This is time that takes to receive a response after sending a request.
+	 * Specifies the response timeout duration.
```
violetagg, line 78:
```suggestion
	 * Specifies the maximum interval allowed between read operations (resolution: ms). {@link io.netty.handler.timeout.ReadTimeoutHandler} is added to the channel pipeline after sending the request and is removed when the response is fully received. 
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1309,25 +1309,25 @@ public RequestSender request(HttpMethod method) {
...
 	 * If the {@code timeout} is {@code null}, any previous setting will be removed and no response timeout
```
violetagg, line 1313:
```suggestion
	 * If the {@code maxReadOperationInterval} is {@code null}, any previous setting will be removed and no {@code maxReadOperationInterval}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1309,25 +1309,25 @@ public RequestSender request(HttpMethod method) {
...
 	 * will be applied.
 	 * If the {@code timeout} is less than {@code 1ms}, then {@code 1ms} will be the response timeout.
```
violetagg, line 1315:
```suggestion
	 * If the {@code maxReadOperationInterval} is less than {@code 1ms}, then {@code 1ms} will be the {@code maxReadOperationInterval}.
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1309,25 +1309,25 @@ public RequestSender request(HttpMethod method) {
...
 	 * The response timeout setting on {@link HttpClientRequest} level overrides any response timeout
```
violetagg, line 1316:
```suggestion
	 * The {@code maxReadOperationInterval} setting on {@link HttpClientRequest} level overrides any {@code maxReadOperationInterval}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientRequest.java
```diff
@@ -75,17 +75,17 @@
...
 	 * If the {@code timeout} is {@code null}, any previous setting will be removed and no response timeout
```
violetagg, line 79:
```suggestion
	 * If the {@code maxReadOperationInterval} is {@code null}, any previous setting will be removed and no {@code maxReadOperationInterval}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientRequest.java
```diff
@@ -75,17 +75,17 @@
...
 	 * will be applied.
 	 * If the {@code timeout} is less than {@code 1ms}, then {@code 1ms} will be the response timeout.
```
violetagg, line 81:
```suggestion
	 * If the {@code maxReadOperationInterval} is less than {@code 1ms}, then {@code 1ms} will be the {@code maxReadOperationInterval}.
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientRequest.java
```diff
@@ -75,17 +75,17 @@
...
 	 * The response timeout setting on {@link HttpClientRequest} level overrides any response timeout
```
violetagg, line 82:
```suggestion
	 * The {@code maxReadOperationInterval} setting on {@link HttpClientRequest} level overrides any {@code maxReadOperationInterval}
```

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
violetagg: I added some comment, please check them

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
hisener: LGTM.

violetagg (approved):

hisener:
> Also, does it make sense to introduce `#readTimeout` and deprecate these methods as their name is misleading?

^ Similar to this question, after these changes, does it make sense to deprecate the `responseTimeout` method and introduce `maxReadOperationInterval` instead? :)

violetagg:
> > Also, does it make sense to introduce `#readTimeout` and deprecate these methods as their name is misleading?
> 
> ^ Similar to this question, after these changes, does it make sense to deprecate the `responseTimeout` method and introduce `maxReadOperationInterval` instead (at some point)? :)

Please open a new PR and let's discuss it there.

hisener:
> > > Also, does it make sense to introduce `#readTimeout` and deprecate these methods as their name is misleading?
> > 
> > 
> > ^ Similar to this question, after these changes, does it make sense to deprecate the `responseTimeout` method and introduce `maxReadOperationInterval` instead (at some point)? :)
> 
> Please open a new PR and let's discuss it there.

Sure. Then, I take this as yes to my question. :)

violetagg:
@reactor/netty-team PTAL

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1309,25 +1309,28 @@ public RequestSender request(HttpMethod method) {
...
-	 * Specifies the response timeout duration in milliseconds.
-	 * This is time that takes to receive a response after sending a request.
-	 * If the {@code timeout} is {@code null}, any previous setting will be removed and no response timeout
-	 * will be applied.
-	 * If the {@code timeout} is less than {@code 1ms}, then {@code 1ms} will be the response timeout.
-	 * The response timeout setting on {@link HttpClientRequest} level overrides any response timeout
-	 * setting on {@link HttpClient} level.
-	 *
-	 * @param timeout the response timeout duration (resolution: ms)
+	 * Specifies the maximum interval allowed between read operations (resolution: ms).
```
simonbasle, line 1312: This is still a bit unclear to me. Especially, we should clarify what the "read operations" are.

```suggestion
	 * Specifies the maximum duration allowed between each network-level read operation
	 * while reading a given response (resolution: ms).
```

(this also occurs below, making the same suggestion there)

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientRequest.java
```diff
@@ -75,17 +75,20 @@
...
 
 	/**
-	 * Specifies the response timeout duration in milliseconds.
-	 * This is time that takes to receive a response after sending a request.
-	 * If the {@code timeout} is {@code null}, any previous setting will be removed and no response timeout
-	 * will be applied.
-	 * If the {@code timeout} is less than {@code 1ms}, then {@code 1ms} will be the response timeout.
-	 * The response timeout setting on {@link HttpClientRequest} level overrides any response timeout
-	 * setting on {@link HttpClient} level.
+	 * Specifies the maximum interval allowed between read operations (resolution: ms).
```
simonbasle, line 78:
```suggestion
	 * Specifies the maximum duration allowed between each network-level read operation
	 * while reading a given response (resolution: ms).
```

simonbasle (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
hisener: I wrapped at 120 characters as stated [here](https://github.com/reactor/reactor-netty/blob/main/CONTRIBUTING.md). Let me know if that's changed.

simonbasle (approved):

violetagg:
@hisener @Stephan202 Thanks for the discussion and the PR!

violetagg:

PR #1623, opened 2021-05-06
Print in the logs SSLEngine instance in use

violetagg:
Print in the logs SSLEngine instance in use and not only the name of the class.


simonbasle (approved):

violetagg:

PR #1638, opened 2021-05-20
Add HttpDecoderSpec#allowDuplicateContentLengths(boolean)

### LINKED ISSUE #1636: Allow HttpClient to configure HttpResponseDecoder.allowDuplicateContentLengths
It would be nice to configure the field `allowDuplicateContentLengths` directly from the HttpClient builder. For example:

```
HttpClient
            .create()
            .httpResponseDecoder (options -> options
                .allowDuplicateContentLengths(true)
            )
```

## Motivation
Every other option of [HttpResponseDecoder](https://github.com/netty/netty/blob/d34212439068091bcec29a8fad4df82f0a82c638/codec-http/src/main/java/io/netty/handler/codec/http/HttpResponseDecoder.java) is configurable from that API

## Desired solution
Add a config method in [HttpDecoderSpec](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/HttpDecoderSpec.java)

## Additional context
Option introduced in July 2020 https://github.com/netty/netty/pull/10349

violetagg:
API for configuring whether or not to allow duplicate `Content-Length` headers with the same value.
By default this is not allowed and such messages are rejected.
When this is configured to allow duplicate `Content-Length` headers, they have to have the same value,
otherwise such messages are rejected. The duplicated `Content-Length` headers are replaced
with a single valid `Content-Length` header.

Fixes #1636

### DIFF
--- a/reactor-netty-http/src/test/java/reactor/netty/http/HttpDecoderSpecTest.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/HttpDecoderSpecTest.java
@@ -32,16 +32,17 @@ class HttpDecoderSpecTest {
 
 	@Test
 	void maxInitialLineLength() {
-		assertThat(conf.maxInitialLineLength()).as("default initial line length").isEqualTo(HttpDecoderSpec.DEFAULT_MAX_INITIAL_LINE_LENGTH);
+		checkDefaultMaxInitialLineLength(conf);
 
 		conf.maxInitialLineLength(123);
 
 		assertThat(conf.maxInitialLineLength()).as("initial line length").isEqualTo(123);
 
-		assertThat(conf.maxHeaderSize()).as("default header size").isEqualTo(HttpDecoderSpec.DEFAULT_MAX_HEADER_SIZE);
-		assertThat(conf.maxChunkSize()).as("default chunk size").isEqualTo(HttpDecoderSpec.DEFAULT_MAX_CHUNK_SIZE);
-		assertThat(conf.validateHeaders()).as("default validate headers").isEqualTo(HttpDecoderSpec.DEFAULT_VALIDATE_HEADERS);
-		assertThat(conf.initialBufferSize()).as("default initial buffer sizes").isEqualTo(HttpDecoderSpec.DEFAULT_INITIAL_BUFFER_SIZE);
+		checkDefaultMaxHeaderSize(conf);
+		checkDefaultMaxChunkSize(conf);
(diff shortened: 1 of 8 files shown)


reactor-netty-http/src/test/java/reactor/netty/http/HttpDecoderSpecTest.java
```diff
@@ -142,6 +152,20 @@ void initialBufferSizeBadValues() {
 				.withMessage("initialBufferSize must be strictly positive");
 	}
 
+	@Test
+	void allowDuplicateContentLengths() {
+		assertThat(conf.allowDuplicateContentLengths()).as("default allow duplicate Content-Length headers").isFalse();
```
simonbasle, line 157: same as above, check vs the constant introduced by this PR

reactor-netty-http/src/test/java/reactor/netty/http/HttpDecoderSpecTest.java
```diff
@@ -32,6 +32,8 @@ void init() {
 
 	@Test
 	void maxInitialLineLength() {
+		assertThat(conf.maxInitialLineLength()).as("default initial line length").isEqualTo(HttpDecoderSpec.DEFAULT_MAX_INITIAL_LINE_LENGTH);
```
simonbasle, line 35: NB: if we want to assert the _content_ of the constants, we can chain a `.isEqualTo(literal)` (or `isTrue()`/`isFalse()` for booleans). that way we assert the default is the constant AND a change to the codebase that modifies the constant's value would get flagged by unit test

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -842,13 +842,15 @@ void httpServerRequestConfigInjectAttributes() {
 		AtomicReference<Channel> channelRef = new AtomicReference<>();
 		AtomicReference<Boolean> validate = new AtomicReference<>();
 		AtomicReference<Integer> chunkSize = new AtomicReference<>();
+		AtomicReference<Boolean> allowDuplicateContentLengths = new AtomicReference<>();
```
simonbasle, line 845: same remark as in the HttpClientTest above

violetagg:
@simonbasle Please check the new commit

simonbasle (approved):

violetagg:

PR #1640, opened 2021-05-25
Add system property for disabling strict Forwarded header validation

### LINKED ISSUE #1637: IPV6 Spring Boot 2.4.5 (gateway) and reactor netty 1.0.6 stopped working
After upgrade to Spring Boot 2.4.5 with netty reactor 1.0.6 we are not able to use IPV6 anymore. The problem is that strict decoding were added that throws an exception (#1589).

Right now we had to disable Forwarded filter using:
spring.cloud.gateway.forwarded.enabled=false

## Expected Behavior
DefaultHttpForwardedHeaderHandler should fallback to parseXForwardedInfo in case strict decoding fails. Otherwise server can stop working just because of incorrect implementation of any gateway vendor in IPV6 environment. As fixing these gateways gets time I do not think strict decoding of forwarded information is a good idea.

## Actual Behavior
```
java.lang.IllegalArgumentException: Invalid IPv4 address 0:0:0:0:0:0:0:1:50532
    at reactor.netty.transport.AddressUtils.parseAddress(AddressUtils.java:132)
    at reactor.netty.http.server.DefaultHttpForwardedHeaderHandler.parseForwardedInfo(DefaultHttpForwardedHeaderHandler.java:66)
    at reactor.netty.http.server.DefaultHttpForwardedHeaderHandler.apply(DefaultHttpForwardedHeaderHandler.java:47)
```

Root cause is that Spring org.springframework.cloud.gateway.filter.headers.ForwardedHeadersFilter is not encoding "Forwarded" according to standard, but there can be many more gateways.

## Steps to Reproduce
Use spring cloud gateway and spring boot 2.4.5 + netty reactor 1.0.6. Then use curl using IPV6 address to spring cloud gateway (or any other broken gateway) that points final application.

## Possible Solution
```java
    @Override
    public ConnectionInfo apply(ConnectionInfo connectionInfo, HttpRequest request) {
        String forwardedHeader = request.headers().get(FORWARDED_HEADER);
        if (forwardedHeader != null) {
            try {
                return parseForwardedInfo(connectionInfo, forwardedHeader);
            } catch (IllegalArgumentException e) {
                // Fallback to xforwarded info
            }
        }
        return parseXForwardedInfo(connectionInfo, request);
    }
```

Or make strict configurable and not hardcoded as "true".

## Your Environment
* Spring Boot 2.4.5
* Spring Cloud Gateway 3.0.2
* Reactor netty 1.0.6

violetagg:
Fixes #1637

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpForwardedHeaderHandler.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpForwardedHeaderHandler.java
@@ -40,6 +40,17 @@ final class DefaultHttpForwardedHeaderHandler implements BiFunction<ConnectionIn
 	static final Pattern FORWARDED_PROTO_PATTERN  = Pattern.compile("proto=\"?([^;,\"]+)\"?");
 	static final Pattern FORWARDED_FOR_PATTERN    = Pattern.compile("for=\"?([^;,\"]+)\"?");
 
+	/**
+	 * Specifies whether the Http Server applies a strict {@code Forwarded} header validation.
+	 * By default it is enabled and strict validation is used.
+	 * @since 1.0.8
+	 * @deprecated The system property is used for backwards compatibility and will be removed in version 1.2.0.
+	 */
+	@Deprecated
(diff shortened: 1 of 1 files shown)


simonbasle (approved):

violetagg:

PR #1642, opened 2021-05-28
Access log must contain only the remote IP without a port

violetagg:
- Common log format must contain the IP address of the client (remote host)
which made the request to the server.
- Response duration must be without "ms".


simonbasle:
for reference: https://en.wikipedia.org/wiki/Common_Log_Format

simonbasle (approved):

violetagg:

Issue #1641, opened 2021-05-28
Cookie decoding for multiple cookies with same name has issues


betterclever:
## Expected Behavior

`Cookies.getCachedCookies()` should return multiple cookies with same name if present in header

## Actual Behavior

`Cookies.getCachedCookies()` returns only the first parsed cookie with a particular name

## Steps to Reproduce

Send a request parsed by rector-netty with a cookie header like
`test-cookie=ABC;test-cookie=DEF`

the getCookies() returns only the first cookie, the second is ignored

## Possible Solution

The bug is in line: https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/Cookies.java#L117

this should be replaced with call to: 
`decode = ((ServerCookieDecoder) decoder).decodeAll(aCookieHeader);`

## Your Environment

* Reactor version(s) used:
* Other relevant libraries versions (eg. `netty`, ...): 0.9.3 but the bug is present even in 1.0.7
* JVM version (`java -version`): 11
* OS and version (eg. `uname -a`): Linux

betterclever:
I can create a PR for the above proposed fix.

violetagg:
@betterclever yep go for it. If you need this for 0.9.x the target branch for the PR should be 0.9.x.
Thanks!

PR #1645, opened 2021-05-28
Fix multiple cookies with same name being ignored

betterclever:
Fixes: #1641


violetagg:
@betterclever This is superseded by #1662

PR #1646, opened 2021-05-31
Add API for ContextView aware metrics recorder

### LINKED ISSUE #1628: Add Per-Channel MetricsRecorder builder with a ContextView or similar
Currently Metric Recorders are client/server wide which might not be enough in some metrics captures scenarios.

We would like to capture informations in clients from the context for instance (to add extra micrometer tags). I think this could take the shape of of a client/server builder `metric(boolean enable, Function<Context, ChannelMetricsRecorder>)`.
It could use the `connectionObserver.currentContext()` as an argument.

violetagg:
New abstract classes are exposed so that the custom metrics recorder
can access the ContextView associated with the Mono/Flux
- ContextAwareChannelMetricsRecorder
- ContextAwareHttpClientMetricsRecorder
- ContextAwareHttpServerMetricsRecorder

Fixes #1628

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2011-Present VMware, Inc. or its affiliates, All Rights Reserved.
+ *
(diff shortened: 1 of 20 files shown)


simonbasle (approved):
I would add a bit more documentation, but otherwise looks good

simonbasle (approved):

ctlove0523 (approved):

PR #1659, opened 2021-06-07
Ensure custom AddressResolverGroup is preserved when changing LoopResources and there is no NameResolverProvider

### LINKED ISSUE #1655: HttpClient resolver gets changed after configuring an event loop with `runOn`
When configuring a custom event loop for the http client, it overrides a previously configured resolver, see `reactor.netty.transport.ClientTransport#runOn`

## Expected Behavior
Keep the configured resolver when a custom event loop is used. Additionally, I would like to avoid instantiating a default resolver that's outside of the client control.

## Actual Behavior
The resolver gets replaced by a default implementation.

## Steps to Reproduce
```java
@Test
public void repoCase() {
        final EventLoopGroup eventLoopGroup = new NioEventLoopGroup(1);
       HttpClient.create(connectionProvider)
                 .resolver(DefaultAddressResolverGroup.INSTANCE)
                 .runOn(eventLoopGroup); --> This creates a new resolver
}
```

## Possible Solution
`reactor.netty.transport.ClientTransport#runOn`
```java
	@Override
	public T runOn(LoopResources loopResources, boolean preferNative) {
		T dup = super.runOn(loopResources, preferNative);
		CONF conf = dup.configuration();
                 if (conf.resolver != null) {
                     return dup;
                 }
		if (conf.nameResolverProvider != null) {
			conf.resolver = conf.nameResolverProvider.newNameResolverGroup(conf.loopResources(), conf.preferNative);
		}
		else {
			conf.resolver = ClientTransportConfig.DEFAULT_NAME_RESOLVER_PROVIDER
					.newNameResolverGroup(conf.loopResources(), conf.preferNative);
		}
		return dup;
	}

```

## Your Environment

* Reactor version(s) used: `1.0.6`

violetagg:
Fixes #1655

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
@@ -307,7 +307,8 @@ public abstract class ClientTransport<T extends ClientTransport<T, CONF>,
 		if (conf.nameResolverProvider != null) {
 			conf.resolver = conf.nameResolverProvider.newNameResolverGroup(conf.loopResources(), conf.preferNative);
 		}
-		else {
+		else if (conf.resolver == null) {
+			conf.nameResolverProvider = ClientTransportConfig.DEFAULT_NAME_RESOLVER_PROVIDER;
 			conf.resolver = ClientTransportConfig.DEFAULT_NAME_RESOLVER_PROVIDER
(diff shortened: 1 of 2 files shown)


reactorbot (approved):

simonbasle (approved):

violetagg:

PR #1662, opened 2021-06-09
Expose an API for obtaining all cookies with one and the same name

### LINKED ISSUE #1641: Cookie decoding for multiple cookies with same name has issues

violetagg:
https://datatracker.ietf.org/doc/html/rfc6265#section-4.2.2

Although cookies are serialized linearly in the `Cookie` header,
servers SHOULD NOT rely upon the serialization order.  In particular,
if the `Cookie` header contains two cookies with the same name (e.g.,
that were set with different `Path` or `Domain` attributes), servers
SHOULD NOT rely upon the order in which these cookies appear in the
header.

Fixes #1641

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/Cookies.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/Cookies.java
@@ -34,9 +34,10 @@ import io.netty.handler.codec.http.cookie.ServerCookieDecoder;
 
 /**
  * Store cookies for the http channel.
+ *
  * @since 0.6
  */
-public final class Cookies {
+public class Cookies {
 
 	/**
 	 * Return a new cookies holder from client response headers.
@@ -53,7 +54,11 @@ public final class Cookies {
 	 *
 	 * @param headers server request headers
 	 * @return a new cookies holder from server request headers
+	 * @deprecated as of 1.0.8.
+	 * Prefer {@link reactor.netty.http.server.ServerCookies#newServerRequestHolder(HttpHeaders, ServerCookieDecoder)}.
+	 * This method will be removed in version 1.2.0.
 	 */
+	@Deprecated
 	public static Cookies newServerRequestHolder(HttpHeaders headers, ServerCookieDecoder decoder) {
 		return new Cookies(headers, HttpHeaderNames.COOKIE, false, decoder);
 	}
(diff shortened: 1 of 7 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/Cookies.java
```diff
@@ -104,27 +111,35 @@ private Cookies(HttpHeaders nettyHeaders, CharSequence cookiesHeaderName, boolea
...
+		markReadCookies();
 		return cachedCookies;
 	}
+
+	protected List<String> allCookieHeaders() {
+		return nettyHeaders.getAll(cookiesHeaderName);
+	}
+
+	protected final boolean hasReadCookies() {
+		return state == READ;
```
simonbasle, line 135: the atomic field updater can only guarantee atomicity and memory-ordering of the write operations if all write operations are performed through it rather than the volatile variable.

for that reason, this should be `STATE.set(this, READ)`.

(reading from the volatile is always ok)

simonbasle (changes requested):
small improvement to existing code requested (see comment, sorry I posted it outside the review)

reactor-netty-http/src/main/java/reactor/netty/http/Cookies.java
violetagg: but here `hasReadCookies` we want just to check the state we do not want to set it or I miss something?

simonbasle: oh gods, that's right 🤦 I was looking on the left hand side seeing `state = READ` and here `state == READ` and my brain jumped to conclusions 😆

all good, nothing to see here, move along

simonbasle (approved):

reactor-netty-http/src/main/java/reactor/netty/http/Cookies.java
violetagg: thanks a lot

Issue #1668, opened 2021-06-15
epoll/kqueue transport bind exception detection does not work on linux with LANGUAGE != en


roggenbrot:
When trying to start a server with a port that is already bind on a epoll/kqueue transport based linux system with `LANGUAGE` not set to `en` a `io.netty.channel.unix.Errors$NativeIoException` instead of a `ChannelBindException` is thrown.

This leads to failed test cases when trying to build `reactor-netty`

```
UdpServerTests > portBindingException() FAILED
    reactor.core.Exceptions$ReactiveException: io.netty.channel.unix.Errors$NativeIoException: bind(..) failed: Die Adresse wird bereits verwendet
        at reactor.core.Exceptions.propagate(Exceptions.java:392)
        at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:137)
        at reactor.core.publisher.Mono.block(Mono.java:1727)
        at reactor.netty.udp.UdpServer.bindNow(UdpServer.java:120)
        at reactor.netty.udp.UdpServerTests.portBindingException(UdpServerTests.java:266)

        Caused by:
        io.netty.channel.unix.Errors$NativeIoException: bind(..) failed: Die Adresse wird bereits verwendet

264 tests completed, 4 failed, 3 skipped

```

## Expected Behavior
`ChannelBindException` for e.g. test case [UdpServerTests.portBindingException](https://github.com/reactor/reactor-netty/blob/4671394337279f529a0cb0a0a7bf0b0589f93999/reactor-netty-core/src/test/java/reactor/netty/udp/UdpServerTests.java#L256-L275)

[](url)

## Actual Behavior

`io.netty.channel.unix.Errors$NativeIoException` is handled by `reactor-netty` and wrapped in `ChannelBindException`

## Steps to Reproduce

Execute [UdpServerTests.portBindingException](https://github.com/reactor/reactor-netty/blob/4671394337279f529a0cb0a0a7bf0b0589f93999/reactor-netty-core/src/test/java/reactor/netty/udp/UdpServerTests.java#L256-L275) on an epoll/kqueue based linux

## Possible Solution

Root problem seems to be the [current solution](https://github.com/reactor/reactor-netty/blob/4671394337279f529a0cb0a0a7bf0b0589f93999/reactor-netty-core/src/main/java/reactor/netty/resources/NewConnectionProvider.java#L130-L141) which tries to detect and wrap such exception based on the message string provided by the underlying OS.

[...]

PR #1670, opened 2021-06-15
Issue-1668: ChannelBindException detection fails

roggenbrot:
Original issue: #1668

This PR fixes the ChannelBindException detection by using a non i18n part of the exception message

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/NewConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/NewConnectionProvider.java
@@ -132,7 +132,7 @@ final class NewConnectionProvider implements ConnectionProvider {
 					// With epoll/kqueue transport it is
 					// io.netty.channel.unix.Errors$NativeIoException: bind(..) failed: Address already in use
 					(t instanceof IOException && t.getMessage() != null &&
-							t.getMessage().contains("Address already in use")))) {
+							t.getMessage().contains("bind(..)")))) {
 				sink.error(ChannelBindException.fail(bindAddress.get(), null));
 			}
 			else {
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
@@ -57,7 +57,7 @@ final class DefaultHttpServerRoutes implements HttpServerRoutes {
 			                   .getPath()
 			                   .replaceFirst(uri, "");
 
-			if (prefix.charAt(0) == '/') {
+			if (!prefix.isEmpty() && prefix.charAt(0) == '/') {
 				prefix = prefix.substring(1);
 			}
 
--- a/reactor-netty-http/src/test/java/reactor/netty/http/HttpCookieHandlingTests.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/HttpCookieHandlingTests.java
@@ -134,6 +134,7 @@ class HttpCookieHandlingTests extends BaseHttpTest {
 		doTestServerCookiesDecodingMultipleCookiesSameName(HttpServerInfos::allCookies, " value1 value2");
 	}
 
+	@SuppressWarnings("CollectionUndefinedEquality")
 	private void doTestServerCookiesDecodingMultipleCookiesSameName(
 			Function<HttpServerRequest, Map<CharSequence, ? extends Collection<Cookie>>> cookies,
 			String expectedResponse) {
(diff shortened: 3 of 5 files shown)


violetagg (approved):

violetagg:
@reactor/netty-team PTAL

simonbasle (changes requested):
that part of the code should be unit-testable. the main challenge being the `Supplier<SocketAdress>` I guess

violetagg:
@@roggenbrot Will you be able to add a test where the exception is constructed with `bind(..) failed: Die Adresse wird bereits verwendet` for example and then we can check that we still work as expected?

roggenbrot:
I will add a test case.

But i think the part I fixed is just copy&paste from ` ChannelBindException.fail`.

I will fix ` ChannelBindException.fail`, change NewConnectionProvider to use `ChannelBindException.fail` and add a test case for ` ChannelBindException fail`

simonbasle (approved):

simonbasle:
@roggenbrot good find! The PR needs some formatting polish (see checkstyle complains) but should otherwise be good to go

reactor-netty-core/src/test/java/reactor/netty/ChannelBindExceptionTest.java
```diff
@@ -0,0 +1,61 @@
...
+	@Test
+	@SuppressWarnings("unchecked")
+	void testFailBindExceptionCause() {
+
+		UnsupportedOperationException cause = new UnsupportedOperationException();
+		ChannelBindException ex = ChannelBindException.fail(new InetSocketAddress("test", 4956), cause);
+		assertThat(ex.getCause()).isEqualTo(cause);
+		assertThat(ex.localHost()).isEqualTo("test");
+		assertThat(ex.localPort()).isEqualTo(4956);
+		assertThat(ex.localPort()).isEqualTo(4956);
```
violetagg, line 37: nit: Duplicating row?

reactor-netty-core/src/test/java/reactor/netty/resources/NewConnectionProviderTest.java
```diff
@@ -0,0 +1,52 @@
...
+		Mockito.verify(sink).error(Mockito.argThat(a -> a instanceof UnsupportedOperationException));
+
+		connect = new NewConnectionProvider.DisposableConnect(sink, () -> new InetSocketAddress("test", 4956));
+		// Not possible to mock io.netty.channel.unix.Errors.NativeIoException or create a new instance because of Jni errors
+		// java.lang.UnsatisfiedLinkError: 'int io.netty.channel.unix.ErrorsStaticallyReferencedJniMethods.errnoENOENT()'
+		connect.onError(new IOException("bind(..) failed: Address already in use"));
+		Mockito.verify(sink).error(Mockito.argThat(a -> a instanceof ChannelBindException && ((ChannelBindException) a).localHost().equals("test") && ((ChannelBindException) a).localPort() == 4956));
+
+		// Issue-1668
+		connect = new NewConnectionProvider.DisposableConnect(sink, () -> new InetSocketAddress("tset", 6956));
```
violetagg, line 46: nit: typos -> `tset` vs `test`

violetagg (changes requested):
nit: There is no strong requirement but aim to wrap code at 120 characters

reactor-netty-core/src/test/java/reactor/netty/ChannelBindExceptionTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.io.IOException;
+import java.net.BindException;
+import java.net.InetSocketAddress;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ChannelBindExceptionTest {
+
+	@Test
+	@SuppressWarnings("unchecked")
```
violetagg, line 29: Do we need this suppression?

reactor-netty-core/src/test/java/reactor/netty/resources/NewConnectionProviderTest.java
roggenbrot: This typos are by purpose to test other string on second iteration. I will change it to test1..

violetagg: thanks

simonbasle (approved):

simonbasle:
@roggenbrot nothing to do with the PR content itself, but for future reference having these kind of commits for a PR is not great for the review process. looking at the log:

image

The merge commits are very weird, and I would expect no rebase nor merges during the review phase of a PR unless absolutely necessary (ie. you depend on a feature that is newly introduced in the base branch).

So I would have expected multiple commits that each represent a meaningful set of changes, ie. addressing review comments made since the last commit - and only that.
I would also expect commit message titles that are more descriptive (rather than a repetition of the first commit message).

Something like:

```
original commit title
review: add a test simulating german in exception message
review: add since tag to javadoc
review: fix formatting
review: fix formatting again 
```

The pattern of commits in this PR blurs what changed when coming back to the PR for review. fortunately it is small in size and can be compared as a whole to `main`, but had it been larger, it would have been very hard to review incrementally.

roggenbrot:
@simonbasle Sorry, struggling/fighting a little bit with amend/IntelliJ/gitflow which I'm not used to. I'm trying to do it better next time. Thanks for feedback

violetagg (approved):

Issue #1669, opened 2021-06-15
Add option to remove a added route from HttpServerRoutes


roggenbrot:
At the moment it's not possible to remove a previously added route from `HttpServerRoutes`

## Motivation
I'm currently working on an "Webflux light" for OSGI. Because of the dynamic nature of OSGI (Components are dynamic added/removed) I need a way to remove a previous added route when the component that added the route is removed.

While the OSGI side is pretty easy to implement, the problem is that there is no way currently to remove a previous added route ...

## Desired solution

Extend [HttpServerRoutes](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java) to a remove option. Remove option should be overloaded to be able to remove by:

(Examples, not thread safe)

- Predicate instance
```java
  public HttpServerRoutes remove(Predicate<? super HttpServerRequest> condition) {
      for(HttpRouteHandler h: handlers){
          if(h.condition.equals(condition)){
              handlers.remove(h);
          }
      }
      return this;
  }
```
- Handler instance
```java
  public HttpServerRoutes remove(BiFunction<? super HttpServerRequest, ? super HttpServerResponse, ? extends Publisher<Void>> handler) {
      for(HttpRouteHandler h: handlers){
          if(h.handler.equals(handler)){
              handlers.remove(h);
          }
      }
      return this;
  }
```
- Path
 ```java
  public HttpServerRoutes remove(String path) {
      for(HttpRouteHandler h: handlers){
          if(h.path != null && h.path.equals(path)){
              handlers.remove(h);
          }
      }
      return this;
  }
```

## Considered alternatives

Extend [HttpServerRoutes](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java) and add a new variant for each method that uses a publisher as a argument

```java
  public interface HttpServerRoute {

      Predicate<? super HttpServerRequest> condition();

      BiFunction<? super HttpServerRequest, ? super HttpServerResponse, ? extends Publisher<Void>> handler()

  }

public void route(Publisher<HttpServerRoute> publisher) {
  publisher.subscribe(new Subscriber<HttpServerRoute>() {
  
	  private final List<HttpRouteHandler> publisherHandlers = new CopyOnWriteArrayList<>();
  
	  @Override
	  public void onSubscribe(Subscription subscription) {
[...]

PR #1676, opened 2021-06-17
#1669  for discussion only

ctlove0523:
In current version,reactor netty hasn't really concept about ‘route’,we can't define route directly use code.In reactory netty,`HttpRouteHandler` is the route,when one request come in,we iterate `HttpRouteHandler` list to find one `HttpRouteHandler` can handle income request,if no `HttpRouteHandler` can handle request return http status code 404(mean no route). In the iterate process we use `Predicate<? super HttpServerRequest>` to match request one by one,when `Predicate<? super HttpServerRequest>` return true,it means there is one route for this request,so i think in reactory netty `Predicate<? super HttpServerRequest>` is  the "route".

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
@@ -30,6 +30,7 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
+import io.netty.handler.codec.http.HttpMethod;
 import org.reactivestreams.Publisher;
 import reactor.core.Exceptions;
 import reactor.core.publisher.Mono;
@@ -78,10 +79,10 @@ final class DefaultHttpServerRoutes implements HttpServerRoutes {
 	}
 
 	@Override
-	public HttpServerRoutes removeRoute(Predicate<? super HttpServerRequest> condition) {
+	public HttpServerRoutes removeIf(Predicate<? super HttpRouteHandlerMetadata> condition) {
 		Objects.requireNonNull(condition, "condition");
 
-		removedRoutes.add(condition);
+		handlers.removeIf(condition);
 
 		return this;
 	}
@@ -95,14 +96,14 @@ final class DefaultHttpServerRoutes implements HttpServerRoutes {
 		if (condition instanceof HttpPredicate) {
 			HttpRouteHandler httpRouteHandler = new HttpRouteHandler(condition,
 					handler,
-					(HttpPredicate) condition, ((HttpPredicate) condition).uri);
+					(HttpPredicate) condition, ((HttpPredicate) condition).uri, ((HttpPredicate) condition).method);
 
 			handlers.add(httpRouteHandler);
 			initialOrderHandlers.add(httpRouteHandler);
 
 		}
 		else {
-			HttpRouteHandler httpRouteHandler = new HttpRouteHandler(condition, handler, null, null);
+			HttpRouteHandler httpRouteHandler = new HttpRouteHandler(condition, handler, null, null, null);
 			handlers.add(httpRouteHandler);
 			initialOrderHandlers.add(httpRouteHandler);
 		}
@@ -131,21 +132,6 @@ final class DefaultHttpServerRoutes implements HttpServerRoutes {
 
 	@Override
 	public Publisher<Void> apply(HttpServerRequest request, HttpServerResponse response) {
-		// check if this request need process or not
-		final Iterator<Predicate<? super HttpServerRequest>> removedRouteIterator = removedRoutes.iterator();
-		Predicate<? super HttpServerRequest> removedRoute;
-		try {
-			while (removedRouteIterator.hasNext()) {
-				removedRoute = removedRouteIterator.next();
-				if (removedRoute.test(request)) {
-					return response.sendNotFound();
-				}
-			}
-		} catch (Throwable throwable) {
-			Exceptions.throwIfJvmFatal(throwable);
-			return Mono.error(throwable);
-		}
-
 		// find I/0 handler to process this request
 		final Iterator<HttpRouteHandler> iterator = handlers.iterator();
 		HttpRouteHandler cursor;
@@ -177,14 +163,18 @@ final class DefaultHttpServerRoutes implements HttpServerRoutes {
 
 		final String path;
 
+		final HttpMethod method;
+
 		HttpRouteHandler(Predicate<? super HttpServerRequest> condition,
 				BiFunction<? super HttpServerRequest, ? super HttpServerResponse, ? extends Publisher<Void>> handler,
 				@Nullable Function<? super String, Map<String, String>> resolver,
-				@Nullable String path) {
+				@Nullable String path,
+				@Nullable HttpMethod method) {
 			this.condition = Objects.requireNonNull(condition, "condition");
 			this.handler = Objects.requireNonNull(handler, "handler");
 			this.resolver = resolver;
 			this.path = path;
+			this.method = method;
 		}
 
 		@Override
@@ -201,6 +191,11 @@ final class DefaultHttpServerRoutes implements HttpServerRoutes {
 		@Override
 		public String getPath() {
 			return path;
+		}
+
+		@Override
+		public HttpMethod getMethod() {
+			return method;
 		}
 
 		@Override
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
@@ -92,9 +92,9 @@ final class DefaultHttpServerRoutes implements HttpServerRoutes {
 		Objects.requireNonNull(handler, "handler");
 
 		if (condition instanceof HttpPredicate) {
+			HttpPredicate predicate = (HttpPredicate) condition;
 			HttpRouteHandler httpRouteHandler = new HttpRouteHandler(condition,
-					handler,
-					(HttpPredicate) condition, ((HttpPredicate) condition).uri, ((HttpPredicate) condition).method);
+					handler, predicate, predicate.uri, predicate.method);
 
 			handlers.add(httpRouteHandler);
 			initialOrderHandlers.add(httpRouteHandler);
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
@@ -48,8 +48,6 @@ final class DefaultHttpServerRoutes implements HttpServerRoutes {
 
 	private Comparator<HttpRouteHandlerMetadata> comparator;
 
-	private List<Predicate<? super HttpServerRequest>> removedRoutes = new ArrayList<>();
-
 	@Override
 	public HttpServerRoutes directory(String uri, Path directory,
 			@Nullable Function<HttpServerResponse, HttpServerResponse> interceptor) {
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
@@ -47,6 +47,8 @@ final class DefaultHttpServerRoutes implements HttpServerRoutes {
 
 	private Comparator<HttpRouteHandlerMetadata> comparator;
 
+	private List<Predicate<? super HttpServerRequest>> removedRoutes = new ArrayList<>();
+
 	@Override
 	public HttpServerRoutes directory(String uri, Path directory,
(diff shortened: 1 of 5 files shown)


violetagg:
@ctlove0523 I converted this PR to draft.

04-30yan:
How to code my dns with other Parsing logic

violetagg:
@04-30yan May be you are mixing the topics? I'll encourage you to use [Gitter](https://gitter.im/reactor/reactor-netty) for questions related to Reactor Netty.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpPredicate.java
```diff
@@ -36,7 +36,7 @@
  *
  * @author Stephane Maldini
  */
-final class HttpPredicate
+public final class HttpPredicate
```
OlegDokuka, line 39: Not sure we can open publically that class

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
```diff
@@ -252,6 +252,15 @@ default HttpServerRoutes put(String path,
...
 	}
 
+	/**
+	 * A generic route predicate that if matched not invoke already register I/O handler use
+	 * {@link HttpServerRoutes#route(Predicate, BiFunction)}.
+	 *
+	 * @param condition a predicate given each inbound request
+	 * @return this {@link HttpServerRoutes}
+	 */
+	HttpServerRoutes removeRoute(Predicate<? super HttpServerRequest> condition);
```
OlegDokuka, line 262: how about to name that method as `removeIf`

OlegDokuka: also, will it be sufficient to remove handler know just path? @ctlove0523 can you expand on your use case and how you wanna remove routes? 

Can it be enough to have `Predicate<? super HttpRouteHandlerMetad>` to find the required route?

OlegDokuka:
@ctlove0523 just wanna add the general comment that the logic sounds like a filter logic where you based on the incoming request filters it or not. 

Though it can work for some use cases, it cannot help in scenarios where one needs to replace a handler with an equivalent one.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
OlegDokuka: I'm thinking of something like 
```
HttpServerRoutes removeIf(Predicate<? super HttpRouteHandlerMetadata> predicate);
```

where the default implementation will utilize it as follows: 

```
@Override
	public HttpServerRoutes removeIf(Predicate<? super HttpRouteHandlerMetadata> predicate) {
		Objects.requireNonNull(predicate, "predicate");

		handlers.removeIf(predicate);

		return this;
	}
	```

File: reactor-netty-http/src/main/java/reactor/netty/http/server/HttpPredicate.java (hunk 1 above)
Comment (ctlove0523) [reply reactor-netty-http/src/main/java/reactor/netty/http/server/HttpPredicate.java]: I make it  package-private again.

File: reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java (hunk 2 above)
Comment (ctlove0523) [reply reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java]: yes it has enough info to remove http route handler,i already change code like this.

File: reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java [hunk 3]
```diff
@@ -47,6 +48,8 @@
 
 	private Comparator<HttpRouteHandlerMetadata> comparator;
 
+	private List<Predicate<? super HttpServerRequest>> removedRoutes = new ArrayList<>();
```
OlegDokuka, line 51: this can be removed

OlegDokuka (approved):
LGTM in general with some minor polishes

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpServerRoutes.java
```diff
@@ -84,14 +94,14 @@ public HttpServerRoutes route(Predicate<? super HttpServerRequest> condition,
 		if (condition instanceof HttpPredicate) {
 			HttpRouteHandler httpRouteHandler = new HttpRouteHandler(condition,
 					handler,
-					(HttpPredicate) condition, ((HttpPredicate) condition).uri);
+					(HttpPredicate) condition, ((HttpPredicate) condition).uri, ((HttpPredicate) condition).method);
```
violetagg, line 97: Is it possible to do the casting `(HttpPredicate) condition` once?

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpRouteHandlerMetadata.java
```diff
@@ -36,4 +37,12 @@
 	 */
 	@Nullable
 	String getPath();
+
+	/**
+	 * Get the http method this handler can handle
+	 *
+	 * @return the http method {@link HttpMethod}
+	 */
```
violetagg, line 45: Add `@since 1.0.10`

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
```diff
@@ -252,6 +252,15 @@ default HttpServerRoutes put(String path,
 		return route(HttpPredicate.put(path), handler);
 	}
 
+	/**
+	 * A generic route predicate that if matched already register I/O handler use
+	 * {@link HttpServerRoutes#route(Predicate, BiFunction)} will be removed.
+	 *
+	 * @param condition a predicate given each http route handler {@link HttpRouteHandlerMetadata}
+	 * @return this {@link HttpServerRoutes}
+	 */
```
violetagg, line 261: Add `@since 1.0.10`

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2613,4 +2614,46 @@ private void doTestConnectionClosePropagatedAsError(String request) throws Excep
 				.isInstanceOf(AbortedException.class)
 				.hasMessage("Connection has been closed");
 	}
+
+	@Test
+	void testRemoveRoutes() {
```
violetagg, line 2619: Please add a suppression for the warning below as this is expected

```
/.../reactor-netty/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java:2643: warning: [ReferenceEquality] Comparison using reference equality instead of value equality
                                && metadata.getMethod() == HttpMethod.GET);
                                                        ^
    (see https://errorprone.info/bugpattern/ReferenceEquality)
  Did you mean '&& Objects.equals(metadata.getMethod(), HttpMethod.GET));' or '&& metadata.getMethod().equals(HttpMethod.GET));'?
/.../reactor-netty/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java:2650: warning: [ReferenceEquality] Comparison using reference equality instead of value equality
                                .expectNextMatches(response -> response.status() == HttpResponseStatus.NOT_FOUND)
                                                                                 ^
    (see https://errorprone.info/bugpattern/ReferenceEquality)
  Did you mean '.expectNextMatches(response -> Objects.equals(response.status(), HttpResponseStatus.NOT_FOUND))' or '.expectNextMatches(response -> response.status().equals(HttpResponseStatus.NOT_FOUND))'?
2 warnings
```

roggenbrot:
Nice work. 

Only Problem I see is that I can only specify the route that should be remove by path/method.

In my use case I have to deal with unknown services that have no knowledge from each other. 

Imagine the following scenario:

1.) First OSGI service is started and adds a new route with path `/test`
2.) Second OSGI service starts and adds a new route with path `/test`
3.) Second OSGI  is stopped and tries to remove the route he has added on start (cleanup)

The current solution would remove both routes (Route of OSGI service "First" and "Second"), or did I miss something?

ctlove0523:
> Nice work.
> 
> Only Problem I see is that I can only specify the route that should be removed by path/method.
> 
> In my use case I have to deal with unknown services that have no knowledge of each other.
> 
> Imagine the following scenario:
> 
> 1.) First OSGI service is started and adds a new route with path `/test`
> 2.) Second OSGI service starts and adds a new route with path `/test`
> 3.) Second OSGI is stopped and tries to remove the route he has added on start (cleanup)
> 
> The current solution would remove both routes (Route of OSGI service "First" and "Second"), or did I miss something?

In one HttpServer you can add many same route,but if you not order same routes,request will match first added route.From your description you want a concept like route owner,owner can only remove routes which added by the owner,but HttpServer has't this concept.

roggenbrot:
> In one HttpServer you can add many same route,but if you not order same routes,request will match first added route

That's correct. But adding a duplicate route will not harm the first added route. The first registered route will work, even if you register millions of routes with the same path/predicate.

Problem I see is that if you not aware of all other routes (e.g. App with Plugin Concept, OSGI ......), you can remove a existing route accidentally when you try to remove "your" route.

You will remove the first (working) route and "harm" the application since a route which was previously working suddenly does not to work, although your intention was to remove just "your" route. 

Debugging such "suddenly disappearing" routes could be a nightmare .....

With wildcard predicate handlers it might be even worse:

If you register a wildcard predicate handler, that should return a nice 404 page on every request which was not matched by a previous handler, using `removeIf` will always remove also such a handler. There is no chance to prevent this, which makes the `removeIf` unusable in such scenarios. 

Don't get me wrong. I really like you work and I really appreciate all the effort you put into it.

All I'm saying is that there might be uses cases where this solution might cause trouble since it's only safe to use it if you are aware of all added (Which might be even a problem when multiple teams work on the same project). 

I don't know/ can not decide if these use cases are valid and should be supported by reactor-netty, someone else has to make this decision.


>From your description you want a concept like route owner,owner can only remove routes which added by the owner,but >HttpServer has't this concept.

You are right, but maybe it should have such concept (Returning `Subscription`, add route via publisher .........)

ctlove0523:
@roggenbrot 
Thanks,the core issue is how to remove routes.If there are three routes,call remove method once,only remove one route?There are two ways may helpful.
**Not allowed to add the same route**
add same route more than once may be code issue and it doesn't make much sense .

**when add route return an object which can identity added route**
If we can get route's identity then can remove the route we want.

violetagg (approved):

violetagg:
> @roggenbrot
> Thanks,the core issue is how to remove routes.If there are three routes,call remove method once,only remove one route?There are two ways may helpful.
> **Not allowed to add the same route**
> add same route more than once may be code issue and it doesn't make much sense .
> 
> **when add route return an object which can identity added route**
> If we can get route's identity then can remove the route we want.

@roggenbrot Do you think that if we add something like a route identity, it will solve your use case?

violetagg:
@ctlove0523 @roggenbrot I'll postpone this for 1.0.11 so that we clarify the comment https://github.com/reactor/reactor-netty/pull/1676#issuecomment-887431351

violetagg:
@ctlove0523 @roggenbrot I'm going to commit this change.

@roggenbrot We can reopen your issue if you think that https://github.com/reactor/reactor-netty/pull/1676#issuecomment-887431351 may help for your use case.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpRouteHandlerMetadata.java
```diff
@@ -36,4 +37,13 @@
 	 */
 	@Nullable
 	String getPath();
+
+	/**
+	 * Get the http method this handler can handle
+	 *
+	 * @return the http method {@link HttpMethod}
+	 * @since 1.0.10
```
violetagg, line 45:
```suggestion
	 * @since 1.0.11
```

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRoutes.java
```diff
@@ -252,6 +252,16 @@ default HttpServerRoutes put(String path,
 		return route(HttpPredicate.put(path), handler);
 	}
 
+	/**
+	 * A generic route predicate that if matched already register I/O handler use
+	 * {@link HttpServerRoutes#route(Predicate, BiFunction)} will be removed.
+	 *
+	 * @param condition a predicate given each http route handler {@link HttpRouteHandlerMetadata}
+	 * @return this {@link HttpServerRoutes}
+	 * @since 1.0.10
```
violetagg, line 261:
```suggestion
	 * @since 1.0.11
```

violetagg (approved):

ctlove0523:
> @ctlove0523 @roggenbrot I'm going to commit this change.
> 
> @roggenbrot We can reopen your issue if you think that [#1676 (comment)](https://github.com/reactor/reactor-netty/pull/1676#issuecomment-887431351) may help for your use case.

Ok,if there are other requirements, i will continue to this.

violetagg:
Thank you all for the discussion and the PR!

PR #1687, opened 2021-06-24
[polish] Update date range in copyright header

violetagg:
The date range has been modified across the board to end at 2021.
For now start dates have been kept as is.

see reactor/reactor#682


simonbasle (approved):

violetagg:

PR #1688, opened 2021-06-24
[polish] Harmonise license URL indentation in headers

violetagg:
We keep 3 spaces: one for the comment block marker, 2 for indentation
(as in the ASF example).

see reactor/reactor#682

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *    https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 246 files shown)


simonbasle (approved):

simonbasle:
there are 4 spaces, but I don't really care... I don't even know if it is that useful. I just did it in reactor-core because I knew there were several different spacing issues. up to you @violetagg

violetagg:
> there are 4 spaces, but I don't really care... I don't even know if it is that useful. I just did it in reactor-core because I knew there were several different spacing issues. up to you @violetagg

oops I fixed that
thanks for the review

PR #1705, opened 2021-07-02
[build] Switch to Spotless plugin for license check

violetagg:


codequality/spotless/licenseSlashstarStyle.txt
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2011-2021 VMware, Inc. or its affiliates, All Rights Reserved.
```
simonbasle, line 2: the range here should be replaced with $YEAR. this will be interpreted by spotless as follows:

 - single 2021 is present -> all good
 - single year is present, not current year, eg. `2017` -> `2017-2021`
 - range is present ending in 2021 -> all good
 - range is present not ending in 2021, eg. 2017-2019` -> `2017-2021`

simonbasle (changes requested):
the template should have a variable

I would also at least locally apply `./gradlew spotlessApply -PspotlessFrom=ALL` and sanity check it to ensure no elements between the license header and the `package` keyword exist in any file (otherwise the above command would simply remove these, replacing everything between the top of the file and the `package` keyword with the license from the template)

simonbasle (approved):

codequality/spotless/licenseSlashstarStyle.txt
violetagg: fixed

violetagg:

PR #1713, opened 2021-07-05
Share the name resolver when the NameResolverProvider and the default LoopResources are the same

violetagg:
Every name resolver keeps an opened channel that is used for the name resolution.
The name resolver should be shared when the `NameResolverProvider` and
the default `LoopResources` are the same, thus the opened channels will be kept to a minimum.

This is related to the Gitter discussion
https://gitter.im/reactor/reactor-netty?at=60ddeafb258ebf6d92fc9c50

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransportConfig.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransportConfig.java
@@ -18,6 +18,7 @@ package reactor.netty.transport;
 import java.net.SocketAddress;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentMap;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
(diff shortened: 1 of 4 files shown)


OlegDokuka (approved):
LGTM

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransportConfig.java
```diff
@@ -226,6 +230,16 @@ protected void proxyProvider(@Nullable ProxyProvider proxyProvider) {
 		}
 	}
 
+	static final ConcurrentMap<Integer, DnsAddressResolverGroup> RESOLVERS_CACHE = PlatformDependent.newConcurrentHashMap();
+
+	static DnsAddressResolverGroup getOrCreateResolver(
+			NameResolverProvider nameResolverProvider,
+			LoopResources loopResources,
+			boolean preferNative) {
+		return RESOLVERS_CACHE.computeIfAbsent(Objects.hash(nameResolverProvider, loopResources, preferNative),
```
simonbasle, line 239: would it be less brittle to at least enforce `Objects.identityHashCode` on the `nameResolverProvider` and `loopResources`, just in case they get a hash code that is more prone to collisions at some point? (that would introduce a subtle bug here)

violetagg: `Objects.identityHashCode` should be `System.identityHashCode`?

Why don't we want to check the `NameResolverProvider#hashCode()`?

simonbasle: hashcodes _can_ have collisions, so technically if the classe's hashcode is something that is more likely to collision than the pseudo "memory pointer" from identityHashcode, you could get the wrong resolver out of the map, couldn't you ?

simonbasle: I missed the fact that `NameResolverProvider` already has a well-defined hashcode which reflects the configuration (and we want similarly configured resolvers to be shared).

simonbasle (approved):

violetagg:
@OlegDokuka @simonbasle Thanks for the review!

PR #1717, opened 2021-07-05
Obtain the metrics resolver on the configuration step when the metrics are enabled

### LINKED ISSUE #1707: Reactor Netty does calculate the poolkey hash for every supplied MetricsRecorder
In https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java#L83, the supplied metrics recorder instance is used to calculate the channelHash() value. This is causing the pool key cardinality to explode beyond whats expected.

As a workaround, implementing hashcode/equals or returning the same metricsRecorder is possible. However it seems much safer to calculate the hash off the supplier itself instead of the supplied instance.

violetagg:
This will guarantee the channel hash is calculated correctly as there are cases
when the metrics recorder supplier returns always different instances of the metrics recorder.

Fixes #1707

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java
@@ -80,7 +80,7 @@ public abstract class TransportConfig {
 
 	public int channelHash() {
 		return Objects.hash(attrs, bindAddress != null ? bindAddress.get() : 0, channelGroup, doOnChannelInit,
-				loggingHandler, loopResources, metricsRecorder != null ? metricsRecorder.get() : 0, observer, options, preferNative);
+				loggingHandler, loopResources, metricsRecorder, observer, options, preferNative);
 	}
 
 	/**
@@ -169,7 +169,7 @@ public abstract class TransportConfig {
 	 */
 	@Nullable
 	public final Supplier<? extends ChannelMetricsRecorder> metricsRecorder() {
-		return this.metricsRecorder;
+		return this.metricsRecorderSupplier;
 	}
 
 	/**
@@ -193,7 +193,8 @@ public abstract class TransportConfig {
 	ChannelPipelineConfigurer                  doOnChannelInit;
 	LoggingHandler                             loggingHandler;
 	LoopResources                              loopResources;
-	Supplier<? extends ChannelMetricsRecorder> metricsRecorder;
+	ChannelMetricsRecorder                     metricsRecorder;
+	Supplier<? extends ChannelMetricsRecorder> metricsRecorderSupplier;
 	ConnectionObserver                         observer;
 	Map<ChannelOption<?>, ?>                   options;
 	boolean                                    preferNative;
@@ -232,6 +233,7 @@ public abstract class TransportConfig {
 		this.loggingHandler = parent.loggingHandler;
 		this.loopResources = parent.loopResources;
 		this.metricsRecorder = parent.metricsRecorder;
+		this.metricsRecorderSupplier = parent.metricsRecorderSupplier;
 		this.observer = parent.observer;
 		this.options = parent.options;
 		this.preferNative = parent.preferNative;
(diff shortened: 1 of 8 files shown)


reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java
```diff
@@ -169,7 +169,7 @@ public final LoopResources loopResources() {
 	 */
 	@Nullable
 	public final Supplier<? extends ChannelMetricsRecorder> metricsRecorder() {
-		return this.metricsRecorder;
+		return this.metricsRecorderSupplier;
```
simonbasle, line 172: what is the goal of this method? imagine the `Supplier` initially provided creates a new `ChannelMetricsRecorder` on each call to `get()`. Do we want this method to return such a `Supplier`, or on the contrary an artificial supplier that always returns the current `metricsRecorder` (keeping the `Supplier` return type purely for backward compatibility)?

simonbasle (changes requested):

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java
violetagg: yeah it is for backwards compatibility ... should we deprecate it?

simonbasle: yeah, probably... although this brings the question of naming the replacement method.
another thing that might need a bit of rework is the setter (will comment there)

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java
```diff
@@ -311,8 +313,13 @@ protected void loggingHandler(LoggingHandler loggingHandler) {
 		this.loggingHandler = loggingHandler;
 	}
 
-	protected void metricsRecorder(@Nullable Supplier<? extends ChannelMetricsRecorder> metricsRecorder) {
-		this.metricsRecorder = metricsRecorder;
+	protected void metricsRecorder(@Nullable Supplier<? extends ChannelMetricsRecorder> metricsRecorderSupplier) {
+		this.metricsRecorderSupplier = metricsRecorderSupplier;
+		this.metricsRecorder = metricsRecorderSupplier != null ? metricsRecorderSupplier.get() : null;
```
simonbasle, line 318: here we completely remove the benefit of passing a `Supplier` by eagerly calling it if it is provided. the method is `protected` so we can't really remove it though.

but if we deprecate `Supplier metricsRecorder()` above, we should probably deprecate that one too and document the eager resolving for implementors?

actually, I would probably deprecate that one in priority.

simonbasle (changes requested):

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java
violetagg: Changed it to return a `Supplier` with the resolved metrics recorder

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java
simonbasle: also, if we don't return the original `Supplier`, there's no real need for keeping it around in `this.metricsRecorderSupplier` ;)

violetagg: yeah ... true

violetagg: > but if we deprecate Supplier metricsRecorder() above, we should probably deprecate that one too and document the eager resolving for implementors?

On the other hand this is used as a setter for the metrics recorder, so we need just to remove `this.metricsRecorderSupplier`

see here
https://github.com/reactor/reactor-netty/blob/d1f8f10a2dcb2805cfc70c84b1ab2b0ba08d9b1c/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java#L1197
and here
https://github.com/reactor/reactor-netty/blob/d1f8f10a2dcb2805cfc70c84b1ab2b0ba08d9b1c/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java#L567

Wdyt?

simonbasle: yeah, fine by me. we can come back to the `Supplier` part of the API later and refine 👍

simonbasle (approved):

violetagg:

PR #1729, opened 2021-07-08
Ensure SimpleCompressionHandler#decode can handle FullHttpRequest with freed content

violetagg:
In case of `H2C`, `FullHttpRequest` is received.
When the `FullHttpRequest` has content with 0 readableBytes, the content is released immediately.
Ensure `SimpleCompressionHandler#decode` can handle `FullHttpRequest` with freed content,
this can happen in cases when the response is delayed.

Related to https://github.com/spring-projects/spring-boot/issues/27176


simonbasle (approved):

violetagg:

PR #1731, opened 2021-07-08
Add the missing deprecated tag

### LINKED ISSUE #1730: DefaultConfigurationType marked as deprecated, but still used in public, non-deprecated API
[Non-deprecated usage](https://github.com/reactor/reactor-netty/blob/2ea7d622242f8750b70844c0a69a49446e8872f1/reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java#L97)

[Deprecation](https://github.com/reactor/reactor-netty/blob/2ea7d622242f8750b70844c0a69a49446e8872f1/reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java#L269)

Looks like `SslProvider#updateDefaultConfiguration` should be deprecated as well...

* Reactor version(s) used: 1.0.7

violetagg:
Fixes #1730

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
@@ -94,6 +94,12 @@ public final class SslProvider {
 		return new SslProvider(provider, handlerConfigurator);
 	}
 
+	/**
+	 * @deprecated as of 1.0.6. Prefer {@link SslProvider.SslContextSpec#sslContext(ProtocolSslContextSpec)},
+	 * where the default configuration is applied before any other custom configuration.
+	 * This will be removed in version 1.2.0.
+	 */
+	@Deprecated
 	public static SslProvider updateDefaultConfiguration(SslProvider provider, DefaultConfigurationType type) {
 		Objects.requireNonNull(provider, "provider");
 		Objects.requireNonNull(type, "type");


reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
```diff
@@ -94,6 +94,12 @@ public static SslProvider addHandlerConfigurator(
 		return new SslProvider(provider, handlerConfigurator);
 	}
 
+	/**
+	 * @deprecated as of 1.0.6. Prefer {@link SslProvider.SslContextSpec#sslContext(ProtocolSslContextSpec)},
+	 * where the default configuration is applied before any other custom configuration.
+	 * This will be removed in version 1.2.0.
```
simonbasle, line 100: should it state "at the earliest"? in the deprecation policy, we state that such versions are only a "minimal" date and that deprecated methods can be kept around for longer than that if they don't hurt too much (with the caveat that past the stated version, the method can be removed without particular notice in any patch release)

simonbasle (approved):

reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
violetagg: well this is something that we want to remove as soon as possible, isn't it?

violetagg:

PR #1736, opened 2021-07-13
Added logging for exception while creating default SSL provider

dnagarajan89:
In some cases,  a 'null' is returned while calling 'SslProvider.defaultClientProvider()' and this is causing a problem with [this LoC ](https://github.com/reactor/reactor-netty/blob/v1.0.3/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientSecure.java#L97)where it expects not null for SslProvider. This logging will help identify the cause of the problem in such cases.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/tcp/TcpClientSecure.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/tcp/TcpClientSecure.java
@@ -15,6 +15,9 @@
  */
 package reactor.netty.tcp;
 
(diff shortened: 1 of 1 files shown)


violetagg (approved):

violetagg:
@reactor/netty-team PTAL

simonbasle (approved):

violetagg:

violetagg:
@dnagarajan89 Thanks for the PR!

PR #1744, opened 2021-07-18
Fix invalid version number of getting-started doc.

phiysng:


violetagg:
@phiysng Thanks for the PR
I think that we need more general solution so that the version is always up-to-date.
Please take a look at #1745

PR #1768, opened 2021-08-09
Fix UnresolvedAddressException when no system proxy settings

### LINKED ISSUE #1765: setting `proxyWithSystemProperties()` without defining  jvm arguments like `HTTP_PROXY_HOST/HTTPS_PROXY_HOST/SOCKS_PROXY_HOST` causes `UnresolvedAddressException`
setting `proxyWithSystemProperties()` without defining  jvm arguments like `HTTP_PROXY_HOST/HTTPS_PROXY_HOST/SOCKS_PROXY_HOST` causes `UnresolvedAddressException`

## Expected Behavior

just works as no proxy set

## Actual Behavior

throws UnresolvedAddressException

## Steps to Reproduce

``` java
var httpClient = HttpClient.create().proxyWithSystemProperties();
var connector = new ReactorClientHttpConnector(httpClient);
var webClient = WebClient.builder().clientConnector(connector) .build();
```
if there is no proxy defined by jvm arguments, requesting a url with webClient causes UnresolvedAddressException

## Your Environment

Windows 10 and CentOS 7

* Reactor version(s) used: 1.0.9
* JVM version (`java -version`): openjdk 11.0.10 2021-01-19 LTS

violetagg:
Fixes #1765

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
@@ -240,11 +240,15 @@ public abstract class ClientTransport<T extends ClientTransport<T, CONF>,
 	 */
 	public T proxy(Consumer<? super ProxyProvider.TypeSpec> proxyOptions) {
 		Objects.requireNonNull(proxyOptions, "proxyOptions");
-		T dup = duplicate();
 		ProxyProvider.Build builder = (ProxyProvider.Build) ProxyProvider.builder();
 		proxyOptions.accept(builder);
+		return proxyWithProxyProvider(builder.build());
+	}
+
+	final T proxyWithProxyProvider(ProxyProvider proxy) {
+		T dup = duplicate();
 		CONF conf = dup.configuration();
-		conf.proxyProvider = builder.build();
+		conf.proxyProvider = proxy;
 		if (conf.resolver == null) {
 			conf.resolver = NoopAddressResolverGroup.INSTANCE;
 		}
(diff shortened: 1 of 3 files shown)


reactor-netty-core/src/test/java/reactor/netty/tcp/TcpClientTests.java
```diff
@@ -1372,4 +1372,33 @@ private void doTestSharedNameResolver(TcpClient client, boolean sharedClient) th
...
+				         .handle((req, res) -> res.sendString(Mono.just("noSystemProxySettings")))
+				         .bindNow();
+
+		AtomicReference<AddressResolverGroup<?>> resolver = new AtomicReference<>();
+		Connection conn = null;
+		try {
+			conn = TcpClient.create()
+			                .host("localhost")
+			                .port(disposableServer.port())
+			                .proxyWithSystemProperties()
```
simonbasle, line 1391: should there be an assertion that the current test environment doesn't have any proxy-related system properties ? Alternatively we can test `proxyWithSystemProperties(Properties)` instead of `proxyWithSystemProperties()`.

reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
```diff
@@ -240,34 +240,41 @@ public T port(int port) {
...
+	 * If a {@link ClientTransport} instance already has a proxy set via {@link ClientTransport#proxy(Consumer)}
+	 * the new instance created by this method has all proxy settings replaced
+	 * with proxy settings from the system properties only.
+	 * <p>
+	 * If a {@link ClientTransport} instance already has a proxy set via {@link ClientTransport#proxy(Consumer)}
+	 * but the system properties do not have a configuration for a proxy, the new
+	 * instance returned by this method behaves as there is no proxy settings.
+	 * <p>
+	 * If the system properties do not have a configuration for a proxy, the new
+	 * instance returned by this method behaves as there is no proxy settings.
```
simonbasle, line 276:
```suggestion
	 * instance returned by this method behaves as if there is no proxy settings.
```

simonbasle: the two last paragraphs can be merged together, I think. Something like:

```
	 * If the system properties do not have a configuration for a proxy, the new
	 * instance returned by this method behaves as there is no proxy settings,
	 * regardless of configuration of the original {@link ClientTransport} instance.
```

simonbasle (changes requested):

simonbasle (approved):

violetagg:

reactor-netty-core/src/test/java/reactor/netty/tcp/TcpClientTests.java
```diff
@@ -1372,4 +1374,37 @@ private void doTestSharedNameResolver(TcpClient client, boolean sharedClient) th
 			    .block();
 		}
 	}
+
+	/* https://github.com/reactor/reactor-netty/issues/1765 */
+	@Test
+	void noSystemProxySettings() {
+		Properties props = System.getProperties();
+		assumeThat(!(props.containsKey("http.proxyHost") || props.containsKey("https.proxyHost")
```
simonbasle, line 1382: for tracking purpose, it is a bit of a stretch but having an `assumeThat` here could hide issues in the future if a regression is introduced to `proxyWithSystemProperties()`-without-actual-properties AND the test environment does contain a proxy configuration.

PR #1776, opened 2021-08-17
Provide methods to configure the prefix for metrics (#1775)

### LINKED ISSUE #1775: Provide a util method to configure the prefix for metrics
## Motivation

It's cumbersome to create a new class extending `MicrometerChannelMetricsRecorder` just for a custom prefix.

## Desired solution

```java
            Connection connection = TcpClient.create()
                    .host(server.host())
                    .port(server.port())
                    .metrics(true, "my-custom-prefix") // what we want
                    .handle((in, out) -> Mono.never())
                    .connectNow();
````

### PR DESCRIPTION
**Pull request by JamesChenX**
Fixes #1775

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/tcp/TcpClientConfig.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/tcp/TcpClientConfig.java
@@ -142,12 +142,16 @@ public final class TcpClientConfig extends ClientTransportConfig<TcpClientConfig
 	 */
 	static final boolean SSL_DEBUG = Boolean.parseBoolean(System.getProperty(ReactorNetty.SSL_CLIENT_DEBUG, "false"));
 
-	static final class MicrometerTcpClientMetricsRecorder extends MicrometerChannelMetricsRecorder {
+	public static class MicrometerTcpClientMetricsRecorder extends MicrometerChannelMetricsRecorder {
 
 		static final MicrometerTcpClientMetricsRecorder INSTANCE = new MicrometerTcpClientMetricsRecorder();
 
-		MicrometerTcpClientMetricsRecorder() {
-			super(reactor.netty.Metrics.TCP_CLIENT_PREFIX, "tcp");
+		public MicrometerTcpClientMetricsRecorder(String prefix) {
+			super(prefix, "tcp");
+		}
+
+		public MicrometerTcpClientMetricsRecorder() {
+			this(reactor.netty.Metrics.TCP_CLIENT_PREFIX);
 		}
 	}
 
--- a/reactor-netty-core/src/main/java/reactor/netty/tcp/TcpServerConfig.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/tcp/TcpServerConfig.java
@@ -122,12 +122,16 @@ public final class TcpServerConfig extends ServerTransportConfig<TcpServerConfig
 	 */
 	static final boolean SSL_DEBUG = Boolean.parseBoolean(System.getProperty(ReactorNetty.SSL_SERVER_DEBUG, "false"));
 
-	static final class MicrometerTcpServerMetricsRecorder extends MicrometerChannelMetricsRecorder {
+	public static class MicrometerTcpServerMetricsRecorder extends MicrometerChannelMetricsRecorder {
 
 		static final MicrometerTcpServerMetricsRecorder INSTANCE = new MicrometerTcpServerMetricsRecorder();
 
-		MicrometerTcpServerMetricsRecorder() {
-			super(reactor.netty.Metrics.TCP_SERVER_PREFIX, "tcp");
+		public MicrometerTcpServerMetricsRecorder(String prefix) {
+			super(prefix, "tcp");
+		}
+
+		public MicrometerTcpServerMetricsRecorder() {
+			this(reactor.netty.Metrics.TCP_SERVER_PREFIX);
 		}
 	}
 
--- a/reactor-netty-core/src/main/java/reactor/netty/udp/UdpClientConfig.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/udp/UdpClientConfig.java
@@ -126,12 +126,16 @@ public final class UdpClientConfig extends ClientTransportConfig<UdpClientConfig
 			AdvancedByteBufFormat.HEX_DUMP
 					.toLoggingHandler(UdpClient.class.getName(), LogLevel.DEBUG, Charset.defaultCharset());
 
-	static final class MicrometerUdpClientMetricsRecorder extends MicrometerChannelMetricsRecorder {
+	public static class MicrometerUdpClientMetricsRecorder extends MicrometerChannelMetricsRecorder {
 
 		static final MicrometerUdpClientMetricsRecorder INSTANCE = new MicrometerUdpClientMetricsRecorder();
 
-		MicrometerUdpClientMetricsRecorder() {
-			super(reactor.netty.Metrics.UDP_CLIENT_PREFIX, "udp");
+		public MicrometerUdpClientMetricsRecorder(String prefix) {
+			super(prefix, "udp");
+		}
+
+		public MicrometerUdpClientMetricsRecorder() {
+			this(reactor.netty.Metrics.UDP_CLIENT_PREFIX);
 		}
 	}
 }
--- a/reactor-netty-core/src/main/java/reactor/netty/udp/UdpServerConfig.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/udp/UdpServerConfig.java
@@ -170,12 +170,16 @@ public final class UdpServerConfig extends TransportConfig {
 			AdvancedByteBufFormat.HEX_DUMP
 					.toLoggingHandler(UdpServer.class.getName(), LogLevel.DEBUG, Charset.defaultCharset());
 
-	static final class MicrometerUdpServerMetricsRecorder extends MicrometerChannelMetricsRecorder {
+	public static class MicrometerUdpServerMetricsRecorder extends MicrometerChannelMetricsRecorder {
 
 		static final MicrometerUdpServerMetricsRecorder INSTANCE = new MicrometerUdpServerMetricsRecorder();
 
-		MicrometerUdpServerMetricsRecorder() {
-			super(reactor.netty.Metrics.UDP_SERVER_PREFIX, "udp");
+		public MicrometerUdpServerMetricsRecorder(String prefix) {
+			super(prefix, "udp");
+		}
+
+		public MicrometerUdpServerMetricsRecorder() {
+			this(reactor.netty.Metrics.UDP_SERVER_PREFIX);
 		}
 	}
 
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerHttpClientMetricsRecorder.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerHttpClientMetricsRecorder.java
@@ -34,12 +34,16 @@ import static reactor.netty.Metrics.URI;
  * @author Violeta Georgieva
  * @since 0.9
  */
-final class MicrometerHttpClientMetricsRecorder extends MicrometerHttpMetricsRecorder implements HttpClientMetricsRecorder {
+public class MicrometerHttpClientMetricsRecorder extends MicrometerHttpMetricsRecorder implements HttpClientMetricsRecorder {
 
-	final static MicrometerHttpClientMetricsRecorder INSTANCE = new MicrometerHttpClientMetricsRecorder();
+	static final MicrometerHttpClientMetricsRecorder INSTANCE = new MicrometerHttpClientMetricsRecorder();
 
-	private MicrometerHttpClientMetricsRecorder() {
-		super(HTTP_CLIENT_PREFIX, "http");
+	public MicrometerHttpClientMetricsRecorder(String prefix) {
+		super(prefix, "http");
+	}
+
+	public MicrometerHttpClientMetricsRecorder() {
+		this(HTTP_CLIENT_PREFIX);
 	}
 
 	@Override
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
@@ -34,12 +34,16 @@ import static reactor.netty.Metrics.URI;
  * @author Violeta Georgieva
  * @since 0.9
  */
-final class MicrometerHttpServerMetricsRecorder extends MicrometerHttpMetricsRecorder implements HttpServerMetricsRecorder {
+public class MicrometerHttpServerMetricsRecorder extends MicrometerHttpMetricsRecorder implements HttpServerMetricsRecorder {
 
-	final static MicrometerHttpServerMetricsRecorder INSTANCE = new MicrometerHttpServerMetricsRecorder();
+	static final MicrometerHttpServerMetricsRecorder INSTANCE = new MicrometerHttpServerMetricsRecorder();
 
-	private MicrometerHttpServerMetricsRecorder() {
-		super(HTTP_SERVER_PREFIX, "http");
+	public MicrometerHttpServerMetricsRecorder(String prefix) {
+		super(prefix, "http");
+	}
+
+	public MicrometerHttpServerMetricsRecorder() {
+		this(HTTP_SERVER_PREFIX);
 	}
 
 	@Override

### CODE REVIEW

File: reactor-netty-core/src/main/java/reactor/netty/transport/Transport.java [hunk 1]
```diff
@@ -155,6 +156,48 @@ else if (configuration().metricsRecorder != null) {
...
+		if (enable) {
+			Objects.requireNonNull(prefix, "prefix");
+			Objects.requireNonNull(protocol, "protocol");
+			if (!Metrics.isInstrumentationAvailable()) {
+				throw new UnsupportedOperationException(
+						"To enable metrics, you must add the dependency `io.micrometer:micrometer-core`" +
+								" to the class path first");
+			}
+			T dup = duplicate();
+			dup.configuration().metricsRecorder(() -> new MicrometerChannelMetricsRecorder(prefix, protocol));
```
violetagg, line 186: For generic `Transport`s, this may be OK, but for example `HTTP`, we have a specific integration with Micrometer.

You may want to check 
https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerHttpClientMetricsRecorder.java

Also I don't want to extend the scope of the feature request, but if I think in perspective, there might be other requests for customising the default integration with Micrometer. Isn't is better if we consider some kind of `Builder` for configuring this default integration?

violetagg (changes requested):

reactor-netty-core/src/main/java/reactor/netty/transport/Transport.java
violetagg: `Transport` uses this contract for the default integration with Micrometer https://github.com/reactor/reactor-netty/blob/17d3c512f7a4ebff609b1298aa30fd027a8482c2/reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java#L294

JamesChenX: I have updated the commit to ensure they use their own `TCP/UDP/HTTP` recorders.
As you have mentioned `builder`, I think a more practical way is to make these metrics recorders `public` without `final` currently so users can extend them as they want. Using `builder` needs to consider what properties should be allowed to build, and it can be another ticket if someones else really need it.

violetagg: @JamesChenX Agree. Let's make `MicrometerHttpServerMetricsRecorder` and `MicrometerHttpClientMetricsRecorder` public classes with a corresponding constructor for the `prefix`.
Then the users can utilise:
- `reactor.netty.transport.Transport#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>)`
- `reactor.netty.http.client.HttpClient#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>)`
- `reactor.netty.http.client.HttpClient#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>, java.util.function.Function<java.lang.String,java.lang.String>)`
- `reactor.netty.http.server.HttpServer#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>)`
- `reactor.netty.http.server.HttpServer#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>, java.util.function.Function<java.lang.String,java.lang.String>)`

JamesChenX: I have added the constructor for the `prefix` in `MicrometerHttpServerMetricsRecorder` and `MicrometerHttpClientMetricsRecorder`
https://github.com/reactor/reactor-netty/pull/1776/files#diff-c82d769d5cd42c0cd7042f2294d0c6a10b360abd667551c3610c156c76440f8cR41-R43

violetagg: oops sorry :woman_facepalming:

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpClient.java
```diff
@@ -240,6 +240,10 @@ public TcpClient metrics(boolean enable) {
 		return super.metrics(enable);
 	}
 
+	public TcpClient metrics(boolean enable, String prefix) {
+		return metrics(enable, () -> new TcpClientConfig.MicrometerTcpClientMetricsRecorder(prefix));
+	}
+
```
violetagg, line 246: Is this needed? One can use 
`reactor.netty.transport.Transport#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>)` with `MicrometerChannelMetricsRecorder`

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpClientConfig.java
```diff
@@ -142,12 +142,16 @@ protected ChannelPipelineConfigurer defaultOnChannelInit() {
...
 
-		MicrometerTcpClientMetricsRecorder() {
-			super(reactor.netty.Metrics.TCP_CLIENT_PREFIX, "tcp");
+		public MicrometerTcpClientMetricsRecorder(String prefix) {
+			super(prefix, "tcp");
+		}
+
+		public MicrometerTcpClientMetricsRecorder() {
+			this(reactor.netty.Metrics.TCP_CLIENT_PREFIX);
 		}
```
violetagg, line 155: Is there a value of making this public? One can directly use `MicrometerChannelMetricsRecorder`.

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpServer.java
```diff
@@ -127,6 +127,10 @@ public TcpServer metrics(boolean enable) {
 		return super.metrics(enable);
 	}
 
+	public TcpServer metrics(boolean enable, String prefix) {
+		return metrics(enable, () -> new TcpServerConfig.MicrometerTcpServerMetricsRecorder(prefix));
+	}
+
```
violetagg, line 133: Is this needed? One can use 
`reactor.netty.transport.Transport#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>)` with `MicrometerChannelMetricsRecorder`

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpServerConfig.java
```diff
@@ -122,12 +122,16 @@ protected ChannelPipelineConfigurer defaultOnChannelInit() {
...
 		static final MicrometerTcpServerMetricsRecorder INSTANCE = new MicrometerTcpServerMetricsRecorder();
 
-		MicrometerTcpServerMetricsRecorder() {
-			super(reactor.netty.Metrics.TCP_SERVER_PREFIX, "tcp");
+		public MicrometerTcpServerMetricsRecorder(String prefix) {
+			super(prefix, "tcp");
+		}
+
+		public MicrometerTcpServerMetricsRecorder() {
+			this(reactor.netty.Metrics.TCP_SERVER_PREFIX);
```
violetagg, line 134: Is there a value of making this public? One can directly use `MicrometerChannelMetricsRecorder`.

reactor-netty-core/src/main/java/reactor/netty/udp/UdpClientConfig.java
```diff
@@ -126,12 +126,16 @@ protected ChannelMetricsRecorder defaultMetricsRecorder() {
...
 
-		MicrometerUdpClientMetricsRecorder() {
-			super(reactor.netty.Metrics.UDP_CLIENT_PREFIX, "udp");
+		public MicrometerUdpClientMetricsRecorder(String prefix) {
+			super(prefix, "udp");
+		}
+
+		public MicrometerUdpClientMetricsRecorder() {
+			this(reactor.netty.Metrics.UDP_CLIENT_PREFIX);
 		}
```
violetagg, line 139: Is there a value of making this public? One can directly use `MicrometerChannelMetricsRecorder`.

reactor-netty-core/src/main/java/reactor/netty/udp/UdpServerConfig.java
```diff
@@ -170,12 +170,16 @@ protected EventLoopGroup eventLoopGroup() {
...
 
-		MicrometerUdpServerMetricsRecorder() {
-			super(reactor.netty.Metrics.UDP_SERVER_PREFIX, "udp");
+		public MicrometerUdpServerMetricsRecorder(String prefix) {
+			super(prefix, "udp");
+		}
+
+		public MicrometerUdpServerMetricsRecorder() {
+			this(reactor.netty.Metrics.UDP_SERVER_PREFIX);
 		}
```
violetagg, line 183: Is there a value of making this public? One can directly use `MicrometerChannelMetricsRecorder`.

reactor-netty-core/src/main/java/reactor/netty/udp/UdpClient.java
```diff
@@ -135,6 +135,10 @@ public final UdpClient metrics(boolean enable) {
 		return super.metrics(enable);
 	}
 
+	public final UdpClient metrics(boolean enable, String prefix) {
+		return metrics(enable, () -> new UdpClientConfig.MicrometerUdpClientMetricsRecorder(prefix));
+	}
+
```
violetagg, line 141: Is this needed? One can use 
`reactor.netty.transport.Transport#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>)` with `MicrometerChannelMetricsRecorder`

reactor-netty-core/src/main/java/reactor/netty/udp/UdpServer.java
```diff
@@ -200,6 +200,10 @@ public final UdpServer metrics(boolean enable) {
 		return super.metrics(enable);
 	}
 
+	public final UdpServer metrics(boolean enable, String prefix) {
+		return metrics(enable, () -> new UdpServerConfig.MicrometerUdpServerMetricsRecorder(prefix));
+	}
+
 	@Override
```
violetagg, line 207: Is this needed? One can use 
`reactor.netty.transport.Transport#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>)` with `MicrometerChannelMetricsRecorder`

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1210,6 +1210,10 @@ else if (configuration().metricsRecorder() != null) {
 		}
 	}
 
+	public HttpClient metrics(boolean enable, String prefix) {
+		return metrics(enable, () -> new MicrometerHttpClientMetricsRecorder(prefix));
+	}
+
```
violetagg, line 1216: Is this needed? One can use 
- `reactor.netty.http.client.HttpClient#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>)`
- `reactor.netty.http.client.HttpClient#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>, java.util.function.Function<java.lang.String,java.lang.String>)` with `MicrometerHttpClientMetricsRecorder`

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerHttpClientMetricsRecorder.java
```diff
@@ -34,12 +34,16 @@
  * @author Violeta Georgieva
  * @since 0.9
  */
-final class MicrometerHttpClientMetricsRecorder extends MicrometerHttpMetricsRecorder implements HttpClientMetricsRecorder {
+public class MicrometerHttpClientMetricsRecorder extends MicrometerHttpMetricsRecorder implements HttpClientMetricsRecorder {
```
violetagg, line 37: Let's keep the class final and make it just public with the corresponding constructor. Wdyt?

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -34,12 +34,16 @@
  * @author Violeta Georgieva
  * @since 0.9
  */
-final class MicrometerHttpServerMetricsRecorder extends MicrometerHttpMetricsRecorder implements HttpServerMetricsRecorder {
+public class MicrometerHttpServerMetricsRecorder extends MicrometerHttpMetricsRecorder implements HttpServerMetricsRecorder {
```
violetagg, line 37: Let's keep the class final and make it just public with the corresponding constructor. Wdyt?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -580,6 +580,10 @@ else if (configuration().metricsRecorder() != null) {
 		}
 	}
 
+	public final HttpServer metrics(boolean enable, String prefix) {
+		return metrics(enable, () -> new MicrometerHttpServerMetricsRecorder(prefix));
+	}
+
```
violetagg, line 586: Is this needed? One can use

- `reactor.netty.http.server.HttpServer#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>)`
- `reactor.netty.http.server.HttpServer#metrics(boolean, java.util.function.Supplier<? extends reactor.netty.channel.ChannelMetricsRecorder>, java.util.function.Function<java.lang.String,java.lang.String>)` with `MicrometerHttpServerMetricsRecorder`

violetagg (changes requested):
Basically I think that if we have these classes public, then we have already the needed API
- MicrometerChannelMetricsRecorder
- MicrometerHttpServerMetricsRecorder
- MicrometerHttpClientMetricsRecorder

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpClient.java
JamesChenX: Removed. It just helps the user configure a prefix without the need to know the exact metrics recorder they use.

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpClientConfig.java
JamesChenX: Yes. The reason is the same for why we created a class `MicrometerTcpClientMetricsRecorder`, that's for abstraction.
If someday the implementation of `MicrometerTcpClientMetricsRecorder` changes, users don't need to know it because they can extend `MicrometerTcpClientMetricsRecorder`.

reactor-netty-core/src/main/java/reactor/netty/tcp/TcpServerConfig.java
JamesChenX: Yes. The reason is the same for why we created a class `MicrometerTcpServerMetricsRecorder`, that's for abstraction.
If someday the implementation of `MicrometerTcpServerMetricsRecorder` changes, users don't need to know it because they can extend `MicrometerTcpServerMetricsRecorder`.

reactor-netty-core/src/main/java/reactor/netty/udp/UdpClientConfig.java
JamesChenX: ditto

reactor-netty-core/src/main/java/reactor/netty/udp/UdpServerConfig.java
JamesChenX: Removed. It just helps the user configure a prefix without the need to know the exact metrics recorder they use.

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerHttpClientMetricsRecorder.java
JamesChenX: I think it's more practical to keep it public without `final`. The reason is the same for why we created a class `MicrometerHttpClientMetricsRecorder`, that's for abstraction.
If someday the implementation of `MicrometerHttpClientMetricsRecorder` changes, users don't need to know it because they can extend `MicrometerHttpClientMetricsRecorder`.

violetagg:
@JamesChenX I see what you mean, but I still think that we can keep those classes `final` and if there is a use case we can remove the `final`.

JamesChenX:
@violetagg Thanks for your support and I close the PR because I haven't plan to override these specific metrics recorder (e.g. HTTP) because they are just empty wrapper of `ChannelMetricsRecorder`, and I have implemented the custom prefix by extending from `MicrometerChannelMetricsRecorder`, something similar to the following code:
```java
.metrics(true, () -> new TurmsMicrometerChannelMetricsRecorder(MetricsConstant.CLIENT_NETWORK, "tcp"))
```
The code may help other developers to implement their own recorder.

PR #1799, opened 2021-09-10
Ensure HTTP/2 pool owner is taken from the connection and not from the stream

violetagg:
HTTP/2 pool owner is stored in the connection and not in the individual stream.


simonbasle (approved):

violetagg:

PR #1797, opened 2021-09-07
Remove WeakHashMap cache

violetagg:
The pool and the metrics are one and the same object.
The implementation already keeps all pools in a ConcurrentMap.
Add test to keep track that the pool and the metrics are the same object.

Fixes #1790

PR #1806, opened 2021-09-13
MicrometerPooledConnectionProviderMeterRegistrar works directly with the pool

violetagg:
MicrometerPooledConnectionProviderMeterRegistrar does not implement
ConnectionProvider.MeterRegistrar as we want to work directly with the pool
otherwise we need to keep another map with weak references which is not acceptable
Additional change for #1797


simonbasle (approved):

violetagg:
@simonbasle thanks for the review

PR #1817, opened 2021-09-21
Emission stuck test case

matyasberry:
Please have a look on the test case in the commit below.
Unless the `doOnError` line is uncommented, the test never finished.
I'd appreciate for any advice related to the reason why doOnError helps or why the test by itself does not finish.


violetagg:
@matyasberry Thanks a lot for the test case, it helped a lot! If you are able to test 1.0.12-SNAPSHOT, it will be great!

matyasberry:
thanks for solving it this quickly.

PR #1821, opened 2021-09-23
Fixes SYNC fusion message draining

OlegDokuka:


reactor-netty-core/src/main/java/reactor/netty/channel/MonoSendMany.java
```diff
@@ -578,6 +578,13 @@ public boolean trySuccess(Void result) {
...
+				if (sourceMode == SYNC) {
+					if (requested <= REFILL_SIZE) {
+						int u = MAX_SIZE - requested;
+						requested += u;
+						nextRequest += u;
+					}
+				}
 				trySchedule();
 				return true;
 			}
```
violetagg, line 590:
```suggestion
			if (checkTerminated() && sourceMode != SYNC) {
				trySchedule();
				return true;
			}
```

Can we do it like this in order to reduce duplications?

reactor-netty-core/src/test/java/reactor/netty/TcpEmissionTest.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.IntStream;
+
+import io.netty.buffer.Unpooled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+import reactor.core.publisher.Flux;
+import reactor.netty.tcp.TcpServer;
+
+public class TcpEmissionTest {
```
violetagg, line 30: Please revert this, I will merge the the other PR that provides the test.

violetagg (changes requested):

violetagg:
@OlegDokuka Sorry your first fix was the correct one so I picked it. Thanks for the PR!

PR #1823, opened 2021-09-24
[documentation] Add an example with ExtendedSocketOptions.TCP_KEEP* options for NIO transport

violetagg:
These socket options are available with Java 11


simonbasle (approved):

violetagg:

PR #1841, opened 2021-10-05
Support sending trailer headers on the server and receiving such on the client

### LINKED ISSUE #1837: Support for trailing headers
As of today, there is no way to pass through trailing headers into any response, because of that `Http2StreamFrameToHttpObjectCodec` will not be able, when encoding the last content, to send the End of stream.

This is a problem when using `gRPC` since the protocol relies on trailing headers.

## Considered alternatives
Not really an alternative but, adding a new `DefaultHttp2HeadersFrame` when `markSentBody`, will lead to the right code-path.

```
diff --git a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
index f48a2276..6490447c 100644
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
@@ -57,6 +57,8 @@ import io.netty.handler.codec.http.multipart.HttpData;
 import io.netty.handler.codec.http.multipart.HttpPostRequestDecoder;
 import io.netty.handler.codec.http.websocketx.CloseWebSocketFrame;
 import io.netty.handler.codec.http.websocketx.WebSocketCloseStatus;
+import io.netty.handler.codec.http2.DefaultHttp2Headers;
+import io.netty.handler.codec.http2.DefaultHttp2HeadersFrame;
 import io.netty.handler.codec.http2.HttpConversionUtil;
 import io.netty.util.AsciiString;
 import io.netty.util.ReferenceCountUtil;
@@ -629,7 +631,10 @@ class HttpServerOperations extends HttpOperations<HttpServerRequest, HttpServerR
                        f = channel().writeAndFlush(newFullBodyMessage(EMPTY_BUFFER));
                }
                else if (markSentBody()) {
-                       f = channel().writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT);
+//                     f = channel().writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT);
+                       f = channel()
+                                       .writeAndFlush(new DefaultHttp2HeadersFrame(new DefaultHttp2Headers(false)
+                                                       .add("grpc-status", "0"), true));
                }
                else {
                        discard();

```

## Additional context
https://github.com/netty/netty/blob/201f6b6bb2dcdff07df6e30552b96fedbcf21759/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java#L134-L137

violetagg:
Fixes #1837

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
@@ -19,7 +19,9 @@ import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -33,9 +35,12 @@ import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.DefaultHeaders;
 import io.netty.handler.codec.TooLongFrameException;
 import io.netty.handler.codec.http.DefaultFullHttpResponse;
+import io.netty.handler.codec.http.DefaultHttpHeaders;
 import io.netty.handler.codec.http.DefaultHttpResponse;
+import io.netty.handler.codec.http.DefaultLastHttpContent;
 import io.netty.handler.codec.http.FullHttpRequest;
 import io.netty.handler.codec.http.FullHttpResponse;
 import io.netty.handler.codec.http.HttpContent;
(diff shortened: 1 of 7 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
```diff
@@ -629,7 +643,26 @@ protected void onOutboundComplete() {
 			f = channel().writeAndFlush(newFullBodyMessage(EMPTY_BUFFER));
 		}
 		else if (markSentBody()) {
-			f = channel().writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT);
+			LastHttpContent lastHttpContent = LastHttpContent.EMPTY_LAST_CONTENT;
+			if (trailerHeadersConsumer != null && isTransferEncodingChunked(nettyResponse)) {
```
simonbasle, line 647: there are two cases where any `Consumer` defined by the user simply gets silently ignored: `isTransferEncodingChunked` is false or `declaredHeaderNames == null`.

should these cases be documented in the javadoc? or should they lead to an exception? (especially if a `Consumer<TrailerHeaders>` is provided in the setup but no `declaredHeaderNames`/`HttpHeaderNames.TRAILER` entry is provided at all?)

simonbasle (approved):
one question on the `markOutboundComplete` if/else and handling of cases where consumer is ignored. otherwise looks good. I'll let you decide if something must be changed in regard to that single comment, or if you want to merge despite that, so I'm Approving

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
violetagg: Please check it now, I added additional comments and javadoc

simonbasle (approved):

PR #1842, opened 2021-10-06
Support graceful shutdown for the ConnectionProvider

violetagg:


violetagg:
@Agraphie This is Reactor Netty's support for graceful shutdown for the `ConnectionProvider`

reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -138,7 +141,12 @@ protected PooledConnectionProvider(Builder builder) {
 			List<Mono<Void>> pools;
 			pools = channelPools.values()
 			                    .stream()
-			                    .map(Pool::disposeLater)
+			                    .map(pool -> {
+			                        if (pool instanceof GracefulShutdownInstrumentedPool) {
+			                            return ((GracefulShutdownInstrumentedPool<T>) pool).disposeGracefully(disposeTimeout);
```
simonbasle, line 146: maybe we want to `onErrorResume` this, since unlike `pool.disposeLater()` the graceful shutdown can "fail" (it just notifies that the resources were not all `released()` within the provided `Duration`, but does so after having actually shut down the pool anyway).

reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -143,7 +143,9 @@ protected PooledConnectionProvider(Builder builder) {
 			                    .stream()
 			                    .map(pool -> {
 			                        if (pool instanceof GracefulShutdownInstrumentedPool) {
-			                            return ((GracefulShutdownInstrumentedPool<T>) pool).disposeGracefully(disposeTimeout);
+			                            return ((GracefulShutdownInstrumentedPool<T>) pool)
+			                                    .disposeGracefully(disposeTimeout)
+			                                    .onErrorResume(t -> Mono.empty());
```
simonbasle, line 148: you might want to just log the fact that the pool didn't shut down gracefully within the grace period, up to you.

note that acquire `Mono`s that are still _pending_ at this point will terminate with an `onError` anyway.

what doesn't get notified is already acquired `Mono` for which the pool awaits a `release()`. once the pool is shut down, that release should still happen but it will lead to a destruction of the resource.

simonbasle (approved):

simonbasle (approved):

Issue #1854, opened 2021-10-13
ContextView in default implementations before ContextAwareHttpServerMetricsRecorder


HaloFour:
I am attempting to make use of the new APIs `ContextAwareHttpServerMetricsRecorder` to capture and record HTTP server metrics.  I am finding that the `ContextView` being passed to any of the methods is always empty.  This seems to be the result of the default implementation of `ConnectionObserver#currentContext` returning `Context.empty()` and some of the delegating implementations aren't implementing that method.  For example, `ServerTransport.ChildObserver` never delegates the method to `this.childObs.currentContext()` for the configured child observer.

## Expected Behavior
The `ContextView` should contain some context entries, at least those configured on the child ConnectionObservers.

## Actual Behavior
The `ContextView` is always empty.

## Steps to Reproduce
[Gist](https://gist.github.com/HaloFour/a08c5db29af2df18a27b849430d2eed9)

## Possible Solution
Work through the implementations of `ConnectionObserver` to find those that should delegate to a child instance of the interface.

## Your Environment

* Reactor version(s) used: `io.projectreactor.netty:reactor-netty-http:1.0.11`
* JVM version (`java -version`): 17

PR #1855, opened 2021-10-13
Delegate the Context from the child ConnectionObserver

HaloFour:
Delegates the `ConnectionObserver#currentContext` to the wrapped `childObs` in `ServerTransport.ChildObserver` so that context on the child observer should be accessible.

See: https://github.com/reactor/reactor-netty/issues/1854

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
@@ -477,6 +477,11 @@ public abstract class ServerTransport<T extends ServerTransport<T, CONF>,
 
 			childObs.onStateChange(connection, newState);
 		}
+
+		@Override
+		public Context currentContext() {
+			return childObs.currentContext();
+		}
 	}
 
 	static class DisposableBind implements CoreSubscriber<Channel>, DisposableServer, Connection {


pderop:
Hi HaloFour ,

I looked into your proposed patch and I also tend to think like you that the CompositeConnectionObserver also needs to merge the key/values from the multiple child observers. Indeed, currently, the last observer is used from the array:

```
	static final class CompositeConnectionObserver implements ConnectionObserver {

		final ConnectionObserver[] observers;

		CompositeConnectionObserver(ConnectionObserver[] observers) {
			this.observers = observers;
		}

		@Override
		public Context currentContext() {
			return observers[observers.length - 1].currentContext();
		}
```
And the last observer is always the HttpServer.HttpServerHandle observer.

Using the below patch seems to work for me:

```
	static final class CompositeConnectionObserver implements ConnectionObserver {

		final ConnectionObserver[] observers;
		volatile Context context;

		CompositeConnectionObserver(ConnectionObserver[] observers) {
			this.observers = observers;
			Map<Object, Object> mergedContext = new HashMap<>();
			Stream.of(observers)
					.flatMap(obs -> obs.currentContext().stream())
					.forEach(e -> mergedContext.put(e.getKey(), e.getValue()));
			this.context = Context.of(mergedContext);
		}

		@Override
		public Context currentContext() {
			return new Context() {
				@Override
				@SuppressWarnings({"unchecked", "TypeParameterUnusedInFormals"})
				public <T> T get(Object key) {
					return context.get(key);
				}

				@Override
				public boolean hasKey(Object key) {
					return context.hasKey(key);
				}

				@Override
				public int size() {
					return context.size();
				}

				@Override
				public Stream<Map.Entry<Object, Object>> stream() {
					return context.stream();
				}

				@Override
				public Context put(Object key, Object value) {
					context = context.put(key, value);
					return this;
				}

				@Override
				public Context delete(Object key) {
					context = context.delete(key);
					return this;
				}
			};
		}
...
```

Notice that in the above patch, the context is not final and volatile because for delete and put methods, we need to store the context with the contexts returned from the put/delete methods (because Contexts are immutable objects).

@violetagg, please comment on this ? 

Using the patched proposed by @HaloFour, as well as the patch in the CompositeConnectionObserver seems to unblock me for the other PR #1956

thanks

violetagg:
This is superseded by #2002

PR #1893, opened 2021-11-23
Add support for Macos M1 architecture

pderop:
As discussed this morning, please review this PR for adding support for macos M1 arch (osx-aarch_64).
thanks.

### DIFF
--- a/reactor-netty-core/build.gradle
+++ b/reactor-netty-core/build.gradle
@@ -55,7 +55,12 @@ dependencies {
 	api "io.netty:netty-handler-proxy:$nettyVersion"
 	api "io.netty:netty-resolver-dns:$nettyVersion"
 	if (!"$nettyVersion".endsWithAny("SNAPSHOT")) {
-		api "io.netty:netty-resolver-dns-native-macos:$nettyVersion:osx-x86_64"
+		if (osdetector.classifier == "osx-x86_64" || osdetector.classifier == "osx-aarch_64") {
+			api "io.netty:netty-resolver-dns-native-macos:$nettyVersion$os_suffix"
+		}
+		else {
+			api "io.netty:netty-resolver-dns-native-macos:$nettyVersion:osx-x86_64"
+		}
 	}
 	else {
 		// MacOS binaries are not available for Netty SNAPSHOT version
--- a/reactor-netty-http/build.gradle
+++ b/reactor-netty-http/build.gradle
@@ -63,7 +63,12 @@ dependencies {
 	api "io.netty:netty-resolver-dns:$nettyVersion"
 	// MacOS binaries are not available for Netty SNAPSHOT version
 	if (!"$nettyVersion".endsWithAny("SNAPSHOT")) {
-		api "io.netty:netty-resolver-dns-native-macos:$nettyVersion:osx-x86_64"
+		if (osdetector.classifier == "osx-x86_64" || osdetector.classifier == "osx-aarch_64") {
+			api "io.netty:netty-resolver-dns-native-macos:$nettyVersion$os_suffix"
+		}
+		else {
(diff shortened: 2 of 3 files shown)


violetagg (approved):

violetagg:
@reactor/netty-team PTAL

reactor-netty-core/build.gradle
```diff
@@ -55,7 +55,12 @@ dependencies {
 	api "io.netty:netty-handler-proxy:$nettyVersion"
 	api "io.netty:netty-resolver-dns:$nettyVersion"
 	if (!"$nettyVersion".endsWithAny("SNAPSHOT")) {
-		api "io.netty:netty-resolver-dns-native-macos:$nettyVersion:osx-x86_64"
+		if (osdetector.classifier == "osx-x86_64" || osdetector.classifier == "osx-aarch_64") {
+			api "io.netty:netty-resolver-dns-native-macos:$nettyVersion$os_suffix"
+		}
+		else {
+			api "io.netty:netty-resolver-dns-native-macos:$nettyVersion:osx-x86_64"
```
simonbasle, line 62: I wonder if that is still relevant if we cannot detect any `osx` from `osdetector` ?
ie should we default to pure `$nettyVersion` in that case ?

reactor-netty-http/build.gradle
```diff
@@ -63,7 +63,12 @@ dependencies {
 	api "io.netty:netty-resolver-dns:$nettyVersion"
 	// MacOS binaries are not available for Netty SNAPSHOT version
 	if (!"$nettyVersion".endsWithAny("SNAPSHOT")) {
-		api "io.netty:netty-resolver-dns-native-macos:$nettyVersion:osx-x86_64"
+		if (osdetector.classifier == "osx-x86_64" || osdetector.classifier == "osx-aarch_64") {
+			api "io.netty:netty-resolver-dns-native-macos:$nettyVersion$os_suffix"
+		}
+		else {
+			api "io.netty:netty-resolver-dns-native-macos:$nettyVersion:osx-x86_64"
```
simonbasle, line 70: same question as above

violetagg:
There are two important things of this change:

1. **Running a local build:** When somebody builds locally Reactor Netty and is running on Mac OS
  - when running on M1 arch, the build will apply the correct artifact which is `io.netty:netty-resolver-dns-native-macos:$nettyVersion$osx-aarch_64` and the name resolution functionality will use the corresponding Mac OS native implementation.
  - when is NOT running on M1 arch, the build will apply the correct artifact which is `io.netty:netty-resolver-dns-native-macos:$nettyVersion$osx-x86_64` and the name resolution functionality will use the corresponding Mac OS native implementation.
2. **Publishing Reactor Netty artifacts:** Reactor Netty publishes artifacts with default transitive dependency to `io.netty:netty-resolver-dns-native-macos:$nettyVersion$osx-x86_64` as this is currently the more popular arch. Doing that, users that use Reactor Netty in their applications on Mac OS will be able to use the corresponding Mac OS native implementation for the the name resolution, which is our preferred solution. This is similar to the fact that Reactor Netty has a transitive dependency to the binaries for the epoll transport.

If we depend only on `io.netty:netty-resolver-dns-native-macos:$nettyVersion` this means only compilation dependency against the classes, then the users must provide the arifacts with the natives by themselves.

My preference is to provide a solution for the most of the users where the Mac OS native implementation will be used for the the name resolution.

@reactor/netty-team Wdyt?

simonbasle:
Sounds good to me :+1:

reactor-netty-core/build.gradle
violetagg: @simonbasle Having the comment below, do we still need to address this?

simonbasle: no, sorry, I answered as a main comment instead of resolving this conversion. go for merge.

simonbasle (approved):

Issue #1891, opened 2021-11-23
WebsocketServerOperations sends WebSocketCloseStatus 1005 which is not accepted by CloseWebSocketFrame


Johannes-Rost:
## Expected Behavior
No exception should occur when WebsocketServerOperations closes a Frame.

## Actual Behavior
reactor.netty.http.server.WebsocketServerOperations#onInboundNext sends a WebSocketCloseStatus.EMPTY (1005) which is not accepted by CloseWebSocketFrame created in reactor.netty.http.client.WebsocketClientOperations#sendClose(int, java.lang.String), because it is not considered valid by io.netty.handler.codec.http.websocketx.WebSocketCloseStatus#isValidStatusCode

## Steps to Reproduce
Error appears in our logs. I could not reproduce it localy. The stacktrace looks like this:

```
Stack Trace:
  reactor.core.Exceptions$ErrorCallbackNotImplemented: java.lang.IllegalArgumentException: WebSocket close status code does NOT comply with RFC-6455: 1005
Caused by: java.lang.IllegalArgumentException: WebSocket close status code does NOT comply with RFC-6455: 1005
        at io.netty.handler.codec.http.websocketx.CloseWebSocketFrame.requireValidStatusCode(CloseWebSocketFrame.java:209)
        at io.netty.handler.codec.http.websocketx.CloseWebSocketFrame.<init>(CloseWebSocketFrame.java:69)
        at reactor.netty.http.client.WebsocketClientOperations.sendClose(WebsocketClientOperations.java:213)
        at org.springframework.web.reactive.socket.adapter.ReactorNettyWebSocketSession.close(ReactorNettyWebSocketSession.java:124)
        at reactor.core.publisher.MonoFlatMap$FlatMapMain.onNext(MonoFlatMap.java:125)
        at reactor.core.publisher.FluxFilter$FilterSubscriber.onNext(FluxFilter.java:113)
        at reactor.core.publisher.FluxMap$MapConditionalSubscriber.onNext(FluxMap.java:220)
        at reactor.core.publisher.FluxFirstWithSignal$FirstEmittingSubscriber.onNext(FluxFirstWithSignal.java:330)
        at org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber.onNext(ScopePassingSpanSubscriber.java:88)
        at reactor.core.publisher.Operators$MonoInnerProducerBase.complete(Operators.java:2634)
[...]

PR #1896, opened 2021-11-24
Clarify in javadoc that some status codes must not be used as such in websocket close frame

violetagg:
Related to #1891


simonbasle (approved):

violetagg:

Issue #1872, opened 2021-11-05
Implement additional metrics from reactor-pool


epicoxymoron:
[reactor-pool supports](https://github.com/reactor/reactor-pool/blob/main/src/main/java/reactor/pool/InstrumentedPool.java) metrics beyond those which reactor-netty understands: `#getMaxAllocatedSize()` and `#getMaxPendingAcquireSize()`.  It would be great if these could be made available to [reactor-netty's ConnectionPoolMetrics](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java).

## Motivation
I would like to monitor connection pool saturation.  I'm absolutely open to other ways of doing this, but first instinct is to monitor the ratio of active connections to the connection pool limit; I have a gauge only for the former.  If a gauge were made available for the latter, I could monitor both and get the ratio I'm looking for.

I don't have a use case for a gauge on the max pending acquire size, so I'll leave it to the project to consider whether adding that makes sense.

## Desired solution
The ConnectionPoolMetrics interface would be extended with the new field(s), and [DelegatingConnectionPoolMetrics](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/resources/DelegatingConnectionPoolMetrics.java) would be updated to add the new field(s) from the delegate.

## Considered alternatives
1. Fill the "max allocated size" in as a constant in my monitoring; this is problematic because when the true value changes, the monitoring system has to be updated separately, which is obviously error-prone.

## Additional context
None.

violetagg:
@epicoxymoron Do you configure the `ConnectionProvider` by yourself? Basically `#getMaxAllocatedSize()` is `maxConnections` configuration https://github.com/reactor/reactor-pool/blob/1fc4387fa7863c65b53b9bebc19a8f306465dd47/src/main/java/reactor/pool/AbstractPool.java#L99.
`#getMaxPendingAcquireSize()` is `pendingAcquireMaxCount` configuration https://github.com/reactor/reactor-pool/blob/1fc4387fa7863c65b53b9bebc19a8f306465dd47/src/main/java/reactor/pool/AbstractPool.java#L104

epicoxymoron:
Yes, I do.  I understand where the values are coming from.

[...]

PR #1897, opened 2021-11-25
#1872: Implement additional metrics from reactor-pool

pderop:
Please review this PR, which adds two methods in the ConnectionPoolMetrics interface:

The added methods (maxAllocatedSize, and maxPendingAcquireSize) simply expose the delegated methods from the original InstrumentedPool.PoolMetrics. This will most likely resolve the #1872 issue.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
@@ -54,4 +54,21 @@ public interface ConnectionPoolMetrics {
 	 */
 	int pendingAcquireSize();
 
+	/**
+	 * Get the maximum number of live resources this Pool will allow.
+	 * A Pool might be unbounded, in which case this method returns Integer.MAX_VALUE.
+	 * @return the maximum number of live resources that can be allocated by this Pool
+	 */
+	int maxAllocatedSize();
+
+	/**
+	 * Get the maximum number of Pool.acquire() this Pool can queue in a pending state
+	 * when no available resource is immediately handy (and the Pool cannot allocate more
+	 * resources).
+	 * A Pool pending queue might be unbounded, in which case this method returns
+	 * Integer.MAX_VALUE.
+	 * @return the maximum number of pending acquire that can be enqueued by this Pool
+	 */
+	int maxPendingAcquireSize();
+
 }
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
@@ -56,18 +56,22 @@ public interface ConnectionPoolMetrics {
 
 	/**
 	 * Get the maximum number of live resources this Pool will allow.
-	 * A Pool might be unbounded, in which case this method returns Integer.MAX_VALUE.
+	 * <p>
+	 * A Pool might be unbounded, in which case this method returns {@code Integer.MAX_VALUE}.
 	 * @return the maximum number of live resources that can be allocated by this Pool
+	 * @since 1.0.14
 	 */
 	int maxAllocatedSize();
 
 	/**
-	 * Get the maximum number of Pool.acquire() this Pool can queue in a pending state
+	 * Get the maximum number of {@code Pool.acquire()} this Pool can queue in a pending state
 	 * when no available resource is immediately handy (and the Pool cannot allocate more
 	 * resources).
+	 * <p>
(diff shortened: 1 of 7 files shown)


reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
```diff
@@ -54,4 +54,21 @@
 	 */
 	int pendingAcquireSize();
 
+	/**
+	 * Get the maximum number of live resources this Pool will allow.
+	 * A Pool might be unbounded, in which case this method returns Integer.MAX_VALUE.
```
violetagg, line 59: nit:

```suggestion
	 * A Pool might be unbounded, in which case this method returns {@code Integer.MAX_VALUE}.
```

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
```diff
@@ -54,4 +54,21 @@
...
+	 * @return the maximum number of live resources that can be allocated by this Pool
```
violetagg, line 60: Add `@since 1.0.14`
When a functionality is added after GA, we explicitly indicate in which version this functionality is added.

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
```diff
@@ -54,4 +54,21 @@
...
+	 */
+	int maxAllocatedSize();
+
+	/**
+	 * Get the maximum number of Pool.acquire() this Pool can queue in a pending state
```
violetagg, line 65: nit:

```suggestion
	 * Get the maximum number of {@code Pool.acquire()} this Pool can queue in a pending state
```

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
```diff
@@ -54,4 +54,21 @@
...
+	 * when no available resource is immediately handy (and the Pool cannot allocate more
+	 * resources).
+	 * A Pool pending queue might be unbounded, in which case this method returns
+	 * Integer.MAX_VALUE.
```
violetagg, line 69: nit:

```suggestion
	 * {@code Integer.MAX_VALUE}.
```

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionPoolMetrics.java
```diff
@@ -54,4 +54,21 @@
...
+	 * @return the maximum number of pending acquire that can be enqueued by this Pool
```
violetagg, line 70: Add `@since 1.0.14`

violetagg (changes requested):

violetagg (approved):

violetagg:
@reactor/netty-team PTAL

epicoxymoron:
If I'm understanding this PR correctly, the metrics wouldn't be pushed to Micrometer.  Correct me if I'm wrong, but wouldn't there need to also be changes [here](https://github.com/reactor/reactor-netty/blob/137cc47005c58dfbd375df62a9b0ba68760916c3/reactor-netty-core/src/main/java/reactor/netty/resources/MicrometerPooledConnectionProviderMeterRegistrar.java#L50)?

pderop:
@epicoxymoron , thanks for your comment, I have updated the PR in order to publish the added meters to Micrometers. Also updated the documentation.

violetagg (approved):

epicoxymoron:
This looks like it would solve my issue in #1872.  Thank you!

As a small nit, you may consider reusing the `@return` commentary from the [InstrumentedPool Javadocs](https://github.com/reactor/reactor-pool/blob/main/src/main/java/reactor/pool/InstrumentedPool.java) as the descriptions; personally I think it's a little more clear.

* .max.connections: the maximum number of live resources that can be allocated by this pool
* .max.pending.connections: the maximum number of pending acquire that can be enqueued by this pool

I'm not a member of the project team, but this has my approval with or without those changes.  Thank you again!

simonbasle (approved):

pderop:
I did not exactly copy the same descriptions from the reactor pool which is generic, but I slightly clarified the descriptions.

violetagg (approved):

epicoxymoron:
Verbiage LGTM.  You may want to propagate it to the docs + Javadocs on Metrics.java as well.

violetagg (approved):

simonbasle (approved):

PR #1909, opened 2021-12-01
Introduce functionality for mutating the ConnectionProvider configuration

### LINKED ISSUE #1800: HTTP client metrics - HTTP2 pool name should include underlying connection provider name
We leverage almost all of your micrometer-collected metrics.  We name the connection providers and this gives us a nice view of what's going on for our various clients (since you set create tag name=connProviderName).  We're just starting with HTTP2 and would like to have a view of both the underlying pool (that we named) and HTTP2 pool that sits on top of it.  However, the HTTP2 connection provider is named 'HTTP2'.  I'm actually not sure what the behavior is when you have more, but what we were thinking is it should be something like `name="http2." + underlyingConnProviderName`.

## Motivation

We need to monitor all the client pools so that we understand the behavior of the clients.

This will likely become even more important to us if you start adding metrics like 'open.streams' to an HTTP2 'connection'.

violetagg:
HTTP/2 connection pool inherits the configuration from its parent pool.
HTTP/2 connection pool name is a combination of `http2` and the parent pool's name.

Fixes #1800

### DIFF
--- a/reactor-netty-core/build.gradle
+++ b/reactor-netty-core/build.gradle
@@ -193,6 +193,12 @@ task japicmp(type: JapicmpTask) {
 	txtOutputFile = file("${project.buildDir}/reports/japi.txt")
 	ignoreMissingClasses = true
 	includeSynthetic = true
+	methodExcludes = [
+			// New methods with default implementation
+			// reactor.netty.resources
+			"reactor.netty.resources.ConnectionProvider#mutate()",
+			"reactor.netty.resources.ConnectionProvider#name()"
+	]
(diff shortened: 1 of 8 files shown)


pderop (commented):
LGTM, but maybe someone else should approve ?

simonbasle (approved):

violetagg:
@simonbasle I added a new commit to preserve the current behaviour

simonbasle (approved):
ah yeah, sorry I missed that the old implementation used to force `pendingAcquireMaxCount(-1)`

violetagg:

PR #1917, opened 2021-12-03
[documentation] Use NMT to track native memory leaks

pderop:
Added a note in the "How can I debug a memory leak" FAQ, on how to use Native Memory Tracking jvm tool.


pderop:
using NMT does not help because buffer pool reserves memory space in advance ...
however, some metrics that can be obtained from the memory pool, and it is then possible to monitor the current number of allocated direct buffers.

docs/asciidoc/faq.adoc
```diff
@@ -102,6 +102,25 @@ logger (`_reactor.netty.channel.LeakDetection`) that helps to identify where the
...
+        // display the total number of buffers allocated from direct memory arenas:
+        while (true) {
+             Thread.sleep(1000);
+             System.out.println(metrics.directArenas().stream()
+                 .collect(Collectors.summingLong(PoolArenaMetric::numAllocations)));
+        }
+```
+
+If the above code dislays some growing values, it's likely that there is a direct buffer memory leak.
+
```
violetagg, line 123: May be it is better to extend our Memory metrics. Wdyt?

violetagg: This discussion is interesting https://github.com/netty/netty/pull/11642

pderop: yes !, better is to add more meter(s).
ok, I'll work on this ...

pderop:
as suggested, I added two new gauges:

- reactor.netty.bytebuf.allocator.active.heap.buffers which provides the current number of active (acquired or released) heap buffers
- reactor.netty.bytebuf.allocator.active.direct.buffers which provides the current number of active (acquired or released) direct buffers

note that initially, I used a wrong meter from netty metrics: I used PoolArenaMetric::numAllocations, but it's actually PoolArenaMetric::numActiveAllocations which must be used.

pderop:
Changed comments for USED_HEAP_MEMORY and USED_DIRECT_MEMORY meters, which are actually reporting the amount of memory committed to the buffer allocator.

Renamed active.heap.buffers to active.heap.memory, and active.direct.buffers to active.direct.memory.

Clarified the documentation about this two new meters, which report the
actual bytes allocated for in-use buffers. Reported bytes may be greater than the buffer capacity because what is reported is the the amount of bytes allocated from internal buffer pool arenas for in-use buffers. This precisely corresponds to what has been suggested in https://github.com/netty/netty/pull/11642

docs/asciidoc/faq.adoc
```diff
@@ -102,6 +102,13 @@ logger (`_reactor.netty.channel.LeakDetection`) that helps to identify where the
 or whether Reactor Netty already forwarded the ownership of the buffers to the application/framework.
 By default, this logger is disabled. To enable it, increase the log level to `DEBUG`.
 
+Another way to detect memory leaks is to monitor _reactor.netty.bytebuf.allocator.active.heap.memory_ and _reactor.netty.bytebuf.allocator.active.direct.memory_ meters:
+
+- The _reactor.netty.bytebuf.allocator.active.heap.memory_ provides the actual bytes consumed by in-use buffers allocated from heap buffer pools
+- The _reactor.netty.bytebuf.allocator.active.direct.memory_ provides the actual bytes consumed by in-use buffers allocated from direct buffer pools
```
violetagg, line 108: It is more visible if we use `something` instead of _something_


```suggestion
Another way to detect memory leaks is to monitor `reactor.netty.bytebuf.allocator.active.heap.memory` and `reactor.netty.bytebuf.allocator.active.direct.memory` meters:

- The `reactor.netty.bytebuf.allocator.active.heap.memory` provides the actual bytes consumed by in-use buffers allocated from heap buffer pools
- The `reactor.netty.bytebuf.allocator.active.direct.memory` provides the actual bytes consumed by in-use buffers allocated from direct buffer pools
```

violetagg (changes requested):

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2022 VMware, Inc. or its affiliates, All Rights Reserved.
```
simonbasle, line 2:
```suggestion
 * Copyright (c) 2019-2022 VMware, Inc. or its affiliates, All Rights Reserved.
```

docs/asciidoc/alloc-metrics.adoc
```diff
@@ -3,12 +3,14 @@
 [width="100%",options="header"]
 |=======
 | metric name | type | description
-| reactor.netty.bytebuf.allocator.used.heap.memory | Gauge | The number of the bytes of the heap memory
-| reactor.netty.bytebuf.allocator.used.direct.memory | Gauge | The number of the bytes of the direct memory
+| reactor.netty.bytebuf.allocator.used.heap.memory | Gauge | The number of bytes committed to heap buffer allocator
+| reactor.netty.bytebuf.allocator.used.direct.memory | Gauge | The number of bytes committed to direct buffer allocator
```
simonbasle, line 7: how about "reserved by" rather than "committed to"?

```suggestion
| reactor.netty.bytebuf.allocator.used.heap.memory | Gauge | The number of bytes reserved by heap buffer allocator
| reactor.netty.bytebuf.allocator.used.direct.memory | Gauge | The number of bytes reserved by direct buffer allocator
```

if accepted, please reflect this change in wording in the various other files (javadocs)

simonbasle (changes requested):

chrisvest:
> my experiments showed that some times the pinnedDirectMemory() (or pinnedHeapMemory) is returning 0

Pinned is memory that is used by buffers, but you can have chunks that are allocated - so the memory is used by the process from the perspective of the OS - but is not used by any buffers.

pderop:
Hi Chris,

Thanks for your last comment. The intent of this PR is to try to report an estimate of the amount of memory used by in-use buffers (buffers currently allocated by applications, and not yet released to the buffer pool).

So, if I understand you correctly, you suggest to simply report the value returned by PooledByteBufAllocator.pinnedXXXMemory()  instead of using the code below ?

	double bytesAllocatedForInUseBuffers(List<PoolArenaMetric> arenas) {
		double totalUsed = 0;
		for (PoolArenaMetric arenaMetrics : arenas) {
			for (PoolChunkListMetric arenaMetric : arenaMetrics.chunkLists()) {
				for (PoolChunkMetric chunkMetric : arenaMetric) {
					// chunkMetric.chunkSize() returns maximum of bytes that can be served out of the chunk
					// and chunkMetric.freeBytes() returns the bytes that are not yet allocated by in-use buffers
					totalUsed += chunkMetric.chunkSize() - chunkMetric.freeBytes();
				}
			}
		}
		return totalUsed;
	}

thank you.

chrisvest:
@pderop Yeah, that's right. If you are seeing pinned memory returning zero when you know for a fact that pooled buffers of that memory type are still in use, then that is something we'd need to investigate.

pderop:
thanks Chris. So, I have created an issue here, with a sample project that is reproducing the issue where sometimes, the  pinnedDirectMemory is returning 0 even if some buffers are currently allocated and in-use.

https://github.com/netty/netty/issues/11984

simonbasle (approved):

violetagg (approved):

simonbasle (approved):

Issue #1844, opened 2021-10-07
Ensure #1807 and #1799 are covered with the new Http 2 connection pool


violetagg:
This is related to #1813

Issue #1818, opened 2021-09-21
HTTP/2 maxLifeTime of the connection


AmandeepSingh97:
While testing with maxLifeTime settings restrictions in the connection provider, it is observed that HTTP2 connections are not recycled when the connection is actively used for serving requests.

## Expected Behavior

HTTP/2 connection should not accept new streams after the maxLifeTime has been reached.

## Actual Behavior

HTTP/2 connection is actively used if there are incoming streams and maxLifeTime is not respected.

## Steps to Reproduce

Reactor netty testcase which depicts the [behaviour](https://github.com/AmandeepSingh97/reactor-netty/blob/http2maxlifetime/reactor-netty-http/src/test/java/reactor/netty/http/client/ConnectionPoolTests.java#L433)

## Your Environment

* Reactor version(s) used: 1.0.10
* Other relevant libraries versions (eg. `netty`, ...): 4.1.66.Final
* JVM version (`java -version`): openjdk 11.0.10
* OS and version (eg. `uname -a`): 19.6.0 Darwin Kernel Version

violetagg:
A change about this issue is available in 1.0.14-SNAPSHOT. It will be great if you are able to test it.

AmandeepSingh97:
> A change about this issue is available in 1.0.14-SNAPSHOT. It will be great if you are able to test it.

Sure @violetagg Me and my team will test it out and get back with the results.

dnagarajan89:
@violetagg Thanks for addressing this issue. We have validated with the test case that we have as well as the [reproducer](https://github.com/AmandeepSingh97/reactor-netty/blob/http2maxlifetime/reactor-netty-http/src/test/java/reactor/netty/http/client/ConnectionPoolTests.java#L433) that is shared in this issue earlier and we observed that the HTTP/2 connection is now respecting the maxLifeTime configuration.

violetagg:
> @violetagg Thanks for addressing this issue. We have validated((with the latest SNAPSHOT version) with the test case that we have as well as the [reproducer](https://github.com/AmandeepSingh97/reactor-netty/blob/http2maxlifetime/reactor-netty-http/src/test/java/reactor/netty/http/client/ConnectionPoolTests.java#L433) that is shared in this issue earlier and we observed that the HTTP/2 connection is now respecting the maxLifeTime configuration.

thanks

PR #1923, opened 2021-12-06
New HTTP/2 connection pool implementation

### LINKED ISSUE #1789: HTTP2: Use/Create new connections when maximum active streams is reached
the netty reactor does not use/create a new connection when the max concurrent stream is reached.

gitter discussion: https://gitter.im/reactor/reactor-netty?at=6126f8df5cfd665e52edf858

Netty throws `Max active streams is reached.; nested exception is java.io.IOException: Error while acquiring from reactor.netty.internal.shaded.reactor.pool.SimpleDequePool@8ae0ed. Max active streams is reached.` while the max connection is not reached.

## Expected Behavior
a new connection to be used.

## Actual Behavior
Netty throws `Max active streams is reached.

## Steps to Reproduce
I added the client/server code here: https://github.com/siavashta/sia-h2

The max connection on the client-side is set to 100
```java
    @Bean
    public HttpClient h2Client() {

        var provider = ConnectionProvider.builder("sia")
                .maxConnections(100)
                .build();

        return HttpClient.create(provider)
                .protocol(HttpProtocol.H2C)
                .wiretap(true);
    }
```
the max concurrent stream for the server is set to `1`
```java
		DisposableServer server =
				HttpServer.create()
						.protocol(HttpProtocol.H2C)
						.http2Settings(h2 -> h2.maxConcurrentStreams(1))
						.port(9192)
						.handle((req, res) -> res.sendString(Mono
								.just("sia")
								.delayElement(Duration.ofSeconds(5))))
						.wiretap(true)
						.bindNow();

		server.onDispose()
				.block();
```
running server and client
```
stahmureszadeh@TOR-MBP-7724 siaserver % mvn spring-boot:run
stahmureszadeh@TOR-MBP-7724 siaclient % mvn spring-boot:run
```
open jmeter and start the thread group with 5 concurrent requests to client
```
stahmureszadeh@TOR-MBP-7724 sia-h2 % jmeter -t siaclient/client.jmx
```

clients encountered `Max active streams is reached.` exception
```
org.springframework.web.reactive.function.client.WebClientRequestException: Error while acquiring from reactor.netty.internal.shaded.reactor.pool.SimpleDequePool@8ae0ed. Max active streams is reached.; nested exception is java.io.IOException: Error while acquiring from reactor.netty.internal.shaded.reactor.pool.SimpleDequePool@8ae0ed. Max active streams is reached.
```

## Possible Solution
1- Create the max connections in initialization and use them properly
2- retry better on creating a new connection

violetagg:
The connection is removed from the pool when:
- The connection is closed.
- The connection has reached its life time and there are no active streams.
- The connection has no active streams.
- When the client is in one of the two modes: 1) `H2` and `HTTP/1.1` or 2) `H2C` and `HTTP/1.1`,
and the negotiated protocol is `HTTP/1.1`.

The connection is filtered out when:
- The connection has reached its life time and there are active streams. In this case, the connection stays
in the pool, but it is not used. Once there are no active streams, the connection is removed from the pool.
- The connection has reached its max active streams configuration. In this case, the connection stays
in the pool, but it is not used. Once the number of the active streams is below max active streams configuration,
the connection can be used again.

This pool always invalidate the `PooledRef`, there is no release functionality.
- `PoolMetrics#acquiredSize()` and `PoolMetrics#allocatedSize()` always return the number of
the active streams from all connections currently in the pool.
- `PoolMetrics#idleSize()` always returns `0`.

Fixes #1789, #1818, #1844

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
@@ -59,11 +59,9 @@ import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
-import static reactor.netty.ConnectionObserver.State.DISCONNECTING;
 import static reactor.netty.ReactorNetty.format;
 import static reactor.netty.http.client.HttpClientState.STREAM_CONFIGURED;
 import static reactor.netty.http.client.HttpClientState.UPGRADE_REJECTED;
-import static reactor.netty.http.client.HttpClientState.UPGRADE_SUCCESSFUL;
 
 /**
  * A HTTP/2 implementation for pooled {@link ConnectionProvider}.
@@ -120,11 +118,7 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 			DisposableAcquire da = (DisposableAcquire) owner;
 			da.pooledRef
 			  .invalidate()
-			  .subscribe(null, null, () -> {
-			      if (log.isDebugEnabled()) {
-			          logPoolState(channel, da.pool, "Channel removed from the pool");
-			      }
-			  });
+			  .subscribe();
 		}
 	}
 
@@ -136,7 +130,7 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 	}
 
 	static void registerClose(Channel channel) {
-		ConnectionObserver owner = channel.attr(OWNER).get();
+		ConnectionObserver owner = channel.parent().attr(OWNER).get();
 		channel.closeFuture()
 		       .addListener(f -> {
 		           Channel parent = channel.parent();
@@ -146,10 +140,7 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 		               logStreamsState(channel, localEndpoint, "Stream closed");
 		           }
 
-		           if (localEndpoint.numActiveStreams() == 0) {
-		               channel.attr(OWNER).set(null);
-		               invalidate(owner, parent);
-		           }
+		           invalidate(owner, parent);
 		       });
 	}
 
@@ -259,10 +250,6 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 			this.pooledRef = pooledRef;
 			Channel channel = pooledRef.poolable().channel();
 
-			if (log.isDebugEnabled()) {
-				logPoolState(channel, pool, "Channel activated");
-			}
-
 			ConnectionObserver current = channel.attr(OWNER)
 			                                    .getAndSet(this);
 
@@ -294,10 +281,7 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 
 		@Override
 		public void onStateChange(Connection connection, State newState) {
-			if (newState == UPGRADE_SUCCESSFUL || newState == DISCONNECTING) {
-				release(connection.channel());
-			}
-			else if (newState == UPGRADE_REJECTED) {
+			if (newState == UPGRADE_REJECTED) {
 				invalidate(connection.channel().attr(OWNER).get(), connection.channel());
 			}
 
@@ -357,8 +341,6 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 			else {
 				sink.error(future.cause());
 			}
-
-			release(this, channel);
 		}
 
 		boolean isH2cUpgrade() {
@@ -411,23 +393,6 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 			}
 			return false;
 		}
-
-		static void release(Channel channel) {
-			release(channel.attr(OWNER).get(), channel);
-		}
-
-		static void release(@Nullable ConnectionObserver owner, Channel channel) {
-			if (owner instanceof DisposableAcquire) {
-				DisposableAcquire da = (DisposableAcquire) owner;
-				da.pooledRef
-				  .release()
-				  .subscribe(null, null, () -> {
-				      if (log.isDebugEnabled()) {
-				          logPoolState(channel, da.pool, "Channel deactivated");
-				      }
-				  });
-			}
-		}
 	}
 
 	static final class PendingConnectionObserver implements ConnectionObserver {
@@ -473,42 +438,18 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 			this.config = (HttpClientConfig) config;
 			this.remoteAddress = remoteAddress;
 			this.resolver = resolver;
-			this.pool = poolFactory.newPool(connectChannel(), null, DEFAULT_DESTROY_HANDLER, DEFAULT_EVICTION_PREDICATE);
+			this.pool = poolFactory.newPool(connectChannel(), null, DEFAULT_DESTROY_HANDLER, DEFAULT_EVICTION_PREDICATE,
+					poolConFig -> new Http2Pool(poolConFig, poolFactory.maxLifeTime()));
 		}
 
 		Publisher<Connection> connectChannel() {
 			return parent.acquire(config, new DelegatingConnectionObserver(), remoteAddress, resolver)
-				         .map(conn -> {
-				             if (log.isDebugEnabled()) {
-				                 logPoolState(conn.channel(), pool, "Channel acquired from the parent pool");
-				             }
-
-				             return conn;
-				         });
+				         .map(conn -> conn);
 		}
 
 		static final BiPredicate<Connection, PooledRefMetadata> DEFAULT_EVICTION_PREDICATE =
-				(connection, metadata) -> !connection.channel().isActive() || !connection.isPersistent();
+				(connection, metadata) -> false;
 
-		static final Function<Connection, Publisher<Void>> DEFAULT_DESTROY_HANDLER =
-				connection -> {
-					Channel channel = connection.channel();
-					if (channel.isActive()) {
-						Http2FrameCodec frameCodec = channel.pipeline().get(Http2FrameCodec.class);
-						if (frameCodec != null && frameCodec.connection().local().numActiveStreams() == 0) {
-							ChannelOperations<?, ?> ops = connection.as(ChannelOperations.class);
-							if (ops != null) {
-								ops.listener().onStateChange(ops, ConnectionObserver.State.DISCONNECTING);
-							}
-							else if (connection instanceof ConnectionObserver) {
-								((ConnectionObserver) connection).onStateChange(connection, ConnectionObserver.State.DISCONNECTING);
-							}
-							else {
-								connection.dispose();
-							}
-						}
-					}
-					return Mono.empty();
-				};
+		static final Function<Connection, Publisher<Void>> DEFAULT_DESTROY_HANDLER = connection -> Mono.empty();
 	}
 }
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
@@ -1,0 +1,829 @@
+/*
+ * Copyright (c) 2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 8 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -0,0 +1,829 @@
...
+			Http2FrameCodec frameCodec = connection.channel().pipeline().get(Http2FrameCodec.class);
+			if (frameCodec != null) {
+				if (pooledRef.slot.concurrency() == 0 && pooledRef.slot.invalidate()) {
+					releaseConnection(connection);
+				}
+			}
+			else {
+				// no need to guard here with invalidate()
+				// #findConnection will never give a connection that does not have Http2FrameCodec
+				poolConfig.allocationStrategy().returnPermits(1);
```
simonbasle, line 272: I need to have a harder look at this part, but I think I'm getting my head around it. I wonder if `returnPermit` outside of the one done in `invalidate` really makes sense. Would appreciate if you can detail the logic a bit more, though.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -0,0 +1,829 @@
...
+						// we are ready here, the connection can be used for opening another stream
+						slot.deactivate();
+						poolConfig.acquisitionScheduler().schedule(() -> borrower.deliver(new Http2PooledRef(slot)));
+					}
+					else {
+						borrowers.offerFirst(borrower);
+						continue;
+					}
+				}
+				else {
```
simonbasle, line 336: I need to look back at this part of the drain loop more in details. one thing I noticed is that it implements warmup.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -0,0 +1,829 @@
...
+
+			if (WIP.decrementAndGet(this) == 0) {
+				recordInteractionTimestamp();
+				break;
+			}
+		}
+	}
+
+	@Nullable
+	Slot findConnection(ConcurrentLinkedQueue<Slot> resources) {
```
simonbasle, line 429: would it make sense for `findConnection` to kinda "reserve" a `Slot` it found with `activate()`? I'm just wondering if by the time we `activate()` it in the drain loop, the various criteria checked below could have changed, and if so does activating early help... 🤔

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -0,0 +1,829 @@
...
+
+		@Override
+		public void subscribe(CoreSubscriber<? super PooledRef<Connection>> actual) {
+			Objects.requireNonNull(actual, "subscribing with null");
+			Borrower borrower = new Borrower(actual, parent, acquireTimeout);
+			actual.onSubscribe(borrower);
+		}
+	}
+
+	static final class Http2PooledRef extends AtomicBoolean implements PooledRef<Connection>, PooledRefMetadata {
```
simonbasle, line 675: the metrics are not exposed, but we still expose an `acquire` `Mono<PooledRef<Connection>>`. could the various counters, timestamps, etc... methods be improved?

simonbasle (changes requested):
looking promising. I have a few comments and questions, and I need to look again at some parts later, but sharing a first review round nonetheless

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -0,0 +1,829 @@
...
+ *     there are no idle resources in the pool. Once the connection does not have active streams, it
+ *     is returned to the parent pool.</li>
+ *     <li>{@link PoolConfig#evictionPredicate()} - the eviction predicate cannot be used as more complex
+ *     checks have to be done. Also the pool uses filtering for the connections (a connection might not be able
+ *     to be used but is required to stay in the pool).</li>
+ *     <li>{@link PoolConfig#metricsRecorder()} - no pool instrumentation.</li>
+ *     <li>{@link PoolConfig#releaseHandler()} - release functionality works as invalidate.</li>
+ *     <li>{@link PoolConfig#reuseIdleResourcesInLruOrder()} - FIFO is used when checking the connections.</li>
+ *     <li>FIFO is used when obtaining the pending borrowers</li>
+ *     <li>Warm up functionality is not supported</li>
```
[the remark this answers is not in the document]
violetagg: I meant this

```
@Override
public Mono<Integer> warmup() {
	return Mono.just(0);
}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
```diff
@@ -473,42 +438,18 @@ public void onUncaughtException(Connection connection, Throwable error) {
...
 		Publisher<Connection> connectChannel() {
 			return parent.acquire(config, new DelegatingConnectionObserver(), remoteAddress, resolver)
-				         .map(conn -> {
-				             if (log.isDebugEnabled()) {
-				                 logPoolState(conn.channel(), pool, "Channel acquired from the parent pool");
-				             }
-
-				             return conn;
-				         });
+				         .map(conn -> conn);
```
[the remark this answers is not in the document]
violetagg: > I'm guessing this is to force downcasting to Mono<Connection> rather than Mono<? extends Connection>?

yes

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
violetagg: This is for handling the cases where the negotiated protocol is HTTP/1.1

There is a test for this use case
reactor.netty.http.client.Http2PoolTest#nonHttp2ConnectionEmittedOnce

invalidate() makes sense only when we are in HTTP/2 use case.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -0,0 +1,829 @@
...
+		if (!isDisposed()) {
+			ConcurrentLinkedDeque<Borrower> q = pending;
+			q.remove(borrower);
+		}
+	}
+
+	Mono<Void> destroyPoolable(Http2PooledRef ref) {
+		Mono<Void> userProvidedDestroy;
+		try {
+			userProvidedDestroy = Mono.from(destroyHandler().apply(this, ref));
```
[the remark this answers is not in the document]
violetagg: inlined

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
simonbasle: nevermind the warmup comment, this was explained in header javadoc

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -0,0 +1,829 @@
...
+		void decrementConcurrency() {
+			CONCURRENCY.decrementAndGet(this);
+		}
+
+		void incrementConcurrency() {
+			CONCURRENCY.incrementAndGet(this);
+		}
+
+		boolean invalidate() {
+			if (STATE.compareAndSet(this, STATE_DEACTIVATED, STATE_INVALIDATED)) {
```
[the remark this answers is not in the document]
violetagg: If `Htt2PooledRef` tries to invalidate(), but the slot has been already activated then there is no need to proceed with invalidation, let us use that connection for opening a new stream. We want to return to the parent pool only the connections that doesn't have any active streams or are not in a process of having an active stream.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
violetagg: `Http2PooledRef` is always invalidated, there is no release functionality, because of that I made most of the metrics NA.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
violetagg: we have 3 main things checked in `findConnection`

1. channel has been closed
  * if the channel is closed, this cannot be changed
  * if the channel is opened and then closed at some time - this will fail on opening a stream and I don't see how we can prevent this with `activate()`
2. channel has reached max life time
  * if the channel reached max life time, this cannot be changed
  * if the channel can still be used and then later it reaches max life time - then this will be the last stream that will be opened on this channel and then the channel will be closed. I don't see how we can prevent this with `activate()`.
3. channel has reached max number of active streams - the channel will not be used and will stay in the pool. When some active stream finishes its work and is closed we will decrease the concurrency and drain will be invoked (destroyPoolable) at that time the channel will be able to be used again.  I don't see how we can prevent this with `activate()`.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
simonbasle: ok. I was assuming too much that the `Http2PooledRef` could somehow be presented to other components in the library as a generic `PooledRef` (or even to users). But it is hidden away in the `Http2ConnectionProvider`, so even `lifetime()` information might not be that useful.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -0,0 +1,820 @@
...
+		void deactivate() {
+			if (STATE.compareAndSet(this, STATE_ACTIVATED, STATE_DEACTIVATED)) {
+				if (log.isDebugEnabled()) {
+					log.debug(format(connection.channel(), "Channel deactivated"));
+				}
+				offerSlot(this);
+			}
+		}
+
+		void decrementConcurrency() {
```
simonbasle, line 792: I'm still a bit worried about these methods, and the fact that there is a separate `STATE` and `CONCURRENCY`... Here the `incrementConcurrency()`+`decrementConcurrency()`+`concurrency()` break the atomic benefit of `CONCURRENCY`. Can the STATE and CONCURRENCY be collapsed into a single int and the methods reworked to use the CAS / incrementAndGet / decrementAndGet better ?

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
simonbasle: ok, so it boils downs to the fact that we can still end up serving "slots" to HTTP/1.1 clients, in which case we never increment the slot usage nor do we want to keep it around, correct?

but since  `invalidate` does returnPermits + removeSlot under the cover, wouldn't it simplify the code to always call `invalidate` here ?

that is, provided `concurrency` is kept at `1` for the HTTP/1.1 case...

something like:

```java
Http2FrameCodec frameCodec = connection.channel().pipeline().get(Http2FrameCodec.class);
if (ref.slot.decrementConcurrencyAndGet() == 0 //more on this below
  && ref.slot.invalidate()) {
	if (frameCodec != null) {
		//case serving HTTP/2 client
		releaseConnection(connection);
	} //else serving HTTP/1.1 client, no need to releaseConnection
}
```

I'm a bit worried that calls to `decrementConcurrency` and `concurrency()` are split, as this negates the benefits of the underlying atomic: you increment and check the value separately. I think it would be preferable to expose more atomic operations, like prototyped in the snippet above (see my other comment in `Slot` too).

simonbasle (changes requested):

simonbasle (approved):
Approving, with a side note:

Since there is no warmup-allocation in `drainLoop` at all anymore, I would suggest to add a "security belt" just in case an allocationStrategy is introduced by mistake which has a min size > 0. This can be a test, a check in the `Http2Pool` constructor (eg. `config.allocationStrategy.getPermits(0) == 0`) or in the drainLoop when we call `getPermits(1)`. since the usage is internal, I'm ok with setting that aside, but it can lead to subtle bugs.

violetagg:
> Approving, with a side note:
> 
> Since there is no warmup-allocation in `drainLoop` at all anymore, I would suggest to add a "security belt" just in case an allocationStrategy is introduced by mistake which has a min size > 0. This can be a test, a check in the `Http2Pool` constructor (eg. `config.allocationStrategy.getPermits(0) == 0`) or in the drainLoop when we call `getPermits(1)`. since the usage is internal, I'm ok with setting that aside, but it can lead to subtle bugs.

I added a check in the constructor + test

simonbasle (approved):
perfect, good to go!

violetagg:

PR #1930, opened 2021-12-12
Add a pre-screen before locking inside computeIfAbsent

zimatars:
computeIfAbsent pessimistically locks instead of optimistically trying to return an existing entry.


simonbasle:
@zimatars @pderop has made some measurements, and from what I understand in newer versions of java there's an integrated pre-screen that is even better because it can accomodate parrallel deletions.

this particular map is not necessarily a great candidate for pre-screening that way, precisely because entries are regularly deleted from it (and the pre-screen isn't atomic)

there are however other instances in the code of reactor-netty where maps are more like pre-established dictionaries with very little deletion, which could benefit from your approach. Do you want to explore that in a separate PR?

@reactor/netty-team I don't think we can merge this one

simonbasle (changes requested):
(see my comment above)
I'll delegate the final choice to @violetagg

simonbasle:
@zimatars, @violetagg is still thinking about whether or not to apply this PR, we'll keep you posted for that particular case.

violetagg:
@zimatars I'm going to accept this PR. Thanks a lot!

violetagg (approved):

PR #1948, opened 2021-12-20
Ensure HttpConnectionProvider exposes correct maxConnections/maxConnectionsPerHost information

### LINKED ISSUE #1941: `maxConnections` value of a HttpClient returns -1
`HttpClient` returns maxConnections as `-1` even when the `maxConnections` property is set to a non-negative number.

## Expected Behavior

The maxConnections() method should return the actual value that was configured when the client was created.

## Actual Behavior

Always returns `-1`

## Steps to Reproduce

```java
@Test
public void reproCase() {
    HttpClient httpClientWithoutProvider = HttpClient.create();
    System.out.println("Max connections without a connection provider: "
            + httpClientWithoutProvider.configuration().connectionProvider().maxConnections()); // this should ideally print a non-negative  value - the maxConnections used by the client.

    int maxConnections = 200;
    ConnectionProvider connectionProvider = ConnectionProvider.create("test", maxConnections);
    HttpClient httpClientWithProvider = HttpClient.create(connectionProvider);
    System.out.println("Max connections with a connection provider: "
            + httpClientWithProvider.configuration().connectionProvider().maxConnections()); // this should print 200
}
```
Running the above code prints the following results:

```
Max connections without a connection provider: -1
Max connections with a connection provider: -1
```

## Possible Solution

It looks like `HttpConnectionProvider` does not implement the `maxConnections()` method and falls back to the default implementation provided by `ConnectionProvider` interface which [always returns -1](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java#L221). Possible solution is to implement this method in `HttpConnectionProvider` and read the maxConnections value from `http1ConnectionProvider`.

## Your Environment

* Reactor version(s) used: 3.4.12
* Other relevant libraries versions (eg. `netty`, ...): reactor-netty 1.0.13
* OS and version (eg. `uname -a`): Windows 11

violetagg:
Fixes #1941

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpConnectionProvider.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpConnectionProvider.java
@@ -24,6 +24,7 @@ import reactor.netty.transport.TransportConfig;
 import reactor.util.annotation.Nullable;
 
 import java.net.SocketAddress;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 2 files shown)


simonbasle (approved):

pderop (approved):
LGTM

violetagg:

Issue #1366, opened 2020-11-05
HTTP server connection Pool metrics


jtorkkel:
HTTP server Connection Pool metrics missing.

SpringBoot 1 / Jetty (and Tomcat) provides HTTP server connection pool metrics
 - jetty_threads_busy
- jetty_threads_config_max
- jetty_threads_config_min
- jetty_threads_current
- jetty_threads_idle
- jetty_threads_jobs

In Springboot 2 WebFlux provides Pooled ConnectionProvider metrics
https://projectreactor.io/docs/netty/release/reference/index.html#_metrics_5
- reactor.netty.connection.provider.total.connections
- reactor.netty.connection.provider.active.connections
- reactor.netty.connection.provider.idle.connections

Similar metrics should be available for HTTP server

violetagg:
@jtorkkel 
> HTTP server Connection Pool metrics missing.
> 
> SpringBoot 1 / Jetty (and Tomcat) provides HTTP server connection pool metrics
> 
> * jetty_threads_busy
> * jetty_threads_config_max
> * jetty_threads_config_min
> * jetty_threads_current
> * jetty_threads_idle
> * jetty_threads_jobs

These are not connections but threads?
Reactor Netty uses event loop, it works with a very few threads (typically equal to the number of cores).

> 
> In Springboot 2 WebFlux provides Pooled ConnectionProvider metrics
> https://projectreactor.io/docs/netty/release/reference/index.html#_metrics_5
> 
> * reactor.netty.connection.provider.total.connections
> * reactor.netty.connection.provider.active.connections
> * reactor.netty.connection.provider.idle.connections
> 
> Similar metrics should be available for HTTP server

HTTP server does not use connection pool. Can you clarify a bit?

violetagg:
@jtorkkel In order to address this enhancement we need an answers to the questions above.

jtorkkel:
Hi, sorry for the delay.

I did not mean actual pools but have similar metrics how many sockets/channels netty server has open, how many are active and how many are inactive.

I found that extending Netty ChannelInboundHandlerAdapter methods channelActive and channelInactive it might be possible to tract the active connections. I hope the number of open socket can be found somewhere so with this data metrics could be created.
Article referring how incoming connection can be limited using this approach. https://stackoverflow.com/questions/46849964/how-to-set-maximum-concurrent-connections-with-netty

violetagg:
[...]

PR #1956, opened 2021-12-23
Add more server connection metrics

pderop:
This PR which adds the following metrics for the #1366 issue:

- reactor.netty.tcp.server.total.connections: this new meter (Counter) reflects the number of currently opened and accepted tcp server connections. TAGS supported: uri (always "tcp"), local.address (corresponds to server listening address, in the format ip:port)

- reactor.netty.http.server.total.connections: this new meter (Counter) reflects the number of currently opened and accepted http server connections. TAGS supported: uri (always "http"), local.address (corresponds to server listening address, in the format ip:port)

- reactor.netty.http.server.active.connections: this new meter (Counter) reflects the number of http connections currently processing requests. TAGS supported: uri (the http request URI, or the uri transformed by the uri mapper function, if any), local.address (corresponds to server listening address, in the format ip:port)

Note: for the ContextAwareHttpServerMetricsRecorder, the ContextView being passed to any of the methods is always empty, this is caused by another pending related issue from #1854


pderop:
please do not review for the moment, it seems some tests are failing, I need to investigate further.

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2019-2022 VMware, Inc. or its affiliates, All Rights Reserved.
```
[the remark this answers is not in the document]
pderop: thanks Violeta, applied suggestion.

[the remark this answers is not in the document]
simonbasle: mmh I saw 2022 only in the other PR (which will likely clash with that one btw) and this file _was_ created in 2019, so it was correct

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2022 VMware, Inc. or its affiliates, All Rights Reserved.
```
simonbasle, line 2:
```suggestion
 * Copyright (c) 2019-2022 VMware, Inc. or its affiliates, All Rights Reserved.
```

simonbasle (approved):
~looks good, with the small nitpick of the Metrics license header date range.~

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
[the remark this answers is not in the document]
violetagg: @simonbasle ooo, my bad. You are right. I was thinking that this is a new file.

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelMetricsRecorder.java
```diff
@@ -76,4 +76,14 @@
...
 	void recordResolveAddressTime(SocketAddress remoteAddress, Duration time, String status);
+
+	/**
+	 * Updates the number of the channels currently opened and accepted by the server
+	 *
+	 * @param localAddress the server local address
+	 * @param amount The amount of server connection to add in the meter
+	 * @since 1.0.15
+	 */
+	default void incrementServerConnections(SocketAddress localAddress, int amount) { }
```
violetagg, line 87: @simonbasle What do you think about the API? I'm still thinking that this is a bit ... yeah I don't know.

Increment
`recorder().incrementServerConnections(ctx.channel().localAddress(), -1);`

Decrement
`recorder().incrementServerConnections(ctx.channel().localAddress(), -1);`

simonbasle: I guess a single method called increment which accepts a negative number is a bit weird 🤔 

in fact, how did I miss this: `Counter` in micrometer **doesn't support decrementing**:

> Counters report a single metric, a count. The Counter interface allows you to increment by a fixed amount, which must be positive.

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -37,9 +43,24 @@
...
+	protected final ConcurrentMap<MeterKey, Counter> activeConnectionsCache = new ConcurrentHashMap<>();
+
+	protected final Counter.Builder connectionsBuilder;
+	protected final ConcurrentMap<MeterKey, Counter> connectionsCache = new ConcurrentHashMap<>();
 
 	private MicrometerHttpServerMetricsRecorder() {
-		super(HTTP_SERVER_PREFIX, "http");
+		super(HTTP_SERVER_PREFIX, HTTP);
+
+		this.activeConnectionsBuilder =
```
simonbasle, line 57: same as in TCP, use a `Gauge` over `AtomicInteger` (unless we think integer max value is not enough in practice, in which case `AtomicLong`)

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -114,6 +135,32 @@ public void incrementErrorsCount(SocketAddress remoteAddress, String uri) {
...
+		}
+	}
+
+	@Override
+	public void incrementServerConnections(SocketAddress localAddress, int amount) {
+		String address = Metrics.formatSocketAddress(localAddress);
+		MeterKey meterKey = new MeterKey(null, address, null, null);
+		Counter connections = connectionsCache.get(meterKey);
+		connections = connections != null ? connections : connectionsCache.computeIfAbsent(meterKey,
+						key -> filter(connectionsBuilder.tags(URI, HTTP, LOCAL_ADDRESS, address).register(REGISTRY)));
```
simonbasle, line 157: should we replace `URI` with a new `PROTOCOL` tag?

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -37,9 +43,24 @@
 final class MicrometerHttpServerMetricsRecorder extends MicrometerHttpMetricsRecorder implements HttpServerMetricsRecorder {
 
 	final static MicrometerHttpServerMetricsRecorder INSTANCE = new MicrometerHttpServerMetricsRecorder();
+	private final static String HTTP = "http";
```
simonbasle, line 46: maybe rename to `PROTOCOL_VALUE_HTTP` ? consider also adding a `PROTOCOL` tag name constant.

simonbasle (changes requested):

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelMetricsRecorder.java
simonbasle: indeed, maybe 2 methods would make sense (especially given that amount is either -1 or 1). So I suggest:

```java
recordConnectionOpened(SocketAddress localAddress);
recordConnectionClosed(SocketAddress localAddress);
```

see also the Http recorder below

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerMetricsRecorder.java
```diff
@@ -54,4 +54,14 @@
...
+
+	/**
+	 * Updates the number of active in-use http connections
+	 *
+	 * @param uri the requested URI
+	 * @param method the HTTP method
+	 * @param amount the number of connections to add in the meter
+	 * @since 1.0.15
+	 */
+	default void incrementActiveConnections(String uri, String method, int amount) { }
```
simonbasle, line 66: same as in tcp, split into 2 methods

```java
recordConnectionActive(String uri, String method);
recordConnectionInactive(String uri, String method);
```

regarding `uri`, as stated in previous review, this seems to be more of a constant `http` closer to a protocol. I don't think we need/want to keep track of number of active connections per uri. Tagging the method might make sense (ie. look up how many active PUT requests are being served).

simonbasle (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
simonbasle: note that if this was actually an URI, there would again be a potential for a high number of variations. not as much as the remote address, but still in some case there can be a huge number of URIs served by the server.

but in any case, this doesn't seem to be used as uri, more as a protocol tag ✅

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerMetricsRecorder.java
simonbasle: disregard that comment about `uri`. For active connections, it is indeed correctly passed (and the `tag mapper` can be used to reduce the cardinality) 👍

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
simonbasle: oh I see... the URI tag is set to the protocol via the constructor and meter builders, for cases where the actual URI is unknown / non-applicable.

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -37,9 +43,24 @@
...
 
 	private MicrometerHttpServerMetricsRecorder() {
-		super(HTTP_SERVER_PREFIX, "http");
+		super(HTTP_SERVER_PREFIX, HTTP);
+
+		this.activeConnectionsBuilder =
+				Counter.builder(reactor.netty.Metrics.HTTP_SERVER_PREFIX + ACTIVE_CONNECTIONS)
+						.description("The number of http connections currently processing requests");
+
+		this.connectionsBuilder =
```
simonbasle, line 61: this one is not needed. `super(HTTP_SERVER_PREFIX, HTTP)` creates a counter builder with http both in the name and as a `URI` tag. actually, I don't think you need to override `incrementServerConnections` thanks to that fact

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
simonbasle: disregard these comments. the common API with tcp base doesn't allow passing of an URI and so the URI tag doesn't make sense here. our meters use `URI = protocol` as a convention when the uri is unknown hence this tag.

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -37,9 +43,24 @@
 final class MicrometerHttpServerMetricsRecorder extends MicrometerHttpMetricsRecorder implements HttpServerMetricsRecorder {
 
 	final static MicrometerHttpServerMetricsRecorder INSTANCE = new MicrometerHttpServerMetricsRecorder();
+	private final static String HTTP = "http";
+
+	protected final Counter.Builder activeConnectionsBuilder;
+	protected final ConcurrentMap<MeterKey, Counter> activeConnectionsCache = new ConcurrentHashMap<>();
+
+	protected final Counter.Builder connectionsBuilder;
+	protected final ConcurrentMap<MeterKey, Counter> connectionsCache = new ConcurrentHashMap<>();
```
simonbasle, line 52: these are redundant with the similar ones in super class and can be removed
```suggestion
```

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -37,9 +43,24 @@
...
-		super(HTTP_SERVER_PREFIX, "http");
+		super(HTTP_SERVER_PREFIX, HTTP);
+
+		this.activeConnectionsBuilder =
+				Counter.builder(reactor.netty.Metrics.HTTP_SERVER_PREFIX + ACTIVE_CONNECTIONS)
+						.description("The number of http connections currently processing requests");
+
+		this.connectionsBuilder =
+				Counter.builder(HTTP_SERVER_PREFIX + TOTAL_CONNECTIONS)
+						.description("The number of all opened connections");
```
simonbasle, line 63: this one is not needed. `super(HTTP_SERVER_PREFIX, HTTP)` creates a counter builder with http both in the name and as a `URI` tag. actually, I don't think you need to override `incrementServerConnections` thanks to that fact

```suggestion
```

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -114,6 +135,32 @@ public void incrementErrorsCount(SocketAddress remoteAddress, String uri) {
...
+		String address = Metrics.formatSocketAddress(localAddress);
+		MeterKey meterKey = new MeterKey(null, address, null, null);
+		Counter connections = connectionsCache.get(meterKey);
+		connections = connections != null ? connections : connectionsCache.computeIfAbsent(meterKey,
+						key -> filter(connectionsBuilder.tags(URI, HTTP, LOCAL_ADDRESS, address).register(REGISTRY)));
+
+		if (connections != null) {
+			connections.increment(amount);
+		}
+	}
```
simonbasle, line 162: this override is not needed. the super method already does the same thing, since `super(HTTP_SERVER_PREFIX, HTTP)` correctly configures the meter builder name and URI tag. as a result, `connectionsCache` and `connectionsBuilder` are also redundant

```suggestion
```

simonbasle (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
pderop: good catch !! 
thanks for the review, I'll work on applying all this.

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
```diff
@@ -142,11 +142,21 @@
 	 */
 	public static final String TOTAL_CONNECTIONS = ".total.connections";
```
simonbasle, line 143: these are now unused, right?

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
```diff
@@ -142,11 +142,21 @@
...
 
+	/**
+	 * The number of all connections, active or idle
+	 */
+	public static final String CONNECTIONS_TOTAL = ".connections.total";
+
 	/**
 	 * The number of the connections that have been successfully acquired and are in active use
 	 */
 	public static final String ACTIVE_CONNECTIONS = ".active.connections";
```
simonbasle, line 153: same here, unused constant ?

simonbasle (changes requested):
excellent, I think that addresses everything. I might be mistaken but old metric names constants are still there and should probably be removed, but that's it

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
violetagg: These are used by the `ConnectionProvider` and cannot be changes as this is a backwards incompatible change.

simonbasle (approved):

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
```diff
@@ -142,11 +142,21 @@
...
 
+	/**
+	 * The number of the connections that have been successfully acquired and are in active use
+	 */
```
violetagg, line 157: the java doc needs correction

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
```diff
@@ -142,11 +142,21 @@
 	 */
 	public static final String TOTAL_CONNECTIONS = ".total.connections";
 
+	/**
+	 * The number of all connections, active or idle
+	 */
```
violetagg, line 147: the java doc needs correction

reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
```diff
@@ -43,6 +43,18 @@ protected AbstractChannelMetricsHandler(@Nullable SocketAddress remoteAddress, b
 		this.onServer = onServer;
 	}
 
+	@Override
+	public void channelActive(ChannelHandlerContext ctx) throws Exception {
```
violetagg, line 47:
```suggestion
	public void channelActive(ChannelHandlerContext ctx) {
```

reactor-netty-core/src/main/java/reactor/netty/channel/MicrometerChannelMetricsRecorder.java
```diff
@@ -162,6 +169,22 @@ public void recordResolveAddressTime(SocketAddress remoteAddress, Duration time,
 		}
 	}
 
+	@Override
+	public void recordServerConnectionOpened(SocketAddress serverAddr) {
+		LongAdder totalConnectionAdder = getTotalConnectionsAdder(serverAddr);
+		if (totalConnectionAdder != null) {
+			totalConnectionAdder.add(1);
```
violetagg, line 176: nit: `increment` instead of `add(1)`

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -133,6 +142,22 @@ public void incrementErrorsCount(SocketAddress remoteAddress, String uri) {
 		}
 	}
 
+	@Override
+	public void recordServerConnectionActive(String uri, String method) {
+		LongAdder adder = getServerConnectionAdder(uri, method);
+		if (adder != null) {
+			adder.add(1);
```
violetagg, line 149: the same as channel recorder

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -133,6 +142,22 @@ public void incrementErrorsCount(SocketAddress remoteAddress, String uri) {
...
+		}
+	}
+
+	@Override
+	public void recordServerConnectionInactive(String uri, String method) {
+		LongAdder adder = getServerConnectionAdder(uri, method);
+		if (adder != null) {
+			adder.add(-1);
```
violetagg, line 157: the same as channel recorder

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerMetricsRecorder.java
```diff
@@ -54,4 +54,22 @@
...
 	void recordResponseTime(String uri, String method, String status, Duration time);
+
+	/**
+	 * Record a new active in-use http connection
+	 *
+	 * @param uri the requested URI
+	 * @param method the HTTP method
+	 * @since 1.0.15
+	 */
+	default void recordServerConnectionActive(String uri, String method) { }
```
simonbasle, line 65: as discussed off-band, let's avoid tagging with uri/method, and align input parameters with `recordServiceConnectionOpened|Closed`: only provide `SocketAddress localAddress`.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerMetricsRecorder.java
```diff
@@ -54,4 +54,22 @@
...
+
+	/**
+	 * Record an inactive http connection
+	 *
+	 * @param uri the requested URI
+	 * @param method the HTTP method
+	 * @since 1.0.15
+	 */
+	default void recordServerConnectionInactive(String uri, String method) { }
```
simonbasle, line 74: same as above: align input parameters with `recordServiceConnectionOpened|Closed`: only provide `SocketAddress localAddress`.

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -162,4 +187,20 @@ public void recordConnectTime(SocketAddress remoteAddress, Duration time, String
 	public void recordResolveAddressTime(SocketAddress remoteAddress, Duration time, String status) {
 		throw new UnsupportedOperationException();
 	}
+
+	private LongAdder getServerConnectionAdder(String uri, String method) {
```
simonbasle, line 191: transitively, this should also only have a single input parameter `SocketAddress localAddress`.

reactor-netty-http/src/main/java/reactor/netty/http/server/MicrometerHttpServerMetricsRecorder.java
```diff
@@ -162,4 +187,20 @@ public void recordConnectTime(SocketAddress remoteAddress, Duration time, String
...
+		MeterKey meterKey = new MeterKey(uri, null, method, null);
+		LongAdder adder = activeConnectionsCache.get(meterKey);
+		adder = adder != null ? adder : activeConnectionsCache.computeIfAbsent(meterKey,
+				key -> {
+					Gauge gauge = filter(Gauge.builder(reactor.netty.Metrics.HTTP_SERVER_PREFIX + CONNECTIONS_ACTIVE,
+							activeConnectionsAdder, LongAdder::longValue)
+							.tags(URI, uri, METHOD, method)
```
simonbasle, line 198: the tags would become:
`.tags(URI, PROTOCOL_VALUE_HTTP, LOCAL_ADDRESS, localAddress)`

simonbasle: NB: amend the tests accordingly

simonbasle (changes requested):
the active connection metric tags and recorder signatures should be changed

pderop:
thanks for the review, applied feedbacks. 
have reverted the import reorganise that I did in the TcpMetricsTests class.

violetagg (approved):

PR #1976, opened 2022-01-11
Additional read operation is needed when websocket handshake

violetagg:
Additional read operations are needed on the server and the client when websocket handshake
is performed, this is a result of behaviour change in Netty https://github.com/netty/netty/pull/11966


OlegDokuka (approved):

simonbasle (approved):

violetagg:
@simonbasle @OlegDokuka Thanks for the review!

PR #1989, opened 2022-01-18
fixes #1940: adds support to basic auth (username and password) to pr…

### LINKED ISSUE #1940: Add support to authentification in http proxy configuration created from system properties
When using HttpClient.proxyWithSystemProperties, it is not currently possible to define a proxy configuration to a proxy server that requires authentification.

## Motivation

I'm running a program that needs to make http calls through a corporate proxy. That proxy requires basic authentification (username/password). All the proxy settings that I need to use (host, port, username, password and non-proxy hosts) are defined as environment variables on the system on which I run my program. I would like to be able to define all those settings as java system properties (using -D options on the command line) and then use HttpClient.proxyWithSystemProperties to configure the proxy in my http client.

## Desired solution

We should add support to the following system properties:
- http.proxyUser
- http.proxyPassword
- https.proxyUser
- https.proxyPassword

into ProxyProvider.createHttpProxyFrom(Properties).
If these properties are defined, then we should add calls to ProxyProvider.Builder.username and ProxyProvider.Builder.password.

## Considered alternatives

Another way of achieving the same result would be to define a custom proxy configuration using HttpClient.proxy(), and then inject the relevant system properties into this configuration. This solution somehow defeats the purpose of HttpClient.proxyWithSystemProperties() though.

## Additional context

The system properties related to http proxy authentication are not listed in the [Oracle docs](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html), but they are often mentioned in topics related to http proxy authentication, for instance https://stackoverflow.com/questions/1626549/authenticated-http-proxy-with-java

sebphil:
…oxy providers created from system properties

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
@@ -304,7 +304,8 @@ public final class ProxyProvider {
 
 			if (properties.containsKey(passwordProperty)) {
 				proxy = proxy.password(u -> properties.getProperty(passwordProperty));
-			} else {
+			}
+			else {
 				throw new NullPointerException("Proxy username is set via '" + userProperty + "', but '" + passwordProperty + "' is not set.");
 			}
 		}
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
@@ -232,8 +232,12 @@ public final class ProxyProvider {
 
 	static final String HTTP_PROXY_HOST = "http.proxyHost";
 	static final String HTTP_PROXY_PORT = "http.proxyPort";
+	static final String HTTP_PROXY_USER = "http.proxyUser";
+	static final String HTTP_PROXY_PASSWORD = "http.proxyPassword";
 	static final String HTTPS_PROXY_HOST = "https.proxyHost";
 	static final String HTTPS_PROXY_PORT = "https.proxyPort";
+	static final String HTTPS_PROXY_USER = "https.proxyUser";
+	static final String HTTPS_PROXY_PASSWORD = "https.proxyPassword";
 	static final String HTTP_NON_PROXY_HOSTS = "http.nonProxyHosts";
 	static final String DEFAULT_NON_PROXY_HOSTS = "localhost|127.*|[::1]";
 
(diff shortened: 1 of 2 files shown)


reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -265,28 +269,39 @@ static ProxyProvider createFrom(Properties properties) {
...
+			userProperty = HTTP_PROXY_USER;
+			passwordProperty = HTTP_PROXY_PASSWORD;
 			defaultPort = "80";
 		}
 
 		String hostname = Objects.requireNonNull(properties.getProperty(hostProperty), hostProperty);
 		int port = parsePort(properties.getProperty(portProperty, defaultPort), portProperty);
 
+		String username = properties.getProperty(userProperty);
+		String password = properties.getProperty(passwordProperty);
```
simonbasle, line 294: I would check that the password is non-null IF AND ONLY IF the user property is set. If username non-null but password null, fail with an NPE with a message like `"Proxy username is set via " + userProperty + ", but no password is set".

reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -265,28 +269,39 @@ static ProxyProvider createFrom(Properties properties) {
...
+
 		String nonProxyHosts = properties.getProperty(HTTP_NON_PROXY_HOSTS, DEFAULT_NON_PROXY_HOSTS);
 		RegexShouldProxyPredicate transformedNonProxyHosts = RegexShouldProxyPredicate.fromWildcardedPattern(nonProxyHosts);
 
 		return ProxyProvider.builder()
 				.type(ProxyProvider.Proxy.HTTP)
 				.host(hostname)
 				.port(port)
+				.username(username)
```
simonbasle, line 303: here I think we need to conditionally invoke `username` and `password` methods? (the same way it is done for the SOCKS proxy configuration below)

reactor-netty-core/src/test/java/reactor/netty/transport/ProxyProviderTest.java
```diff
@@ -368,6 +369,42 @@ void proxyFromSystemProperties_customNonProxyHostsWithWildcardSetForHttpsProxy()
...
+		assertThat(provider).isNotNull();
+
+		ProxyHandler handler = provider.newProxyHandler();
+		assertThat(handler.getClass()).isEqualTo(HttpProxyHandler.class);
+
+		HttpProxyHandler httpHandler = (HttpProxyHandler) handler;
+		assertThat(httpHandler.username()).isEqualTo("user");
+		assertThat(httpHandler.password()).isEqualTo("password");
+	}
+
```
simonbasle, line 407: add two tests to check the NPE reflects the property used (HTTP vs HTTPS) + at least one test to check what happens when `username(null)` is invoked?

simonbasle (changes requested):
this is looking good, thanks for the PR @sebphil !
a few improvements to deal with partially set username/password pairs and it should be good to go 👍

simonbasle:
ah, and all touched java files should have their license header updated @sebphil. This is most easily done by running `./gradlew spotlessApply` 🙇

reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
sebphil: I thought about doing that, but it appears that the username and password methods are just "setters", so passing null is just a no-op and doesn't break anything. If you think this behavior is subject to change, it might be worthwile to add a nullity check. Please let me know what you think is best.

reactor-netty-core/src/test/java/reactor/netty/transport/ProxyProviderTest.java
sebphil: I added tests for the NPE cases. I don't think it's useful to add specific tests for when username is null, as it's already the case in all other tests.

reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
sebphil: I refactored my code to be more in line with the SOCKS part

simonbasle (approved):
looks good to me. @violetagg you can quickly go through that one and merge at your convenience

violetagg (approved):

reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -283,12 +293,23 @@ static ProxyProvider createHttpProxyFrom(Properties properties) {
...
+				.nonProxyHostsPredicate(transformedNonProxyHosts);
+
+		if (properties.containsKey(userProperty)) {
+			proxy = proxy.username(properties.getProperty(userProperty));
+
+			if (properties.containsKey(passwordProperty)) {
+				proxy = proxy.password(u -> properties.getProperty(passwordProperty));
+			} else {
+				throw new NullPointerException("Proxy username is set via '" + userProperty + "', but '" + passwordProperty + "' is not set.");
+			}
```
violetagg, line 309:
```suggestion
			}
			else {
				throw new NullPointerException("Proxy username is set via '" + userProperty + "', but '" + passwordProperty + "' is not set.");
			}
```

violetagg:
@sebphil Thanks for the review!

sebphil:
@violetagg thanks for the merge (and the additional style commit)!

PR #1996, opened 2022-01-25
[build] Hardening of GitHub actions

violetagg:
This commit makes several modifications to our workflows to make them
more secure:

- explicitly have GITHUB_TOKEN read-only by default
- split tagging steps in their own jobs, limiting the scope of needed
write permission to these short-lived jobs

### DIFF
--- a/.github/workflows/check_netty_snapshots.yml
+++ b/.github/workflows/check_netty_snapshots.yml
@@ -3,7 +3,7 @@ name: Check Netty SNAPSHOTS
 on:
   schedule:
     - cron: "0 14 * * *"
-
+permissions: read-all
 jobs:
   build:
 
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -19,6 +19,10 @@ jobs:
     name: Analyze
     runs-on: ubuntu-latest
 
+    permissions:
+      # required for all workflows
+      security-events: write
+
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,7 +13,7 @@ on:
     branches: [ main ]
   schedule:
     - cron: '0 12 * * 6'
-
+permissions: read-all
 jobs:
   analyze:
     name: Analyze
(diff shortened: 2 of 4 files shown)


.github/workflows/check_netty_snapshots.yml
```diff
@@ -3,7 +3,7 @@ name: Check Netty SNAPSHOTS
 on:
   schedule:
     - cron: "0 14 * * *"
-
+permissions: read-all
```
simonbasle, line 6: this could also be enabled on `codeql` workflow. that said, ultimately the org-wide setting will probably be used on top of explicit read-only `permissions` clauses

violetagg: mmm for CodeQL we need `security-events: write`
see https://github.com/github/codeql-action#usage

simonbasle: right. that doesn't seem too bad and probably better than the (current) default of write-all so I'd say go for it

violetagg: ok we have now

```
GITHUB_TOKEN Permissions
  Metadata: read
  SecurityEvents: write
```

https://github.com/reactor/reactor-netty/runs/4940728231?check_suite_focus=true#step:1:13

violetagg:
@simonbasle PTAL

.github/workflows/codeql-analysis.yml
```diff
@@ -19,6 +19,10 @@ jobs:
     name: Analyze
     runs-on: ubuntu-latest
 
+    permissions:
+      # required for all workflows
+      security-events: write
```
simonbasle, line 24: I don't think that will work without a `permissions: read-all` at top level, will it ?
(if any specific permission is set, anything that is unset defaults to `none`).

I guess this doesn't come up in the PR's checks because CodeQL is only activated on main branch 🤔

simonbasle (changes requested):

.github/workflows/codeql-analysis.yml
violetagg: I added a top level permissions

simonbasle (approved):

violetagg:

PR #1998, opened 2022-01-25
HttpServerOperations#isFormUrlencoded handles header value with charset

### LINKED ISSUE #1988: HttpServerRequest.receiveForm() can't handle Content-Type application/x-www-form-urlencoded;charset=UTF-8
`HttpServerRequest.receiveForm()` uses `HttpServerOperations.isFormUrlencoded()` to check if the content is `application/x-www-form-urlencoded` but uses an exact match eg: `HttpHeaderValues.APPLICATION_X_WWW_FORM_URLENCODED.contentEqualsIgnoreCase(contentType)` but i have a web page attaching charset to the content type (`;charset=UTF-8`) and it throws an exception instead:

```java.lang.IllegalStateException: Request is not POST or does not have Content-Type with value 'application/x-www-form-urlencoded' or 'multipart/form-data'```

for `multipart/form-data` nettys `HttpPostRequestDecoder.isMultipart(nettyRequest)` is used instead which only checks if the Content-Type starts with `multipart/form-data`

## Possible Solution

Make `HttpServerOperations.isFormUrlencoded()` use `startsWith` instead of `contentEqualsIgnoreCase` 

* Reactor version(s) used: 4.1.70
* Other relevant libraries versions (eg. `netty`, ...): netty 1.0.13
* JVM version (`java -version`): java 11
* OS and version (eg. `uname -a`): windows 11

### PR DESCRIPTION
**Pull request by violetagg**
Fixes #1988

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2011-2022 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -295,7 +295,8 @@ class HttpServerOperations extends HttpOperations<HttpServerRequest, HttpServerR
 	@Override
 	public boolean isFormUrlencoded() {
 		String contentType = requestHeaders().get(HttpHeaderNames.CONTENT_TYPE);
-		return HttpHeaderValues.APPLICATION_X_WWW_FORM_URLENCODED.contentEqualsIgnoreCase(contentType);
+		return contentType != null &&
+				contentType.startsWith(HttpHeaderValues.APPLICATION_X_WWW_FORM_URLENCODED.toString());
 	}
 
 	@Override
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
@@ -294,9 +294,9 @@ class HttpServerOperations extends HttpOperations<HttpServerRequest, HttpServerR
 
 	@Override
 	public boolean isFormUrlencoded() {
-		String contentType = requestHeaders().get(HttpHeaderNames.CONTENT_TYPE);
-		return contentType != null &&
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java [hunk 1]
```diff
@@ -295,7 +295,8 @@ public HttpServerResponse headers(HttpHeaders headers) {
 	@Override
 	public boolean isFormUrlencoded() {
 		String contentType = requestHeaders().get(HttpHeaderNames.CONTENT_TYPE);
-		return HttpHeaderValues.APPLICATION_X_WWW_FORM_URLENCODED.contentEqualsIgnoreCase(contentType);
+		return contentType != null &&
+				contentType.startsWith(HttpHeaderValues.APPLICATION_X_WWW_FORM_URLENCODED.toString());
```
simonbasle, line 299: I would suggest avoiding `startsWith` and instead maybe introducing some sort of headers sanitization or comparison util which removes/ignores directives... as you said offband, the `;` can be separated by whitespace so it is not a simple matter of checking the n-th character, unfortunately :(

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2717,4 +2717,40 @@ void testMaxKeepAliveRequestsBadValues(int maxKeepAliveRequests) {
...
+		doTestIsFormUrlencoded("application/x-www-form-urlencoded;charset=UTF-8", true);
+	}
+
+	@Test
+	void testIsFormUrlencodedWithoutCharset() {
+		doTestIsFormUrlencoded("application/x-www-form-urlencoded", true);
+	}
+
+	@Test
+	void testIsNotFormUrlencoded() {
```
simonbasle, line 2732: maybe add a test that `application/x-www-form-urlencoded-bad` is not `isFormUrlencoded`?

simonbasle:
probably minor but we need to avoid false positives on top of avoiding the false negative of the original issue :)

simonbasle (changes requested):
probably minor but we need to avoid false positives on top of the original false negative :)

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
violetagg: good catch, let me fix that

violetagg: please check now

simonbasle (approved):
Nice thinking of using the getMimetype util!

violetagg:

PR #2002, opened 2022-01-26
Ensure the correct Context is provided to ContextAwareHttpServerMetricsRecorder

### LINKED ISSUE #1854: ContextView in default implementations before ContextAwareHttpServerMetricsRecorder

violetagg:
Fixes #1854

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2011-2022 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -49,6 +49,7 @@ import reactor.netty.transport.ServerTransport;
 import reactor.util.Logger;
 import reactor.util.Loggers;
 import reactor.util.Metrics;
+import reactor.util.context.Context;
 
 import static reactor.netty.ReactorNetty.format;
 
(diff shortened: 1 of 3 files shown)


pderop (approved):
LGTM

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -955,7 +956,9 @@ public void onStateChange(Connection connection, State newState) {
 						log.debug(format(connection.channel(), "Handler is being applied: {}"), handler);
 					}
 					HttpServerOperations ops = (HttpServerOperations) connection;
-					Mono<Void> mono = Mono.fromDirect(handler.apply(ops, ops));
+					Mono<Void> mono = Mono.deferContextual(Mono::just)
+					                      .doOnNext(ctx -> ops.currentContext = Context.of(ctx))
+					                      .then(Mono.fromDirect(handler.apply(ops, ops)));
```
simonbasle, line 961: I would avoid extra couple operators (doOnNext, then) by setting the `ops.currentContext` and returning the `fromDirect` inside the `deferContextual` lambda.

simonbasle (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
simonbasle: as discussed, this suggestion doesn't really work because the `catch` clause below depends on `handler#apply` throwing at assembly time. applying the function outside of `deferContextual` should work though.

violetagg: fixed

violetagg:

PR #2005, opened 2022-01-27
Remove japicmp exclusions

violetagg:
When we release a version there is no need to keep this exclusions.
The backwards compatibility is checked always against the previous
version and not againts GA version.


simonbasle (approved):

pderop (approved):
LGTM

violetagg:

PR #2014, opened 2022-02-02
Use the correct owner when invalidating the stream on close

### LINKED ISSUE #1982: HTTP2 - Active Connections and Streams Leak when maxLifeTime is set for the ConnectionProvider
The netty reactor does not shut down/reduce the active connections/streams when the max lifetime is reached. Hence, The active connections and active streams leak. Please note that the maxLifeTime and evictionInbackground are set. 
It looks like MaxLifeTime is not respected.

## Expected Behavior
1-  When there is no request, we should see the active streams and active connection go back to 0.
2- When there is a constant request, we should see valid # active connections and streams.

## Actual Behavior
1- When there is no constant request, we see that no effort has been made by netty to shut down the active connections and streams gracefully after the maxLifeTime is reached. They remain in a bad state (i.e active_connection: 17, active streams: 40)
2- When there is a constant request, we see that active connections and streams keep going up. Although the leak is more visible for active streams.

## Steps to Reproduce

1-  checkout `h2ConnAndStremLeak` branch from [sia-h2](https://github.com/siavashta/sia-h2) repo
```
git clone -b h2ConnAndStremLeak git@github.com:siavashta/sia-h2.git
```
2- run the server and client
```
siaserver % mvn spring-boot:run
siaclient % mvn spring-boot:run
```
3- open JMeter and send 200reqs/sec(Number of Threads) with 3 times (Loop Count)
```
stahmureszadeh@TOR-MBP-7724 sia-h2 % jmeter -t siaclient/client.jmx
```
please note that maxLifeTime is set to 20 secs and evictInBackground set 5 secs.
```
        var provider = ConnectionProvider.builder("sia")
                .fifo()
                .maxConnections(100)
                .maxLifeTime(Duration.ofSeconds(20))
                .evictInBackground(Duration.ofSeconds(5))
                .metrics(true)
                .build();
 
```
4- open prometheus metrics page after 2-3mins (http://localhost:9066/prometheus) or check the logs and look for `active_connections` and `active_streams`.
```
reactor_netty_connection_provider_active_streams{id="1624003311",name="http2.sia",remote_address="localhost:9192",} 131.0
reactor_netty_connection_provider_total_connections{id="1624003311",name="sia",remote_address="localhost:9192",} 26.0
```
**We expected maxLifetime to kick in and reset the ActiveConnections and ActiveStrems to 0, but they remained at the mentioned numbers without any sign of recovery**

if you keep the Jmeter sending the constant number of requests (set Loop count to infinite), you will see the active_connections, and specifically `active_streams` will constantly increase without any recovery.
i.e 200 reqs/sec should be handled by around 10 connections each accepting 20 streams (set by server).
Math: 
10 conns * 20 streams 200 reqs. 
So, we expect to handle 200req/sec with active conns around 10-20 and streams around 20-50 all the time, but active streams keep going up forever. after 30 mins we saw the active stream reached 150

violetagg:
Both `Http2FrameCodec` and `Http2MultiplexHandler` has to be available
in the pipeline when checking whether max active streams limit has been reached.
In case of `H2C` upgrade these handlers are added only on a successful upgrade.

Fixes #1982

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
@@ -135,8 +135,7 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 				localEndpoint.maxActiveStreams());
 	}
 
-	static void registerClose(Channel channel) {
-		ConnectionObserver owner = channel.parent().attr(OWNER).get();
+	static void registerClose(Channel channel, ConnectionObserver owner) {
 		channel.closeFuture()
 		       .addListener(f -> {
 		           Channel parent = channel.parent();
@@ -281,7 +280,7 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 				return;
 			}
 
-			HttpClientConfig.openStream(channel, obs, opsFactory, acceptGzip)
+			HttpClientConfig.openStream(channel, this, obs, opsFactory, acceptGzip)
 			                .addListener(this);
 		}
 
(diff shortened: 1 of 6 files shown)


simonbasle (approved):
LGTM

pderop (approved):
LGTM

Issue #1978, opened 2022-01-12
memory leak if webfiler return response,  when client send headframe and dataframe in separated TCP packets


tigersean:
## Actual Behavior

When upgrade reacotr-netty to  > = 1.0.5,
  1.  Client send HTTP2  POST message with headframe and dataframe in separated TCP packets (like: jmeter plugin: Blazemeter-1.6.1)
![Screenshot from 2022-01-12 11-23-01](https://user-images.githubusercontent.com/7018121/149058333-5279284c-5b1a-47a7-bc9f-dca43a6b2539.png)

  2.  sprint-webflux's webfilter  return response  
  3.  some objects cannot GC (inlcude SimpleConnection, DefaultChannelHandlerContext and etc.)
![image](https://user-images.githubusercontent.com/7018121/149057010-77d1882a-d7ab-4be9-93c5-84b855af5685.png)

## Steps to Reproduce

```java

```

## Possible Solution

we rollback https://github.com/reactor/reactor-netty/pull/1514/commits/90ea1fb77f3129f2b2c71802d4afcb55d2d22333. It work

## Your Environment

* Reactor version(s) used:
* Other relevant libraries versions (eg. `netty`, ...):
* JVM version (`java -version`):
* OS and version (eg. `uname -a`):

violetagg:
@tigersean It seems that you can reproduce it. Please provide a reproducible example.

tigersean:
That is test project : https://github.com/tigersean/webflux_test  and  webfilter.jmx is jmeter scripts.

5%--10% traffic can cause memory leak

pderop:
Hi @tigersean,

Could you give a try to the 1.0.16-SNAPSHOT ? it contains a candidate fix for your issue.
(and thanks again for your reproducer project, it helped a lot !)

PR #2015, opened 2022-02-02
Additional read operation needs to be scheduled on writing LastHttpContent

violetagg:
This fix handles the use case when everything is received on the server and the response is delayed,
then additional read operation needs to be scheduled on writing `LastHttpContent`.
Extend the test to handle all compatible `H2` and `H2C` combinations.

Additional fix for #1978


pderop (approved):
nice patch ! LGTM

simonbasle (approved):

violetagg:

PR #2019, opened 2022-02-04
Updated to the latest API, documented a distribution summary

marcingrzejszczak:


violetagg:
Let me merge that and then I'll rebase on top of main

There is no need of review as this is a feature branch for experimenting with the new Micrometer API

PR #2020, opened 2022-02-04
Ensure HttpClient uses the new ConnectionProvider after HttpResources#reset

### LINKED ISSUE #1943: Calling HttpResources.reset() breaks Http2Resources
## Issue

Http2 requests via the HttpClient leverage the Http2Resources static get() interface, which on the first call atomically instantiates Http2Resources by wrapping HttpResources.get().   This all works fine until HttpResources is reset, which leaves Http2Resources holding on to a broken/disposed HttpResources instance.   In particular, the reset HttpResources object holds on to a disposed AddressResolverGroup.

## Possible Solution

An easy fix would be to add a public static reset() on Http2Resources and broaden the accessibility of Http2Resources, but this might not be the cleanest approach.  Alternatives welcome.

violetagg:
Fixes #1943


pderop (approved):
LGTM

simonbasle (approved):

violetagg:

PR #2023, opened 2022-02-07
Run spotless/japicmp as a preliminary job in ci

simonbasle:
This commit changes the `check_transport.yml` workflow so that it runs a
first job to check spotless and japicmp. The test matrix, which run longer,
depend on that first job succeeding.

We output a few tips to fix spotless failures and to indicate japicmp
failures need to be discussed with the maintainers, as pointers intended
for contributors.

### DIFF
--- a/.github/workflows/check_transport.yml
+++ b/.github/workflows/check_transport.yml
@@ -4,8 +4,36 @@ on:
   pull_request: {}
 permissions: read-all
 jobs:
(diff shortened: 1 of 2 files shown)


simonbasle:
similar to https://github.com/reactor/reactor-core/pull/2895

violetagg (approved):

PR #2034, opened 2022-02-16
Upgrade to Gradle v7.4

violetagg:
According to the documentation the command below should be used for upgrade
`./gradlew wrapper --gradle-version=7.4`


simonbasle (changes requested):
I'd expect the `gradle-wrapper.jar` to be part of this upgrade (see for instance reactor/reactor-core#2922 which also bumped from 7.3.3)

violetagg:
I used `./gradlew wrapper --gradle-version=7.4` and no jar update :(

simonbasle (approved):

PR #2035, opened 2022-02-16
Adapt the workflows to run against 1.0.x branch

violetagg:


simonbasle:
@violetagg LGTM, understanding that we don't want to produce snapshots just yet for the `main` branch, correct? not until the `version` is changed to `1.1.0-SNAPSHOT` at least

violetagg:
> @violetagg LGTM, understanding that we don't want to produce snapshots just yet for the `main` branch, correct? not until the `version` is changed to `1.1.0-SNAPSHOT` at least

I'm gonna change the version in main in a while :)

pderop (approved):
LGTM

build.gradle
```diff
@@ -154,7 +154,7 @@ spotless {
 		enforceCheck false
 	}
 	else {
-		String spotlessBranch = "origin/main"
+		String spotlessBranch = "origin/1.0.x"
```
simonbasle, line 157: NB: this one will need to be brought back to `main` once forward-merged

simonbasle (approved):

violetagg:

Issue #1952, opened 2021-12-21
Integration with the new Micrometer 2.0.x Observation API


violetagg:
https://github.com/micrometer-metrics/micrometer/wiki/Migrating-to-new-2.0.x-Observation-API
https://micrometer.io/docs/tracing

violetagg:
Context propagation is still in progress #2051

PR #2041, opened 2022-02-17
Integrated with Micrometer Docs Generator

violetagg:
Use DocumentedMeter for describing Reactor Netty metrics
and add gradle task for generating documentation for these metrics.

Related to #1952

Co-authored-by: Marcin Grzejszczak <mgrzejszczak@vmware.com>

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelMeters.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelMeters.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright (c) 2022 VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty.channel;
+
+import io.micrometer.api.instrument.Meter;
+import io.micrometer.api.instrument.docs.DocumentedMeter;
+import io.micrometer.api.instrument.docs.TagKey;
+
+/**
+ * Channel meters.
+ *
+ * @author Marcin Grzejszczak
+ * @author Violeta Georgieva
+ * @since 1.1.0
+ */
+public enum ChannelMeters implements DocumentedMeter {
+
+	/**
+	 * The number of all opened connections on the server.
+	 */
+	CONNECTIONS_TOTAL {
+		@Override
+		public String getName() {
+			return "%s";
+		}
+
+		@Override
+		public TagKey[] getTagKeys() {
+			return ConnectionsTotalMeterTags.values();
+		}
+
+		@Override
+		public Meter.Type getType() {
+			return Meter.Type.GAUGE;
+		}
+	},
+
+	/**
+	 * Amount of the data received, in bytes.
+	 */
+	DATA_RECEIVED {
+		@Override
+		public String getBaseUnit() {
+			return "bytes";
+		}
+
+		@Override
+		public String getName() {
+			return "%s";
+		}
+
+		@Override
+		public TagKey[] getTagKeys() {
+			return ChannelMetersTags.values();
+		}
+
+		@Override
+		public Meter.Type getType() {
+			return Meter.Type.DISTRIBUTION_SUMMARY;
+		}
+	},
+
+	/**
+	 * Amount of the data sent, in bytes.
+	 */
+	DATA_SENT {
+		@Override
+		public String getBaseUnit() {
+			return "bytes";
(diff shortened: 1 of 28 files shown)


pderop (changes requested):
Please correct Copyright year in gradle/asciidoc.gradle

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelMeters.java
```diff
@@ -0,0 +1,166 @@
...
+	 */
+	DATA_RECEIVED {
+		@Override
+		public String getBaseUnit() {
+			return "bytes";
+		}
+
+		@Override
+		public String getName() {
+			return "%s";
```
simonbasle, line 62: can you clarify the `%s`?
from what I understand:
 -  the name of the metric is either explicitly set in the enum (like in `reactor.netty.connection.provider.active.connections`)
 - OR the name of the metric is set to `"%s"` and somehow it will resolve what's configured as name in the builder? (eg in `MicrometerChannelMetricsRecorder`, still uses the `Metrics.DATA_RECEIVED` suffix + a variable `name` for prefix, resulting in eg. `reactor.netty.tcp.client.data.received`)
 - the link to the generated adoc section uses the enum's name? (`DATA_RECEIVED` => `<<observability-metrics-data-received>>`, `ERROR_COUNT` => `<<observability-metrics-error-count>>`, `ACTIVE_CONNECTIONS` => `observability-metrics-active-connections`...)

reactor-netty-core/src/main/java/reactor/netty/transport/ByteBufAllocatorMetrics.java
```diff
@@ -65,59 +55,49 @@ private ByteBufAllocatorMetrics() {
 
 	void registerMetrics(String allocType, ByteBufAllocatorMetric metrics, ByteBufAllocator alloc) {
 		MapUtils.computeIfAbsent(cache, metrics.hashCode() + "", key -> {
-			String[] tags = new String[] {ID, key, TYPE, allocType};
+			Tags tags = Tags.of(ID.getKey(), key, TYPE.getKey(), allocType);
 
-			Gauge.builder(BYTE_BUF_ALLOCATOR_PREFIX + USED_HEAP_MEMORY, metrics, ByteBufAllocatorMetric::usedHeapMemory)
-			     .description(USED_HEAP_MEMORY_DESCRIPTION)
+			Gauge.builder(USED_HEAP_MEMORY.getName(), metrics, ByteBufAllocatorMetric::usedHeapMemory)
```
simonbasle, line 60: ah I see, it's the reverse: the builder uses the enum's `getName()` unless it needs something more dynamic (in which case the enum's `getName()` doesn't really matter?)

reactor-netty-core/src/main/java/reactor/netty/channel/MicrometerChannelMetricsRecorder.java
```diff
@@ -86,9 +81,9 @@ public void recordDataReceived(SocketAddress remoteAddress, long bytes) {
 		String address = reactor.netty.Metrics.formatSocketAddress(remoteAddress);
 		DistributionSummary ds = MapUtils.computeIfAbsent(dataReceivedCache, address,
 				key -> filter(DistributionSummary.builder(name + DATA_RECEIVED)
```
simonbasle, line 83: just to be sure: if my interpretation is correct from my other two comments, then this would have been equivalent (although concatenation of constants is more concise and prob more efficient):
```suggestion
				key -> filter(DistributionSummary.builder(name + DATA_RECEIVED)
				key -> filter(DistributionSummary.builder(ChannelMeters.DATA_RECEIVED.getName(name, DATA_RECEIVED))
```

simonbasle (commented):
just a quick question to verify I understood the change correctly. LGTM so far

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelMeters.java
violetagg: When we know in advance the name we can provide it directly in the `DocumentedMeter`, for example this is the case with `reactor.netty.connection.provider.active.connections`. Then in the generated documentation we have

```
**Metric name** `reactor.netty.connection.provider.active.connections`. **Type** `gauge`.
```

When the name is dynamic in the `DocumentedMeter` we specify `%s`, then the generated documentation looks like this

```
**Metric name** `%s` - since it contains `%s`, the name is dynamic and will be resolved at runtime. **Type** `gauge`.
```

such metrics are 
`reactor.netty.tcp.client.data.received`
`reactor.netty.tcp.server.data.received`
`reactor.netty.http.client.data.received`
`reactor.netty.http.server.data.received`

Reactor Netty provides `data.reveived` metric for every server/client type.

reactor-netty-core/src/main/java/reactor/netty/transport/ByteBufAllocatorMetrics.java
violetagg: yes, when it is dynamic it doesn't matter

simonbasle (approved):

reactor-netty-core/src/main/java/reactor/netty/channel/MicrometerChannelMetricsRecorder.java
violetagg: yes exactly

violetagg:

PR #2048, opened 2022-02-22
Create the Tags once as they are the same for every Gauge

violetagg:


pderop (approved):
LGTM

violetagg:

PR #2064, opened 2022-03-01
[test] Refactor the test in order to be able to test H2/H2C variations

violetagg:


reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
```diff
@@ -584,6 +662,13 @@ void checkGauge(String name, boolean exists, double expectedCount, String... tag
 		}
 	}
 
+	static Stream<Arguments> httpCompatibleProtocols() {
+		return Stream.of(
+				Arguments.of(null, null, new HttpProtocol[]{HttpProtocol.HTTP11}, new HttpProtocol[]{HttpProtocol.HTTP11}),
```
simonbasle, line 667: I would suggest using `Named` here to give a descriptive name to each invocation

violetagg: The current configuration will produce this output, isn't it more descriptive than `Named`? We know exactly the combination that is executed. (The screenshot is from a change where I will introduce HTTP/2 combinations)
Screenshot 2022-03-01 at 12 20 24

simonbasle: yeah, I didn't remember that the array would be well represented.
maybe just for the serverCtx and clientCtx? eg.  `Named.of("server with SSL", someSslContext)` ?
but the above should be fine

violetagg: It is important to know with what kind of SslContext it is running - see sometimes it is Http11 sometimes Http2.
I can move the server/client protocols as first arguments as this is more important than actual SslContext and because the context might be null.
Wdyt?

violetagg: Such change

```
	static Stream<Arguments> httpCompatibleProtocols() {
		return Stream.of(
				Arguments.of(new HttpProtocol[]{HttpProtocol.HTTP11}, new HttpProtocol[]{HttpProtocol.HTTP11}, null, null),
				Arguments.of(new HttpProtocol[]{HttpProtocol.HTTP11}, new HttpProtocol[]{HttpProtocol.HTTP11},
						Named.of("Http11SslContextSpec", serverCtx11), Named.of("Http11SslContextSpec", clientCtx11))
		);
	}
```

Will result in the following output

Screenshot 2022-03-01 at 12 54 56

simonbasle: yeah, that looks good 👍  the non-null ssl context have a short meaningful name, which is definitely not mandatory but nice to have. I'd go with that one 😄

violetagg: fixed

pderop (approved):
LGTM

violetagg:

PR #2069, opened 2022-03-04
Update Reactor Core version to 3.5.0-SNAPSHOT

violetagg:
Update Reactor Addons version to 3.5.0-SNAPSHOT
Update Reactor Pool version to 1.0.0-SNAPSHOT


violetagg:
Rebased in order to pick up #2070 so that the build can pass.

PR #2078, opened 2022-03-07
Wrong metric value for TLS handshake time

pderop:
This PR corrects a bad value for the TLS handshake duration time metric.
Before, the handshake start time was wrongly starting when the channel was registered, but the handshake start time must corresponds to the time just before the initial hanshake message is sent out by the netty SslHandler.

To correct the problem, we now register two handlers: 

- a tls metric handler which is registered just before the netty SslHandler. The start time of the handshake starts when the tls metric handler is activated (just before the ssl handler will send the initial handler)
- an ssl read handler, which is registered just after the netty SslHandler, which is used to just control reads, as before.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelMetricsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelMetricsHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2019-2022 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


simonbasle (approved):
look good to me. just wondering if there is an opportunity to _remove_ the handshake metric handler once it has recorded the handshake? (as, if I'm not mistaken, there will be no further handshakes for the same handler)

simonbasle:
Ah, my bad

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelMetricsHandler.java
```diff
@@ -74,4 +81,30 @@ public void connect(ChannelHandlerContext ctx, SocketAddress remoteAddress,
...
 	}
+
+	static class TlsMetricsHandler extends ChannelInboundHandlerAdapter {
+		protected final ChannelMetricsRecorder recorder;
+		TlsMetricsHandler(ChannelMetricsRecorder recorder) {
+			this.recorder = recorder;
+		}
+
+		@Override
+		public void channelActive(ChannelHandlerContext ctx) throws Exception {
```
violetagg, line 92: `Exception 'java.lang.Exception' is never thrown in the method`

reactor-netty-core/src/main/java/reactor/netty/channel/ContextAwareChannelMetricsHandler.java
```diff
@@ -123,4 +128,22 @@ void recordConnectTime(ChannelHandlerContext ctx, SocketAddress address, long co
 			}
 		}
 	}
+
+	static class TlsMetricsHandler extends ChannelMetricsHandler.TlsMetricsHandler {
```
violetagg, line 132: Let's make the class `final`

violetagg (changes requested):

violetagg (approved):

PR #2082, opened 2022-03-08
[doc] Extend the example for disabling the connection pool

violetagg:
Show that with one and the same `HttpClient` instance, one can
initiate many requests that always will use a new connection.

Related to comment https://github.com/reactor/reactor-netty/issues/1828#issuecomment-1060399563


violetagg:

PR #2090, opened 2022-03-09
Optimize Http2Pool by reducing calls to CLDQ size method

pderop:
The PR provides an enhancement for the Http2Pool in order to reduce  _ConcurrentLinkedDeque.size()_ method calls, which is not a constant time operation. 

Indeed, it has been observed that the CLDQ size method may consume around 18% of CPU when there is a high load of http client requests while all streams are currently unavailable.


simonbasle (approved):

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -493,6 +505,30 @@ void recordInteractionTimestamp() {
...
+	void removePending(ConcurrentLinkedDeque<Borrower> borrowers, Borrower borrower) {
+		if (borrowers.remove(borrower)) {
+			PENDING_SIZE.decrementAndGet(this);
+		}
+	}
+
+	int addPending(ConcurrentLinkedDeque<Borrower> borrowers, Borrower borrower, boolean first) {
+		if (first) {
+			borrowers.offerFirst(borrower);
+		} else {
```
violetagg, line 526: checkstyle: else on a new row

```suggestion
		}
		else {
```

violetagg (changes requested):

violetagg (approved):

Issue #1808, opened 2021-09-14
HTTP2 - Round-robining streams across a minimum setting of connections


crankydillo:
We leverage L4 load balancers for distributing load, resiliency needs, etc.  We are thinking 2 additions (likely opt-in) to the HTTP2 pool could help us in this regard:

1. The HTTP2 pool maintains a minimum number of connections before opening additional streams.
2. As it opens additional streams, it round-robins between the connections in the HTTP2 pool.  A fancier technique where it chooses the conn with the least number of open streams is likely better, but it's unclear yet if that is critical.

Some of the above was initially discussed [here](https://gitter.im/reactor/reactor-netty?at=6127c20913ac9b6b83969047).

violetagg:
@crankydillo Can you elaborate more on this `The HTTP2 pool maintains a minimum number of connections before opening additional streams.`

crankydillo:
Currently, you only get a new connection from the underlying pool if requests are made in parallel in certain ways or if you have exhausted the max streams on a connection.  In the latter case, that can result in max streams requests all getting sent to one server that sits behind an L4 load balancer.  What we want is to specify at the client level that the some minimum number of connections should be used to round-robin the streams.  With that, we get some client-side load-balancing of requests.  We also close connections from the server and are assuming that graceful handling of that will happen in the client.

crankydillo:
> @crankydillo Can you elaborate more on this `The HTTP2 pool maintains a minimum number of connections before opening additional streams.`

If the above doesn't help, my precise definition of that prhase is:

Up to some number (the minimum) the client will open a new connection when creating a new stream as opposed to creating an additional stream on a connection that is already in the HTTP/2 pool.

crankydillo:
@violetagg There is another aspect of this that we need to get clarity on.  We are hoping that new connections will be gated by this minimum pool size until max streams is reached.

For example, if we have:

```
HttpClient client = Http.builderStuff().maxConnections(256)).build();
final int concurrency = 100;
Flux.range(0, N)  // N often, but not guaranteed to, exceed concurrency level
   .flatMap(i -> clientCall(client), concurrency)
   .subscribe();
```

[...]

PR #2091, opened 2022-03-09
[1808] For HTTP/2 Use a minimum num of socket connections..

crankydillo:
before spawning multiple streams on any given socket connection.

Just shooting for some early feedback.

 #1808

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/Http2SettingsSpec.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/Http2SettingsSpec.java
@@ -85,6 +85,14 @@ public final class Http2SettingsSpec {
 		 * @return {@code this}
 		 */
 		//Builder pushEnabled(boolean pushEnabled);
+
+		/**
+		 * Sets the {@code SETTINGS_MIN_CONNECTIONS} value.
+		 *
+		 * @param minConnections the {@code SETTINGS_MIN_CONNECTIONS} value
+		 * @return {@code this}
+		 */
+		Builder minConnections(int minConnections);
 	}
 
 	/**
@@ -157,6 +165,16 @@ public final class Http2SettingsSpec {
 		return pushEnabled;
 	}
 
+	/**
+	 * Returns the configured minimum connections to create before adding more than one stream.
+	 * Defaults to 0.
+	 *
+	 * @return TODO
+	 */
+	public int minConnections() {
+		return minConnections;
+	}
+
 	@Override
 	public boolean equals(Object o) {
 		if (this == o) {
@@ -171,7 +189,8 @@ public final class Http2SettingsSpec {
 				Objects.equals(maxConcurrentStreams, that.maxConcurrentStreams) &&
 				Objects.equals(maxFrameSize, that.maxFrameSize) &&
 				maxHeaderListSize.equals(that.maxHeaderListSize) &&
-				Objects.equals(pushEnabled, that.pushEnabled);
+				Objects.equals(pushEnabled, that.pushEnabled) &&
+				minConnections == that.minConnections;
 	}
 
 	@Override
@@ -185,6 +204,7 @@ public final class Http2SettingsSpec {
 	final Integer maxFrameSize;
 	final Long maxHeaderListSize;
 	final Boolean pushEnabled;
+	final int minConnections;
 
 	Http2SettingsSpec(Build build) {
 		Http2Settings settings = build.http2Settings;
@@ -194,10 +214,12 @@ public final class Http2SettingsSpec {
 		maxFrameSize = settings.maxFrameSize();
 		maxHeaderListSize = settings.maxHeaderListSize();
 		pushEnabled = settings.pushEnabled();
+		minConnections = build.minConnections;
 	}
 
 	static final class Build implements Builder {
 		final Http2Settings http2Settings = Http2Settings.defaultSettings();
+		private int minConnections;
 
 		@Override
 		public Http2SettingsSpec build() {
@@ -241,5 +263,12 @@ public final class Http2SettingsSpec {
 			return this;
 		}
 		*/
+
+		@Override
+		public Builder minConnections(int minConnections) {
+			this.minConnections = minConnections;
+			return this;
+		}
+
 	}
 }
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
@@ -132,18 +132,29 @@ final class Http2Pool implements InstrumentedPool<Connection>, InstrumentedPool.
 
 	long lastInteractionTimestamp;
 
-	Http2Pool(PoolConfig<Connection> poolConfig, long maxLifeTime) {
-		if (poolConfig.allocationStrategy().getPermits(0) != 0) {
-			throw new IllegalArgumentException("No support for configuring minimum number of connections");
-		}
+	/**
+	 * The minimum amount of connections to create before creating more than 1 stream
+	 * on any given connection.
+	 */
+	int minConnections;
+
+	Http2Pool(PoolConfig<Connection> poolConfig, long maxLifeTime, int minConnections) {
 		this.clock = poolConfig.clock();
 		this.connections = new ConcurrentLinkedQueue<>();
 		this.lastInteractionTimestamp = clock.millis();
 		this.maxLifeTime = maxLifeTime;
 		this.pending = new ConcurrentLinkedDeque<>();
 		this.poolConfig = poolConfig;
+		if (minConnections < 0) {
+			throw new IllegalArgumentException("HTTP/2 minimum connections must be non-negative.");
+		}
+		this.minConnections = minConnections;
 
 		recordInteractionTimestamp();
+	}
+
+	Http2Pool(PoolConfig<Connection> poolConfig, long maxLifeTime) {
+		this(poolConfig, maxLifeTime, 0);
 	}
 
 	@Override
@@ -384,10 +395,15 @@ final class Http2Pool implements InstrumentedPool<Connection>, InstrumentedPool.
 		}
 	}
 
+	// TODO think on better names for this.  Some kind of signal that this only
+	// returns a connection from the
 	@Nullable
 	Slot findConnection(ConcurrentLinkedQueue<Slot> resources) {
 		int resourcesCount = resources.size();
-		while (resourcesCount > 0) {
+		// TODO this is the KEY idea.  I'm curious on reactor-netty's team
+		// thought on this.  Completely wrong place even if we do some name changes,
+		// cleanup, etc?
+		while (resourcesCount > (minConnections == 0 ? 0 : minConnections + 1)) {
 			// There are connections in the queue
 
 			resourcesCount--;
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
@@ -444,7 +444,7 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 			this.remoteAddress = remoteAddress;
 			this.resolver = resolver;
 			this.pool = poolFactory.newPool(connectChannel(), null, DEFAULT_DESTROY_HANDLER, DEFAULT_EVICTION_PREDICATE,
-					poolConFig -> new Http2Pool(poolConFig, poolFactory.maxLifeTime()));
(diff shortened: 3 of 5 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -384,10 +395,15 @@ else if (sig.isOnError()) {
...
+	// TODO think on better names for this.  Some kind of signal that this only
+	// returns a connection from the
 	@Nullable
 	Slot findConnection(ConcurrentLinkedQueue<Slot> resources) {
 		int resourcesCount = resources.size();
-		while (resourcesCount > 0) {
+		// TODO this is the KEY idea.  I'm curious on reactor-netty's team
+		// thought on this.  Completely wrong place even if we do some name changes,
+		// cleanup, etc?
+		while (resourcesCount > (minConnections == 0 ? 0 : minConnections + 1)) {
```
crankydillo, line 406: @violetagg This is the crux of my proposal.  Definitely needs cleanup, but what do you think about this being a good place to force the creation of socket connections?  Honestly, I want to look a little higher up for this, but I am curious if you are fundamentally opposed to this:)

reactor-netty-http/src/main/java/reactor/netty/http/Http2SettingsSpec.java
```diff
@@ -241,5 +263,12 @@ public Builder pushEnabled(boolean pushEnabled) {
 			return this;
 		}
 		*/
+
+		@Override
+		public Builder minConnections(int minConnections) {
```
crankydillo, line 268: I'm a big fan of where I put this, but just going fast in the config area.  I really wanted config to be able to test different scenarios.  Otherwise, I would have just hard-coded in `Http2Pool`.

violetagg:
I was thinking more in the direction of utilising the AllocationStrategy, although I haven't looked this proposed solution in details.

reactor.pool.PoolBuilder#sizeBetween
reactor.pool.AllocationStrategies.SizeBasedAllocationStrategy

crankydillo:
My colleague will be happy to hear that.  My first, super quick attempt was based on `AllocationStrategy`, but it failed.  In hindsight, I could have had some problems with my test, as I had to make several changes to it.  I planned on looking at `allocationStrategy` again, but I'm kind of operating on my own time at this point.  I did feel badly about using `findConnection`, but I did plan on some name changes to smooth over some of that.  Regardless, if an existing algo/design solves the problem, it should be leveraged!

I may be able to squeeze another look at this tomorrow before work starts.

And of course, if you submit a PR, I will gladly just consume it:P

violetagg:
> And of course, if you submit a PR, I will gladly just consume it:P

Unfortunately I don't have anything on my side for the moment.

crankydillo:
No worries @violetagg.  It wasn't expected;)  The `sizeBetween` minimum does sound exactly like what we want.  I sure thought I had that correctly done for the (socket) pool, but I must have done something silly.  Just one more question, what would be the deadline for a PR that has a chance of getting into the 1.0.17 release?

violetagg:
> No worries @violetagg. It wasn't expected;) The `sizeBetween` minimum does sound exactly like what we want. I sure thought I had that correctly done for the (socket) pool, but I must have done something silly. Just one more question, what would be the deadline for a PR that has a chance of getting into the 1.0.17 release?

today and tomorrow

violetagg:
Just to summarise that I understand the requirement correctly:

We want to have minimum size for the cached resources in the pool. When acquiring if we are below that size instead of using a cached resource we want to take a new one so that we can again reach this minimum size of cached resources?

In addition to that if there are no at all resources in the pool (empty), we want to create only resource at the minimum size and not beyond?

crankydillo:
Ultimately, we are trying to get some HTTP/2 load spread in L4 load balancer scenarios.

> Just to summarise that I understand the requirement correctly:
> 
> We want to have minimum size for the cached resources in the pool. When acquiring if we are below that size instead of using a cached resource we want to take a new one so that we can again reach this minimum size of cached resources?

I think this is right, but I do want to speak in concrete terms.  For HTTP/2 requests, we want to have a minimum size for socket connections that will be maintained before creating additional streams on any given socket connection.  In other words, you'd never (ignoring existing HTTP/2 streams) have a socket connection with multiple streams created on it until you have reached the minimum socket connection size.

Honestly, I think we want this minimum socket connection even for HTTP/1 requests, because it would allow some startup optimizations, but I may be misremembering.  Regardless, it is not a requirement for our current HTTP/2 need.

> In addition to that if there are no at all resources in the pool (empty), we want to create only resource at the minimum size and not beyond?

For this, you are referring to maintaining a constant set of resources (whether socket connection or HTTP/2 stream)?  If so, that would be nice, but I'm not sure it's a hard requirement.

If so, I really need to talk to my team about that.  I think we are OK with that, but want to double check.  Personally, I don't think it's a _hard_ requirement for what we need.

Hope this makes sense.

/cc @anilgursel

crankydillo:
If you use reactor-pool's sizeBetween(5, 10), does that mean it will ‘always' keep the pool populated with 5 'resources’? Or will it only do it if there is demand for at least 1 resource and non exist in the pool? In other words, imagine you are pooling socket connections in reactor-netty's HTTP client and tried to use sizeBetween. If there are no requests and all existing connections get evicted due to server severing them, will the sizeBetween-created pool respond to that and create new connections even if there are no current requests? My read of the javadoc for sizeBetween suggests the answers are No/Yes (doc talks about being driven by acquisition).  Honestly, I should look at the code.  I'll try to create time for that:)

We can take this to gitter if you prefer.  However, I don't think I have many more questions like this.

violetagg:
> Ultimately, we are trying to get some HTTP/2 load spread in L4 load balancer scenarios.
> 
> > Just to summarise that I understand the requirement correctly:
> > We want to have minimum size for the cached resources in the pool. When acquiring if we are below that size instead of using a cached resource we want to take a new one so that we can again reach this minimum size of cached resources?
> 
> I think this is right, but I do want to speak in concrete terms. For HTTP/2 requests, we want to have a minimum size for socket connections that will be maintained before creating additional streams on any given socket connection. In other words, you'd never (ignoring existing HTTP/2 streams) have a socket connection with multiple streams created on it until you have reached the minimum socket connection size.

Then I don't think your PR handles that because the current Http2Pool implementation NEVER keeps connections without active streams. If a connection is without active streams it is immediately returned to the HTTP11 pool (the parent pool).
See reactor.netty.http.client.Http2Pool#destroyPoolable

Most probably you need minimum size on the HTTP11 pool + warmup so that the pool will have this minimum resources at the beginning and when Http2 pool need a connection it will be there. (this is just guessing)

See more in the reactor pool javadoc
https://projectreactor.io/docs/pool/release/api/reactor/pool/PoolBuilder.html#sizeBetween-int-int-
https://projectreactor.io/docs/pool/release/api/reactor/pool/Pool.html#warmup--

Why I do not typically recommend something like this: most of the servers, load balancer etc. have idle timeout by default. For example Tomcat has `keepAliveTimeout` (The number of milliseconds this Connector will wait for another HTTP request before closing the connection. The default value is to use the value that has been set for the connectionTimeout attribute.), so even though that you will instruct the pool to open some number of connections, they might be close pretty soon because of this idle timeout setting.

> 
> Honestly, I think we want this minimum socket connection even for HTTP/1 requests, because it would allow some startup optimizations, but I may be misremembering. Regardless, it is not a requirement for our current HTTP/2 need.
> 
> > In addition to that if there are no at all resources in the pool (empty), we want to create only resource at the minimum size and not beyond?
> 
> For this, you are referring to maintaining a constant set of resources (whether socket connection or HTTP/2 stream)? If so, that would be nice, but I'm not sure it's a hard requirement. However, I really need to talk to my team about that. I think we are OK with that, but want to double check. Personally, I don't think it's a _hard_ requirement for what we need.
> 
> Hope this makes sense.
> 
> /cc @anilgursel

crankydillo:
Sorry for updating/deleting comments.  Trying to reduce burden on you.  I hope I didn't make it worse!

violetagg:
Take a look here (This is what HTTP11 can be implemented to support very easy however note the warning from my comment above for idle timeout)

https://github.com/reactor/reactor-pool/blob/31f839bf989563412e35e35f41745f9cea91a187/src/main/java/reactor/pool/SimpleDequePool.java#L227

https://github.com/reactor/reactor-pool/blob/31f839bf989563412e35e35f41745f9cea91a187/src/main/java/reactor/pool/SimpleDequePool.java#L443

crankydillo:
> > Ultimately, we are trying to get some HTTP/2 load spread in L4 load balancer scenarios.
> > > Just to summarise that I understand the requirement correctly:
> > > We want to have minimum size for the cached resources in the pool. When acquiring if we are below that size instead of using a cached resource we want to take a new one so that we can again reach this minimum size of cached resources?
> > 
> > 
> > I think this is right, but I do want to speak in concrete terms. For HTTP/2 requests, we want to have a minimum size for socket connections that will be maintained before creating additional streams on any given socket connection. In other words, you'd never (ignoring existing HTTP/2 streams) have a socket connection with multiple streams created on it until you have reached the minimum socket connection size.
> 
> Then I don't think your PR handles that because the current Http2Pool implementation NEVER keeps connections without active streams. If a connection is without active streams it is immediately returned to the HTTP11 pool (the parent pool). See reactor.netty.http.client.Http2Pool#destroyPoolable

I made a mistake.  I don't think we have to 'maintain' the minimum size.  The key aspect is to not create 2 active streams on any given socket (HTTP11) connection before reaching a minimum number of socket connections.  That is the most important thing.

> Most probably you need minimum size on the HTTP11 pool + warmup so that the pool will have this minimum resources at the beginning and when Http2 pool need a connection it will be there. (this is just guessing)

I'll look at this.  I definitely didn't try anything with warmup.  I just hacked my way to using `sizeBetween` on the HTTP11 pool and even there I might have made a mistake.  Ultimately, I couldn't get my test to pass.

> 
> See more in the reactor pool javadoc https://projectreactor.io/docs/pool/release/api/reactor/pool/PoolBuilder.html#sizeBetween-int-int- https://projectreactor.io/docs/pool/release/api/reactor/pool/Pool.html#warmup--

Will do.  I did not not notice `warmup`.

> Why I do not typically recommend something like this: most of the servers, load balancer etc. have idle timeout by default. For example Tomcat has `keepAliveTimeout` (The number of milliseconds this Connector will wait for another HTTP request before closing the connection. The default value is to use the value that has been set for the connectionTimeout attribute.), so even though that you will instruct the pool to open some number of connections, they might be close pretty soon because of this idle timeout setting.

That will be fine.  I hate that I may have turned this towards 'maintaining a pool size at all time', that is not required.  Our servers typically will be closing connections after a period of inactivity.

> > Honestly, I think we want this minimum socket connection even for HTTP/1 requests, because it would allow some startup optimizations, but I may be misremembering. Regardless, it is not a requirement for our current HTTP/2 need.
> > > In addition to that if there are no at all resources in the pool (empty), we want to create only resource at the minimum size and not beyond?
> > 
> > 
> > For this, you are referring to maintaining a constant set of resources (whether socket connection or HTTP/2 stream)? If so, that would be nice, but I'm not sure it's a hard requirement. However, I really need to talk to my team about that. I think we are OK with that, but want to double check. Personally, I don't think it's a _hard_ requirement for what we need.
> > Hope this makes sense.
> > /cc @anilgursel

Thanks again Violeta!  We'll see if we can get a PR in today or tomorrow.

violetagg:
@crankydillo @anilgursel Do you have progress on this PR?

violetagg:
@crankydillo @anilgursel Did you discuss the exact requirements that you have for this feature request?

crankydillo:
> @crankydillo @anilgursel Did you discuss the exact requirements that you have for this feature request?

Hello @violetagg .  Sort of hate to specify this here, given we have the open issue, but the requirements are based on the desire to have some HTTP/2 request load spread when using HTTP/2 and an L4 load balancer.  We are thinking it requires these features when making HTTP/2 requests:

- Some configurable, minimum number of socket connections are created before ever creating more than one stream on any given socket connection.
- When creating streams, choosing a socket connection is done in a round-robin fashion.

While thinking about this, I had a concern about the almost opposite problem, which revolves around creating more socket connections than needed for HTTP/2 requests.  That was mentioned [here](https://github.com/reactor/reactor-netty/issues/1808#issuecomment-931626065).  Dealing with this is not technically required for the L4 load-spread feature we request above, but it does seem related.

Hope this helps.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
crankydillo: Looking at [PR](https://github.com/reactor/reactor-netty/pull/2155/files), I now understand why this is not a good place.  Honestly, I should have looked at the calling code more.  I'm still a little confused by `drainLoop`, but I'm slowly getting there:)

PR #2098, opened 2022-03-11
Ensure the Pool implementation for the test class implements the new abstract method config()

violetagg:
This is related to https://github.com/reactor/reactor-pool/pull/156


simonbasle (approved):

simonbasle:
it could also be `throw new UnsupportedOperationException("...")`, but for tests I guess this doesn't really matter.

violetagg:

PR #2119, opened 2022-03-23
[test] Do not use Thread.sleep(1000) but check the results once the last stream is closed

violetagg:


simonbasle (approved):

pderop (approved):
LGTM.

violetagg:

PR #2147, opened 2022-04-08
Fix: README.md tags

devlap2:


violetagg:
@devlap2 Can you please explain in more detail what the intent of this change is?

violetagg:
@devlap2 Looking at the change - it is removing 2 empty rows, I think we can close this PR.

PR #2155, opened 2022-04-19
Add API for configuring minimum connections for HTTP/2 connection pool

### LINKED ISSUE #1808: HTTP2 - Round-robining streams across a minimum setting of connections

violetagg:
Fixes #1808

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
@@ -80,6 +80,10 @@ import static reactor.netty.ReactorNetty.format;
  *     <li>{@link PoolMetrics#idleSize()} always returns {@code 0}.</li>
  * </ul>
  * <p>
+ * If minimum connections is specified, the cached connections with active streams will be kept at that minimum
+ * (can be the best effort). However, if the cached connections have reached max concurrent streams,
+ * then new connections will be allocated up to the maximum connections limit.
+ * <p>
  * Configurations that are not applicable
  * <ul>
  *     <li>{@link PoolConfig#destroyHandler()} - the destroy handler cannot be used as the destruction is more complex.</li>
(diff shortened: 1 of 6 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -80,6 +80,10 @@
  *     <li>{@link PoolMetrics#idleSize()} always returns {@code 0}.</li>
  * </ul>
  * <p>
+ * If minimum connections is specified, the cached connections with active streams will be kept at that minimum
+ * (can be the best effort). However, if the cached connections have reached max concurrent streams,
+ * then new connections will be allocated up to the maximum connections limit.
```
crankydillo, line 85: Hmm, so the code now will do this, right:
For any new HTTP/2 request, if the number of socket connections is below the minimum number of connections, a new socket connection will get created and the HTTP/2 stream will get created from that channel, right?

I don't really get that from the javadoc, but that may just be me.

What does the best effort part mean?  I think I'm missing that part in the code and want to make sure I'm not misunderstanding something?

crankydillo (commented):
I understand a decent bit, and it looks good to me.  Have a question about the doc though.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
violetagg: > Hmm, so the code now will do this, right: For any new HTTP/2 request, if the number of socket connections is below the minimum number of connections, a new socket connection will get created and the HTTP/2 stream will get created on that new socket connection, right?

yes

> I don't really get that from the javadoc, but that may just be me.
> 
> What does the best effort part mean? I think I'm missing that part in the code and want to make sure I'm not misunderstanding something?

As we do not have a strong synchronisation between when we return the connection and we acquire the connection we may not be able to stay exactly on the min number that is specified.

violetagg:
I need to think about the proper place for this configuration. Most probably on the level of `ConnectionProvider` as we can have connection pools per remote host.

samueldlightfoot:
Thanks for this - we are very keen to consume this feature to aid our client-side load balancing too!

violetagg:
This is superseded by #2257

A new API `Http2AllocationStrategy` is exposed so that minimum connections and maximum concurrent streams can be configured.

PR #2162, opened 2022-04-23
support dynamic ssl load

ashraf-revo:
support dynamic ssl load


violetagg:
This is superseded by #2163

PR #2168, opened 2022-04-26
Ensure PooledConnectionProvider check only for event loop in the thread local

violetagg:
`EventLoopGroup.next()` should be called only in cases it will be used
for example establishing a connection.
`Pool.acquire(...)` in most cases will reuse the connections and in
these cases `EventLoopGroup.next()` invocation is not needed.


simonbasle (approved):
LGTM 👍

pderop (approved):
LGTM

violetagg:

PR #2175, opened 2022-05-03
Allow to configure connection pool aquire timers

pderop:
This draft PR is an improvement to allow http client users to specify a hook function that can be used to configure a custom scheduler to schedule timers for pending connections acquisition.

By default, it's the **Schedulers.parallel** scheduler which is used, as before, but the intent is to let users to specify their own implementation for the timer service used by reactor-netty (timer scheduled for pending connections pool acquisition).

Example usage of a client that is configured with a netty hashed wheel timer service:
```
        HashedWheelTimer timer = new HashedWheelTimer(10, TimeUnit.MILLISECONDS, 1024);

        ConnectionProvider provider = ConnectionProvider
                .builder(POOL_NAME)
                .pendingAcquireTimeout(Duration.ofMillis(10000))
                .pendingAcquireTimer((r, d) -> {
                    Timeout t = timer.newTimeout(timeout -> r.run(), d.toNanos(), TimeUnit.NANOSECONDS);
                    return () -> t.cancel();
                })
                .build();
```

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
@@ -593,10 +593,10 @@ public interface ConnectionProvider extends Disposable {
 		}
 
 		/**
-		 * Set the function to applly when scheduling timers for connection acquisitions.
+		 * Set the function to apply when scheduling pending acquisition timers.
 		 * By default, the {@link Schedulers#parallel()} will be used.
 		 *
-		 * @param acquireTimer the function to applly when scheduling timers for connection acquisitions
+		 * @param acquireTimer the function to apply when scheduling pending acquisition timers
 		 * @return {@literal this}
 		 * @since 1.0.19
 		 */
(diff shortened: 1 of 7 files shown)


reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -440,6 +442,10 @@ public ConnectionProvider build() {
 		static final Duration EVICT_IN_BACKGROUND_DISABLED       = Duration.ZERO;
 		static final int PENDING_ACQUIRE_MAX_COUNT_NOT_SPECIFIED = -2;
 
+		/**
+		 * Default timer service used for scheduling connection acquisition timers.
+		 */
+
```
violetagg, line 448: Is this javadoc needed?

violetagg (changes requested):

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
pderop: This is a mistake, I'm removing it, sorry.

violetagg (approved):
@reactor/netty-team PTAL

simonbasle (approved):

simonbasle:
I wonder if the netty hashwheel code should be offered as a convenience method ?

simonbasle:
ok, lgtm then 👍

PR #2178, opened 2022-05-07
Add support for secured http proxy

### LINKED ISSUE #2165: reactor\netty\tcp\ProxyProvider.java  no HTTPS support?
reactor\netty\tcp\ProxyProvider.java  
```
public enum Proxy 
{ 	
HTTP, SOCKS4, SOCKS5 
} 
```
Why would you not add HTTPS support for proxy ? 

While using netty in spring cloud gateway. I am trying to connect to HTTPS type of proxy (Both with and without authentication). The proxy settings are provided using VM args like -Dhttps.proxyHost=myproxy.com -Dhttps.proxyPort=<secured port>. I have also created a trust manager and set the same in client but while connecting to destination host like google.com am getting 
"Caused by: io.netty.handler.proxy.ProxyConnectException: http, none, myproxy.com /xx.xxx.xxx.xxx:xxxx => www.google.com:443, java.io.IOException: Connection reset by peer"

Could you please point me to what am I doing wrong.

pderop:
This PR provides the ability to configure an http client proxy in secure mode. 
You can now specify an ssl context in the proxy spec of the http client, and a secured channel will then be established between the client and the proxy. Then, on top of the tunnel, the client can also establish a second secured connection between itself and the remote server.

You only have to call the secure method on the proxy options, see the following example:

```
	SslContext proxySsl = SslContextBuilder.forClient()
			.trustManager(InsecureTrustManagerFactory.INSTANCE).build();

	Http11SslContextSpec clientSsl =
			Http11SslContextSpec.forClient()
					.configure(builder -> builder.trustManager(InsecureTrustManagerFactory.INSTANCE));

	HttpClient.create()
			.proxy(ops -> ops
					.type(ProxyProvider.Proxy.HTTP)
					.host("localhost")
					.port(proxy.port())
					.secure(spec -> spec.sslContext(proxySsl)))
			.secure(spec -> spec.sslContext(clientSsl))
			...
```
Fixes #2165

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/NettyPipeline.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/NettyPipeline.java
@@ -108,6 +108,7 @@ public interface NettyPipeline {
 	String OnChannelReadIdle     = LEFT + "onChannelReadIdle";
 	String OnChannelWriteIdle    = LEFT + "onChannelWriteIdle";
 	String ProxyHandler          = LEFT + "proxyHandler";
+	String ProxySslHandler       = LEFT + "proxySslHandler";
 	String UnvoidHandler         = LEFT + "unvoidHandler";
 	String ProxyLoggingHandler   = LEFT + "proxyLoggingHandler";
 	String ProxyProtocolDecoder  = LEFT + "proxyProtocolDecoder";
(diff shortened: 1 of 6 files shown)


pderop:
close the PR, which is actually not needed.

PR #2184, opened 2022-05-10
Support 303 redirect

AlexejTimonin:
This PR enables HttpClient to automatically GET resource from Location URI on 303 response, given followRedirect is enabled.

From rfc2616 (https://datatracker.ietf.org/doc/html/rfc2616#section-10.3.4):
>    10.3.4 303 See Other
   The response to the request can be found under a different URI and
   SHOULD be retrieved using a GET method on that resource. This method
   exists primarily to allow the output of a POST-activated script to
   redirect the user agent to a selected resource. The new URI is not a
   substitute reference for the originally requested resource. The 303
   response MUST NOT be cached, but the response to the second
   (redirected) request might be cacheable.

A client configured as shown below will on received 303 response issue a GET request to location:
```java
var nettyHttpClient = HttpClient.create()
            .compress(true)
            .followRedirect(true);
```

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -905,7 +905,7 @@ public abstract class HttpClient extends ClientTransport<HttpClient, HttpClientC
 	}
 
 	/**
-	 * Specifies whether HTTP status 301|302|307|308 auto-redirect support is enabled.
+	 * Specifies whether HTTP status 301|302|303|307|308 auto-redirect support is enabled.
 	 *
 	 * <p><strong>Note:</strong> The sensitive headers {@link #followRedirect(boolean, Consumer) followRedirect}
 	 * are removed from the initialized request when redirecting to a different domain, they can be re-added
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java
@@ -637,7 +637,7 @@ public final class HttpClientConfig extends ClientTransportConfig<HttpClientConf
 		return bootstrap.open();
 	}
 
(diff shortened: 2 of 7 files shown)


violetagg:
@AlexejTimonin Thanks for the PR!

The specification says 

```
The response to the request can be found under a different URI and
SHOULD be retrieved using a GET method on that resource.
```

So we should use `GET` as a replacement for the method.
Please revert the new API `HttpClientRequest method`.
You can change the `method` to `GET`, in case of `303`, similar to how the `uri` is changed. You need to make a change here:

https://github.com/reactor/reactor-netty/blob/fa466c1472fe9e56e1852d7a5dd995bdd5578176/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L676-L679

AlexejTimonin:
> @AlexejTimonin Thanks for the PR!
> 
> The specification says
> 
> ```
> The response to the request can be found under a different URI and
> SHOULD be retrieved using a GET method on that resource.
> ```
> 
> So we should use `GET` as a replacement for the method. Please revert the new API `HttpClientRequest method`. You can change the `method` to `GET`, in case of `303`, similar to how the `uri` is changed. You need to make a change here:
> 
> https://github.com/reactor/reactor-netty/blob/fa466c1472fe9e56e1852d7a5dd995bdd5578176/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L676-L679

Thank you for your quick and detailed feedback. I will make the changes you've proposed today.

AlexejTimonin:
@violetagg Would this work, or am I completely off here?

violetagg:
> @violetagg Would this work, or am I completely off here?

Yep this is what I meant.

violetagg (approved):

violetagg:
@reactor/netty-team PTAL

pderop (approved):
LGTM, thanks for the PR !

simonbasle (approved):

violetagg:
@AlexejTimonin I applied the change to `1.0.x` branch and then merged it into `main`. Thanks!

PR #2195, opened 2022-05-16
[tests] Add bouncycastle dependency

violetagg:
Add `bouncycastle` dependency for `noMicrometer` tests similar to the standard tests.

This change is added for `1.0.x` because one may decide to build Reactor Netty with java version where one won't be able to create a `SelfSignedCertificate`


pderop (approved):
lgtm

violetagg:

Issue #2217, opened 2022-05-18
Adapt DefaultLoopIOUring to Netty 5 changes


violetagg:
This is depending on Netty IOUring for Netty 5

The current integration is temporary excluded with #2218

PR #2218, opened 2022-05-18
Temporary exclude IOUring as there is no version for Netty 5

violetagg:
The new integration will be handled with #2217


pderop (approved):
lgtm

violetagg:

PR #2226, opened 2022-05-19
Netty's native code is separated from the java code

violetagg:
`*-native-*` is the artifact for the native code
`*-classes-*` is the artifact for the java code

Related to #1873


pderop (approved):
lgtm

PR #2231, opened 2022-05-20
Move from ChannelOutboundHandlerAdapter to ChannelHandlerAdapter

violetagg:
- There is no `VoidPromise` anymore so no need to handle this use case in `ProxyProvider`

Related to #1873


pderop (approved):
LGTM

Issue #2187, opened 2022-05-11
Recording metrics could cause errors in operational code


ChristianLMI:
## Motivation
I recently ran into an issue where an exception thrown in Micrometer caused my request to break. The actual issue I ran into is not in reactor although reactor was used. However as reactor uses Micrometer as well and e.g. [here|https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java] there also doesn't seem to be any protection of the main functionality against errors in metrics recording, this might be good to look at in reactor as well.

## Desired solution
IMHO metrics should be of lower priority than production results. Therefor a failing metric registration should write a log entry but not abort the current operation.

## Additional context
In my case a misconfigured a MeterFilter that caused NullPointerExceptions when recording a filtered metric.

violetagg:
@ChristianLMI Would you like to provide a PR?

ChristianLMI:
I'll try to find some time to at least get a first step done. However as I'm not too familiar with the project I might miss edges ;)

violetagg:
@ChristianLMI It will be great if you can find time. If you decide to work on this please do the changes against `1.0.x` branch.

The classes that need to have this enhancement are:
`reactor-netty-core`
- reactor.netty.channel.AbstractChannelMetricsHandler
- reactor.netty.channel.ChannelMetricsHandler
- reactor.netty.channel.ContextAwareChannelMetricsHandler
- reactor.netty.transport.AddressResolverGroupMetrics

`reactor-netty-http`
- reactor.netty.http.client.AbstractHttpClientMetricsHandler
- reactor.netty.http.client.ContextAwareHttpClientMetricsHandler
- reactor.netty.http.server.AbstractHttpServerMetricsHandler
- reactor.netty.http.server.ContextAwareHttpServerMetricsHandler

ChristianLMI:
I created a PR. However now I need to figure out how to get an approval for the CLA.

PR #2237, opened 2022-05-23
Issue 2187

ChristianLMI:
Fixing issues described in #2187 by wrapping only metrics related code in try-catch and logging issues on warn level.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
@@ -27,9 +27,6 @@ import reactor.util.Loggers;
 import reactor.util.annotation.Nullable;
 
 import java.net.SocketAddress;
-import java.util.function.BiConsumer;
-import java.util.function.Function;
-import java.util.function.Supplier;
 
 /**
  * Base {@link ChannelHandler} for collecting metrics on protocol level.
@@ -96,7 +93,24 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 
 	@Override
 	public void channelRead(ChannelHandlerContext ctx, Object msg) {
-		fromPayload(msg, DatagramPacket::sender, (address, bytes) -> recordRead(ctx, address, bytes));
+		try {
+			if (msg instanceof ByteBuf) {
+				ByteBuf buffer = (ByteBuf) msg;
+				if (buffer.readableBytes() > 0) {
+					recordRead(ctx, remoteAddress, buffer.readableBytes());
+				}
+			}
+			else if (msg instanceof DatagramPacket) {
+				DatagramPacket p = (DatagramPacket) msg;
+				ByteBuf buffer = p.content();
+				if (buffer.readableBytes() > 0) {
+					recordRead(ctx, remoteAddress != null ? remoteAddress : p.sender(), buffer.readableBytes());
+				}
+			}
+		} catch (RuntimeException e) {
+			log.warn("Exception caught while recording metrics.", e);
+			// Allow request-response exchange to continue, unaffected by metrics problem
+		}
 
 		ctx.fireChannelRead(msg);
 	}
@@ -104,7 +118,24 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 	@Override
 	@SuppressWarnings("FutureReturnValueIgnored")
 	public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
-		fromPayload(msg, DatagramPacket::recipient, (address, bytes) -> recordWrite(ctx, address, bytes));
+		try {
+			if (msg instanceof ByteBuf) {
+				ByteBuf buffer = (ByteBuf) msg;
+				if (buffer.readableBytes() > 0) {
+					recordWrite(ctx, remoteAddress, buffer.readableBytes());
+				}
+			}
+			else if (msg instanceof DatagramPacket) {
+				DatagramPacket p = (DatagramPacket) msg;
+				ByteBuf buffer = p.content();
+				if (buffer.readableBytes() > 0) {
+					recordWrite(ctx, remoteAddress != null ? remoteAddress : p.recipient(), buffer.readableBytes());
+				}
+			}
+		} catch (RuntimeException e) {
+			log.warn("Exception caught while recording metrics.", e);
+			// Allow request-response exchange to continue, unaffected by metrics problem
+		}
 
 		//"FutureReturnValueIgnored" this is deliberate
 		ctx.write(msg, promise);
@@ -126,29 +157,6 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 	public abstract ChannelHandler tlsMetricsHandler();
 
 	public abstract ChannelMetricsRecorder recorder();
-
-	private void fromPayload(Object msg, Function<DatagramPacket, ? extends SocketAddress> addressSelector,
-	                         BiConsumer<SocketAddress, Integer> recordFn) {
-		try {
-			if (msg instanceof ByteBuf) {
-				ByteBuf buffer = (ByteBuf) msg;
-				if (buffer.readableBytes() > 0) {
-					recordFn.accept(remoteAddress, buffer.readableBytes());
-				}
-			}
-			else if (msg instanceof DatagramPacket) {
-				DatagramPacket p = (DatagramPacket) msg;
-				ByteBuf buffer = p.content();
-				if (buffer.readableBytes() > 0) {
-					SocketAddress address = remoteAddress != null ? remoteAddress : addressSelector.apply(p);
-					recordFn.accept(address, buffer.readableBytes());
-				}
-			}
-		} catch (RuntimeException e) {
-			log.warn("Exception caught while recording metrics.", e);
-			// Allow request-response exchange to continue, unaffected by metrics problem
-		}
-	}
 
 	protected void recordException(ChannelHandlerContext ctx, SocketAddress address) {
 		recorder().incrementErrorsCount(address);
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
@@ -52,7 +52,8 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 		if (onServer) {
 			try {
 				recorder().recordServerConnectionOpened(ctx.channel().localAddress());
-			} catch (RuntimeException e) {
+			}
+			catch (RuntimeException e) {
 				log.warn("Exception caught while recording metrics.", e);
 				// Allow request-response exchange to continue, unaffected by metrics problem
 			}
@@ -65,7 +66,8 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 		if (onServer) {
 			try {
 				recorder().recordServerConnectionClosed(ctx.channel().localAddress());
-			} catch (RuntimeException e) {
+			}
+			catch (RuntimeException e) {
 				log.warn("Exception caught while recording metrics.", e);
 				// Allow request-response exchange to continue, unaffected by metrics problem
 			}
@@ -107,7 +109,8 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 					recordRead(ctx, remoteAddress != null ? remoteAddress : p.sender(), buffer.readableBytes());
 				}
 			}
-		} catch (RuntimeException e) {
+		}
+		catch (RuntimeException e) {
 			log.warn("Exception caught while recording metrics.", e);
 			// Allow request-response exchange to continue, unaffected by metrics problem
 		}
@@ -132,7 +135,8 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 					recordWrite(ctx, remoteAddress != null ? remoteAddress : p.recipient(), buffer.readableBytes());
 				}
 			}
-		} catch (RuntimeException e) {
+		}
+		catch (RuntimeException e) {
 			log.warn("Exception caught while recording metrics.", e);
 			// Allow request-response exchange to continue, unaffected by metrics problem
 		}
@@ -145,7 +149,8 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 	public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
 		try {
 			recordException(ctx, remoteAddress != null ? remoteAddress : ctx.channel().remoteAddress());
-		} catch (RuntimeException e) {
+		}
+		catch (RuntimeException e) {
 			log.warn("Exception caught while recording metrics.", e);
 			// Allow request-response exchange to continue, unaffected by metrics problem
 		}
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
@@ -22,9 +22,14 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.socket.DatagramPacket;
 import reactor.netty.NettyPipeline;
+import reactor.util.Logger;
+import reactor.util.Loggers;
 import reactor.util.annotation.Nullable;
 
 import java.net.SocketAddress;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
 
 /**
  * Base {@link ChannelHandler} for collecting metrics on protocol level.
@@ -33,6 +38,8 @@ import java.net.SocketAddress;
  * @since 1.0.8
  */
 public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler {
+
+	private static final Logger log = Loggers.getLogger(AbstractChannelMetricsHandler.class);
 
 	final SocketAddress remoteAddress;
 
@@ -46,7 +53,12 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 	@Override
 	public void channelActive(ChannelHandlerContext ctx) {
 		if (onServer) {
-			recorder().recordServerConnectionOpened(ctx.channel().localAddress());
+			try {
+				recorder().recordServerConnectionOpened(ctx.channel().localAddress());
+			} catch (RuntimeException e) {
+				log.warn("Exception caught while recording metrics.", e);
+				// Allow request-response exchange to continue, unaffected by metrics problem
+			}
 		}
 		ctx.fireChannelActive();
 	}
@@ -54,7 +66,12 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 	@Override
 	public void channelInactive(ChannelHandlerContext ctx) {
 		if (onServer) {
-			recorder().recordServerConnectionClosed(ctx.channel().localAddress());
+			try {
+				recorder().recordServerConnectionClosed(ctx.channel().localAddress());
+			} catch (RuntimeException e) {
+				log.warn("Exception caught while recording metrics.", e);
+				// Allow request-response exchange to continue, unaffected by metrics problem
+			}
 		}
 		ctx.fireChannelInactive();
 	}
@@ -79,19 +96,7 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 
 	@Override
 	public void channelRead(ChannelHandlerContext ctx, Object msg) {
-		if (msg instanceof ByteBuf) {
-			ByteBuf buffer = (ByteBuf) msg;
-			if (buffer.readableBytes() > 0) {
-				recordRead(ctx, remoteAddress, buffer.readableBytes());
-			}
-		}
-		else if (msg instanceof DatagramPacket) {
-			DatagramPacket p = (DatagramPacket) msg;
-			ByteBuf buffer = p.content();
-			if (buffer.readableBytes() > 0) {
-				recordRead(ctx, remoteAddress != null ? remoteAddress : p.sender(), buffer.readableBytes());
-			}
-		}
+		fromPayload(msg, DatagramPacket::sender, (address, bytes) -> recordRead(ctx, address, bytes));
 
 		ctx.fireChannelRead(msg);
 	}
@@ -99,19 +104,7 @@ public abstract class AbstractChannelMetricsHandler extends ChannelDuplexHandler
 	@Override
 	@SuppressWarnings("FutureReturnValueIgnored")
 	public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
-		if (msg instanceof ByteBuf) {
-			ByteBuf buffer = (ByteBuf) msg;
-			if (buffer.readableBytes() > 0) {
-				recordWrite(ctx, remoteAddress, buffer.readableBytes());
-			}
-		}
-		else if (msg instanceof DatagramPacket) {
-			DatagramPacket p = (DatagramPacket) msg;
-			ByteBuf buffer = p.content();
-			if (buffer.readableBytes() > 0) {
-				recordWrite(ctx, remoteAddress != null ? remoteAddress : p.recipient(), buffer.readableBytes());
-			}
-		}
+		fromPayload(msg, DatagramPacket::recipient, (address, bytes) -> recordWrite(ctx, address, bytes));
 
 		//"FutureReturnValueIgnored" this is deliberate
 		ctx.write(msg, promise);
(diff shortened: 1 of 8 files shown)


violetagg (changes requested):

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java
```diff
@@ -376,15 +376,20 @@ protected void initChannel(Channel channel) {
...
+							ByteBufAllocatorMetrics.INSTANCE.registerMetrics("pooled", ((PooledByteBufAllocator) alloc).metric(), alloc);
+						}
+						else if (alloc instanceof UnpooledByteBufAllocator) {
+							ByteBufAllocatorMetrics.INSTANCE.registerMetrics("unpooled", ((UnpooledByteBufAllocator) alloc).metric(), alloc);
+						}
+
+						MicrometerEventLoopMeterRegistrar.INSTANCE.registerMetrics(channel.eventLoop());
+					} catch (RuntimeException e) {
+						log.warn("Exception caught while recording metrics.", e);
+						// Allow request-response exchange to continue, unaffected by metrics problem
```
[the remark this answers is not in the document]
ChristianLMI: As far as I understand it does, as there could be customization inside the metrics registration. If I remember correctly I only found this by a failing test.

reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
```diff
@@ -79,47 +96,28 @@ public void channelRegistered(ChannelHandlerContext ctx) {
...
-			}
-		}
-		else if (msg instanceof DatagramPacket) {
-			DatagramPacket p = (DatagramPacket) msg;
-			ByteBuf buffer = p.content();
-			if (buffer.readableBytes() > 0) {
-				recordRead(ctx, remoteAddress != null ? remoteAddress : p.sender(), buffer.readableBytes());
-			}
-		}
+		fromPayload(msg, DatagramPacket::sender, (address, bytes) -> recordRead(ctx, address, bytes));
```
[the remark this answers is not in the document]
ChristianLMI: I'm not really a JVM internal expert, so I might easily overlook s.th. here, but I don't really see additional object creation. There is of course a slight overhead from addition function invocation, but I'd assume the JIT to cover for that - especially in a hop path.
But if you like that better I can of course build the change back.

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java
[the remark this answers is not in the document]
violetagg: Both `ByteBufAllocatorMetrics` and `MicrometerEventLoopMeterRegistrar` are Reactor Netty implementations and  we do not provide API for any user's recorders.

reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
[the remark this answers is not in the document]
violetagg: yep let use the try/catch as with the rest of the code

violetagg (changes requested):

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java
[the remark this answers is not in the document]
ChristianLMI: Please have a look at `reactor.netty.transport.TransportEventLoopMetricsTest#testEventLoopMetricsFailure` this test will break if I remove the error handing. A MeterFilter throwing an exception was the issue that trigger me to take a look at reactor and actuator. Do you think the assumption as wrong here?

reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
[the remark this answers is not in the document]
violetagg: ok I understand, let's keep the try/catch then

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java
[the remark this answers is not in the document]
violetagg: ok I understand, let's keep the try/catch then

reactor-netty-core/src/main/java/reactor/netty/channel/AbstractChannelMetricsHandler.java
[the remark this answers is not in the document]
ChristianLMI: I updated the PR. How's your policy do I resolve the conversation or are you taking care of that?

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
```diff
@@ -83,94 +88,112 @@ public void channelInactive(ChannelHandlerContext ctx) {
...
-						// This metric is not applicable for HTTP/2
-						// ops.hostAddress() == null when request decoding failed, in this case
-						// we do not report active connection, so we do not report inactive connection
-						recordInactiveConnection(ops);
+			dataSent += extractProcessedDataFromBuffer(msg);
+
+			if (msg instanceof LastHttpContent) {
+				promise.addListener(future -> {
+					try {
+						withHttpServerOperations(ctx, ops -> {
```
violetagg, line 105: Please use `try/catch` without introducing a `Consumer`

violetagg (changes requested):
Please fix the checkstyle warnings.

reactor-netty-core/src/test/java/reactor/netty/transport/TransportEventLoopMetricsTest.java
```diff
@@ -115,6 +117,51 @@ void testEventLoopMetrics() throws InterruptedException {
 		}
 	}
 
+	// https://github.com/reactor/reactor-netty/issues/2187
+	@Test
+	void testEventLoopMetricsFailure() throws InterruptedException {
+		registry.config().meterFilter(new MeterFilter() {
+			public Meter.Id map(Meter.Id id) {
```
violetagg, line 124: Missing `@Override`

reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
```diff
@@ -178,18 +178,18 @@ void testExistingEndpoint(HttpProtocol[] serverProtocols, HttpProtocol[] clientP
...
-		            .expectComplete()
-		            .verify(Duration.ofSeconds(30));
+						.uri("/1")
+						.send(body)
+						.responseContent()
+						.aggregate()
+						.asString())
+				.expectNext("Hello World!")
+				.expectComplete()
+				.verify(Duration.ofSeconds(30));
```
violetagg, line 192: Please revert the formatting changes in this class.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractHttpClientMetricsHandler.java
```diff
@@ -76,64 +80,88 @@ protected AbstractHttpClientMetricsHandler(AbstractHttpClientMetricsHandler copy
...
+						recordWrite(address);
+					} catch (RuntimeException e) {
+						log.warn("Exception caught while recording metrics.", e);
+						// Allow request-response exchange to continue, unaffected by metrics problem
+					}
+				});
+			}
+		} catch (RuntimeException e) {
+			log.warn("Exception caught while recording metrics.", e);
+			// Allow request-response exchange to continue, unaffected by metrics problem
```
violetagg, line 103: Do we need this `try/catch`? Is it possible to have a `RuntimeException` not related to metrics?

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractHttpClientMetricsHandler.java
```diff
@@ -76,64 +80,88 @@ protected AbstractHttpClientMetricsHandler(AbstractHttpClientMetricsHandler copy
...
-		if (msg instanceof LastHttpContent) {
-			recordRead(ctx.channel().remoteAddress());
-			reset();
+			if (msg instanceof LastHttpContent) {
+				recordRead(ctx.channel().remoteAddress());
+				reset();
+			}
+		} catch (RuntimeException e) {
+			log.warn("Exception caught while recording metrics.", e);
+			// Allow request-response exchange to continue, unaffected by metrics problem
```
violetagg, line 126: Do we need this `try/catch` for the whole block or it is more appropriate to have it only for the recording? Is it possible to have a `RuntimeException` not related to metrics?

reactor-netty-core/src/test/java/reactor/netty/resources/PooledConnectionProviderCustomMetricsTest.java
```diff
@@ -25,6 +25,10 @@
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.config.MeterFilter;
```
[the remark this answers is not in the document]
ChristianLMI: Yes, sure that slipped.

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
```diff
@@ -83,94 +88,112 @@ public void channelInactive(ChannelHandlerContext ctx) {
...
+						// Allow request-response exchange to continue, unaffected by metrics problem
 					}
-				}
 
-				dataSent = 0;
-			});
+					dataSent = 0;
+				});
+			}
+		} catch (RuntimeException e) {
```
violetagg, line 123: Do we need this `try/catch`? Is it possible to have a `RuntimeException` not related to metrics?

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
```diff
@@ -83,94 +88,112 @@ public void channelInactive(ChannelHandlerContext ctx) {
...
-			}
+			if (msg instanceof LastHttpContent) {
+				withHttpServerOperations(ctx, ops -> {
+					recordRead(ops, uriTagValue == null ? ops.path : uriTagValue.apply(ops.path), ops.method().name());
+				});
 
-			dataReceived = 0;
+				dataReceived = 0;
+			}
+		} catch (RuntimeException e) {
```
violetagg, line 154: Do we need this `try/catch` for the whole block or it is more appropriate to have it only for the recording? Is it possible to have a `RuntimeException` not related to metrics?

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
ChristianLMI: I'll revert that change, still a reason would be much appreciated.

reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
ChristianLMI: Interesting, this seems to be either not in the .editorConfig, or not picked up by IntelliJ automatically.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractHttpClientMetricsHandler.java
ChristianLMI: No, this is not necessary. I likely introduced this before I found that the actual recording is done in a listener here.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractHttpClientMetricsHandler.java
ChristianLMI: This is a good question. I'm not very familiar with the objects used here. `status = ((HttpResponse) msg).status().codeAsText().toString();` does look like it could result in a `NullPointerException` also other invocation don't clearly identify as safe to me. All of the enclosed code is only contributing to metrics recording and should IMHO not terminate data processing. I'd be glad to reduce the scope if it's certain that there will be no issues with those calls.

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
ChristianLMI: No, also the actual handling is in a listener I likely only discovered later.

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
ChristianLMI: Similar to the other handler. If it's save to assume the other operations will not throw I'd gladly move the try/catch around the actual record operations.

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
violetagg: Try to create JMH test and share the results from the two approaches.

ChristianLMI: I created a very generic setup here https://github.com/ChristianLMI/lambda-bench. The repo also includes some results.

violetagg: Thanks for taking the time to setup this. It is very interesting. Can you now extend the lambda to use outer class members as it is in our case. (e.g. uriTagValue)

ChristianLMI: I added another benchmark. Please have a look if that matches what you had in mind.

violetagg: yes

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractHttpClientMetricsHandler.java
```diff
@@ -77,63 +81,82 @@ protected AbstractHttpClientMetricsHandler(AbstractHttpClientMetricsHandler copy
...
 		ctx.fireChannelRead(msg);
 	}
 
 	@Override
 	public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
-		recordException(ctx);
-
+		try {
+			recordException(ctx);
+		} catch (RuntimeException e) {
```
violetagg, line 130:
```suggestion
		}
		catch (RuntimeException e) {
```

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
```diff
@@ -93,26 +98,26 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+						recordWrite(ops, uriTagValue == null ? ops.path : uriTagValue.apply(ops.path),
+								ops.method().name(), ops.status().codeAsText().toString());
+						if (!ops.isHttp2() && ops.hostAddress() != null) {
+							// This metric is not applicable for HTTP/2
+							// ops.hostAddress() == null when request decoding failed, in this case
+							// we do not report active connection, so we do not report inactive connection
+							recordInactiveConnection(ops);
+						}
 					}
+				} catch (RuntimeException e) {
```
violetagg, line 118:
```suggestion
				}
				catch (RuntimeException e) {
```

violetagg (changes requested):

reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
```diff
@@ -233,6 +233,59 @@ else if (clientProtocols.length == 2 &&
 				numWrites[1], bytesWrite[1]);
 	}
 
+	// https://github.com/reactor/reactor-netty/issues/2187
+	@ParameterizedTest
+	@MethodSource("httpCompatibleProtocols")
+	void testRecordingFailsServerSide(HttpProtocol[] serverProtocols, HttpProtocol[] clientProtocols,
+	                                  @Nullable ProtocolSslContextSpec serverCtx, @Nullable ProtocolSslContextSpec clientCtx) throws Exception {
```
violetagg, line 240: `Exception` is never thrown in the method

```suggestion
	                                  @Nullable ProtocolSslContextSpec serverCtx, @Nullable ProtocolSslContextSpec clientCtx) {
```

reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
```diff
@@ -233,6 +233,59 @@ else if (clientProtocols.length == 2 &&
...
+				    .expectNext("Hello World!")
+				    .expectComplete()
+				    .verify(Duration.ofSeconds(2));
+	}
+
+	// https://github.com/reactor/reactor-netty/issues/2187
+	@ParameterizedTest
+	@MethodSource("httpCompatibleProtocols")
+	void testRecordingFailsClientSide(HttpProtocol[] serverProtocols, HttpProtocol[] clientProtocols,
+	                                  @Nullable ProtocolSslContextSpec serverCtx, @Nullable ProtocolSslContextSpec clientCtx) throws Exception {
```
violetagg, line 267: `Exception` is never thrown in the method

```suggestion
	                                  @Nullable ProtocolSslContextSpec serverCtx, @Nullable ProtocolSslContextSpec clientCtx) {
```

reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
```diff
@@ -233,6 +233,59 @@ else if (clientProtocols.length == 2 &&
...
+		disposableServer = customizeServerOptions(httpServer, serverCtx, serverProtocols)
+				.bindNow();
+
+		AtomicReference<SocketAddress> serverAddress = new AtomicReference<>();
+		httpClient = customizeClientOptions(httpClient, clientCtx, clientProtocols).doAfterRequest((req, conn) ->
+            serverAddress.set(conn.channel().remoteAddress())
+		).metrics(true, id -> {
+			throw new IllegalArgumentException("Testcase injected Exception");
+		});
```
violetagg, line 276: `serverAddress` is never used

```suggestion
		httpClient = customizeClientOptions(httpClient, clientCtx, clientProtocols).metrics(true, id -> {
			throw new IllegalArgumentException("Testcase injected Exception");
		});
```

reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
```diff
@@ -233,6 +233,59 @@ else if (clientProtocols.length == 2 &&
...
+		disposableServer = customizeServerOptions(httpServer, serverCtx, serverProtocols)
+				.metrics(true, id -> {
+					throw new IllegalArgumentException("Testcase injected Exception");
+				})
+				.bindNow();
+
+		AtomicReference<SocketAddress> serverAddress = new AtomicReference<>();
+		httpClient = customizeClientOptions(httpClient, clientCtx, clientProtocols).doAfterRequest((req, conn) ->
+            serverAddress.set(conn.channel().remoteAddress())
+		);
```
violetagg, line 250: `serverAddress` is never used

```suggestion
		httpClient = customizeClientOptions(httpClient, clientCtx, clientProtocols);
```

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractHttpClientMetricsHandler.java
violetagg: Actually we do need here this because `uriTagValue` is provided by the user and may throw an exception - see the failing CI build

violetagg (changes requested):
I think that we need to handle also the recordings 

reactor.netty.http.server.AbstractHttpServerMetricsHandler#channelActive
reactor.netty.http.server.AbstractHttpServerMetricsHandler#channelInactive

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
```diff
@@ -93,27 +97,28 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+							// ops.hostAddress() == null when request decoding failed, in this case
+							// we do not report active connection, so we do not report inactive connection
+							recordInactiveConnection(ops);
+						}
 					}
 				}
+				catch (RuntimeException e) {
+					log.warn("Exception caught while recording metrics.", e);
+					// Allow request-response exchange to continue, unaffected by metrics problem
+				}
```
violetagg, line 121: We have reactor.netty.http.HttpMetricsHandlerTests#testServerConnectionsMicrometer failing on the CI

The problem is that `uriTagValue` throws an exception which fails both recordings `recordWrite` and `recordInactiveConnection`.

Should we leave it like this or have a try/catch per recording, basically `recordInactiveConnection` doesn't depend on `uriTagValue` and can provide correct metric if we introduce separate try/catch, otherwise both metrics will be incorrect.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractHttpClientMetricsHandler.java
ChristianLMI: Added this back

reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
ChristianLMI: Removed throws declaration

reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
ChristianLMI: Removed unused variable

reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
ChristianLMI: Removed unused variable

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
ChristianLMI: Added this back

violetagg:
> The 1.0.x branch has advanced. Should I rebase to keep this in sync?

Yes please rebase. Thanks

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
```diff
@@ -67,110 +71,145 @@ public void channelActive(ChannelHandlerContext ctx) {
...
+									ops.method().name(), ops.status().codeAsText().toString());
+							if (!ops.isHttp2() && ops.hostAddress() != null) {
+								// This metric is not applicable for HTTP/2
+								// ops.hostAddress() == null when request decoding failed, in this case
+								// we do not report active connection, so we do not report inactive connection
+								recordInactiveConnection(ops);
+							}
+						}
+					}
+          catch (RuntimeException e) {
```
violetagg, line 129:
```suggestion
					catch (RuntimeException e) {
```

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
```diff
@@ -67,110 +71,145 @@ public void channelActive(ChannelHandlerContext ctx) {
...
 
-		if (msg instanceof ByteBufHolder) {
-			dataSent += ((ByteBufHolder) msg).content().readableBytes();
+					dataSent = 0;
+				});
+			}
 		}
-		else if (msg instanceof ByteBuf) {
-			dataSent += ((ByteBuf) msg).readableBytes();
+    catch (RuntimeException e) {
```
violetagg, line 138:
```suggestion
		catch (RuntimeException e) {
```

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
```diff
@@ -67,110 +71,145 @@ public void channelActive(ChannelHandlerContext ctx) {
...
+			log.warn("Exception caught while recording metrics.", e);
+			// Allow request-response exchange to continue, unaffected by metrics problem
 		}
+    finally {
```
violetagg, line 142:
```suggestion
		finally {
```

violetagg (approved):

violetagg:
@reactor/netty-team PTAL

pderop (approved):
LGTM

PR #2244, opened 2022-05-23
Replace LastHttpContent.EMPTY_LAST_CONTENT with EmptyLastHttpContent

violetagg:
Related to #1873

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/HttpOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/HttpOperations.java
@@ -34,6 +34,7 @@ import io.netty5.channel.ChannelHandler;
 import io.netty5.channel.CombinedChannelDuplexHandler;
 import io.netty5.handler.codec.ByteToMessageCodec;
 import io.netty5.handler.codec.ByteToMessageDecoder;
+import io.netty5.handler.codec.http.EmptyLastHttpContent;
 import io.netty5.handler.codec.http.FullHttpMessage;
 import io.netty5.handler.codec.http.HttpHeaderNames;
 import io.netty5.handler.codec.http.HttpMessage;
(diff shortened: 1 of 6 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
```diff
@@ -638,7 +639,7 @@ protected void onOutboundComplete() {
 			f = channel().writeAndFlush(newFullBodyMessage(EMPTY_BUFFER));
 		}
 		else if (markSentBody()) {
-			LastHttpContent lastHttpContent = LastHttpContent.EMPTY_LAST_CONTENT;
+			LastHttpContent<?> lastHttpContent = new EmptyLastHttpContent(channel().bufferAllocator());
```
pderop, line 642: same question: can we cache the EmptyLastHttpContent in a static variable and reuse it ?

reactor-netty-http/src/main/java/reactor/netty/http/HttpOperations.java
```diff
@@ -423,7 +424,7 @@ else if (!SCHEME_PATTERN.matcher(tempUri).matches()) {
 						ByteBuf bb = ((ByteBufHolder) msg).content();
 						ctx.fireChannelRead(bb);
 						if (msg instanceof LastHttpContent) {
-							ctx.fireChannelRead(LastHttpContent.EMPTY_LAST_CONTENT);
+							ctx.fireChannelRead(new EmptyLastHttpContent(ctx.bufferAllocator()));
```
pderop, line 427: Is it possible to cache the EmptyLastHttpContent in some static variables and reuse it instead of allocating one each time ?

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -534,7 +535,7 @@ protected void onOutboundComplete() {
 		}
 		else if (markSentBody()) {
 			//"FutureReturnValueIgnored" this is deliberate
-			channel().writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT);
+			channel().writeAndFlush(new EmptyLastHttpContent(channel().bufferAllocator()));
```
pderop, line 538: same question here, can we cache the EmptyLastHttpContent in a static variable ?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
violetagg: EmptyLastHttpContent is Resource, once you close it you cannot reuse it

pderop (approved):
LGTM.
Indeed, you are right, we can't reuse a resource once it is closed, so we need to allocate ...

PR #2257, opened 2022-05-31
Http2 connection pool optimisation

violetagg:
HTTP2 stream channel pipeline optimisation

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -89,6 +89,7 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 	final Duration poolInactivity;
 	final Duration disposeTimeout;
 	final Map<SocketAddress, Integer> maxConnections = new HashMap<>();
+	Mono<Void> onDispose;
 
 	protected PooledConnectionProvider(Builder builder) {
 		this(builder, null);
@@ -106,6 +107,7 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 			poolFactoryPerRemoteHost.put(entry.getKey(), new PoolFactory<>(entry.getValue(), builder.disposeTimeout));
 			maxConnections.put(entry.getKey(), entry.getValue().maxConnections);
 		}
+		this.onDispose = Mono.empty();
 		scheduleInactivePoolsDisposal();
 	}
 
@@ -190,10 +192,10 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 			                    })
 			                    .collect(Collectors.toList());
 			if (pools.isEmpty()) {
-				return Mono.empty();
+				return onDispose;
 			}
 			channelPools.clear();
-			return Mono.when(pools);
+			return onDispose.and(Mono.when(pools));
 		});
 	}
 
@@ -241,6 +243,10 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 	@Override
 	public String name() {
 		return name;
+	}
+
+	public void onDispose(Mono<Void> disposeMono) {
+		onDispose = onDispose.and(disposeMono);
 	}
 
 	protected abstract CoreSubscriber<PooledRef<T>> createDisposableAcquire(
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -455,6 +455,10 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 			return poolBuilder;
 		}
 
+		public long maxIdleTime() {
+			return this.maxIdleTime;
+		}
+
 		public long maxLifeTime() {
 			return maxLifeTime;
 		}
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -30,7 +30,6 @@ import reactor.netty.ConnectionObserver;
 import reactor.netty.ReactorNetty;
 import reactor.netty.transport.TransportConfig;
 import reactor.netty.internal.util.MapUtils;
-import reactor.pool.AllocationStrategy;
 import reactor.pool.InstrumentedPool;
 import reactor.pool.Pool;
 import reactor.pool.PoolBuilder;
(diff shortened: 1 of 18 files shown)


reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -466,6 +487,45 @@ public String toString() {
 					", pendingAcquireTimeout=" + pendingAcquireTimeout +
 					'}';
 		}
+
+		static final class DelegatingAllocationStrategy implements reactor.pool.AllocationStrategy {
```
simonbasle, line 491: not sure why a `DelegatingAllocationStrategy` is necessary, if the strategies can be deep-copied. can you clarify?

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2AllocationStrategy.java
```diff
@@ -0,0 +1,209 @@
...
+			if (minConnections > maxConnections) {
+				throw new IllegalArgumentException("minConnections must be less than or equal to maxConnections");
+			}
+			return new Http2AllocationStrategy(this);
+		}
+
+		@Override
+		public Builder maxConcurrentStreams(long maxConcurrentStreams) {
+			if (maxConcurrentStreams < -1) {
+				throw new IllegalArgumentException("maxConcurrentStreams must be grater than or equal to -1");
```
simonbasle, line 185: typo
```suggestion
				throw new IllegalArgumentException("maxConcurrentStreams must be greater than or equal to -1");
```

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2AllocationStrategy.java
```diff
@@ -0,0 +1,209 @@
...
+		 * Build a new {@link Http2AllocationStrategy}
+		 *
+		 * @return a new {@link Http2AllocationStrategy}
+		 */
+		Http2AllocationStrategy build();
+
+		/**
+		 * Configures the maximum number of the concurrent streams that can be opened to the remote peer.
+		 * When evaluating how many streams can be opened to the remote peer,
+		 * the minimum of this configuration and the remote peer configuration is taken.
```
simonbasle, line 45: maybe indicate that `-1` is a valid value for which this statement doesn't hold?
something like:
```suggestion
		 * the minimum of this configuration and the remote peer configuration is taken (unless -1 is used).
```
the sentence right below clarifies what happens when -1 is used, on top of stating it is the default.

simonbasle (changes requested):
overall LGTM, though I added a few small comments

reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
violetagg: The current `reactor.netty.resources.ConnectionProvider.AllocationStrategy` is not `reactor.pool.AllocationStrategy` so we need to provide `reactor.pool.AllocationStrategy`.

simonbasle: ah, sorry for missing that 🤦

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2AllocationStrategy.java
```diff
@@ -0,0 +1,209 @@
...
+		static final int DEFAULT_MIN_CONNECTIONS = 0;
+
+		long maxConcurrentStreams = DEFAULT_MAX_CONCURRENT_STREAMS;
+		int maxConnections = DEFAULT_MAX_CONNECTIONS;
+		int minConnections = DEFAULT_MIN_CONNECTIONS;
+
+		@Override
+		public Http2AllocationStrategy build() {
+			if (minConnections > maxConnections) {
+				throw new IllegalArgumentException("minConnections must be less than or equal to maxConnections");
```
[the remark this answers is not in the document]
violetagg: fixed

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2AllocationStrategy.java
violetagg: fixed

simonbasle (approved):

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -140,23 +149,33 @@ final class Http2Pool implements InstrumentedPool<Connection>, InstrumentedPool.
 			AtomicIntegerFieldUpdater.newUpdater(Http2Pool.class, "wip");
 
 	final Clock clock;
+	final Long maxConcurrentStreams;
+	final long maxIdleTime;
 	final long maxLifeTime;
+	final int minConnections;
 	final PoolConfig<Connection> poolConfig;
+	final LongAdder totalMaxConcurrentStreams = new LongAdder();
```
pderop, line 158: The totalMaxConcurrentStreams is a LongAdder but it seems to be less frequently updated than read ?
is it correct ? for example, the totalMaxConcurrentStreams is only updated when a slot is created or when the server has sent a SETTINGS with an updated max-concurrent-stream, in this case, the canOpenStream() method will update the totalMaxConcurrentStreams (but only if the diff is != 0):

```
		boolean canOpenStream() {
			ChannelHandlerContext frameCodec = http2FrameCodecCtx();
			if (frameCodec != null && http2MultiplexHandlerCtx() != null) {
				long maxActiveStreams = ((Http2FrameCodec) frameCodec.handler()).connection().local().maxActiveStreams();
				maxActiveStreams = pool.maxConcurrentStreams == -1 ? maxActiveStreams :
						Math.min(pool.maxConcurrentStreams, maxActiveStreams);
				long diff = maxActiveStreams - maxConcurrentStreams;
->				if (diff != 0) {
					maxConcurrentStreams = maxActiveStreams;
					pool.totalMaxConcurrentStreams.addAndGet(diff);
				}
				int concurrency = this.concurrency;
				return concurrency < maxActiveStreams;
			}
			return false;
		}
```
However, the totalMaxConcurrentStreams is read very very frequently by the request() method.

So, what I mean is that even if under low update contentions a LongAdder performance is similar to an AtomicInteger, it however has higher space consumption and is less performant than AtomicInteger in terms of read access.

Then why not using an AtomicInteger instead of a LongAdder for the totalMaxConcurrentStreams ?

violetagg: fixed

violetagg:
@reactor/netty-team I added 3 additional changes PTAL

pderop (approved):
looks good to me.

simonbasle (approved):

violetagg:

violetagg:
Changes were applied to `1.0.x` and merged in `main`

PR #2273, opened 2022-06-09
Adapt TransportConnector to the changes in channel.register()

violetagg:
Related to #1873

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/TransportConnector.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/TransportConnector.java
@@ -27,7 +27,6 @@ import io.netty5.channel.unix.DomainSocketAddress;
 import io.netty5.resolver.AddressResolver;
 import io.netty5.resolver.AddressResolverGroup;
 import io.netty5.util.AttributeKey;
-import io.netty5.util.concurrent.EventExecutor;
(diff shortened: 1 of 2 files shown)


violetagg:
wait with the review I'll update the PR

reactor-netty-core/src/main/java/reactor/netty/transport/TransportConnector.java
```diff
@@ -252,56 +261,84 @@ static void doConnect(
...
+				ChannelFactory<? extends Channel> channelFactory = config.connectionFactory(isDomainSocket);
+				channel = channelFactory.newChannel(eventLoop);
+			}
 		}
 		catch (Throwable t) {
-			if (channel != null) {
-				channel.unsafe().closeForcibly();
-			}
 			return Mono.error(t);
 		}
```
pderop, line 296: why the channel is not closed anymore  in case of an exception ?

violetagg: I something fails it will be channel creation? Do we have a channel if there is an exception?

pderop: oops, yeah, sorry.

violetagg:
@pderop Thanks for the review

PR #2276, opened 2022-06-10
Replace ByteBufFormat with BufferFormat in tests

violetagg:
Related to #1873


violetagg:
@pderop Thanks for the review

PR #2303, opened 2022-06-16
Tests need adaptation to work with Netty 5 changes and the new Buffer API

violetagg:
`WebsocketTest#firefoxConnectionTest`

https://datatracker.ietf.org/doc/html/rfc6455#section-4.1

`The request MUST include a header field with the name Sec-WebSocket-Version.  The value of this header field MUST be 13.`

`The request MUST include a header field with the name Sec-WebSocket-Key.  The value of this header field MUST be a
nonce consisting of a randomly selected 16-byte value that has been base64-encoded (see Section 4 of [RFC4648]).
The nonce MUST be selected randomly for each connection.`


pderop (approved):
LGTM

violetagg:
@pderop Thanks for the review

PR #2310, opened 2022-06-20
Ensure the next request is started once the previous is terminated

violetagg:
The request termination happens also as a scheduled task in a listener, there is no ordering.

Related to #1873


pderop (approved):
LGTM

violetagg:
@pderop Thanks

PR #2331, opened 2022-06-27
Enable the disabled HTTP/2 tests

violetagg:
- `Http2StreamBridgeClientHandler` was missing `@Sharable`
- `Http2Pool` when disposing a channel, `close` should be invoked and not `closeFuture`
- When creating channel pipeline for `H2C`, `read` cannot be invoked on `handlerAdded`.
This is related to the changes in channel registration.

Related to #1873


violetagg:
@pderop Thanks

PR #2332, opened 2022-06-27
SimpleCompressionHandler must invoke super.decodeAndClose(...) only with request meta data when no payload/not accessible

violetagg:
`super.decodeAndClose(...)` is needed only for initialising `acceptEncodingQueue`

Related to #1873


violetagg:
@pderop Thanks

PR #2335, opened 2022-06-28
Ensure active connections is decremented

pderop:
This PR is proposing an additional improvement for the original PR #2237.

While debugging another problem in _HttpMetricsHandlerTest_, I found that when the _AbstractHttpServerMetricsHandler.write_ method is getting an exception thrown by _recordWrite_, then the _recordInactiveConnection_ is never called, see [here](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java#L114-L132).

So, this is too bad because the long adder that is incremented [from here](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java#L158) is then never decremented by the missed call to _recordInactiveConnection_, resulting in leaving the number of active connections to an inconsistent value.

So, in _HttpMetricsHandlerTest_, when the _testRecordingFailsServerSide_ or the _testRecordingFailsClientSide_ are executed, and if the _testServerConnectionsMicrometer_ is executed after, then it will fail because the number of active connections is left in a wrong state (after the _testRecordingFailsServerSide_ or _testRecordingFailsClientSide_ is executed):

```
08:32:38.898 [reactor-http-nio-5] ERROR r.n.http.server.HttpServerOperations - [95393cb0-1, L:/127.0.0.1:58225 - R:/127.0.0.1:58226] Error finishing response. Closing connection
org.opentest4j.AssertionFailedError: 
expected: 1.0
 but was: 5.0
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at reactor.netty.http.HttpMetricsHandlerTests.checkGauge(HttpMetricsHandlerTests.java:895)
	at reactor.netty.http.HttpMetricsHandlerTests.checkServerConnectionsMicrometer(HttpMetricsHandlerTests.java:738)
	at reactor.netty.http.HttpMetricsHandlerTests.lambda$setUp$7(HttpMetricsHandlerTests.java:144)
	at reactor.core.publisher.FluxPeek$PeekSubscriber.onNext(FluxPeek.java:185)
[...]

PR #2338, opened 2022-06-28
Remove netty5 Unsafe class

pderop:
This PR attempts to remove usage of Usafe (see netty PR from https://github.com/netty/netty/pull/12511).

Looking at the recent changes in the netty PR, it seems that calls like "channel.unsafe().closeForcibly()" are replaced by "channel.close();"

So, I simply did the same in this PR, but ... I'm not sure.
@violetagg , can you help on this please ?

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
@@ -437,7 +437,7 @@ public abstract class ServerTransport<T extends ServerTransport<T, CONF>,
 		}
 
 		static void forceClose(Channel child, Throwable t) {
-			child.unsafe().closeForcibly();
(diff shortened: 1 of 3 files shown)


violetagg (approved):

violetagg:
Related to #1873

pderop:
ok, I have temporarily disabled the _testRecordingFailsClientSide_ and _testRecordingFailsClientSide_ tests, and now, we have successful tests for both **ubuntu** and **macos**. So, I tend to think that some more works need to be done for the other #2335 PR, which is most likely to be again related (because if I disable the _testRecordingFailsClientSide_ and the _testRecordingFailsClientSide_, then tests are passing OK).

Now, the remaining failing test is the same one for windows from the #2338.

let's not merge this PR for the moment, I need to get back to the other #2335.

PR #2353, opened 2022-07-04
Make HttpMetricsHandlerTests stable

pderop:
This PR is an attempt to sabilise the HttpMetricsHandlerTests, because now in netty5, channel listeners execution is rescheduled at the end of the current event loop queue, see Netty PR [#9489 Always notify FutureListener via the EventExecutor](https://github.com/netty/netty/pull/9489)
So, this is problem for the HttpMetricsHandlerTests tests, because when we receive a response from a server, we start to verify metrics, while the metrics may be not be already up-to-date because they are updated using channel listeners, which may not be executed yet at the time we receive a response.
See for example [AbstractHttpServerMetricsHandler.java](https://github.com/reactor/reactor-netty/blob/netty5/reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java#L122-L148)

This PR ensures that no pending tasks are currently enqueued in event loops, before decrementing test latches, see new _scheduleCoundDown_ method.

### DIFF
--- a/reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
@@ -869,29 +869,14 @@ class HttpMetricsHandlerTests extends BaseHttpTest {
 	}
 
 	HttpServer customizeServerOptions(HttpServer httpServer, @Nullable ProtocolSslContextSpec ctx, HttpProtocol[] protocols,
-	                                  @Nullable AtomicReference<CountDownLatch> latch, boolean registerLatchHandlerOnConnection) {
+	                                  @Nullable AtomicReference<CountDownLatch> latchRef, boolean registerLatchHandlerOnConnection) {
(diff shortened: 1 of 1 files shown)


pderop:
@violetagg , please do not review for the moment. I'll finalize this PR a bit later. Now I'm working on the multipart PR.

reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
```diff
@@ -877,11 +865,39 @@ private void checkExpectationsBadRequest(String serverAddress, boolean checkTls)
...
+						return ctx.write(msg).addListener(future -> latch.get().countDown());
+					}
+					else {
+						return ctx.write(msg);
+					}
+				}
+
+				@Override
+				public boolean isSharable() {
+					return true;
```
violetagg, line 888: This is declared `sharable`. Can we reuse one and the same instance?

violetagg (changes requested):

violetagg (approved):

pderop:

PR #2379, opened 2022-07-13
[build] Separate publish build on several jobs

violetagg:


violetagg:
@simonbasle PTAL

.github/workflows/publish.yml
```diff
@@ -29,16 +29,61 @@ jobs:
...
         run: ./gradlew qualifyVersionGha
-      - name: run checks
-        id: checks
-        run: ./gradlew check
+      - name: run reactor-netty-core checks
+        id: reactor-netty-core-checks
+        run: ./gradlew reactor-netty-core:check
+  reactor-netty-http:
+    # Notes on reactor-netty-http: this job has no access to secrets, only github token. As a result, all non-core actions are centralized here
+    # This includes the tagging and drafting of release notes. Still, when possible we favor plain run of gradle tasks
```
simonbasle, line 37: nit: these two lines of comments are not entirely correct anymore. could keep only the first sentence

.github/workflows/publish.yml
```diff
@@ -101,7 +146,7 @@ jobs:
 
   tagMilestone:
     name: Tag milestone
-    needs: [ prepare, deployMilestone ]
+    needs: [ reactor-netty-core, reactor-netty-http, reactor-netty-http-brave, reactor-netty-incubator-quic, deployMilestone ]
```
simonbasle, line 149: I think that just `needs: [ deployMilestone, reactor-netty-core ]` would be enough since the deploy job itself needs all these others test-only jobs. it still needs to reference `reactor-netty-core` because it uses its output below in the tag message.

.github/workflows/publish.yml
```diff
@@ -111,12 +156,12 @@ jobs:
...
           git config --local user.name 'reactorbot'
           git config --local user.email '32325210+reactorbot@users.noreply.github.com'
-          git tag -m "Release milestone ${{ needs.prepare.outputs.fullVersion }}" v${{ needs.prepare.outputs.fullVersion }} ${{ github.sha }}
+          git tag -m "Release milestone ${{ needs.reactor-netty-core.outputs.fullVersion }}" v${{ needs.reactor-netty-core.outputs.fullVersion }} ${{ github.sha }}
           git push --tags
 
   tagRelease:
     name: Tag release
-    needs: [ prepare, deployRelease ]
+    needs: [ reactor-netty-core, reactor-netty-http, reactor-netty-http-brave, reactor-netty-incubator-quic, deployRelease ]
```
simonbasle, line 164: same comment here as in `tagMilestone`. limiting the number of needed jobs will also avoid a lot of extra connecting lines in the UI btw 😄

simonbasle (commented):
this looks like an easy path to parallelize the testing phase 👍 suggested a few improvements inline

.github/workflows/publish.yml
violetagg: fixed

.github/workflows/publish.yml
violetagg: fixed

.github/workflows/publish.yml
violetagg: fixed

pderop (dismissed):
LGTM

violetagg:
@pderop @simonbasle Thanks

PR #2380, opened 2022-07-13
[doc] Javadoc link in reference documentation points to a concrete version

violetagg:
With 3 parallel versions of Reactor Netty (1.0.x, 1.1.x and 2.0.x), the reference documentation
has to point to the concrete version of Reactor Netty for each it is built


pderop (approved):
LGTM

violetagg:
@pderop Thanks

PR #2388, opened 2022-07-15
Use the new API Observation#parentObservation

violetagg:
Make context propagation optional dependency

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -84,11 +84,11 @@ ext {
 	}
 
 	//Metrics
-	micrometerVersion = '1.10.0-M3' //optional baseline
-	micrometerTracingVersion = '1.0.0-M6' //optional baseline
-	micrometerDocsVersion = '1.0.0-M5' //optional baseline
+	micrometerVersion = '1.10.0-SNAPSHOT' //optional baseline
+	micrometerTracingVersion = '1.0.0-SNAPSHOT' //optional baseline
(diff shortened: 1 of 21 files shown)


marcingrzejszczak (approved):

pderop (approved):
LGTM

rstoyanchev (approved):
This makes more sense, avoiding the creation of a ThreadLocal. 

I know it's a separate question, but I was just wondering [is this](https://github.com/reactor/reactor-core/blob/5e9b08465f718352913f09f723b41ffae7e3f427/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/MicrometerObservationListener.java#L91) in reactor-core-micrometer a similar pattern that could also change?

violetagg:
> This makes more sense, avoiding the creation of a ThreadLocal.
> 
> I know it's a separate question, but I was just wondering [is this](https://github.com/reactor/reactor-core/blob/5e9b08465f718352913f09f723b41ffae7e3f427/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/MicrometerObservationListener.java#L91) in reactor-core-micrometer a similar pattern that could also change?

Yes we can use the `parentObservation` directly in `reactor-core-micrometer` also. /cc @simonbasle

violetagg:
@pderop @rstoyanchev @marcingrzejszczak Thanks for the review!

PR #2394, opened 2022-07-21
Remove the dependency to the context propagation library

violetagg:
@simonbasle @rstoyanchev @marcingrzejszczak PTAL


reactor-netty-core/build.gradle
```diff
@@ -106,8 +106,6 @@ dependencies {
 	compileOnly "io.micrometer:micrometer-core:$micrometerVersion"
 	compileOnly "io.micrometer:micrometer-tracing:$micrometerTracingVersion"
```
marcingrzejszczak, line 107: Actually `micrometer-tracing` has a compile dependency on context-propagation

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
```diff
@@ -297,4 +308,8 @@ public static String formatSocketAddress(@Nullable SocketAddress socketAddress)
 		}
 		return null;
 	}
+
+	public static Context updateContext(Context context, Object observation) {
+		return context.hasKey(OBSERVATION_KEY) ? context : context.put(OBSERVATION_KEY, observation);
```
marcingrzejszczak, line 313: So you're manually doing what the context propagation library already does, correct?

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerHttpClientMetricsHandler.java
```diff
@@ -113,8 +113,8 @@ protected void startWrite(HttpRequest msg, Channel channel, @Nullable ContextVie
 
 		responseTimeHandlerContext = new ResponseTimeHandlerContext(recorder, msg, path, channel.remoteAddress());
 		responseTimeObservation = Observation.createNotStarted(recorder.name() + RESPONSE_TIME, responseTimeHandlerContext, OBSERVATION_REGISTRY);
-		if (contextView != null && contextView.hasKey(ObservationThreadLocalAccessor.KEY)) {
-			responseTimeObservation.parentObservation(contextView.get(ObservationThreadLocalAccessor.KEY));
+		if (contextView != null && contextView.hasKey(OBSERVATION_KEY)) {
```
marcingrzejszczak, line 116: I would suggest writing a test that actually uses context-propagation (which should in effect update the Reactor context with this specific key) and then you would reuse that entry that is stored by the key you've copy pasted yourself. That way if for some reason anything changes in the context propagation repo in a breaking way then your tests should fail

marcingrzejszczak: Actually you don't even need that because the Key is there in `ObservationThreadLocalAccessor` which is in micrometer-core!

reactor-netty-core/build.gradle
violetagg: Isn't it `implementation` and not `api`?

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
violetagg: Yeah I need only this from the context propagation library and there is no need to bring additional dependency.

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
```diff
@@ -33,6 +36,7 @@
  */
 public class Metrics {
 	public static final MeterRegistry REGISTRY = io.micrometer.core.instrument.Metrics.globalRegistry;
+	public static final String OBSERVATION_KEY = "micrometer.observation";
```
simonbasle, line 39: could use the one from the TLAccessor:
```suggestion
	public static final String OBSERVATION_KEY = ObservationThreadLocalAccessor.KEY;
```
reading a static constant field shouldn't lead to class initialization so it should work without `context-propagation` on the classpath

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerHttpClientMetricsHandler.java
violetagg: `micrometer-observation` has only optional dependency. If I try to build my project without that dependency I'm receiving

```
> Task :reactor-netty-core:compileJava
error: cannot access ThreadLocalAccessor
  class file for io.micrometer.context.ThreadLocalAccessor not found
  Consult the following stack trace for details.
  com.sun.tools.javac.code.Symbol$CompletionFailure: class file for io.micrometer.context.ThreadLocalAccessor not found
1 error
```

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
violetagg: @simonbasle The build fails with the below if I don't have dependency

```
> Task :reactor-netty-core:compileJava
error: cannot access ThreadLocalAccessor
  class file for io.micrometer.context.ThreadLocalAccessor not found
  Consult the following stack trace for details.
  com.sun.tools.javac.code.Symbol$CompletionFailure: class file for io.micrometer.context.ThreadLocalAccessor not found
1 error
```

simonbasle: (I would defer to @rstoyanchev's expertise here though, I might have been wrong about that)

simonbasle: interesting... nevermind then

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerHttpClientMetricsHandler.java
simonbasle: let's go with the route of adding a test sanity checking the key is same as in `ObservationThreadLocalAccessor` 👍 
I'll do the same in reactor-core-micrometer

reactor-netty-core/build.gradle
marcingrzejszczak: It's `implementation`

simonbasle (approved):

violetagg:
@simonbasle PTAL

simonbasle (approved):

PR #2399, opened 2022-07-25
Upgrade to Gradle v7.5

violetagg:
According to the documentation the command below should be used for upgrade
`./gradlew wrapper --gradle-version=7.5`

Some checkstyle warnings are suppressed, other are fixed.


reactor-netty-core/src/main/java/reactor/netty/ChannelBindException.java
```diff
@@ -42,8 +42,10 @@ public static ChannelBindException fail(SocketAddress bindAddress, @Nullable Thr
 		if (cause instanceof java.net.BindException ||
 				// With epoll/kqueue transport it is
 				// io.netty.channel.unix.Errors$NativeIoException: bind(..) failed: Address already in use
+				//CHECKSTYLE:OFF
```
simonbasle, line 45: LGTM overall but wondering if the `UnnecessaryParentheses` checkstyle module can be configured to reduce these false positives (or removed at all) vs adding 10 suppressions.

doesn't `CHECKSTYLE:OFF` potentially suppresses all other checkstyle modules for these blocks?

simonbasle (approved):

violetagg:

violetagg:
This is superseded by #2420

PR #2411, opened 2022-07-29
Use pattern for the field name when checking the classpath

violetagg:
In preparation for better GraalVM support


simonbasle (approved):

violetagg:

Issue #1953, opened 2021-12-22
Transitive dependency on netty-codec-http in reactor-netty-core


shakuzen:
## Expected Behavior
reactor-netty-core does not have HTTP-related dependencies. We want to avoid security scanners picking up HTTP-related CVEs, so we are using reactor-netty-core.

## Actual Behavior
Recently we had [an issue reported](https://github.com/micrometer-metrics/micrometer/issues/2929) due to the inclusion of netty-codec-http classes.
reactor-netty-core has a transitive dependency on netty-codec-http [through netty-handler-proxy](https://github.com/netty/netty/blob/16f31f83f2b17dc9f450943bf84192cd28874e42/handler-proxy/pom.xml#L63-L67)

https://github.com/reactor/reactor-netty/blob/b3a473ef0a3150a0e84c911a0944bc6a28cbb4b3/reactor-netty-core/build.gradle#L55

## Steps to Reproduce

Add a dependency to reactor-netty-core to your project.

## Possible Solution
If reactor-netty-core works without the transitive dependency on netty-codec-http, excluding it in reactor-netty-core could make sense.

## Your Environment

* Reactor version(s) used: reactor-netty-core 1.0.14
* Other relevant libraries versions (eg. `netty`, ...): netty 4.1.72
* JVM version (`java -version`): 
* OS and version (eg. `uname -a`):

violetagg:
@shakuzen It will be backwards incompatible if we do that in 1.0.x as the current `ProxyProvider` needs `HttpHeaders`
https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java#L407-L417

I'm adding this to 2.0.x Backlog where we will be able to do an incompatible change.

violetagg:
@shakuzen This has been fixed with #2547

PR #2413, opened 2022-08-02
Introduce HttpClientProxyProvider with HTTP related configurations

violetagg:
This is in preparation for removing `netty5-codec-http` dependency from `reactor-netty-core`.
This dependency is transitively available via `netty-handler-proxy` dependency.

Related to #1953


pderop (approved):
LGTM

violetagg:

PR #2421, opened 2022-08-08
HttpMetricsHandlerTests testServerConnectionsMicrometer still unstable

### LINKED ISSUE #2368: Revise HttpMetricsHandlerTests#testServerConnectionsRecorder
This test is flaky and it is observed that it fails from time to time on the CI for all versions 1.0.x, 1.1.x, 2.0.x

pderop:
The HttpMetricsHandlerTests testServerConnectionsMicrometer is still unstable. 
Sometimes, the following assertion fails because the SERVER_CONNECTIONS_TOTAL is 1, [see this failing assertion](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java#L626) 

Reproduced the problem in github using netty5 branch while validating the multipart porting draft PR.

What I can imagine is the following probable scenario:

1. the testServerConnectionsMicrometer is finishing an HTTP2 test, so the tearDown is in progress, and the client connection pool is disposing, but the server has not yet seen the close, so the SERVER_CONNECTIONS_TOTAL meter is still set to 1
2. now, the testServerConnectionsMicrometer starts a next test using HTTP1.1, but when it waits for the H1 connection to be closed [here](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java#L625), then the SERVER_CONNECTIONS_TOTAL is still set to 2, because the connection close done in previous test in not yet seen by the server. 
3. So, when the HTTP1.1 socket is closed and see by the server, the latch is counted down, but the SERVER_CONNECTIONS_TOTAL is now set to 1, because the client socket close from previous test is still in progress, and not yet seen by the server.

So, in testServerConnectionsMicrometer, this problem seems to be fully fixed when we dispose the client connection pool in case protocol is HTT2 and when we also wait for the server to see the H2 connection close.

Fixes #2368

### DIFF
--- a/reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/HttpMetricsHandlerTests.java
@@ -591,10 +591,11 @@ class HttpMetricsHandlerTests extends BaseHttpTest {
 		boolean isHttp11 = clientProtocols.length == 1 && clientProtocols[0] == HttpProtocol.HTTP11;
 		HttpServer server = customizeServerOptions(httpServer, serverCtx, serverProtocols)
 				.metrics(true, Function.identity())
-				.doOnConnection(cnx -> responseSentHandler.register(responseSent, cnx.channel().pipeline()));
+				.doOnConnection(cnx -> {
+					serverCloseHandler.register(cnx.channel(), serverClosed, isHttp11);
+					responseSentHandler.register(responseSent, cnx.channel().pipeline());
+				});
 
-		server = isHttp11 ?
-				server.doOnChannelInit((cnxObs, ch, sockAddr) -> serverCloseHandler.register(ch, serverClosed, isHttp11)) : server;
 		disposableServer = server.bindNow();
 
 		AtomicReference<SocketAddress> clientAddress = new AtomicReference<>();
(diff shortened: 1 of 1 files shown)


violetagg:
When we dispose a server we also dispose the connection pool. It happens here https://github.com/reactor/reactor-netty/blob/netty5/reactor-netty5-core/src/main/java/reactor/netty5/transport/ServerTransport.java#L528-L543.

Do you think that the new test starts before server dispose?

pderop:
When _ServerTransport.dispose_  method call is done, the server connection is indeed closed (the server does not listen anymore on the server port),  that's ok. 

But when the dispose method returns, I think that the server has not yet seen the client connection close, it will eventually see it, but not immediately. That's what I remember having observed with the debugger last week.

pderop:
(the failure CI test on windows is unrelated)

violetagg (changes requested):
I'm wondering - do we need something like this:

```
private static final EventExecutor executor = new DefaultEventExecutor();

@AfterClass
public static void afterClass() throws Exception {
	executor.shutdownGracefully()
	        .get(5, TimeUnit.SECONDS);
}
```

pderop:
Indeed, I have updated the PR like in [ITTracingHttpServerDecoratorTest](Https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http-brave/src/test/java/reactor/netty/http/brave/ITTracingHttpServerDecoratorTest.java#L42-L48)
I also have closed the group, like it is done in [ITTracingHttpServerDecoratorTest.close](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http-brave/src/test/java/reactor/netty/http/brave/ITTracingHttpServerDecoratorTest.java#L56-L59) (isn't it also necessary ? else I can revert and just close the executor ?)

PS: in [HttpServerTests](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java#L1653-L1655), the executor and the group does not seem to be closed, does something should be fixed in  HttpServerTests.java (with another PR) ?

violetagg:
yes go ahead and fix them all

pderop:
The failed test on windows is not relevant to this PR.

PR #2432, opened 2022-08-15
Add reflection hints for better native-image support

### LINKED ISSUE #2419: Add native-image support
While working on the [`session-redis-webflux` smoke test](https://github.com/spring-projects/spring-aot-smoke-tests/tree/main/session-redis-webflux), i discovered that there are reflection hints missing to get reactor-netty working. The test uses Lettuce as a redis client and webflux as web framework. The hints I had to add [were these](https://github.com/spring-projects/spring-aot-smoke-tests/blob/main/aot-smoke-test-third-party-hints/src/main/java/org/springframework/aot/smoketest/thirdpartyhints/ReactorNettyHints.java).

It would be nice if reactor-netty could ship these out of the box, for example as `META-INF/native-image/reflect-config.json` files.

I'm also not sure if these cover all use cases, but they are enough to let the sample pass.

Without these hints you see netty complaining about reaching the end of the pipeline without handling the message and then dropping it.

If you don't want these hints in your codebase, please let me know. I will then contribute the hints to the https://github.com/oracle/graalvm-reachability-metadata.

Thanks!

violetagg:
When a `ChannelHandler` is added to the `Netty` channel pipeline and if the `ChannelHandler`
implements `ChannelInboundHandler`/`ChannelOutboundHandler`,
`io.netty.channel.ChannelHandlerMask#mask0` will inspect the `ChannelHandler's` methods for
`io.netty.channel.ChannelHandlerMask.Skip` annotation.
When there are no reflection hints, `NoSuchMethodException` is thrown when trying to query the methods with reflection.
This will lead to methods marked as not skippable, although the program will be able to run
without any problems, when all methods are marked as not skippable, one might observe performance issues.

Fixes #2419

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -122,6 +122,7 @@ ext {
 	junitPlatformLauncherVersion = '1.9.0'
 	mockitoVersion = '4.7.0'
 	blockHoundVersion = '1.0.6.RELEASE'
+	reflectionsVersion = '0.10.2'
 
 	javadocLinks = ["https://docs.oracle.com/javase/8/docs/api/",
 					"https://fasterxml.github.io/jackson-databind/javadoc/2.5/",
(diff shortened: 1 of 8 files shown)


sdeleuze:
This PR looks fine to me, there is no [conditional configuration](https://www.graalvm.org/22.0/reference-manual/native-image/Reflection/#konica-conditional-access-configuration) but since this is just configuration query method access, the impact on footprint should be low.

Also that's great to have tests checking no implementation is missed.

pderop (approved):
sounds good to me.

violetagg:
@sdeleuze @pderop Thanks for the review!

PR #2433, opened 2022-08-15
ServerTransport.Acceptor#exceptionCaught should not propagate the event

violetagg:
There are no other `ChannelHandlers` in the pipeline except `ServerTransport.Acceptor`, thus
it is not needed to propagate the event.


pderop (approved):
LGTM

violetagg:

PR #2441, opened 2022-08-19
ensures flux receiver terminates subscriber on dispose

OlegDokuka:
this fix addresses https://github.com/rsocket/rsocket-java/issues/1050


violetagg:
@OlegDokuka There is a test failure. Can you take a look?
Can you provide a test that shows the issue?

OlegDokuka:
@violetagg fixed

OlegDokuka:
@violetagg added a test that reflects the problem. It fails on the 1.0.x branch now

OlegDokuka:
@violetagg it seems that the other tcp-http related test fails because of the fix and I'm not sure how to clearly fix it...

violetagg (approved):

Issue #2425, opened 2022-08-09
Need HTTP request retry to be configurable


crankydillo:
Due to our infrastructure, we need to retry requests, primarily connection establishment related like connection failed/timed out, SSL handshake timeout, etc..  To date, we have tried doing this by adding retry to the following:

1. The `Mono` we return from the function given to `HttpClient#mapConnect`.
1. The `Mono` returned from the `ConnectionProvider#acquire` (via wrapping)

The first option is more promising (thanks @violetagg !), but it still comes with some negatives.  We're thinking the second will not work for all the cases we want to retry.

At this time, we're thinking the best approach is to make the retry more configurable.  Something like [Apache HTTP client retry interface](https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/), but not quite that opened ended.  I totally get that this is not a slam-dunk on whether or not this should be added given reactor-core's built-in retry concept, but the biggest issue we have using the `mapConnect` approach above is making sure we don't retry if any application data has been sent across the wire (e.g. request line, headers).  The only thing we currently have to solve that problem is to register a `ConnectionObserver`, but this seems a little too coupled with reactor-http implementation.

This was discussed in gitter [here](https://gitter.im/reactor/reactor-netty?at=60c999b08681550d4ef22dba).

The bulk of the retry code I refer to for `mapConnect` is:

```
final AtomicBoolean headersSent = new AtomicBoolean(false);

final Retry retryPolicy = Retry.from(companion -> companion.map(sig -> {
    if (!headersSent.get() && sig.totalRetries() < config.getMaxRequestRetry()) {
        return sig.totalRetries();
    }
    throw Exceptions.propagate(sig.failure());
}));

return httpClient
    .observe((connection, newState) -> {
        if (newState == ConnectionObserver.State.CONNECTED) {
            headersSent.set(true);
        }
    }).mapConnect(conn -> {
        return conn.retryWhen(retryPolicy);
    });
```

## Motivation

We have many cases where infrastructure problems can be 'worked around' by doing an immediate request retry.  These exceptional cases are broader than what reactor-http currently retries.

## Desired solution

Ability to customize retry behavior to suit our infrastructure.  Basically, to mimic what we were doing with Apache HTTP client:)

## Considered alternatives

Detailed above.

crankydillo:
[...]

PR #2453, opened 2022-08-27
Configurable request retry

crankydillo:
Allow more configuration when it comes to retrying requests.

This is currently failing at least 1 test for 2 reasons that need to be addressed.  First, `Retry#max` exhaustion generates the retryexhausted wrapper.  I think that's a good thing, but I understand why you'd go a different route from multiple perspectives (IO checked retry, breaking behavior).  Still thinking on that, but feel free to tell me what to do.

Also, in the error case, this fires the `doOnError` hook maxRetry + 1.  The additional 1 comes from the split out retryWhen for URL redirection.  I will try to deal with this

Need to stop for the day, but figured I'd get a draft out there.  I'll clean up all commits/code/etc. before moving out of draft.

#2425

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -63,6 +63,7 @@ import reactor.netty.tcp.TcpClient;
 import reactor.netty.transport.ClientTransport;
 import reactor.util.Metrics;
 import reactor.util.annotation.Nullable;
+import reactor.util.retry.RetrySpec;
 
 /**
  * An HttpClient allows building in a safe immutable way an http client that is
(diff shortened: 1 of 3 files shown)


violetagg:
@crankydillo I'm gonna check this next week

crankydillo:
Thanks Violeta.  Honestly, I need to work on it.  It was not as easy as initially thought:)  I can pass all the tests with my current code (not pushed yet), but I'm not happy about it, and I think there may be some corner cases I'm missing.  I'll do my best to work on it a bit over the next week.

crankydillo:
I pushed it, but I'm not happy about it.  It's going to be a busy week or two, so I'm not sure when I will come back to it.  Bottom line is I couldn't figure out how to avoid some huge redesign to the observers and still do this in a clean way.  The crux of the problem revolves around getting that done-retrying signal to the observers at the right time.  The only other idea I have is to have an internal, failing always sink-based flux that fails always embedded within `HttpObserver`, but that feels hacky, generates a bunch of unnecessary exceptions, and still has some problems.  The main one being not knowing exactly what prompts retry (since its configurable!).

crankydillo:
On a side note, I thought I'd already signed the Pivotal agreement.  I'll get it signed again I guess.

crankydillo:
@violetagg I'm just going to close this.  I understand the desire to leverage reactor's retry, but I don't see how that can quickly be done.  Given that my team is not demanding this feature, I doubt I'll be able to resolve this.

Issue #2452, opened 2022-08-26
websockt: Method WebSocketOutbound.sendString(Mono.just("asd11")).sendString(Mono.just("asd11")) is a problem


huchaogithup:
We use httpclient.create().websocket(): set the handle method. when calling WebSocketOutbound.sendString(Mono..just ("asd11"). sendString.(Mono.just ("asd11")), an error occurred on the websock service side

## Expected Behavior
The service side accepts ”asd11“ and ”asd11“

## Actual Behavior
The service side accepts ”asd11“ and  an error occurred ”websocket: unexpected reserved bits 0x60“

```java
@Test
public void repoCase() {
           HttpClient.create().websocket().uri("ws://localhost:8080")
                .handle((req,resp)->{
                    return resp.sendString(Mono.just("asd11")).sendString(Mono.just("asd11"));
                }).blockLast();
}
```

## Possible Solution
WebSocketOutbound.sendString(Mono.just("asd11")) return  ReactorNetty.OutboundThen without "stringToWebsocketFrame"

* Reactor version(s) used:  reactor-netty-http-1.0.22.jar
* JVM version (`java -version`): jdk17
* OS and version (eg. `uname -a`): windows10

pderop:
Hi @huchaogithup,

Chaining multiple send operations is discouraged and will not work as expected when send involves scalar or mono publishers, please check javadoc from NettyOutbound.

Now, could you clarify and indicate your use use ? it should be easy to do one single sendString call like _sendString(Flux.concat(…))_ or whatever ?

PS: I see you have created a PR (thank you for that), but as said, chaining/nesting send* calls is discouraged, so let's see if a reactive-chain can resolve your issue, thank you.

huchaogithup:
This is because the client needs to send the first message after the websockt handshake is completed.  for example `        HttpClient.create().websocket().uri("ws://localhost:8080")
                .handle((req,resp)-> resp.sendString(Mono.just("fist message..."))
                            .sendString(req.receive().map(s->{
                                //todo Business processing
                            }))).blockLast();`

huchaogithup:
Flux.concat(…)也是一种解决问题的方式，但是可能链式调用可能更符合一些习惯

huchaogithup:
Flux. Concat (...) is also a way to solve the problem, but chain calling may be more consistent with some habits

pderop:
Indeed, this use case seems to be easily addressed using Flux operators like Concat.
[...]

PR #2455, opened 2022-08-29
Optimize websocket chain call

huchaogithup:
https://github.com/reactor/reactor-netty/issues/2452
a solution to the current problem


pderop:
as said in #2452, I'm closing this PR because chaining/nesting send* method is discouraged and may not work as expected.
Please leverage the Reactor API and use a work around like Concat in order to achieve your use case, thanks.

Issue #2462, opened 2022-09-04
Use HttpConnectionProvider mutate method


yangbongsoo:
## Motivation

I want to make new `ConnectionProvider` using `defaultHttpClient.configuration().connectionProvider().mutate()` operation.
However I can't use `mutate()` method.
```java
  return new ReactorClientHttpConnector(reactorResourceFactory, defaultHttpClient -> {
    // builder value is always null.
    Builder builder = defaultHttpClient.configuration().connectionProvider().mutate();

  }
}
```

Because HttpConnectionProvider(defaultHttpClient type is HttpConnectionProvider) don't override `mutate()` method and default `mutate()` method return null.
```java
@FunctionalInterface
public interface ConnectionProvider extends Disposable {
	@Nullable
	default Builder mutate() {
		return null;
	}
}
```

## Desired solution
```java
private ReactorClientHttpConnector customize(maxIdleTimeoutMillis, maxLifeTimeoutMillis) {
  return new ReactorClientHttpConnector(reactorResourceFactory, defaultHttpClient -> {
    ConnectionProvider provider = defaultHttpClient.configuration().connectionProvider()
        .mutate()
        .name("provider")
        .maxIdleTime(Duration.ofMillis(maxIdleTimeoutMillis))
        .maxLifeTime(Duration.ofMillis(maxLifeTimeoutMillis))
        .build();

    HttpClient httpClient = HttpClient.create(provider)
  }
}
```
I made PR https://github.com/reactor/reactor-netty/pull/2463

## Considered alternatives
I have a question. [Violetagg told me](https://gitter.im/reactor/reactor-netty?at=6311c13c999499629345befb) that `defaultHttpClient.configuration().connectionProvider()` should return the original ConnectionProvider.

So I made below test code. I think that `configuredConnectionProvider.http1ConnectionProvider`  return the original ConnectionProvider. Isn't it? Please tell me if I'm mistaken.
```java
@Test
void configuredConnectionProviderShouldReturnTheOriginalConnectionProvider() {
	ConnectionProvider provider = ConnectionProvider.create("provider");
	try {
		HttpClient client = HttpClient.create(provider);
		HttpConnectionProvider configuredConnectionProvider =
				(HttpConnectionProvider) client.configuration().connectionProvider();
		assertThat(provider.getClass().getName())
				.isEqualTo(configuredConnectionProvider.http1ConnectionProvider.getClass().getName());
	}
	finally {
		provider.disposeLater()
[...]

PR #2463, opened 2022-09-04
Use HttpConnectionProvider mutate method (#2463)

yangbongsoo:
I made mutate method(overriding) in HttpConnectionProvider.
However I can't access `Builder` class directly. 
So I edit mutate method  in ConnectionProvider default method

relates: #2462

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpConnectionProvider.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpConnectionProvider.java
@@ -57,6 +57,15 @@ final class HttpConnectionProvider implements ConnectionProvider {
 	}
 
 	@Override
+	public Builder mutate() {
+		if (this.http1ConnectionProvider == null) {
+			return ConnectionProvider.super.mutate();
+		}
+
+		return this.http1ConnectionProvider.mutate();
+	}
+
+	@Override
 	public void disposeWhen(SocketAddress address) {
 		http1ConnectionProvider().disposeWhen(address);
 	}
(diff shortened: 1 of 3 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/client/HttpConnectionProvider.java
```diff
@@ -56,6 +56,15 @@ else if (http1ConnectionProvider == null) {
 		}
 	}
 
+	@Override
+	public Builder mutate() {
+		if (this.http1ConnectionProvider == null) {
+			return ConnectionProvider.super.mutate();
+		}
+
+		return this.http1ConnectionProvider.mutate();
```
yangbongsoo, line 65: I know that I haven't considered `h2ConnectionProvider` yet.
I'd like to hear the opinion of the maintainer.

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpConnectionProviderTest.java
```diff
@@ -92,4 +92,64 @@ void maxConnectionsPerHostDefaultConnectionProvider() {
...
+
+		ConnectionProvider mergedProvider = client1.configuration().connectionProvider()
+				.mutate()
+				.pendingAcquireTimeout(Duration.ofSeconds(2L))
+				.maxIdleTime(Duration.ofSeconds(2L))
+				.maxLifeTime(Duration.ofSeconds(20L))
+				.build();
+
+		assertThat(beforeProvider.maxConnections()).isEqualTo(mergedProvider.maxConnections());
+		assertThat(beforeProvider.name()).isEqualTo(mergedProvider.name());
```
yangbongsoo, line 133: I wanted to compare more values, but it was inaccessible.

violetagg:
@yangbongsoo My idea was not to implement `mutate` for `HttpConnectionProvider`, but the configuration to return the original `ConnectionProvider`. This means if we have

```
ConnectionProvider provider = ...
HttpClient client = HttpClient.create(provider);
```

Then the code below will return `provider` and not the wrapper `HttpConnectionProvider`

```
client.configuration().connectionProvider()
```

Wdyt?

yangbongsoo:
I made new PR as you recommand. https://github.com/reactor/reactor-netty/pull/2468
I hope we can make a better decision by looking at the code.

violetagg:
This is superseded by #2483

PR #2464, opened 2022-09-05
Replace Buffer/ByteBufHolder#touch with Resource#touch

violetagg:


violetagg:
@pderop Thanks for the review

PR #2470, opened 2022-09-08
Adapt SampleTestRunner change in micrometer tracing

ttddyy:
Accommodate the breaking change in `SampleTestRunner` introduced by https://github.com/micrometer-metrics/tracing/pull/57


violetagg:
@ttddyy Thanks for the PR!

Issue #2480, opened 2022-09-14
H2OrHttp11Codec may cause ALPN to fail in HttpClient in certain circumstances 


j-bahr:
## Expected Behavior

Under certain circumstances one may be required to talk to an HTTP proxy via TLS. When this happens a it's necessary to put a second SSLHandler at the head of the pipeline. This instance of the SslHandler will negotiate a secure session with the proxy. It's transparent to the rest of the http pipeline except when attempting to negotiate ALPN. 

```java
httpClient
.protocol(HttpProtocol.H2, HttpProtocol.HTTP11)
.doOnChannelInit(
            (observer, channel, remoteAddress) -> {
                channel
                    .pipeline()
                    .addFirst("proxy.sslhandler", proxySslContext.newHandler(channel.alloc()));
```

When the http client is configured to support support  HTTP 2 and HTTP1, the `H2OrHttp11Codec` *should* select the named `NettyPipeline.SslHandler` when trying to determine the negotiated protocol version. Selecting the SslHandler by name ensures we are checking the protocol negotiated with the remote endpoint not the local forward proxy. To do this I'd expect the `H2orHttp11Codec` to pull the SslHanlder by name, not by type. Asking for the handler by type will return the first SslHandler in the pipeline, causing ALPN selected protocol to be based on the protocol negotiated with the proxy, not the remote endpoint. 

## Actual Behavior

The actual behavior is shown below. We currently pull the SslHandler by type, which will always pull the first handler in the pipeline, the handler talking to the proxy. This can cause ALPN to select the wrong protocol when trying to talk H2 over a http 1.1 proxy.

**Current (incorrect) Code:**
```java
static final class H2OrHttp11Codec extends ChannelInboundHandlerAdapter {
 ....
		@Override
		public void channelActive(ChannelHandlerContext ctx) {
			SslHandler sslHandler = ctx.pipeline().get(SslHandler.class);		
			....
	}
```

## Steps to Reproduce
Unfortunately this requires a somewhat complex setup with a TLS based forward proxy that is not easy to setup or reproduce outside of my production environment. That being said, I have built a custom jar with the fix outlined below and verified it worked correct in my environment. The fix is fairly harmless and will continue to work as expect for all other standard workflows. 

## Possible Solution

**Expected Code:**
To fix this simply request the SslHandler by name, `NettyPipeline.SslHandler`, to get the right negotiated protocol. 
```java
[...]

PR #2488, opened 2022-09-15
Get SslHandler by name for ALPN in HttpClientChannelInitializer

j-bahr:
Switch to accessing the SslHandler by name in client ALPN. This should always return the SslHandler for remote endpoint. If we don't do this, the pipeline will always return the first SslHandler in the pipeline. When talking through a TLS enabled forward proxy there may be more than one SslHandler present in the pipeline. This change ensures that we always get the SslHandler for the remote HTTP endpoint and not an intermediary, ensuring we set the correct negotiated protocol.

Fixes reactor/reactor-netty#2480

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java
@@ -867,27 +867,30 @@ public final class HttpClientConfig extends ClientTransportConfig<HttpClientConf
 
 		@Override
 		public void channelActive(ChannelHandlerContext ctx) {
-			SslHandler sslHandler = ctx.pipeline().get(SslHandler.class);
-			if (sslHandler == null) {
+			ChannelHandler handler = ctx.pipeline().get(NettyPipeline.SslHandler);
+			if (handler instanceof SslHandler) {
+				SslHandler sslHandler = (SslHandler) handler;
+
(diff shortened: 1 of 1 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java
```diff
@@ -867,27 +867,30 @@ static final class H2OrHttp11Codec extends ChannelInboundHandlerAdapter {
...
+				}
+				else {
+					throw new IllegalStateException("unknown protocol: " + protocol);
+				}
 
-			ctx.channel().pipeline().remove(this);
+				ctx.fireChannelActive();
+
+				ctx.channel().pipeline().remove(this);
+			} else {
```
violetagg, line 891: Make checkstyle happy
`[checkstyle] [ERROR] /.../reactor-netty/reactor-netty/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java:891:25: '}' at column 4 should be alone on a line. [RightCurly]`

```suggestion
			}
			else {
```

violetagg (changes requested):

violetagg:
@reactor/netty-team PTAL

violetagg:
@reactor/netty-team The PR is targeting `main` branch. I will manually apply it to `1.0.x` branch.

pderop (approved):
it looks good to me, thanks

violetagg:
@j-bahr Thanks for the PR.
I've merged the fix to `1.0.x` branch (f8f9a3d) and then forward merged to `main` branch (2662ecc) and `netty5` branch (a51c5b7)

PR #2495, opened 2022-09-21
Set the trailing headers directly when constructing DefaultLastHttpContent

violetagg:
Fix the log message


pderop (approved):
LGTM

violetagg:

PR #2496, opened 2022-09-21
Defer the creation of the request/response object

violetagg:
If there is no subscription we may not need this object.


pderop (approved):
approved.

violetagg:

PR #2505, opened 2022-09-28
[test] Adapt to Netty change to drop the data when it is unexpected

violetagg:
A response with `Content-Length: 0` is already sent so any other additional data will be dropped.
https://github.com/netty/netty/pull/12836


pderop (approved):
LGTM

violetagg:
@pderop Thanks

Issue #2509, opened 2022-09-28
SSLEngine closed already; nested exception is io.netty.handler.ssl.SslClosedEngineException


anmol1vw13:
Hi @pderop,
I'm facing this issue with negligible load. 

Stack: 
spring-boot-starter-parent 2.5.6
spring-boot-starter-webflux 2.5.6

Error: `The connection observed an error io.netty.handler.ssl.SslClosedEngineException: SSLEngine closed already`

Here's the trace

```
[id:31e016c9-2, L:/x.xx.xxx.xx:56499 - R:xxx.payu.in/xy.xyy.xyyy.xyy:443] The connection observed an error
io.netty.handler.ssl.SslClosedEngineException: SSLEngine closed already
    at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:861)
    at io.netty.handler.ssl.SslHandler.wrapAndFlush(SslHandler.java:800)
    at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:781)
    at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:750)
    at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:742)
    at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:728)
    at io.netty.handler.logging.LoggingHandler.flush(LoggingHandler.java:304)
    at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:750)
    at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:742)
    at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:728)
    at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.flush(CombinedChannelDuplexHandler.java:531)
    at io.netty.channel.ChannelOutboundHandlerAdapter.flush(ChannelOutboundHandlerAdapter.java:125)
    at io.netty.channel.CombinedChannelDuplexHandler.flush(CombinedChannelDuplexHandler.java:356)
    at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:750)
[...]

Issue #2499, opened 2022-09-21
io.netty.handler.ssl.SslClosedEngineException: SSLEngine closed already


amit84g:
reactor-netty: 1.0.23
Spring Boot: 2.7.2

## Actual Behavior: Getting below exception under heavy load:
++++++++++++++++++++
io.netty.handler.ssl.SslClosedEngineException: SSLEngine closed already
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:860) ~[netty-handler-4.1.79.Final.jar!/:4.1.79.Final]
	at io.netty.handler.ssl.SslHandler.wrapAndFlush(SslHandler.java:799) ~[netty-handler-4.1.79.Final.jar!/:4.1.79.Final]
	at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:780) ~[netty-handler-4.1.79.Final.jar!/:4.1.79.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:750) ~[netty-transport-4.1.79.Final.jar!/:4.1.79.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:742) ~[netty-transport-4.1.79.Final.jar!/:4.1.79.Final]
	at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:728) ~[netty-transport-4.1.79.Final.jar!/:4.1.79.Final]
	at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.flush(CombinedChannelDuplexHandler.java:531) ~[netty-transport-4.1.79.Final.jar!/:4.1.79.Final]
	at io.netty.channel.ChannelOutboundHandlerAdapter.flush(ChannelOutboundHandlerAdapter.java:125) ~[netty-transport-4.1.79.Final.jar!/:4.1.79.Final]
	at io.netty.channel.CombinedChannelDuplexHandler.flush(CombinedChannelDuplexHandler.java:356) ~[netty-transport-4.1.79.Final.jar!/:4.1.79.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:750) ~[netty-transport-4.1.79.Final.jar!/:4.1.79.Final]
[...]

PR #2518, opened 2022-09-29
Close connection when a TLS close_notify is received and acknowledged.

### LINKED ISSUE #2498: Channel inactive not fired after SslCloseCompletionEvent(SUCCESS) causing connection still in pool hence io.netty.handler.ssl.SslClosedEngineException: SSLEngine closed already
## Expected Behavior

SslCloseCompletionEvent(SUCCESS) not correctly handled is causing connection still present in pool hence "io.netty.handler.ssl.SslClosedEngineException: SSLEngine closed already" thrown on next usage of the connection.
A Channel 'INACTIVE' shoud be fired after SslCloseCompletionEvent(SUCCESS)

Here is a fine example of a multi used (7 times) connection :
At the end, the connection is unregistered as soon as the sslCloseCompletion is fired :
image

Full logs : [OK.log](https://github.com/reactor/reactor-netty/files/9618367/OK.log)

## Actual Behavior

In undetermined circumstances, SslCloseCompletionEvent is not followed by INACTIVE 
In this example, the connection is reused 5min06 after SslCloseCompletionEvent
Contrary to what happened in "good" example above there is only one 'READ COMPLETE' event fired.
May be this is the reason why fireChannelInacctive is not triggered.
image

Full logs :  [NOK.log](https://github.com/reactor/reactor-netty/files/9618373/NOK.log)

## Steps to Reproduce

Unfortunatly, I have no example to provide since this happen randomly.

## Possible Solution

Bad idea, the SslReadHandler is not in the pipeline anymore when this happen.
~~In reactor.netty.tcp.SslProvider.SslReadHandler.userEventTriggered~~
image
~~add~~
```java
                        if (evt instanceof SslCloseCompletionEvent) {
                            ctx.fireChannelInactive();
                        }
```

## Your Environment

* Reactor version(s) used: reactor-netty-http : 1.0.23
* Other relevant libraries versions (eg. `netty`, ...): netty : 4.1.81.Final
* JVM version (`java -version`):  openjdk version "1.8.0_342"
* OS and version (eg. `uname -a`): Linux 18.04.1-Ubuntu SMP Mon Aug 23 23:07:49 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

pderop:
When a _TLS close_notify_ is received and when the _close_notify_ ack is replied, then the Netty _SslHandler_ just does not close the channel. Instead of that, in that case it fires a _SslCloseCompletionEvent.SUCCESS_ event down the pipeline, and the closing of the channel is left to other handlers.

And since the [TLS RFC](https://www.rfc-editor.org/rfc/rfc5246#section-7.2.1) states that _the party receiving a close_notify MUST respond with a close_notify alert of its own and close down the connection immediately_, let's do this in Reactor-Netty.

This patch won't avoid AbortedExceptions or PrematureCloseException exceptions in case a close_notify is being received while a request is being written, or while a response is waited for, but it may address some of the pain highlighted by #2498, #2499, and #2509; and may avoid _SslClosedEngineException: SSLEngine closed already_ " exceptions.

When a close_notify is received and acknowledged, a log is now displayed just before closing the connection (in DEBUG):
```
09:09:16.948 [reactor-http-nio-3] DEBUG r.n.channel.ChannelOperationsHandler - [3a3ece28-1, L:/x.x.x.x:49624 - R:/0:0:0:0:0:0:0:0:49623] Received a TLS close_notify, closing the channel now.
```

There are three tests, which are trying to reproduce some corner cases, for example, like in #2498, where a server sends a close_notify to the client but does not close the connection right after (maybe later).

- test2498_close_notify_after_response_two_clients(): the test simulates a particular situation where a connection is idle and available in the client connection pool, and then the client receives a close_notify, but the server has not yet closed the connection (and may close it later). In this case, the connection is now closed immediately and removed from the pool, in order to avoid any _SslClosedEngineException: SSLEngine closed already exception_ the next time the connection will be acquired and written. So, in the test, a client sends a first request to a server, which replies correctly. But once the last response is flushed, then the server sends a close_notify without closing the connection. In this case, on the client side, when the close_notify is received and acknowledged, then the connection is immediately closed and removed from the pool. Then a second request is sent, using a fresh new connection (because the previous one has been closed when the close_notify was received. The patch in this particular case avoids the _SslClosedEngineException: SSLEngine closed already exception_

- test2498_close_notify_on_connect(): the test simulates a situation where the client is receiving a close_notify while writing request body parts to the connection. To simulate this, the server immediately sends a close_notify when the client is connecting. The request is not read and the client connection is not closed. The client is now closing the connection when receiving the close_notify and the request will be aborted with a PrematureCloseException or an AbortedException while writing the request body parts (but won't  _SslClosedEngineException: SSLEngine closed already_ exceptions anymore).

- test2498_close_notify_on_request(): the test simulates a situation where the client has fully written its request to a connection, but while waiting for the response, then a close_notify is received, and the server has not yet closed the connection (it may close it later). So in the test, the server fully reads the client request, but does not respond and sends a close_notify without closing the socket. The client, when receiving the close_notify, will send the close_notify ack and will close the channel, and the request will be aborted immediately with a PrematureCloseException.

Fixes #2498,#2499,#2509

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2011-2022 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
@@ -94,6 +94,9 @@ final class ChannelOperationsHandler extends ChannelInboundHandlerAdapter {
 			// see https://www.rfc-editor.org/rfc/rfc5246#section-7.2.1, which states that when receiving a close_notify,
 			// then the connection must be closed down immediately.
 			if (sslCloseCompletionEvent.isSuccess() && ctx.channel().isActive()) {
+				if (log.isDebugEnabled()) {
+					log.debug(format(ctx.channel(), "Received a TLS close_notify, closing the channel now."));
+				}
 				ctx.close();
 			}
 		}
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
@@ -84,7 +84,7 @@ final class ChannelOperationsHandler extends ChannelInboundHandlerAdapter {
 
 	@Override
 	@SuppressWarnings("FutureReturnValueIgnored")
-	public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+	public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
 		if (evt instanceof SslCloseCompletionEvent) {
 			SslCloseCompletionEvent sslCloseCompletionEvent = (SslCloseCompletionEvent) evt;
 
@@ -100,7 +100,6 @@ final class ChannelOperationsHandler extends ChannelInboundHandlerAdapter {
 				ctx.close();
 			}
 		}
-		ctx.fireUserEventTriggered(evt);
 	}
 
 	@Override
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
@@ -21,6 +21,7 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.handler.codec.DecoderResult;
 import io.netty.handler.codec.DecoderResultProvider;
+import io.netty.handler.ssl.SslCloseCompletionEvent;
 import io.netty.util.IllegalReferenceCountException;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.ReferenceCounted;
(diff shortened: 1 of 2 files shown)


pderop:
@violetagg , can you please have a look ? thanks.

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -148,6 +153,57 @@ class HttpServerTests extends BaseHttpTest {
...
+	 */
+	final static class IgnoreCloseNotifyHandler extends ChannelInboundHandlerAdapter {
+		final static String NAME = "handler.ignore_close_notify";
+
+		static void register(Connection cnx) {
+			cnx.channel().pipeline().addBefore(NettyPipeline.ReactiveBridge, NAME, new IgnoreCloseNotifyHandler());
+		}
+
+		@Override
+		public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
```
violetagg, line 200: Exception is never thrown in the method

```suggestion
		public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
```

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -148,6 +153,57 @@ class HttpServerTests extends BaseHttpTest {
...
+				promise.addListener(future -> sslHandler.closeOutbound().addListener(f -> latch.countDown()));
+			}
+			ctx.write(msg, promise);
+		}
+	}
+
+	/**
+	 * Handler used by secured servers which don't want to close client connection when receiving a client close_notify ack.
+	 * The handler is placed just before the ReactiveBrigde (HttpOperationsHandler), and will block
+	 * any received SslCloseCompletionEvent events. Hence, HttpOperationsHandler won't get the close_notify ack,
```
violetagg, line 189:
```suggestion
	 * The handler is placed just before the ReactiveBridge (ChannelOperationsHandler), and will block
	 * any received SslCloseCompletionEvent events. Hence, ChannelOperationsHandler won't get the close_notify ack,
```

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
```diff
@@ -81,6 +82,27 @@ final public void channelInactive(ChannelHandlerContext ctx) {
 		}
 	}
 
+	@Override
+	@SuppressWarnings("FutureReturnValueIgnored")
+	public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
```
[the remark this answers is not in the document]
pderop: fixed.

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
```diff
@@ -81,6 +82,27 @@ final public void channelInactive(ChannelHandlerContext ctx) {
...
+			// see https://www.rfc-editor.org/rfc/rfc5246#section-7.2.1, which states that when receiving a close_notify,
+			// then the connection must be closed down immediately.
+			if (sslCloseCompletionEvent.isSuccess() && ctx.channel().isActive()) {
+				if (log.isDebugEnabled()) {
+					log.debug(format(ctx.channel(), "Received a TLS close_notify, closing the channel now."));
+				}
+				ctx.close();
+			}
+		}
+		ctx.fireUserEventTriggered(evt);
```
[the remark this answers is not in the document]
pderop: fixed.

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
pderop: fixed.

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
pderop: fixed.

pderop:
applied feedbacks, can you recheck please ?

violetagg (approved):

violetagg:
ship it

PR #2524, opened 2022-10-03
Use BufferAllocator#compose(...) for creating CompositeBuffer

violetagg:
When possible use `BufferAllocator#compose(...)` for creating `CompositeBuffer` instead of `CompositeBuffer#extendWith` as the first performs better.
More information https://github.com/netty/netty/issues/12757

### DIFF
--- a/reactor-netty5-core/src/main/java/reactor/netty5/BufferFlux.java
+++ b/reactor-netty5-core/src/main/java/reactor/netty5/BufferFlux.java
@@ -34,8 +34,9 @@ import java.nio.channels.FileChannel;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
 import java.util.Objects;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 import static io.netty5.buffer.DefaultBufferAllocators.preferredAllocator;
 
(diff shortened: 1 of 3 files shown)


violetagg:
The failing tests on CI are not related

reactor-netty5-core/src/main/java/reactor/netty5/BufferFlux.java
```diff
@@ -255,24 +255,28 @@ public final Flux<String> asString(Charset charset) {
 	@SuppressWarnings("rawtypes")
 	public final BufferMono aggregate() {
 		return Mono.defer(() -> {
-					CompositeBuffer output = CompositeBuffer.compose(alloc);
+					AtomicReference<Buffer> refOutput = new AtomicReference<>();
```
OlegDokuka, line 258: do we care about allocation overhead? if so how about making a class like 

```java
static class DefferedBlablabla extends AtomicReference<Buffer> implements Supplier<Mono<...>> {

	public Mono<...> get() {
	     return return map(Buffer::send)
							.collectList()
							.doOnDiscard(Send.class, Send::close)
							.handle((list, sink) -> {
								Buffer output = null;
								if (!list.isEmpty()) {
									output = alloc.compose(list);
									set(output);
								}
								if (output != null && output.readableBytes() > 0) {
									sink.next(output);
								}
								else {
									sink.complete();
								}
							})
							.doFinally(signalType -> {
							    var instance =  get()
								if (instance != null) {
									instance.close();
								}
							});
	}

}

**Review by OlegDokuka — APPROVED** [review]
Looks good with minor improvements

File: reactor-netty5-core/src/main/java/reactor/netty5/BufferFlux.java (hunk 1 above)
Comment (OlegDokuka) [reply reactor-netty5-core/src/main/java/reactor/netty5/BufferFlux.java:258]: then it can be used as 

```java
return Mono.defer(new DefferedBlablabla(sink))...
```

violetagg: @OlegDokuka Please check the new commit

reactor-netty5-core/src/main/java/reactor/netty5/BufferFlux.java
```diff
@@ -331,4 +312,46 @@ static BufferFlux maybeFuse(Flux<Buffer> source, BufferAllocator allocator) {
...
+						}
+						if (output != null && output.readableBytes() > 0) {
+							sink.next(output);
+						}
+						else {
+							sink.complete();
+						}
+					})
+					.doFinally(signalType -> {
+						Buffer buffer = COMPOSITE_BUFFER.get(this);
```
OlegDokuka, line 350:
```suggestion
						Buffer buffer = this.compositeBuffer;
```

reactor-netty5-core/src/main/java/reactor/netty5/BufferFlux.java
```diff
@@ -331,4 +312,46 @@ static BufferFlux maybeFuse(Flux<Buffer> source, BufferAllocator allocator) {
...
+		@SuppressWarnings("rawtypes")
+		public Mono<Buffer> get() {
+			return source.map(Buffer::send)
+					.collectList()
+					.doOnDiscard(Send.class, Send::close)
+					.<Buffer>handle((list, sink) -> {
+						Buffer output = null;
+						if (!list.isEmpty()) {
+							output = alloc.compose(list);
+							COMPOSITE_BUFFER.set(this, output);
```
OlegDokuka, line 340:
```suggestion
							this.compositeBuffer = output;
```

OlegDokuka (approved):

violetagg:
@OlegDokuka Thanks for the review!

PR #2547, opened 2022-10-17
Remove transitive dependency netty5-codec-http from reactor-netty5-core

### LINKED ISSUE #1953: Transitive dependency on netty-codec-http in reactor-netty-core

violetagg:
Move HTTP proxy functionality from `ProxyProvider` to `HttpClientProxyProvider`.

Fixes #1953

### DIFF
--- a/reactor-netty5-core/build.gradle
+++ b/reactor-netty5-core/build.gradle
@@ -58,7 +58,9 @@ dependencies {
 	compileOnly "com.google.code.findbugs:jsr305:$jsr305Version"
 
 	api "io.netty:netty5-handler:$nettyVersion"
-	api "io.netty.contrib:netty-handler-proxy:$nettyContribVersion"
+	api("io.netty.contrib:netty-handler-proxy:$nettyContribVersion") {
+		exclude module: "netty5-codec-http"
+	}
 	api "io.netty:netty5-resolver-dns:$nettyVersion"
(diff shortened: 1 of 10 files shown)


pderop (approved):
LGTM

violetagg:
@pderop Thanks

PR #2556, opened 2022-10-26
Fix memory leak in HttpClient#sendForm for file uploads with streams

violetagg:
Create file uploads with streams always via `DefaultHttpDataFactory`. This ensures that they are added to `DefaultHttpDataFactory#requestFileDeleteMap` and thus they will be cleaned at the end of the operation.


pderop (approved):
+1, LGTM !

violetagg:

PR #2557, opened 2022-11-01
Allow configurable connection provider eviction predicate

samueldlightfoot:
Add support for a custom eviction predicate in the ConnectionProvider.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
@@ -34,6 +34,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeoutException;
 import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
@@ -278,6 +279,48 @@ public interface ConnectionProvider extends Disposable {
 		return null;
 	}
 
+	interface ConnectionMetadata {
+
+		/**
+		 * Returns the number of times the connection has been acquired from the pool. Returns 1 the first time the
+		 * connection is allocated.
+		 *
+		 * @return the number of times the connection has been acquired
+		 */
+		int acquireCount();
+
+		/**
+		 * Returns the time in ms that the connection has been idle.
+		 *
+		 * @return connection idle time in ms
+		 */
+		long idleTime();
+
+		/**
+		 * Returns the age of the connection in ms.
+		 *
+		 * @return connection age in ms
+		 */
+		long lifeTime();
+
+		/**
+		 * Returns a timestamp that denotes the order in which the connection was last released, to millisecond
+		 * precision.
+		 *
+		 * @return the connection last release timestamp, or zero if currently acquired
+		 */
+		long releaseTimestamp();
+
+		/**
+		 * Returns a timestamp that denotes the order in which the connection was created/allocated, to millisecond
+		 * precision.
+		 *
+		 * @return the connection creation timestamp
+		 */
+		long allocationTimestamp();
+
+	}
+
 	interface AllocationStrategy<A extends AllocationStrategy<A>> {
 
 		/**
@@ -476,6 +519,7 @@ public interface ConnectionProvider extends Disposable {
 		Supplier<? extends ConnectionProvider.MeterRegistrar> registrar;
 		BiFunction<Runnable, Duration, Disposable> pendingAcquireTimer;
 		AllocationStrategy<?> allocationStrategy;
+		BiPredicate<Connection, ConnectionMetadata> evictionPredicate;
 
 		/**
 		 * Returns {@link ConnectionPoolSpec} new instance with default properties.
@@ -501,6 +545,7 @@ public interface ConnectionProvider extends Disposable {
 			this.registrar = copy.registrar;
 			this.pendingAcquireTimer = copy.pendingAcquireTimer;
 			this.allocationStrategy = copy.allocationStrategy;
+			this.evictionPredicate = copy.evictionPredicate;
 		}
 
 		/**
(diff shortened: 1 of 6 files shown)


reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -586,6 +590,22 @@ public final SPEC maxLifeTime(Duration maxLifeTime) {
...
+		 * Default to null implying connections will not be evaluated for eviction. The eviction predicate works
+		 * alongside {@link #maxLifeTime(Duration)} and {@link #maxIdleTime(Duration)}.
+		 * <p><strong>Note:</strong> This configuration is not applicable for {@link reactor.netty.tcp.TcpClient}.
+		 * A TCP connection is always closed and never returned to the pool.
+		 *
+		 * @param evictionPredicate The predicate function that evaluates whether a connection should be evicted
+		 * @return {@literal this}
+		 * @throws NullPointerException if evictionPredicate is null
+		 */
+		public final SPEC evictionPredicate(BiPredicate<Connection, PooledRefMetadata> evictionPredicate) {
```
violetagg, line 604: @samueldlightfoot We don't expose Reactor Pool API directly so we need something provided by Reactor Netty. Check `AllocationStrategy` ...

violetagg (changes requested):

reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
```diff
@@ -442,12 +444,21 @@ PoolBuilder<T, PoolConfig<T>> newPoolInternal(
...
 					           .maxPendingAcquire(pendingAcquireMaxCount)
 					           .evictInBackground(evictionInterval);
 
+			BiPredicate<T, PooledRefMetadata> poolEvictionPredicate = evictionPredicate
+					.or((poolable, meta) -> (maxIdleTime != -1 && meta.idleTime() >= maxIdleTime)
+							|| (maxLifeTime != -1 && meta.lifeTime() >= maxLifeTime));
+
+			if (this.evictionPredicate != null) {
+				poolBuilder = poolBuilder.evictionPredicate(poolEvictionPredicate
+						.or((poolable, meta) -> this.evictionPredicate.test(poolable, new PooledConnectionMetadata(meta))));
```
samueldlightfoot, line 456: @violetagg let me know if you're happy with the extra allocation

samueldlightfoot: Good point - agree we should fully override the default if user supplied.

samueldlightfoot: Do you agree maxLifeTime and maxIdleTime should be applied in addition to either the default or user-supplied eviction predicate if they are specified?

violetagg: hm yeah ... I don't know ... `AllocationStrategy` for example disables the max connection setting

Also are we targeting HTTP/2 because the pool implementation there is more complicated from connections/streams handling point of view and the eviction is not trivial.

reactor.netty.http.client.Http2Pool#evictInBackground
reactor.netty.http.client.Http2Pool#destroyPoolable
reactor.netty.http.client.Http2Pool#findConnection

samueldlightfoot: It does make sense in my mind to also target HTTP/2. What do you think?

violetagg: yes

You may need to rebase this work on top of the latest changes (in case you see tests failing)

samueldlightfoot: Hooked into HTTP2 pool now.

violetagg:
@samueldlightfoot I'll check this in the next days

violetagg:
@samueldlightfoot For HTTP/1.1 you ignore the default eviction predicate and max idle/life time, but for HTTP/2 you keep them. Why do you introduce this difference?

samueldlightfoot:
> @samueldlightfoot For HTTP/1.1 you ignore the default eviction predicate and max idle/life time, but for HTTP/2 you keep them. Why do you introduce this difference?

I'll get this fixed over the weekend

samueldlightfoot:
@violetagg Do we want to keep separate checks for idle and max lifetime or can we integrate into the eviction predicate like SimpleDequePool?

samueldlightfoot:
> @samueldlightfoot Is this ready for a review?

Yes please

reactor-netty-http/src/test/java/reactor/netty/http/client/Http2PoolTest.java
```diff
@@ -770,6 +856,79 @@ void maxLifeTime() throws Exception {
 		}
 	}
 
+	@Test
+	void evictionPredicate() throws Exception {
```
violetagg, line 860: Exception is never thrown in the method

```suggestion
	void evictionPredicate() {
```

violetagg (changes requested):
Nice PR! Thanks!

I have just two small things which are not related to the functionality.

samueldlightfoot:
Committed both changes - thanks for the review. :-)

violetagg (approved):
@reactor/netty-team PTAL

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -635,12 +603,8 @@ Slot findConnection(ConcurrentLinkedQueue<Slot> resources) {
...
 
-	boolean maxIdleReached(Slot slot) {
-		return maxIdleTime != -1 && slot.idleTime() >= maxIdleTime;
-	}
-
-	boolean maxLifeReached(Slot slot) {
-		return maxLifeTime != -1 && slot.lifeTime() >= maxLifeTime;
+	boolean testEvictionPredicate(Slot slot) {
+		return poolConfig.evictionPredicate().test(slot.connection, new PooledSlotMetadata(slot));
 	}
```
pderop, line 608: @samueldlightfoot , 

First of all, thanks for the hard work !

I have one comment: since I'm currently working on optimizing the Http2Pool class, which is a critical component, is it possible to avoid allocating the `PooledSlotMetadata` from the testEvictionPredicate method line 607 ? 

Instead of that, it seams feasible to manage to make the `Slot` inner class implementing itself the PooledRefMetadata interface, you could then avoid the allocation line 607 and replace it by:

```
		return poolConfig.evictionPredicate().test(slot.connection, slot);
```

WDYT ?

samueldlightfoot: Great suggestion - I have pushed the changes.

violetagg:
@samueldlightfoot I merge this in `1.0.x` branch (c8e30ba) and then forward merged into `main` (3b00086) and `netty5` (adef851)

Issue #2560, opened 2022-11-02
Return FutureMono compatible for Netty5


violetagg:
This PR #2250 needs to be reverted and adapted for Netty5 API

This PR needs to be reverted and disabled test to be enabled again

violetagg:
We need to address the `TODO` around removing listener

PR #2561, opened 2022-11-02
Temporary disable tests that observe CancellationException

violetagg:
More info #2560


violetagg:
This won't be working ...

PR #2565, opened 2022-11-03
Use FutureMono instead of Mono#fromCompletionStage

violetagg:
Reactor Netty does not need the type of cancellation provided by `Mono#fromCompletionStage`

### DIFF
--- a/reactor-netty5-core/src/main/java/reactor/netty5/FutureMono.java
+++ b/reactor-netty5-core/src/main/java/reactor/netty5/FutureMono.java
@@ -1,0 +1,204 @@
+/*
+ * Copyright (c) 2011-2022 VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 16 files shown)


reactor-netty5-core/src/main/java/reactor/netty5/FutureMono.java
```diff
@@ -0,0 +1,204 @@
...
+
+		// add listener to the future to propagate on complete when future is done
+		// addListener likely to be thread safe method
+		future.addListener(fs);
+
+		// check once again if is cancelled to see if we need to removeListener in case addListener racing with
+		// subscription.cancel (which should remove listener)
+		if (fs.cancelled) {
+			// future.removeListener(fs); TODO
+		}
```
pderop, line 148: Did you leave the line 147 commented on purpose ?
I ask this because apparently, listeners can't be removed anymore from Future in netty5. So is there something else that still needs to be implemented in case the subscription is cancelled ? If not, then can we remove the lines between L144-L148 ?

violetagg: For the moment we will keep these TODOs until we are sure that we don't need them. This will happen post M3 release.

pderop (approved):
LGTM

violetagg:
@pderop Thanks for the review

PR #2585, opened 2022-11-23
1.0.x improve logging when connection closed after cancel

### LINKED ISSUE #2583: Improve logging when connection closed due to cancel signal
In a pipeline with multiple client calls, it's possible for a cancel signal to close the connection, e.g. `Mono.zip` where one publisher completes empty. In this case, you might see a log messages for subscribing to the response stream, immediately followed by channel closing:

```
2022-11-18 08:58:52.678 DEBUG 41948 --- [tor-http-nio-16] reactor.netty.channel.FluxReceive        : FluxReceive{pending=0, cancelled=false, inboundDone=false, inboundError=null}: subscribing inbound receiver
2022-11-18 08:58:52.678 DEBUG 41948 --- [tor-http-nio-16] r.n.resources.PooledConnectionProvider   : Channel closed, now: 9 active connections, 136 inactive connections and 0 pending acquire requests.
```

It would have been helpful in this case to see a 3rd log message in between that shows there was a cancel signal. That would make it clear that something within the application pipeline caused the connection to be closed, and that the connection wasn't otherwise lost or closed from the remote end.

As an additional point, the Spring Framework `WebClient` uses `onErrorResume`, after an application `onStatus` handler is called, to ensure that response is consumed. It then ignores any errors that may result, if the was response was in fact consumed by the `onStatus` handler. That means in this scenario, you could also see the following in the logs:

```
2022-11-18 08:58:52.679 ERROR 41948 --- [tor-http-nio-16] reactor.core.publisher.Operators         : Operator called default onErrorDropped

java.lang.IllegalStateException: Only one connection receive subscriber allowed.
	at reactor.netty.channel.FluxReceive.startReceiver(FluxReceive.java:182) ~[reactor-netty-core-1.0.24.jar:1.0.24]
	at reactor.netty.channel.FluxReceive.subscribe(FluxReceive.java:143) ~[reactor-netty-core-1.0.24.jar:1.0.24]
	at reactor.core.publisher.InternalFluxOperator.subscribe(InternalFluxOperator.java:62) ~[reactor-core-3.4.24.jar:3.4.24]
	at reactor.netty.ByteBufFlux.subscribe(ByteBufFlux.java:340) ~[reactor-netty-core-1.0.24.jar:1.0.24]
	at reactor.core.publisher.Mono.subscribe(Mono.java:4455) ~[reactor-core-3.4.24.jar:3.4.24]
	at reactor.core.publisher.FluxOnErrorResume$ResumeSubscriber.onError(FluxOnErrorResume.java:103) ~[reactor-core-3.4.24.jar:3.4.24]
```

As this is a scenario that's expected to occur, maybe the error message could be improved a little to something like "ignoring additional read. Response already consumed (or closed, or whatever the actual state)." to provide a little more context and hopefully guide better as to what happened.

pderop:
This PR attempts to improve logging in the following two scenario:

1) In a pipeline with multiple client calls, it's possible for a cancel signal to close the connection, e.g. Mono.zip where one publisher completes empty. In this case, you might see a log messages for subscribing to the response stream, immediately followed by channel closing:
```
2022-11-18 08:58:52.678 DEBUG 41948 --- [tor-http-nio-16] reactor.netty.channel.FluxReceive        : FluxReceive{pending=0, cancelled=false, inboundDone=false, inboundError=null}: subscribing inbound receiver
2022-11-18 08:58:52.678 DEBUG 41948 --- [tor-http-nio-16] r.n.resources.PooledConnectionProvider   : Channel closed, now: 9 active connections, 136 inactive connections and 0 pending acquire requests.
```
This PR adds a 3rd log message in between that shows there was a cancel signal:

```
2022-11-18 08:58:52.678 DEBUG 41948 --- [tor-http-nio-16] r.n.http.client.HttpClientOperations - [2ac0945a-1, L:/0:0:0:0:0:0:0:1:50589 - R:/0:0:0:0:0:0:0:1:9999] Cancelling inbound stream, closing channel.
```

2) As an additional point, the Spring Framework WebClient uses onErrorResume, after an application onStatus handler is called, to ensure that response is consumed. It then ignores any errors that may result, if the response was in fact consumed by the onStatus handler. See for example this spring-webflux [WebClientDataBufferAllocatingTests.onStatusWithBodyConsumed](https://github.com/spring-projects/spring-framework/blob/5.0.x/spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientDataBufferAllocatingTests.java#L104) test. That means in this scenario, you could also see the following in the logs:

```
2022-11-18 08:58:52.679 ERROR 41948 --- [tor-http-nio-16] reactor.core.publisher.Operators         : Operator called default onErrorDropped

java.lang.IllegalStateException: Only one connection receive subscriber allowed.
	at reactor.netty.channel.FluxReceive.startReceiver(FluxReceive.java:182) ~[reactor-netty-core-1.0.24.jar:1.0.24]
	at reactor.netty.channel.FluxReceive.subscribe(FluxReceive.java:143) ~[reactor-netty-core-1.0.24.jar:1.0.24]
	at reactor.core.publisher.InternalFluxOperator.subscribe(InternalFluxOperator.java:62) ~[reactor-core-3.4.24.jar:3.4.24]
	at reactor.netty.ByteBufFlux.subscribe(ByteBufFlux.java:340) ~[reactor-netty-core-1.0.24.jar:1.0.24]
	at reactor.core.publisher.Mono.subscribe(Mono.java:4455) ~[reactor-core-3.4.24.jar:3.4.24]
	at reactor.core.publisher.FluxOnErrorResume$ResumeSubscriber.onError(FluxOnErrorResume.java:103) ~[reactor-core-3.4.24.jar:3.4.24]
```

The original github issue associated to this PR (see #2583) is suggesting to improve the error message, with something like `ignoring additional read. Response already consumed (or closed, or whatever the actual state).` to provide a little more context and hopefully guide better as to what happened. 
For the moment, this PR is simply replacing the IllegalStateException `Only one connection receive subscriber allowed` message by `Ignoring additional inbound receiver subscriber`; and some useful informations are also included in the message (the same infos actually provided by the `reactor.netty.channel.FluxReceiver` logger if it is enabled):

- cancelled (true/false): true if the flux has been cancelled, false if not
- terminated (true/false). true when the flux is done because either completed or closed because of an error.
- pending: the number of messages that are still pending and not consumed
- error (true/false): true in case the flux has been terminated because of an error, in the case, the IllegalStateException is constructed with the original root cause exception, so you will then see it in the stacktrace

So, for example, with the spring-webflux [WebClientDataBufferAllocatingTests.onStatusWithBodyConsumed](https://github.com/spring-projects/spring-framework/blob/5.0.x/spring-webflux/src/test/java/org/springframework/web/reactive/function/client/WebClientDataBufferAllocatingTests.java#L104) test, we will now get the following IllegalStateException:

```
2022-11-18 08:58:52.679 ERROR 41948 --- [tor-http-nio-16] reactor.core.publisher.Operators         : Operator called default onErrorDropped

java.lang.IllegalStateException: Ignoring additional inbound receiver subscriber. Current flux state=[terminated=false,cancelled=true,pending=0,error=false].
	at reactor.netty.channel.FluxReceive.subscribe(FluxReceive.java:143) ~[reactor-netty-core-1.0.24.jar:1.0.24]
	at reactor.core.publisher.InternalFluxOperator.subscribe(InternalFluxOperator.java:62) ~[reactor-core-3.4.24.jar:3.4.24]
	at reactor.netty.ByteBufFlux.subscribe(ByteBufFlux.java:340) ~[reactor-netty-core-1.0.24.jar:1.0.24]
	at reactor.core.publisher.Mono.subscribe(Mono.java:4455) ~[reactor-core-3.4.24.jar:3.4.24]
	at reactor.core.publisher.FluxOnErrorResume$ResumeSubscriber.onError(FluxOnErrorResume.java:103) ~[reactor-core-3.4.24.jar:3.4.24]
```

For consistency, the message logged by the _reactor.netty.channel.FluxReceiver_ logger has also been replaced by `Ignoring additional inbound receiver subscriber.` instead of `Only one connection receive subscriber allowed`

Maybe @violetagg will propose some more simple and user friendly informations to provide in the IllegalStateException ?

Fixes #2583

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
@@ -172,7 +172,8 @@ final class FluxReceive extends Flux<Object> implements Subscription, Disposable
 			s.onSubscribe(this);
 		}
 		else {
-			if (inboundDone && getPending() == 0) {
+			long pending = getPending();
+			if (inboundDone && pending == 0) {
 				if (inboundError != null) {
 					Operators.error(s, inboundError);
 					return;
@@ -182,11 +183,19 @@ final class FluxReceive extends Flux<Object> implements Subscription, Disposable
 			}
 			else {
 				if (log.isDebugEnabled()) {
-					log.debug(format(channel, "{}: Only one connection receive subscriber allowed."), this);
+					log.debug(format(channel, "{}: Ignoring additional inbound receiver subscriber."), this);
 				}
-				Operators.error(s,
-						new IllegalStateException(
-								"Only one connection receive subscriber allowed."));
+
+				StringBuilder msg = new StringBuilder("Ignoring additional inbound receiver subscriber.");
+				msg.append(" Current flux state=[");
+				msg.append("terminated=").append(inboundDone);
+				msg.append(",cancelled=").append(isCancelled());
+				msg.append(",pending=").append(pending);
+				msg.append(",error=").append(inboundError != null);
+				msg.append("].");
+				IllegalStateException ex = inboundError == null ? new IllegalStateException(msg.toString()) :
+						new IllegalStateException(msg.toString(), inboundError);
+				Operators.error(s, ex);
 			}
 		}
 	}
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
@@ -275,6 +275,9 @@ class HttpClientOperations extends HttpOperations<NettyInbound, NettyOutbound>
 			return;
 		}
 		//"FutureReturnValueIgnored" this is deliberate
+		if (log.isDebugEnabled()) {
+			log.debug(format(channel(), "Cancelling inbound stream, closing channel."));
+		}
 		channel().close();
 	}
 


pderop:
@violetagg , @rstoyanchev, can you have a look ?

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -182,11 +183,19 @@ final void startReceiver(CoreSubscriber<? super Object> s) {
...
+
+				StringBuilder msg = new StringBuilder("Ignoring additional inbound receiver subscriber.");
+				msg.append(" Current flux state=[");
+				msg.append("terminated=").append(inboundDone);
+				msg.append(",cancelled=").append(isCancelled());
+				msg.append(",pending=").append(pending);
+				msg.append(",error=").append(inboundError != null);
+				msg.append("].");
+				IllegalStateException ex = inboundError == null ? new IllegalStateException(msg.toString()) :
+						new IllegalStateException(msg.toString(), inboundError);
```
rstoyanchev, line 197: Maybe the same (and more complete) message should also be used for the log message above?

violetagg (changes requested):
Very small comments to the code

Do you plan another PR for distinguishing between a normal connection close and cancelation from let say `Mono.zip`?

rstoyanchev (commented):
I think exposing additional context makes more sense, and better reflects my request. Thanks for the changes. A couple of minor comments below.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -275,6 +275,9 @@ protected void onInboundCancel() {
 			return;
 		}
 		//"FutureReturnValueIgnored" this is deliberate
+		if (log.isDebugEnabled()) {
+			log.debug(format(channel(), "Cancelling inbound stream, closing channel."));
```
rstoyanchev, line 279: Maybe "cancelled" (passive) better reflects that something external got this stream cancelled, vs "cancelling" (active) which leaves it more unclear, did we cancel or did something else?

```suggestion
			log.debug(format(channel(), "Inbound stream cancelled, closing channel."));
```

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -182,11 +183,19 @@ final void startReceiver(CoreSubscriber<? super Object> s) {
...
-						new IllegalStateException(
-								"Only one connection receive subscriber allowed."));
+
+				StringBuilder msg = new StringBuilder("Ignoring additional inbound receiver subscriber.");
+				msg.append(" Current flux state=[");
+				msg.append("terminated=").append(inboundDone);
+				msg.append(",cancelled=").append(isCancelled());
+				msg.append(",pending=").append(pending);
+				msg.append(",error=").append(inboundError != null);
+				msg.append("].");
```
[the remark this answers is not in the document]
pderop: not exactly, the `FluxReceive.toString()` method is displaying the inboundError Message (see `inboundError=" + inboundError`, so the inboundError.toString() method is displayed):

```
	public String toString() {
		return "FluxReceive{" +
				"pending=" + getPending() +
				", cancelled=" + isCancelled() +
				", inboundDone=" + inboundDone +
				", inboundError=" + inboundError +
				'}';
	}
```
so if we reuse the FluxReceive.toString() method when constructing the IllegalStateException, we would end up with something like this (here inboundError is an IOException for example), which is probably a bit too much verbose:

```
java.lang.IllegalStateException: Ignoring additional inbound receiver subscriber. Current flux state=FluxReceive{pending=1, cancelled=false, inboundDone=true, inboundError=java.io.IOException: Inbound Exception}
	at reactor.netty.channel.FluxReceive.startReceiver(FluxReceive.java:197)
	at reactor.netty.channel.FluxReceive.subscribe(FluxReceive.java:148)
	at reactor.core.publisher.InternalFluxOperator.subscribe(InternalFluxOperator.java:62)
...
Caused by: java.io.IOException: Inbound Exception
	at reactor.netty.channel.FluxReceiveTest.testFluxReceive(FluxReceiveTest.java:188)
	... 86 more
```

so, here, what I mean is that the `inboundError=java.io.IOException: Inbound Exception` is redondant (because it is also displayed in the root cause exception at the bottom of the stacktrace. That's why I'm formatting a bit differently, and if an error is present, only `error=true` will be displayed in the message header, and the real error will be displayed at the bottom of the stacktrace:

```
java.lang.IllegalStateException: Ignoring additional inbound receiver subscriber. Current flux state=[terminated=true,cancelled=false,pending=1,error=true].
	at reactor.netty.channel.FluxReceive.startReceiver(FluxReceive.java:197)
	at reactor.netty.channel.FluxReceive.subscribe(FluxReceive.java:148)
	at reactor.core.publisher.InternalFluxOperator.subscribe(InternalFluxOperator.java:62)
...
Caused by: java.io.IOException: Inbound Exception
	at reactor.netty.channel.FluxReceiveTest.testFluxReceive(FluxReceiveTest.java:188)
	... 86 more
```

WDYT ?

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -182,11 +183,19 @@ final void startReceiver(CoreSubscriber<? super Object> s) {
...
 			else {
 				if (log.isDebugEnabled()) {
-					log.debug(format(channel, "{}: Only one connection receive subscriber allowed."), this);
+					log.debug(format(channel, "{}: Ignoring additional inbound receiver subscriber."), this);
 				}
-				Operators.error(s,
-						new IllegalStateException(
-								"Only one connection receive subscriber allowed."));
+
+				StringBuilder msg = new StringBuilder("Ignoring additional inbound receiver subscriber.");
```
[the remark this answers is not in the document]
pderop: ok, I have replaced "ignoring" by "rejected", see in commit [345404d](https://github.com/reactor/reactor-netty/pull/2585/commits/345404d36d11673562c7ef5ff8c3ead1f6019b80)

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
pderop: the same kind of informations is actually logged in the log message above, using the FluxReceive.toString() method (which displays the same informations a bit differently)

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
pderop: indeed, it seems more appropriate to use "cancelled" instead fo "cancelling", please check the commit https://github.com/reactor/reactor-netty/pull/2585/commits/57527cebed90fcfc8dd41bcbb58718be5a21b3f3

To be consistent, I have also done the same for the following classes which are already implementing the _onInboundCancel_ method and which are are already logging cancels before closing:

- WebSocketClientOperations
- WebSocketServerOperations
- QuicStreamOperartions

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -275,6 +275,9 @@ protected void onInboundCancel() {
...
+		}
```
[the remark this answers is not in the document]
pderop: I see that the following classes are already implementing the _onInboundCancel_ methods:

- WebSocketClientOperations
- WebSocketServerOperations
- QuicStreamOperartions

The above classes are all logging the cancels before performing any closing actions.
Now, one that seems to be not overriding the _onInboundCancel_ method is the _HttpServerOperartions_
Then, should I simply add a log in _HttpServerOperartions_ (I mean should I simply implement the _onInboundCancel_ and just log that the stream has been cancelled). Then I wonder if something else should be done ? should the client connection be closed ? I'm confused by this ?

violetagg:
I would recommend to have tests so that we can see how we behave in the different protocols and server/client + connection close vs local cancellation

pderop:
Added a first test `HttpClientTest.testHttpClientCancelled` which verifies how is behaving an http client which is cancelled before the response is received from the server. The test uses Flux.zip(client, mono) where the second mono succeeds with an empty value when the server has received the request. 
From logs, we can see that the client is cancelled, then the socket is closed:
```
23:16:39.838 [reactor-http-nio-3] DEBUG r.n.http.client.HttpClientOperations - [001356ae-1, L:/0:0:0:0:0:0:0:1:51256 - R:/0:0:0:0:0:0:0:1:51255] Http client inbound stream cancelled, closing channel.
23:16:39.838 [reactor-http-nio-3] DEBUG reactor.netty.http.client.HttpClient - [001356ae-1, L:/0:0:0:0:0:0:0:1:51256 - R:/0:0:0:0:0:0:0:1:51255] CLOSE
23:16:39.839 [Test worker] INFO  fluxzip - onComplete()
23:16:39.839 [reactor-http-nio-4] DEBUG reactor.netty.http.server.HttpServer - [eff7fce7-1, L:/0:0:0:0:0:0:0:1:51255 - R:/0:0:0:0:0:0:0:1:51256] READ COMPLETE
23:16:39.840 [reactor-http-nio-4] DEBUG reactor.netty.http.server.HttpServer - [eff7fce7-1, L:/0:0:0:0:0:0:0:1:51255 ! R:/0:0:0:0:0:0:0:1:51256] INACTIVE
23:16:39.840 [reactor-http-nio-3] DEBUG r.n.r.PooledConnectionProvider - [001356ae-1, L:/0:0:0:0:0:0:0:1:51256 ! R:/0:0:0:0:0:0:0:1:51255] Channel closed, now: 0 active connections, 0 inactive connections and 0 pending acquire requests.
```

Will gradually add more tests for server-side cancel for Http, and similar tests for other protocols ...

pderop:
@violetagg ,

This is interesting, I have added in the previous commit the `HttpServerTests.testHttpServerCancelled` test, where a server cancels a request **before** the request body is **fully** consumed. So, I see that the the client socket is not closed, which seems to be a bug. Indeed, the HttpServerOperations does not implement the `onInboundCancel` method. So, the `HttpServerTests.testHttpServerCancelled` is currently failing.

Adding this method in the `HttpServerOperations` seems to fix the issue:

```
	@Override
	protected void onInboundCancel() {
		if (isInboundDisposed()) {
			return;
		}

		if (log.isDebugEnabled()) {
			log.debug(format(channel(), "Http server inbound cancelled. Closing socket."));
		}
		channel().close();
	}
```

I will propose this patch in a separate PR.

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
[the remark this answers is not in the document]
pderop: forget about my previous comments, you are right and we need to centralize the formatting in one single toString method. 

Please check [3633de8](https://github.com/reactor/reactor-netty/pull/2585/commits/3633de88ef374fd832c6209a6806bfa5f53dece8)

pderop:
forced a push, which is a fixup for the testWebSocketClientCancelled test that is now checking websocket status codes in both client and server.

violetagg:
> I just checked with Violeta on a private call: the `HttpServerTests.testHttpServerCancelled` is not making sense and the patch in `HttpServerTests.onInboundCancel' is invalid and breaks other use cases.
> 
> So I will uncommit `HttpServerTests.testHttpServerCancelled` soon.

When on the HTTP server and the I/O Handler decides to stop consuming the inbound data (cancel), the I/O Handler is responsible to decide what kind of response will be send to the client and whether the connection will be closed or not. Reactor Netty has to guarantee that it will discard the incoming data once the response is sent.

pderop:
forced push, in order to ignore all commits that were made after the [3e2122a](https://github.com/reactor/reactor-netty/pull/2585/commits/3e2122a9b8fdd68907c14a40a931062ada56e949).

Now, tests have been added for all protocols (udp/tcp/http/websocket/quic), and there is one "logical" commit for each protocol.

please do not review for the moment, I need first to see if all tests are passing, then if it goes well, I'll move this PR from draft to ready for review state.

violetagg:
@pderop You have a lot similar code shared between all modules. What do you think to move those parts to `reactor-netty-core` `testFixtures`. `testFixtures` is intended exactly for sharing utilities.

pderop:
ok; I'm currently fixing something which is still unstable (on windows apparently), so I'll probably do a few additional fixup commits , then after that, I'll move the common parts to testFixtures.  Thanks for the suggestion !

pderop:
new attempt: common test classes have been moved to core testFixtures, tests have been finalized.
now, let's see how it goes on the build CI ...

pderop:
@violetagg ,

when you have time, can you play with the various tests and take a look ? I think all is finalized now.

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -182,11 +182,13 @@ final void startReceiver(CoreSubscriber<? super Object> s) {
 			}
 			else {
 				if (log.isDebugEnabled()) {
-					log.debug(format(channel, "{}: Only one connection receive subscriber allowed."), this);
+					log.debug(format(channel, "{}: Rejecting additional inbound receiver subscriber."), this);
```
violetagg, line 185: Wdyt?

```suggestion
					log.debug(format(channel, "{}: Rejecting additional inbound receiver."), this);
```

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -182,11 +182,13 @@ final void startReceiver(CoreSubscriber<? super Object> s) {
...
 				}
-				Operators.error(s,
-						new IllegalStateException(
-								"Only one connection receive subscriber allowed."));
+
+				String msg = "Rejecting additional inbound receiver subscriber. State=" + toString(false);
```
violetagg, line 188: Wdyt?

```suggestion
				String msg = "Rejecting additional inbound receiver. State=" + toString(false);
```

violetagg (changes requested):

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
pderop: yes, thanks, I just applied your feedback.

reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
pderop: yes, thanks, I just applied your feedback.

violetagg (approved):

pderop:
@violetagg ,  can you take a quick look at the last three commits ? then may I go head and merge ?
(all this seems to work fine in netty5 branch).

violetagg (approved):

PR #2589, opened 2022-11-25
Update `Http2Pool` javadoc

violetagg:
When `GO_AWAY` is received and there are no streams the connection is removed from the pool. If there are streams the connection is filtered and not used for creation of new streams.


pderop (approved):
LGTM

violetagg:
@pderop Thanks

PR #2600, opened 2022-12-01
onInboundCancel does not close channel when inbound udpclient receiver is cancelled

### LINKED ISSUE #2583: Improve logging when connection closed due to cancel signal

pderop:
While working on #2585, it has been observed that when an UdpClient inbound receiver is cancelled, then unlike the HttpClient, the udp channel is not closed. in other words, the `onInboundCancel` method is not implemented by UdpOperations class.

I'm not sure, but this PR is to start a discussion whether or not the udp **client** channel should be closed when the inbound receiver is cancelled.

The `UdpClientTest.testUdpClientCancelled` test shows an example usecase.

The `UdpServerTest.testUdpServerCancelled` test also verifies that an udp server will drop any incoming messages in case the server inbound receiver is cancelled.

If it is not making sense to close the udp client channel on inbound receiver cancellation, then I will cancel this PR, but will change the UdpClientTest.testUdpClientCancelled to just check if buffers are dropped when the client inbound receiver is cancelled, and I'll add it to the other #2585 PR.
 
Partially Fixes #2583

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/udp/UdpClientConfig.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/udp/UdpClientConfig.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2020-2022 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,8 @@ import io.netty.channel.unix.DomainDatagramChannel;
 import io.netty.handler.logging.LogLevel;
 import io.netty.handler.logging.LoggingHandler;
 import io.netty.resolver.AddressResolverGroup;
+import reactor.netty.Connection;
+import reactor.netty.ConnectionObserver;
 import reactor.netty.channel.ChannelMetricsRecorder;
 import reactor.netty.channel.ChannelOperations;
 import reactor.netty.channel.MicrometerChannelMetricsRecorder;
@@ -39,6 +41,8 @@ import java.net.SocketAddress;
 import java.nio.charset.Charset;
 import java.util.Map;
 import java.util.function.Supplier;
+
+import static reactor.netty.ReactorNetty.format;
 
 /**
  * Encapsulate all necessary configuration for UDP client transport. The public API is read-only.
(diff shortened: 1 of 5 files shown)


pderop:
I will close this PR because in turns out that there is actually no reasons to close the udp client channel on inbound receiver cancellation.

PR #2605, opened 2022-12-08
Update documentation for `ConnectionProvider.Builder#disposeInactivePoolsInBackground`

violetagg:
Clarify that the connection pool can be disposed only when there are no active connections, idle connections and pending acquisitions.


pderop (approved):

violetagg:

PR #2606, opened 2022-12-08
Include forwarded client IP address in access log

### LINKED ISSUE #2603: Unable to use forwarded client IP in access log
## Expected Behavior
It would be nice if the access log entries contain the forwarded client IP.

## Actual Behavior

When the `HttpServer` is configured to accept forwarded headers (i.e. `forwarded(true)`), the access log contains the IP of the peer that sent the http request (forwarded headers are not respected).

## Steps to Reproduce

Clone and run the app that enable forwarded headers and access logs:
https://github.com/MatthiasWinzeler/reactor-netty-forwarded-access-log

Then, fire a sample request containing a `Forwarded` header:

```bash
curl -v http://localhost:8080/hello -H 'Forwarded: for=192.0.2.60;proto=http;by=203.0.113.43'
/192.0.2.60:60365
```

You will notice that the server has successfully parsed the `Forwarded` header and set it on the `connectionInfo` which is printed as a response.

However, the access log will show the IP of the direct client (in this case 127.0.0.1):
```
[ INFO] (reactor-http-nio-2) 127.0.0.1 - - [06/Dec/2022:22:22:06 +0100] "GET /hello HTTP/1.1" 200 17 21
```

I see that we can pass our own `AccessLogFactory` so that we can customize the log format - however, this only has access to the `DefaultHttpRequest` and not the `ConnectionInfo`. So we'd have to parse the raw `Forwarded` request header again, and unfortunately the `Forwarded` parsing logic used within reactor-netty-http is not accessible.

## Possible Solution
- Could the access log by default respect the remote address from ConnectionInfo? Tomcat is doing a similar thing: https://stackoverflow.com/a/48180469 
- If not, could we make the `ConnectionInfo` accessible in the `AccessLogFactory` so that we can get the adjusted (real) client IP?

## Your Environment

* Reactor version(s) used: `2022.0.0`

Many thanks already in advance!

pderop:
This PR addresses #2603 issue, and include any Forwarded client ip address in the access log.

Fixes #2603

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AbstractAccessLogArgProvider.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AbstractAccessLogArgProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2020-2022 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -37,7 +37,7 @@ abstract class AbstractAccessLogArgProvider<SELF extends AbstractAccessLogArgPro
 			DateTimeFormatter.ofPattern("dd/MMM/yyyy:HH:mm:ss Z");
 	static final String MISSING = "-";
 
-	final SocketAddress remoteAddress;
+	SocketAddress remoteAddress;
 	final String user = MISSING;
 	String zonedDateTime;
 	ZonedDateTime accessDateTime;
@@ -135,6 +135,7 @@ abstract class AbstractAccessLogArgProvider<SELF extends AbstractAccessLogArgPro
 		this.contentLength = -1;
 		this.startTime = 0;
 		this.cookies = null;
+		this.remoteAddress = null;
 	}
 
 	SELF cookies(Map<CharSequence, Set<Cookie>> cookies) {
@@ -154,6 +155,11 @@ abstract class AbstractAccessLogArgProvider<SELF extends AbstractAccessLogArgPro
 			}
 			this.contentLength += contentLength;
 		}
+		return get();
+	}
+
+	SELF remoteAddress(SocketAddress remoteAddress) {
+		this.remoteAddress = remoteAddress;
 		return get();
 	}
 
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogHandlerH1.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogHandlerH1.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2018-2022 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


pderop:
@violetagg ,

Can you take a look ?

PS: maybe we should not include this PR for the next upcoming release, that's why I did not add a milestone.

violetagg:
@pderop Now with this change we have the other problem - always log what is provided with `X-Forwarded`/`Forwarded` header.

pderop:
@violetagg,

indeed, the PR breaks something: if the HttpServer `forwarded` mode is enabled, then accesslog will now log the forwarded client address which is a breaking change, so people may complain about that.

second attempt: 

To let people choose either connection peer address, or forwarded remote client address, the last commit is adding a new `forwardedAddress` method in the `AccessLogArgProvider` interface. So if the forwarded client address needs to be displayed in the accesslog, then the user will need to use the AccessLogFactory and invoke the new forwardedAddress() method from the AccessLogArgProvider argument passed to the factory, and the original AccessLogFactory.remoteAddress is left unchanged (the forwarded mode must be enabled on the HttpServer).

please take a look ?

pderop:
mmm, please wait, sanity checks are failing.

pderop:
The sanity check error is now fixed.

pderop:
Let's just follow the suggestion done in the original #2603: simply expose the `ConnectionInfo` from the `AccessLogFactory`

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogArgProvider.java
```diff
@@ -16,12 +16,15 @@
 package reactor.netty.http.server.logging;
 
 import io.netty.handler.codec.http.cookie.Cookie;
+import reactor.netty.http.server.ConnectionInfo;
```
violetagg, line 19:
```suggestion
```

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogArgProvider.java
```diff
@@ -54,10 +57,26 @@ public interface AccessLogArgProvider {
 	 * Returns the address of the remote peer or {@code null} in case of Unix Domain Sockets.
 	 *
 	 * @return the peer's address
+	 * @deprecated Use {@link ConnectionInformation#connectionRemoteAddress()}
```
violetagg, line 60:
```suggestion
	 * @deprecated as of 1.0.26. Use {@link ConnectionInformation#connectionRemoteAddress()}
```

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/AccessLogArgProvider.java
```diff
@@ -54,10 +57,26 @@ public interface AccessLogArgProvider {
...
+	/**
+	 * Returns the information about the current connection.
+	 * <p> Note that the {@link ConnectionInfo#getRemoteAddress()} will return the forwarded
+	 * remote client address if the server is configured in forwarded mode.
+	 *
+	 * @since 1.0.26
+	 * @return the connection info
+	 * @see reactor.netty.http.server.HttpServer#forwarded(BiFunction)
+	 * @see reactor.netty.http.server.HttpServer#forwarded(BiFunction)
+	 */
```
violetagg, line 76:
```suggestion
	/**
	 * Returns the information about the current connection.
	 * <p> Note that the {@link ConnectionInformation#remoteAddress()} will return the forwarded
	 * remote client address if the server is configured in forwarded mode.
	 *
	 * @since 1.0.26
	 * @return the connection info
	 * @see reactor.netty.http.server.HttpServer#forwarded(BiFunction)
	 */
```

violetagg (changes requested):
Small fixes for the javadoc

violetagg (approved):

PR #2610, opened 2022-12-12
Update `ConnectionProvider.MeterRegistrar` javadoc

violetagg:
Since version `1.0.11`, `MicrometerPooledConnectionProviderMeterRegistrar` is not an implementation of this interface. See #1806


violetagg:

Issue #761, opened 2019-06-12
Provide samples as part of the project


violetagg:

violetagg:
Some very basic samples

1fc98e87 - [http] Hello World example
7d5dbb66 - [http] Echo example
d55fb9a5 - [tcp] Echo example
1d6edd88 - [tcp] Discard example

PR #2617, opened 2022-12-16
Add Reactor Netty version of Telnet example #761

jchenga:
This PR attempts to provide a corresponding Reactor Netty version of Telnet example available in Netty project.

### DIFF
--- a/reactor-netty-examples/src/main/java/reactor/netty/examples/tcp/telnet/TelnetServer.java
+++ b/reactor-netty-examples/src/main/java/reactor/netty/examples/tcp/telnet/TelnetServer.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2023 VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty.examples.tcp.telnet;
+
+import io.netty.handler.codec.DelimiterBasedFrameDecoder;
+import io.netty.handler.codec.Delimiters;
+import io.netty.handler.ssl.util.SelfSignedCertificate;
+import reactor.core.publisher.Flux;
+import reactor.netty.tcp.TcpServer;
+import reactor.netty.tcp.TcpSslContextSpec;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.security.cert.CertificateException;
+import java.util.Date;
+
+public class TelnetServer {
+
+	static final boolean SECURE = System.getProperty("secure") != null;
+	static final int PORT = Integer.parseInt(System.getProperty("port", SECURE ? "8992" : "8993"));
+
+	static final boolean WIRETAP = System.getProperty("wiretap") != null;
+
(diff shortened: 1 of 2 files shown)


violetagg:
@jchenga Thank you for having time to work on this example!

I would like to suggest to make the example such that it shows as much as possible of the Reactor Netty API. What do you think?

For example let's start with `TelnetClient.java`
1. `StringDecoder`/`StringEncoder` are used, Reactor Netty provides such functionality. We can remove these codecs from the pipeline. Once they are removed we can change the following:

-   Instead of `in.receiveObject().ofType(String.class)` we can invoke `in.receive().asString()`
-   Instead of `conn.channel().writeAndFlush(text + '\n');` we can invoke `out.sendString(Mono.just(text + "\n"))`

2. For the client use case it is not very convenient to use `handle` method, we can work directly with the connection `inbound` and `outbound`. (See more here https://projectreactor.io/docs/netty/release/reference/index.html#_consuming_data_2)
  

-   First we need to obtain the connection `Connection conn = client.connectNow();` 
-   Instead of `.handle((in, out) -> in.receiveObject().ofType(String.class).doOnNext(s -> System.out.println(s)).then());` we can invoke `conn.inbound().receive().asString().doOnNext(System.out::println).subscribe();`
-   Instead of `conn.channel().writeAndFlush(text + '\n');` we can invoke `conn.outbound().sendString(Mono.just(text + "\r\n")).then().subscribe();`

Would you like to try these suggestions? Then we can change a bit also the server part of the example.

jchenga:
@violetagg thank you for the feedback. I've made the changes in TelnetClient class as you suggested.

reactor-netty-examples/src/main/java/reactor/netty/examples/tcp/telnet/TelnetServer.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright (c) 2020-2022 VMware, Inc. or its affiliates, All Rights Reserved.
```
violetagg, line 2:
```suggestion
 * Copyright (c) 2023 VMware, Inc. or its affiliates, All Rights Reserved.
```

violetagg (changes requested):

violetagg:
> @violetagg thank you for the feedback. I've made the changes in TelnetClient class as you suggested.

Nice!

Let's now take a look at `TelnetServer.java`

1. Here we can also use the String encoding/decoding API provided by Reactor Netty and remove the StringEncoder/Decoder from the pipeline
2. What do you think about extracting the welcome message outside of the `doOnConnection` callback

Instead of:

```
try {
	connection.channel().writeAndFlush("Welcome to " + InetAddress.getLocalHost().getHostName() + "!\r\n");
	connection.channel().writeAndFlush("It is " + new Date() + " now.\r\n");
}
catch (Exception e) {
	throw Exceptions.propagate(e);
}
```

We can prepare a `Flux` like this:

```
String hostName = InetAddress.getLocalHost().getHostName();
...
Flux<String> welcomeFlux =
		Flux.just("Welcome to " + hostName + "!\r\n", "It is " + new Date() + " now.\r\n");
```

3. Now the `handle` method can be made something like this:

```
.handle((in, out) -> {
	Flux<String> welcomeFlux =
			Flux.just("Welcome to " + hostName + "!\r\n", "It is " + new Date() + " now.\r\n");

	Flux<String> responses =
			in.receive()
				.asString()
				.takeUntil("bye"::equalsIgnoreCase)
				.map(s -> {
					String response = "Did you say '" + s + "'?\r\n";
					if (s.isEmpty()) {
						response = "Please type something.\r\n";
					} else if ("bye".equalsIgnoreCase(s)) {
						response = "Have a good day!\r\n";
					}
					return response;
				});

	return out.sendString(Flux.concat(welcomeFlux, responses));
});
```

jchenga:
This is great. @violetagg I've implemented your suggestion and pushed the changes. I also added some comments to try to explain what happen under the hood, based on my understandings after stepping through the call stack. I think it could be helpful to others.

violetagg (approved):
@reactor/netty-team PTAL

pderop (approved):
LGTM, thanks !

PR #2633, opened 2023-01-05
Ensure `HttpServer` `Active Connections` metric is correct when connection closure before response write

### LINKED ISSUE #2619: Connection closure before response write causes incorrect HttpServer 'Active Connections' value
Closure of a HTTP/1.1 connection when a request has been received, but a response not written causes the servers active connections Gauge to become incorrect as it is not decremented (reactor.netty.http.server.active.connections).

On investigating this is due to the fact we decrement the active connections Gauge once we write the 'LastHttpContent' message, which does not happen if the connection is closed before writing the response.

Looking at the code I assume similar behaviour will happen for H2 stream closure.

## Steps to Reproduce

- Enable Micrometer HttpServer metrics
- Protocol = HTTP/1.1
- Create any scenario that causes a connection to be closed before a response has been written
- Observe server side 'active connections' metric is not decremented 

## Possible Solution

An option would be to figure that on connection closure a request was still being processed, such that its response was not received, and hook into 'channelInactive' which is what decrements the total connections metric.

violetagg:
Fixes #2619


violetagg:
@reactor/netty-team PTAL

pderop (approved):
LGTM

violetagg:

PR #2646, opened 2023-01-11
Ensure `HttpServer` `minCompressionSize` is respected when `Content-Length` is not explicit and response is `Mono`

violetagg:
When a `Mono` is sent as a response or a single `ByteBuf` is sent as a response, `Reactor Netty` will send `FullHttpResponse` and thus is able to set `Content-Length` if such is not already set. Based on the `Content-Length`, `Reactor Netty` can decide whether or not a compression is needed.


violetagg:
@reactor/netty-team PTAL

pderop (approved):
ok, I see ... it is the newFullBodyMessage which is setting the Content-Length header, so the `newFullBodyMessage` method must be called prior to 								`afterMarkSentHeaders()` call.

LGTM

PR #2655, opened 2023-01-16
Allow to configure colocation when creating LoopResources

### LINKED ISSUE #2604: Uneven HTTP/2 connection to event loop distribution
During profiling of Http2Pool I regularly saw a small fraction of webclient threads becoming very hot and processing the majority of the requests, whilst leaving other threads under-utilised.

For example, the below shows one thread (one event loop) doing more than 5 times the work of any other event loop. On investigation this appears to be down to new H2 connections being opened on the 'caller' event loop, rather than round-robin the connections to event loops, which is what the TransportConnector will do if no event loop is provided to 'connect' method.

Note in the below image 16 worker threads were set, which would create an EventLoopGroup with 16 EventLoops.

Note I used a separate client event loop to server loop.

OS: Windows 10, 64-bit

## Expected Behavior

Connections should be distributed evenly across event loops (worker threads) in the event loop group.

## Actual Behavior

Large number of new connections opened assigned to same event loop,

## Steps to Reproduce

Checkout below repository, which contains 3 services and run.

Note I had SSL configured, so it is recommended to run with local keystore and cert.

https://github.com/samueldlightfoot/http2-distribution

1. Load gen
2. Client
3. Server

Hit load generator endpoint to trigger processing (Load Gen -> Client -> Server) using H2.

curl -X GET http://localhost:9000/call/{request-count}/{concurrency}

e.g. curl -X GET http://localhost:9000/call/50000/250

I found the problem gets worse at a higher concurrency factor, most likely due to need for more connections to handle load. So would recommend using 250+.

## Possible Solution

The caller event loop should not be used when opening new connections when a separate client event loop is provided to the  HttpClient. It should instead let the EventLoopGroup _Chooser_ do the selection of the next loop to use. I believe the intention of this 'caller loop' feature was for when client and server event loop is shared, which in this case, it is not.

The troublesome code seems to be in PooledConnectionProvider, where we grab event loop group, which is of type ColocatedEventLoopGroup, therefore we grab the thread local loop via nextInternal. _DefaultLoopResources_  cache*ClientLoops() always colocates client with server loop, hence this will always be a ColocatedEventLoopGroup. I can't quite get my head around if this is the intended behaviour.

```				
EventLoopGroup group = config.loopResources().onClient(config.isPreferNative());
if (group instanceof ColocatedEventLoopGroup) {
	eventLoop = ((ColocatedEventLoopGroup) group).nextInternal();
}
```

and then set into Reactor Context

```			
Mono<PooledRef<T>> mono = pool.acquire(Duration.ofMillis(poolFactory.pendingAcquireTimeout));
if (eventLoop != null) {
	mono = mono.contextWrite(ctx -> ctx.put(CONTEXT_CALLER_EVENTLOOP, eventLoop));
}
```

which then calls TransportConnector::connect with the loop from the Context

```
Publisher<PooledConnection> connectChannel() {
	return Mono.create(sink -> {
			PooledConnectionInitializer initializer = new PooledConnectionInitializer(sink);
		EventLoop callerEventLoop = sink.contextView().hasKey(CONTEXT_CALLER_EVENTLOOP) ?
				sink.contextView().get(CONTEXT_CALLER_EVENTLOOP) : null;
		if (callerEventLoop != null) {
			TransportConnector.connect(config, remoteAddress, resolver, initializer, callerEventLoop, sink.contextView())
					.subscribe(initializer);
		}
		else {
			TransportConnector.connect(config, remoteAddress, resolver, initializer, sink.contextView()).subscribe(initializer);
		}
	});
}
```

## Things I've Tried

As a test I overrode DefaultLoopResources::cacheNioClientLoops to use the NioEventLoopGroup directly rather than a ColocatedEventLoopGroup, and the distribution of connections was much better.

```
EventLoopGroup cacheNioClientLoops() {
	EventLoopGroup eventLoopGroup = clientLoops.get();
	if (null == eventLoopGroup) {
                // Below line was modified
		EventLoopGroup newEventLoopGroup = cacheNioServerLoops();
		if (!clientLoops.compareAndSet(null, newEventLoopGroup)) {
			// Do not shutdown newEventLoopGroup as this will shutdown the server loops
		}
		eventLoopGroup = cacheNioClientLoops();
	}
	return eventLoopGroup;
}
```

## Your Environment

* Reactor version(s) used: 1.1.0
* JVM version (`java -version`): JDK17
* OS and version (eg. `uname -a`): Windows 10

pderop:
When a custom event loop group is configured for an H2/H2C http client (i.e using `HttpClient.runOn(LoopResources.create("client", 16, true))` for example), then in  _Forwarder_ kind of scenario when a JVM forwards a received H2/H2C request using the HttpClient to an upstream server, then in this case, it may happen that the custom event loops are unevenly distributed among the http client channels.

what happen is the following:

- One available Slot is being delivered to a borrower from one of the custom event loop [here](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java#L387-L388), so drain() is called from the custom event loop thread.
- if at this point there are many borrowers but no more streams are available from current acquired connections, then drain will acquire many channels up to the max number of connections in the configuration. And because the current thread is one of the custom eventloop group, then Colocation will take place when connecting (see [here](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java#L160), nextInternal will return the current loop from the threadlocal). 

This explains why we may have a few (maybe one) custom event loop consuming a lot of CPU, compared to the others. 

We can't disable colocation, because there are scenario where we want to share a custom event loop between, for example, one HttpServer and HttpClient, so both HttpServer and HttpClient may share resources using the same EventLoop.

So, let's add a new `LoopResources.create`  method with colocation choice parameter, allowing the user to explicitly enable or disable colocation:

```java
	/**
	 * Create a simple {@link LoopResources} to provide automatically for {@link
	 * EventLoopGroup} and {@link Channel} factories
	 *
	 * @param prefix the event loop thread name prefix
	 * @param selectCount number of selector threads. When -1 is specified, no selectors threads will be created,
	 *                    and worker threads will be also used as selector threads.
	 * @param workerCount number of worker threads
	 * @param daemon should the thread be released on jvm shutdown
	 * @param colocate true means that {@link EventLoopGroup} created for clients will reuse current local event loop if already
	 *                 working inside one.
	 * @return a new {@link LoopResources} to provide automatically for {@link
	 * EventLoopGroup} and {@link Channel} factories
	 *
	 * @since 1.0.28
	 */
	static LoopResources create(String prefix, int selectCount, int workerCount, boolean daemon, boolean colocate) {
```

Fixes #2604

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/DefaultLoopResources.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/DefaultLoopResources.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2011-2023 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -48,16 +48,22 @@ final class DefaultLoopResources extends AtomicLong implements LoopResources {
 	final AtomicReference<EventLoopGroup> cacheNativeServerLoops;
 	final AtomicReference<EventLoopGroup> cacheNativeSelectLoops;
 	final AtomicBoolean                   running;
+	final boolean colocate;
 
 	DefaultLoopResources(String prefix, int workerCount, boolean daemon) {
 		this(prefix, -1, workerCount, daemon);
 	}
 
 	DefaultLoopResources(String prefix, int selectCount, int workerCount, boolean daemon) {
+		this(prefix, selectCount, workerCount, daemon, true);
+	}
+
+	DefaultLoopResources(String prefix, int selectCount, int workerCount, boolean daemon, boolean colocate) {
 		this.running = new AtomicBoolean(true);
 		this.daemon = daemon;
 		this.workerCount = workerCount;
 		this.prefix = prefix;
+		this.colocate = colocate;
 
 		this.serverLoops = new AtomicReference<>();
 		this.clientLoops = new AtomicReference<>();
@@ -170,7 +176,8 @@ final class DefaultLoopResources extends AtomicLong implements LoopResources {
 	EventLoopGroup cacheNioClientLoops() {
 		EventLoopGroup eventLoopGroup = clientLoops.get();
 		if (null == eventLoopGroup) {
-			EventLoopGroup newEventLoopGroup = LoopResources.colocate(cacheNioServerLoops());
+			EventLoopGroup newEventLoopGroup = colocate ? LoopResources.colocate(cacheNioServerLoops()) :
+					cacheNioServerLoops();
 			if (!clientLoops.compareAndSet(null, newEventLoopGroup)) {
 				// Do not shutdown newEventLoopGroup as this will shutdown the server loops
 			}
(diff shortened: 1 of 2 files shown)


violetagg (approved):

pderop:
@violetagg, thanks for your review !

PR #2673, opened 2023-01-27
Fix `NullPointerException` when `followRedirect` is true, but there is no `location` header

### LINKED ISSUE #2670: followRedirect by predicate causes NullPointerException
I want to follow the redirection applying some logic to keep track of redirected urls and also limit the number of redirections.
This is my code:

```
httpClient.followRedirect((req, res)-> {
		totalRedirects = res.redirectedFrom().length;
		log.debug("Redirect URL : " + res.resourceUrl());
		return totalRedirects<_maxRedirects;
```

Nevertheless when I use this approach I have a **NullPointerException** for each redirection step.
This exception come from line: 

> redirecting = new RedirectClientException(response.headers(), response.status()); 

in _reactor.netty.http.client.HttpClientOperations_ class

```
final boolean notRedirected(HttpResponse response) {
	if (isFollowRedirect() && followRedirectPredicate.test(this, this)) {
		if (log.isDebugEnabled()) {
			log.debug(format(channel(), "Received redirect location: {}"),
					httpMessageLogFactory().debug(HttpMessageArgProviderFactory.create(response)));
		}
		redirecting = new RedirectClientException(response.headers(), response.status());
		return false;
	}
	return true;
}
```

and

```
RedirectClientException(HttpHeaders headers, HttpResponseStatus status) {
	location = Objects.requireNonNull(headers.get(HttpHeaderNames.LOCATION));
	this.status = Objects.requireNonNull(status);
}
```
because of `location = Objects.requireNonNull(headers.get(HttpHeaderNames.LOCATION));` statement

No problem if I use `followRedirect(true)`

## Expected Behavior
Both approaches shouldn't raise any **NullPointerException**

## Actual Behavior
See above

## Possible Solution
Remove the check on Location header

## Your Environment
Spring boot 2.2.0-RELEASE

* Reactor version(s) used: 2022.0.2
* JVM version (`java -version`): java 8
* OS and version (eg. `uname -a`): Windows Server 2016

violetagg:
Fixes #2670

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
@@ -741,11 +741,19 @@ class HttpClientOperations extends HttpOperations<NettyInbound, NettyOutbound>
 
 	final boolean notRedirected(HttpResponse response) {
 		if (isFollowRedirect() && followRedirectPredicate.test(this, this)) {
+			try {
+				redirecting = new RedirectClientException(response.headers(), response.status());
+			}
+			catch (RuntimeException e) {
+				if (log.isDebugEnabled()) {
+					log.debug(format(channel(), "The request cannot be redirected"), e);
+				}
+				return true;
+			}
 			if (log.isDebugEnabled()) {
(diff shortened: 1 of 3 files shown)


pderop (approved):
+1

violetagg:

PR #2679, opened 2023-01-30
`Reactor Netty` javadoc generation does not need `jackson-databind` javadoc url

violetagg:
`jackson-databind` is used only in tests, url to `jackson-databind` javadoc is not needed for generating `Reactor Netty` javadoc


pderop (approved):

violetagg:
@pderop Thanks

PR #2680, opened 2023-01-31
Use empty shared `ChannelHandler` for inbound stream handling on the `HttpClient`

violetagg:
Handling inbound streams (server pushes) is not supported and disabled. Use empty shared `ChannelHandler` for configuring `Http2MultiplexHandler`.


pderop (approved):
looks good to me

violetagg:
@pderop Thanks

PR #2688, opened 2023-02-08
Fix `Netty 5` leak detection property name

violetagg:


pderop (approved):
LGTM. 

I will check if the same needs to be done in https://github.com/netty/netty/pull/12764

violetagg:
See this one https://github.com/netty/netty/pull/13198

PR #2694, opened 2023-02-13
Depend on `Netty` `v4.1.89.Final`

violetagg:


violetagg:
The CI fails with the new Netty 4.1.88.Final
https://github.com/netty/netty/issues/13214

violetagg:
Updated the Netty version to 4.1.89.Final

Issue #68, opened 2017-04-06
HttpServerRoutes should support query parameters


lucasponce:
Defining a route [1] like
```
        context = HttpServer.create(8556)
                .newRouter(r -> r
                        .get("/hawkular/alerts", (req, resp) -> {
                            return resp.sendString(Mono.just("Get alerts"));
                        })
                        .route(req -> true, (req, resp) -> {
                            QueryStringDecoder query = new QueryStringDecoder(req.uri());
                            System.out.printf("route: %s %s\n", query.path(), query.parameters());
                            return resp.send();
                        })
                ).block();
```
get should accept query parameters in some way.

For example, a call like
```
curl http://localhost:8556/hawkular/alerts?query=AA
```
should be handled by the "get" route and not the second "route" one, as there is match in the path.

[1] https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/ipc/netty/http/server/HttpServerRoutes.java#L174

P.S.: Testing this on 3.0.5.RELEASE version

odravison:
This feature was implemented?

violetagg:
@odravison No it is not. Are you interested in providing a PR?

odravison:
Maybe I do this soon. Just let me get a free time to do so.

Do you know if this feature (query parameters) was implemented on HttpClient module?

violetagg:
@odravison There is similar feature request for query params on the client https://github.com/reactor/reactor-netty/issues/753

violetagg:
@odravison Also can you clarify the use case? Is it for routing? If yes can't you use the current API for custom predicates for routing?

violetagg:
Note: When implementing this feature, we need to discuss also non URL-encode query parameters - #1561

coditori:
From 2017 to 2022 and still nothing :|

LivingLikeKrillin:
Hi @violetagg,

It seems issue #1561 was closed with a suggested workaround, but the fundamental query parameter handling issue described here (#68) is still unresolved.

I'd like to help implement the solution for #68 by introducing direct support for query parameter-aware routing, thereby addressing the core problem explicitly rather than via workaround.

If you have any additional design considerations or preferred approaches, please let me know. I'm eager to contribute.

Thanks!

violetagg:
@LivingLikeKrillin We have this PR https://github.com/reactor/reactor-netty/pull/2696 which we would like to include in 1.3 version

PR #2696, opened 2023-02-15
HttpServerRequest supports query parameters.

jchenga:
This PR is for issue #68. I have done an implementation to add support for query parameters in HttpServerRequest using Netty's QueryStringDecoder. 

I think more needs to be done to fully flesh out the feature, such as injecting a querystring decoder instead of newing one up. @violetagg and @pderop  any suggestion is appreciated.

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/QueryStringDecoder.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/QueryStringDecoder.java
@@ -1,0 +1,203 @@
+/*
+ * Copyright (c) 2012-2023 VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty.http;
+
+import static io.netty.util.internal.StringUtil.EMPTY_STRING;
+import static io.netty.util.internal.StringUtil.SPACE;
+import static io.netty.util.internal.StringUtil.decodeHexByte;
+
+import io.netty.handler.codec.http.HttpConstants;
+import io.netty.util.internal.PlatformDependent;
+
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Provides utility methods to split an HTTP query string into key-value parameter pairs.
+ * <pre>
+ * {@link Map} parameters = {@link QueryStringDecoder}.decodeParams("/hello?recipient=world&x=1;y=2");
+ * assert parameters.get("recipient").get(0).equals("world");
+ * assert parameters.get.get("x").get(0).equals("1");
+ * assert parameters.get.get("y").get(0).equals("2");
+ * </pre>
+ *
+ *
+ * <h3>HashDOS vulnerability fix</h3>
+ *
+ * As a workaround to the <a href="https://netty.io/s/hashdos">HashDOS</a> vulnerability, the decoder
+ * limits the maximum number of decoded key-value parameter pairs, up to {@literal 1024} by
+ * default. you can configure it when you construct the decoder by passing an additional
+ * integer parameter.
+ *
+ */
+public class QueryStringDecoder {
+
+    private static final int DEFAULT_MAX_PARAMS = 1024;
+
+	public static Map<String, List<String>> decodeParams(final String uri) {
+
+		return decodeParams(uri, HttpConstants.DEFAULT_CHARSET,
+				DEFAULT_MAX_PARAMS, true);
+	}
+
+	public static Map<String, List<String>> decodeParams(final String uri, final boolean semiColonIsNormalChar) {
+
+		return decodeParams(uri, HttpConstants.DEFAULT_CHARSET,
+				DEFAULT_MAX_PARAMS, semiColonIsNormalChar);
+	}
+
+	public static Map<String, List<String>> decodeParams(final String uri, final Charset charset, final int maxParams, final boolean semicolonIsNormalChar) {
+		Objects.requireNonNull(uri, "uri");
+		Objects.requireNonNull(charset, "charset");
+		if (maxParams < 1) {
+			throw new IllegalArgumentException("maxParams must be positive");
+		}
+
+		int from = findPathEndIndex(uri);
+		return decodeParams(uri, from, charset,
+				maxParams, semicolonIsNormalChar);
+	}
+
+    private static Map<String, List<String>> decodeParams(String s, int from, Charset charset, int paramsLimit,
+                                                          boolean semicolonIsNormalChar) {
+        int len = s.length();
+        if (from >= len) {
+            return Collections.emptyMap();
+        }
+        if (s.charAt(from) == '?') {
+            from++;
+        }
+        Map<String, List<String>> params = new LinkedHashMap<String, List<String>>();
+        int nameStart = from;
+        int valueStart = -1;
+        int i;
+        loop:
+        for (i = from; i < len; i++) {
+            switch (s.charAt(i)) {
+            case '=':
+                if (nameStart == i) {
+                    nameStart = i + 1;
+                }
+				else if (valueStart < nameStart) {
+                    valueStart = i + 1;
+                }
+                break;
+            case ';':
+                if (semicolonIsNormalChar) {
+                    continue;
+                }
+                // fall-through
+            case '&':
+                if (addParam(s, nameStart, valueStart, i, params, charset)) {
+                    paramsLimit--;
+                    if (paramsLimit == 0) {
+                        return params;
+                    }
+                }
+                nameStart = i + 1;
+                break;
+            case '#':
+                break loop;
+            default:
+                // continue
+            }
+        }
+        addParam(s, nameStart, valueStart, i, params, charset);
+        return params;
+    }
+
+    private static boolean addParam(String s, int nameStart, int valueStart, int valueEnd,
+                                    Map<String, List<String>> params, Charset charset) {
+        if (nameStart >= valueEnd) {
+            return false;
+        }
+        if (valueStart <= nameStart) {
+            valueStart = valueEnd + 1;
+        }
+        String name = decodeComponent(s, nameStart, valueStart - 1, charset, false);
+        String value = decodeComponent(s, valueStart, valueEnd, charset, false);
+        List<String> values = params.get(name);
+        if (values == null) {
+            values = new ArrayList<String>(1);  // Often there's only 1 value.
+            params.put(name, values);
+        }
+        values.add(value);
+        return true;
+    }
+
+    private static String decodeComponent(String s, int from, int toExcluded, Charset charset, boolean isPath) {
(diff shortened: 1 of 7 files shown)


pderop:
@jchenga ,

thanks for this PR ! I will look into it soon today.

pderop:
@violetagg , 

I have set the target milestone to 1.1.4 since it's an enhancement. Let me know if we shoud backport this to 1.0.x branch ?
thanks

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
```diff
@@ -459,6 +462,19 @@ public HttpHeaders requestHeaders() {
...
+		if (queryParamsMap == null) {
+			queryParamsMap = Collections.unmodifiableMap(parseQueryParams(this.nettyRequest));
+		}
+		return queryParamsMap;
+	}
+
+	private Map<String, List<String>> parseQueryParams(HttpRequest request) {
+		QueryStringDecoder decoder = new QueryStringDecoder(request.uri());
+		return decoder.parameters();
+	}
```
pderop, line 476: in order to avoid mixing public and private methods, what do you think about moving the private `parseQueryParams`  method just before the static inner classes that are declared in the end of the class ?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRequest.java
```diff
@@ -143,6 +144,13 @@ default Flux<HttpContent> receiveContent() {
 	 */
 	HttpHeaders requestHeaders();
 
+	/**
+	 * return parsed and decoded query parameter name value pairs
+	 *
+	 * @return query parameters {@link Map&lt;String,List&lt;String&gt;&gt;}
+	 */
```
pderop, line 151:
```suggestion
	 * @return query parameters {@link Map}
	 */
```

To avoid some compilation warnings, I think it is safe to just remove the generic types in the javadoc link.

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2689,6 +2689,58 @@ void testUseComparatorOrderRoutes() {
...
 	}
 
+	@Test
+	void testRouteQuery() {
+		HttpServerRoutes serverRoutes = HttpServerRoutes.newRoutes()
+				.get("/yes/value",
+						(request, response) -> {
+					StringBuilder sbuf = new StringBuilder();
+					request.queryParams().forEach((key, list) -> {
+
```
pderop, line 2699:
```suggestion
```

this is just a formatting nit: can we remove this blank line ?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRequest.java
pderop, line 151:
```suggestion
	 * @since 1.1.4
	 */
```

can you please add the @since annotation ?
Take care: I have set the milestone for this PR to 1.1.4. But, when Violeta will do her review, maybe she will want to backport this to the 1.0.x; in this case we'll have to use `@since 1.0.29`

pderop (commented):
@jchenga ,

So, I have taken a look, can you please check the following reviews  ?
In addition, I have the following comments:

- can you rebase your PR on top of the latest main branch, because for japicmp, you need to update the reactor-netty-http/build.gradle, which has been updated yesterday when we did the release ? thanks.

- japicmp:
  in order to avoid the japicmp error, can you please update the reactor-netty-http/build.gradle, and exclude the new `queryParams()`method, something like this line 244:
  
```
	methodExcludes = [
			'reactor.netty.http.server.HttpServerRequest#queryParams()'
	]
```

(in the latest branch, the methodExcludes is now empty because we have made the release yesterday, so just add your new method).

- I'm not sure if there is more works to be done, such as injecting a querystring decoder ? Indeed the decoder seems to be immutable, and contains only few objects, so why would we have to inject a reusable decoder ? But maybe I missed your point ?

thank you !

violetagg:
@jchenga Can we limit this feature only to `route` functionality similar to how it is done for parameters resolution (see `reactor.netty.http.server.DefaultHttpServerRoutes.HttpRouteHandler` and `reactor.netty.http.server.HttpPredicate`)? Frameworks typically use `handle` and they may want to use something else for query resolution.

violetagg:
@jchenga Take a look at the parameters resolver. You can provide a resolver so that if you use `handle` we can use this resolver to give you want you want. The point is that frameworks that use `handle` may not want to use the Netty's query resolver but something else (e.g. Spring Framework).
https://github.com/reactor/reactor-netty/blob/906de4bcd67c309640d45b71d162667fe1755ef3/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerRequest.java#L70

I also need to take a look at the Netty's query resolver. It seems it resolves everything? if that's true we will have double parsing of the uri or not? https://github.com/reactor/reactor-netty/blob/906de4bcd67c309640d45b71d162667fe1755ef3/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java#L194

violetagg:
@jchenga I see now...

Can we discuss with Netty maintainers to make `decodeParams` `public` similar to `decodeComponent` as basically we need just this utility or it is not the case?

violetagg:
I mean I don't see why we want to create this object every time, basically `QueryStringDecoder` holds mostly the same information as our HttpServerOperations, isn't it?

jchenga:
There is definitely no need to create an QueryStringDecoder object every time parseQueryParams method is called. I tried to eliminate the need to recreate objects by including the null check on the map.

https://github.com/reactor/reactor-netty/blob/1e693a9b4619e98c8abebc0f277dcbeacff88919/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java#L467-L469

I see your point about having to create an object every time we need to parse query params. I can reach out to Netty's maintainers to see if they are willing to make `decodeParams` `public`. 

another option is to bring in `decodeParams` codes from Netty into `HttpOperations`.  Since `decodeParams` is trivial relative to other logics, perhaps we can just bring in the function into this project? Or maybe reusing the similar function in spring is an option too?

Let me know which approach you prefer.

violetagg:
> another option is to bring in decodeParams codes from Netty into HttpOperations. Since decodeParams is trivial relative to other logics, perhaps we can just bring in the function into this project?

My opinion is that forking/copying comes with a drawback that we need always to check whether there are some changes to the original code.

So my proposal is to ask Netty maintainers. If this suggestion is not ok from their point of view, we can discuss the other two options. Wdyt?

jchenga:
> > another option is to bring in decodeParams codes from Netty into HttpOperations. Since decodeParams is trivial relative to other logics, perhaps we can just bring in the function into this project?
> 
> My opinion is that forking/copying comes with a drawback that we need always to check whether there are some changes to the original code.
> 
> So my proposal is to ask Netty maintainers. If this suggestion is not ok from their point of view, we can discuss the other two options. Wdyt?

Sounds good to me. I'll create a request over at the Netty's repo for this.

jchenga:
@violetagg it looks like Netty folks are not keen on exposing decodeParams as public API. Let's go with bringing the codes into this project's codebase? what do you think?

violetagg:
> @violetagg it looks like Netty folks are not keen on exposing decodeParams as public API. Let's go with bringing the codes into this project's codebase? what do you think?

I will need to take a closer look at this one ...

jchenga:
> > @violetagg it looks like Netty folks are not keen on exposing decodeParams as public API. Let's go with bringing the codes into this project's codebase? what do you think?
> 
> I will need to take a closer look at this one ...

Let me know once you know what you would like to do and I'll jump in to help.

violetagg:
> @violetagg it looks like Netty folks are not keen on exposing decodeParams as public API. Let's go with bringing the codes into this project's codebase? what do you think?

@jchenga Let's go with bringing this code to Reactor Netty

jchenga:
> 

sure thing. let me get started on that.

jchenga:
@violetagg I've finally go time to pick this up again. the PR failed the API compatibility check because one new method was added to the HttpServerRequest interface. What should I do to resolve this failure?

violetagg:
> @violetagg I've finally go time to pick this up again. the PR failed the API compatibility check because one new method was added to the HttpServerRequest interface. What should I do to resolve this failure?

Can you try adding this?

```
diff --git a/reactor-netty-http/build.gradle b/reactor-netty-http/build.gradle
index 02b717a63..3272339ed 100644
--- a/reactor-netty-http/build.gradle
+++ b/reactor-netty-http/build.gradle
@@ -250,6 +250,7 @@ task japicmp(type: JapicmpTask) {
 
 	compatibilityChangeExcludes = [ "METHOD_NEW_DEFAULT" ]
 	methodExcludes = [
+			'reactor.netty.http.server.HttpServerRequest#queryParams()'
 	]
 }
```

reactor-netty-http/src/main/java/reactor/netty/http/QueryStringDecoder.java
```diff
@@ -0,0 +1,203 @@
+/*
+ * Copyright (c) 2012-2023 VMware, Inc. or its affiliates, All Rights Reserved.
```
jchenga, line 2: since the codes came from Netty, should the license statement be replaced with the one from Netty?

jchenga:
> > @violetagg I've finally go time to pick this up again. the PR failed the API compatibility check because one new method was added to the HttpServerRequest interface. What should I do to resolve this failure?
> 
> Can you try adding this?
> 
> ```
> diff --git a/reactor-netty-http/build.gradle b/reactor-netty-http/build.gradle
> index 02b717a63..3272339ed 100644
> --- a/reactor-netty-http/build.gradle
> +++ b/reactor-netty-http/build.gradle
> @@ -250,6 +250,7 @@ task japicmp(type: JapicmpTask) {
>  
>  	compatibilityChangeExcludes = [ "METHOD_NEW_DEFAULT" ]
>  	methodExcludes = [
> +			'reactor.netty.http.server.HttpServerRequest#queryParams()'
>  	]
>  }
> ```

done. that resolved the issue.

reactor-netty-http/src/main/java/reactor/netty/http/QueryStringDecoder.java
jchenga: @violetagg could you have a look and see if the PR is good to go?

violetagg:
@jchenga Can you please rebase?

jchenga:
@violetagg I have rebased the branch. the windows 2019, nio build failed, but it does not seem to be related to this PR

jchenga:
Hi @violetagg  I rebased the feature branch on main and some checks failed. Looking at the errors, I don't think the issue is related to the change in the PR. the failures seem to OS specific as "Check Matrix / build (ubuntu-20.04, native) (pull_request) "  and "Check Matrix / build (ubuntu-20.04, nio) (pull_request) " were successful. 

I am stumped. would you be able to help?

violetagg:
> Hi @violetagg I rebased the feature branch on main and some checks failed. Looking at the errors, I don't think the issue is related to the change in the PR. the failures seem to OS specific as "Check Matrix / build (ubuntu-20.04, native) (pull_request) " and "Check Matrix / build (ubuntu-20.04, nio) (pull_request) " were successful.
> 
> I am stumped. would you be able to help?

I ran again the tests and now they are ok

jchenga:
> > Hi @violetagg I rebased the feature branch on main and some checks failed. Looking at the errors, I don't think the issue is related to the change in the PR. the failures seem to OS specific as "Check Matrix / build (ubuntu-20.04, native) (pull_request) " and "Check Matrix / build (ubuntu-20.04, nio) (pull_request) " were successful.
> > I am stumped. would you be able to help?
> 
> I ran again the tests and now they are ok

Thank you for the help. This PR is ready to be merged if everything looks ok.

jchenga:
@violetagg this PR has been around for a while now. is there anything I can do to help move this PR along?

violetagg:
> @violetagg this PR has been around for a while now. is there anything I can do to help move this PR along?

I'll check this one in the next weeks

PR #2700, opened 2023-02-20
Refactor UriEndpoint

### LINKED ISSUE #829: Revise UriEndpointFactory/UriEndpoint

yuzawa-san:
Clean up UriEndpoint logic to be based around the java.net.URI. Move relevant methods into this class. Construct instances using static method. Only calculate derived values on demand (e.g. toExternalForm) when needed in the various request implementations. Cleaned up redirection logic. Added additional tests. Made conversion from InetSocketAddress more efficient (removed the substrings which trimmed out the ports). Fixed HttpClientTest to support ipv6 (by using NetUtil.toSocketAddressString to get the `[]` around the address, and also fixed FailedHttpClientRequest.uri() from returning full URI rather than just Netty definition of "uri" which is raw path and query.

Fixes #829


violetagg:
@yuzawa-san I'm gonna take a look at this

violetagg:
@yuzawa-san I'm gonna review this for 1.0.33

violetagg:
This is superseded by the PRs linked above

PR #2712, opened 2023-03-01
Handling `X-Forwarded-Host` with value `host:port`

### LINKED ISSUE #2711: Handling of "X-Forwarded-Host" to support "host:port"
After Spring Framework switched to using `HttpServerRequest#hostAddress` first instead of parsing the "Host" header in https://github.com/spring-projects/spring-framework/issues/28601 in order to ensure forwarded headers take effect. 

However, as reported in https://github.com/spring-projects/spring-boot/issues/34395 and discussed under https://github.com/spring-projects/spring-framework/issues/30033, if "X-Forwarded-Host" has host and port, e.g. "localhost:3000" then `HttpServerRequest#hostAddress` returns a `hostString` with the port in it. It would be useful to parse that since it can happen.

violetagg:
Fixes #2711


pderop (approved):

violetagg:

PR #2714, opened 2023-03-02
Add `ConnectionInformation#hostName/hostPort`

violetagg:
- The host name and port are derived from the `Host`/`X-Forwarded-*`/`Forwarded` header associated with this request.
- When `X-Forwarded-*`/`Forwarded` header does not specify a port, add a port as per the scheme


rstoyanchev (approved):

pderop (approved):
it looks good to me.

violetagg:
@rstoyanchev @pderop Thanks for the review!

PR #2720, opened 2023-03-07
Fix test failure on Windows OS

violetagg:


violetagg:
The failures from the tests for QUIC module are not related as this change touches tests for HTTP module.

PR #2731, opened 2023-03-19
Replace Objects.hash()

yuzawa-san:
This call requires an Object[] allocation and does boxing on primitives, so should not be used in high performance environments.

CPU usage before: 
image

CPU usage after:
image


violetagg:
This is superseded by #2731

PR #2732, opened 2023-03-19
Replace Objects.hash()

yuzawa-san:
This call requires an Object[] allocation and does boxing on primitives, so should not be used in high performance environments.

CPU usage before: 
image

CPU usage after:
image

sorry for the duplicate, wrong base branch on the first time


violetagg (approved):
@reactor/netty-team PTAL

pderop (approved):
thanks for the PR !

LGTM

violetagg:
I back ported this to version `1.0.x` and also applied to `netty5` branch

PR #2733, opened 2023-03-19
Smarter usage of `FQDN` in `PoolKey`

yuzawa-san:
The SocketAddres.toString method to generate the FQDN is relatively heavy and it appears to be unnecessary given the socket address is being hashed and equaled already. This should work for unix domain sockets as well.

This code:
````java
System.out.println(AddressUtils.createUnresolved("127.0.0.1", 80));
System.out.println(AddressUtils.createUnresolved("127.0.0.1", 80).equals(AddressUtils.createUnresolved("127.0.0.1", 80)));
System.out.println(AddressUtils.createUnresolved("example.com", 80));
System.out.println(AddressUtils.createUnresolved("example.com", 80).equals(AddressUtils.createUnresolved("example.com", 80)));
System.out.println(AddressUtils.createResolved("example.com", 80));
System.out.println(AddressUtils.createResolved("example.com", 80).equals(AddressUtils.createResolved("example.com", 80)));
System.out.println(AddressUtils.createResolved("example.com", 80).equals(AddressUtils.createUnresolved("example.com", 80)));
System.out.println(UnixDomainSocketAddress.of("/tmp/mysock"));
````
prints out:
````
/127.0.0.1:80
true
example.com/<unresolved>:80
true
example.com/93.184.216.34:80
true
false
/tmp/mysock
````

before:
image

and after this is basically gone
image

I also fixed up the hashCode like I did in https://github.com/reactor/reactor-netty/pull/2732

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -240,7 +240,7 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 		toDispose.forEach(e -> {
 			if (channelPools.remove(e.getKey(), e.getValue())) {
 				if (log.isDebugEnabled()) {
-					log.debug("ConnectionProvider[name={}]: Disposing pool for [{}]", name, e.getKey().fqdn);
+					log.debug("ConnectionProvider[name={}]: Disposing pool for [{}]", name, e.getKey().holder);
 				}
 				String id = e.getKey().hashCode() + "";
 				PoolFactory<T> poolFactory = poolFactory(address);
@@ -364,7 +364,7 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 			toDispose.forEach(e -> {
 				if (channelPools.remove(e.getKey(), e.getValue())) {
 					if (log.isDebugEnabled()) {
-						log.debug("ConnectionProvider[name={}]: Disposing inactive pool for [{}]", name, e.getKey().fqdn);
+						log.debug("ConnectionProvider[name={}]: Disposing inactive pool for [{}]", name, e.getKey().holder);
 					}
 					e.getValue().dispose();
 				}
@@ -633,7 +633,15 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 		final int pipelineKey;
 
 		PoolKey(SocketAddress holder, int pipelineKey) {
-			this.fqdn = holder.toString();
+			String fqdn = null;
+			if (holder instanceof InetSocketAddress) {
+				InetSocketAddress inetSocketAddress = (InetSocketAddress) holder;
+				if (!inetSocketAddress.isUnresolved()) {
+					// Use FDQN as a tie-breaker over IP's
+					fqdn = inetSocketAddress.getHostString().toLowerCase();
+				}
+			}
+			this.fqdn = fqdn;
 			this.holder = holder;
 			this.pipelineKey = pipelineKey;
 		}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by yuzawa-san** [comment]
@violetagg I set a breakpoint in the existing code when my client hits `https://en.wikipedia.org/` and `https://wikipedia.org/` and the socket (toString) is `en.wikipedia.org/<unresolved>:443`  and `wikipedia.org/<unresolved>:443` respectively

````java
System.out.println(AddressUtils.createUnresolved("wikipedia.org", 443).equals(AddressUtils.createUnresolved("en.wikipedia.org", 443)));
````
prints `false` which is expected, indicating those would go into two separate pools which is correct.

is the concern that some resolved InetSocketAddresses could be bucketed in the pool together? like weird mixtures of resolved and unresolved InetSocketAddresses or ip-address only vs hostname only InetSocketAddresses? or would it be inconsistent across different java versions? looks like it would work:  https://github.com/openjdk/jdk/blob/jdk-11%2B0/src/java.base/share/classes/java/net/InetSocketAddress.java#L112

I can post some more mixtures shortly if you need more concrete examples

**Comment by violetagg** [comment]
Can you try this?

```
HttpClient client = HttpClient.create().wiretap(true);

client.remoteAddress(() -> new InetSocketAddress("wikipedia.org", 443)).get().uri("/").responseContent().aggregate().asString().block();

client.remoteAddress(() -> new InetSocketAddress("en.wikipedia.org", 443)).get().uri("/").responseContent().aggregate().asString().block();
```

https://httpwg.org/specs/rfc7540.html#reuse

**Comment by yuzawa-san** [comment]
@violetagg my original traces all involved unresolved InetSocketAddresses, but the wikipedia examples are a prime example of where the resolved addresses are the same and thus equals and hashcode bucket them into the same pool key. i found a solution which only uses FQDN for tiebreaking when the InetSocketAddress has a resolved address.

this has a fast path still for the unresolved addresses coming out of the UriEndpointFactory. since those are unresolved, we can skip the fqdn and lean directly on the [equals](https://github.com/openjdk/jdk/blob/da75f3c4ad5bdf25167a3ed80e51f567ab3dbd01/src/java.base/share/classes/java/net/InetSocketAddress.java#L120-L121) and [hashcode](https://github.com/openjdk/jdk/blob/da75f3c4ad5bdf25167a3ed80e51f567ab3dbd01/src/java.base/share/classes/java/net/InetSocketAddress.java#L131-L132) of InetSocketAddress.

i also added a test case. i found that the `holder.toString()` actually did not mimic the toLowerCase behavior present in those links, so maybe that is an extra bugfix i stumbled upon. see `key3` and `key4`, those were true in the prior version of the code.

File: reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java [hunk 1]
```diff
@@ -633,7 +633,15 @@ static final class PoolKey {
 		final int pipelineKey;
 
 		PoolKey(SocketAddress holder, int pipelineKey) {
-			this.fqdn = holder.toString();
+			String fqdn = null;
+			if (holder instanceof InetSocketAddress) {
+				InetSocketAddress inetSocketAddress = (InetSocketAddress) holder;
+				if (!inetSocketAddress.isUnresolved()) {
+					// Use FDQN as a tie-breaker over IP's
```
violetagg, line 640:
```suggestion
					// Use FQDN as a tie-breaker over IP's
```

reactor-netty-core/src/test/java/reactor/netty/resources/PoolKeyTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class PoolKeyTest {
+
+	@Test
+	void test() throws IllegalAccessException {
```
violetagg, line 30: nit: Throwing `IllegalAccessException` is not needed

```suggestion
	void test() {
```

violetagg (changes requested):
@yuzawa-san Thanks

only two small comments

violetagg:
@reactor/netty-team PTAL

pderop (approved):
@yuzawa-san , thanks for the PR.

violetagg (approved):
Thanks

PR #2736, opened 2023-03-22
When graceful shutdown, ensure the streams are processed before closing connection

### LINKED ISSUE #2735: Server closes the connection with a FIN packet immediately after sending a GO_AWAY frame. All accepted streams upto the lastStreamId are not getting processed
My application server using http2 reactive netty server (version: 1.0.26) closes the connection immediately after sending a GO_AWAY packet, without processing outstanding requests up to the lastStreamId. 

+--- org.springframework.boot:spring-boot-starter-webflux:2.7.7
|    +--- org.springframework.boot:spring-boot-starter:2.7.7 (*)
|    +--- org.springframework.boot:spring-boot-starter-json:2.7.7 (*)
|    +--- org.springframework.boot:spring-boot-starter-reactor-netty:2.7.7
|    |    \--- io.projectreactor.netty:reactor-netty-http:1.0.26

[https://www.rfc-editor.org/rfc/rfc7540#section-6.8](RFC-7540) states that all "Activity on streams numbered lower or equal to the last stream identifier might still complete successfully.  The sender of a GOAWAY frame might gracefully shut down a connection by sending a GOAWAY frame, maintaining the connection in an "open" state until all in-progress streams complete."

Does reactive netty server support processing of all accepted streams upto the lastStreamId sent in GO_AWAY frame before terminating the connection? If yes, then please advise if some configuration is needed to make reactive netty http2 server to ensure the processing of all streams up to the lastStreamId quoted in GO_AWAY frame.

violetagg:
Fixes #2735

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2022 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2020-2023 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -36,6 +37,7 @@ import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.EventLoopGroup;
+import io.netty.channel.ServerChannel;
 import io.netty.channel.unix.DomainSocketAddress;
 import io.netty.handler.codec.DecoderException;
 import io.netty.util.AttributeKey;
@@ -50,6 +52,7 @@ import reactor.netty.ConnectionObserver;
 import reactor.netty.DisposableServer;
 import reactor.netty.channel.AbortedException;
 import reactor.netty.channel.ChannelOperations;
+import reactor.netty.internal.util.MapUtils;
 import reactor.netty.resources.ConnectionProvider;
 import reactor.netty.resources.LoopResources;
 import reactor.util.Logger;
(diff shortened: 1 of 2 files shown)


pderop (approved):
ok, nice patch, looks good to me.

reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
```diff
@@ -532,21 +535,31 @@ public void disposeNow(Duration timeout) {
 			}
 			dispose();
 			Mono<Void> terminateSignals = Mono.empty();
-			if (config.channelGroup != null) {
-				List<Mono<Void>> channels = new ArrayList<>();
+			if (config.channelGroup != null && config.channelGroup.size() > 0) {
+				HashMap<Channel, List<Mono<Void>>> channelsToMono = new HashMap<>(config.channelGroup.size());
```
robotmrv, line 539: https://bugs.openjdk.org/browse/JDK-8186958 
> It seems that people normally assume that the constructor HashMap(int) requires the expected size as the argument.
In fact, the argument is the initial capacity, which should be bigger than the expected size. As a result, unnecessary reallocations can occur while the maps are filled.

violetagg:
`HttpClientTest > testHttpClientCancelled()` is failing on Windows OS. I checked the test and it does not enable graceful shutdown. So my opinion is that this failure is not because of this change.

https://github.com/reactor/reactor-netty/actions/runs/4488211188/jobs/7893282178#step:5:157

reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
violetagg: @robotmrv What is your proposal?

robotmrv: I've just noticed possible misuse of the HashMap constructor.  
Probably it could improved the same way as in jdk or guava analog via utility method 
https://github.com/openjdk/jdk/pull/7928/files#diff-9837dc8a9161163b6a1489098920d47c7fb62f19d9f929c1447f3c1c9bf7aa68R2581

or 
https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Maps.java#L277

robotmrv: I am ok 
but quick search shows that there are several similar cases   
https://github.com/reactor/reactor-netty/blob/143191ae6db99c07e509cb62b60e362b771c2fa5/reactor-netty-core/src/main/java/reactor/netty/transport/TransportConfig.java#L354  
https://github.com/reactor/reactor-netty/blob/058159592584fcc3a66c32b48a5bfe6a8fe09297/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpPredicate.java#L347
https://github.com/reactor/reactor-netty/blob/2ea7d622242f8750b70844c0a69a49446e8872f1/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerBind.java#L44

probably it is a good idea to evaluate whether they require the same fix.  
If so - create another PR and move this logic to the utility method

violetagg: here it is #2739

PR #2744, opened 2023-03-26
Support a custom DnsAddressResolverGroup builder function

samueldlightfoot:
This PR is in draft to get feedback on the signature to allow a user supplied function to create a DnsAddressResolverGroup.

Our use-case is to use our extended version of RoundRobinInetAddressResolver with the HttpClient.

Keen to get your thoughts @violetagg on if this is the most sensible way to achieve this.

Work remaining
- Docs
- JAPI compare


violetagg:
@samueldlightfoot After a quick glance it seems ok

samueldlightfoot:
Non-draft PR open here: https://github.com/reactor/reactor-netty/pull/2745

PR #2750, opened 2023-03-29
Temporarily disable publish action for netty5 branch

pderop:
Currently, we depend now on socks-proxy/codec-multipart snapshots, so for the moment, let's disable publish action.


pderop:
@violetagg , thanks.

PR #2756, opened 2023-03-30
Ensure early dispose is delivered to late subscriber

OlegDokuka:
This PR fixes the issue that appears when late inbound subscriber races with connection dispose

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
@@ -170,7 +170,7 @@ final class FluxReceive extends Flux<Object> implements Subscription, Disposable
 			s.onSubscribe(this);
 		}
 		else {
-			if ((inboundDone && getPending() == 0) || isCancelled()) {
+			if (inboundDone && getPending() == 0) {
 				if (inboundError != null) {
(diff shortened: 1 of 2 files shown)


reactor-netty-core/src/main/java/reactor/netty/channel/FluxReceive.java
```diff
@@ -153,11 +150,12 @@ public void subscribe(CoreSubscriber<? super Object> s) {
...
 
 	final void startReceiver(CoreSubscriber<? super Object> s) {
-		if (once == 0 && ONCE.compareAndSet(this, 0, 1)) {
+		if (!subscribedOnce) {
+			subscribedOnce = true;
 			if (log.isDebugEnabled()) {
 				log.debug(format(channel, "{}: subscribing inbound receiver"), this);
 			}
-			if (inboundDone && getPending() == 0) {
+			if ((inboundDone && getPending() == 0) || isCancelled()) {
```
OlegDokuka, line 158: since cancelation is possible when there is a subscriber, in that case isCancelled would mean dispose happened earlier

OlegDokuka:
with that fix I dont see the rsocket test hanging because of an undelivered completion signal on the inbound side) (need that to wait for all bufs to be delivered and consumed)

violetagg (approved):

PR #2761, opened 2023-04-03
Parse X-Forwarded-Port header even if X-Forwarded-Host is not present

### LINKED ISSUE #2751: `X-Forwarded-Port` is sometimes ignored
## Expected Behavior

It is expected that `reactor-netty` always honors the provided `X-Forwarded-Port` when calling `HttpServerRequest#hostPort`.

## Actual Behavior

The problem is that `reactor-netty` does not honor the `X-Forwarded-Port` header if `X-Forwarded-Host` [isn’t set](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpForwardedHeaderHandler.java#L106), and thus it can report the port of a proxied connection rather than the actual (forwarded) host port.

As a result, `reactor-netty` incorrectly reports port 80, while the scheme is `https` and `X-Forwarded-Port` is set. This becomes a problem in our setup which is as follows:

We have AWS Application Load Balancers (ALBs) which perform TLS termination and proxy requests to pods running within our Kubernetes clusters. ALBs only add the following headers to a proxied request, as per the [documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/x-forwarded-headers.html):

- `X-Forwarded-For`
- `X-Forwarded-Proto`
- `X-Forwarded-Port`

In addition, the `Host` header is set by the original request. However, when we call `HttpServerRequest#hostPort` for a request that requests a URL of the format `https://some-host.domain`, we would expect `hostPort`  to return `443`. Instead, it returns `80`.

The following issue may be related: https://github.com/reactor/reactor-netty/pull/2714.
This became a problem in combination with the following commit from Spring framework: https://github.com/spring-projects/spring-framework/commit/a2b7a907ec8aa5f34a20f533a064067e56ef6967, where they (correctly) prefer the host address rather than parsing the `Host` header themselves.

## Steps to Reproduce

```java
@Test
void xForwardedForAndHost() {
	testClientRequest(
			clientRequestHeaders -> {
				clientRequestHeaders.add("Host", "a.example.com");
				clientRequestHeaders.add("X-Forwarded-For", "192.168.0.1");
				clientRequestHeaders.add("X-Forwarded-Port", "443");
				clientRequestHeaders.add("X-Forwarded-Proto", "https");
			},
			serverRequest -> {
				Assertions.assertThat(serverRequest.hostPort()).isEqualTo(443);
				Assertions.assertThat(serverRequest.hostName()).isEqualTo("a.example.com");
				Assertions.assertThat(serverRequest.scheme()).isEqualTo("https");
			});
}
```
This test fails as the port returned is `80` instead of `443`.

## Possible Solution

Adding the following line results in the expected result:
```
clientRequestHeaders.add("X-Forwarded-Host", "a.example.com");
```

## Your Environment

* Reactor version(s) used: 2020.0.30, we use the `reactor-bom`. 
* Other relevant libraries versions (eg. `netty`, ...): Netty 1.0.30.
```
$ java -version
openjdk version "17.0.3" 2022-04-19 LTS
OpenJDK Runtime Environment Zulu17.34+19-CA (build 17.0.3+7-LTS)
OpenJDK 64-Bit Server VM Zulu17.34+19-CA (build 17.0.3+7-LTS, mixed mode, sharing)
```
```
$ uname -a
Linux rick-ommitted 5.15.0-67-generic #74~20.04.1-Ubuntu SMP Wed Feb 22 14:52:34 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

pderop:
Always takes into account any available `X-Forwarded-Port` header even if the `X-Forwarded-Host` header is missing.

Fixes #2751

### DIFF
--- a/reactor-netty-http/src/test/java/reactor/netty/http/server/ConnectionInfoTests.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/server/ConnectionInfoTests.java
@@ -368,6 +368,30 @@ class ConnectionInfoTests extends BaseHttpTest {
 	}
 
 	@Test
+	void xForwardedForAndHostOnly() throws SSLException {
+		SslContext clientSslContext = SslContextBuilder.forClient()
+				.trustManager(InsecureTrustManagerFactory.INSTANCE).build();
+		SslContext serverSslContext = SslContextBuilder.forServer(ssc.certificate(), ssc.privateKey()).build();
+
+		testClientRequest(
+				clientRequestHeaders -> {
+					clientRequestHeaders.add("Host", "a.example.com");
+					clientRequestHeaders.add("X-Forwarded-For", "192.168.0.1");
+					clientRequestHeaders.add("X-Forwarded-Port", "8443");
+					clientRequestHeaders.add("X-Forwarded-Proto", "https");
+				},
+				serverRequest -> {
+					Assertions.assertThat(serverRequest.remoteAddress().getHostString()).isEqualTo("192.168.0.1");
+					Assertions.assertThat(serverRequest.hostPort()).isEqualTo(8443);
+					Assertions.assertThat(serverRequest.hostName()).isEqualTo("a.example.com");
+					Assertions.assertThat(serverRequest.scheme()).isEqualTo("https");
+				},
+				httpClient -> httpClient.secure(ssl -> ssl.sslContext(clientSslContext)),
+				httpServer -> httpServer.secure(ssl -> ssl.sslContext(serverSslContext)),
+				true);
+	}
+
+	@Test
 	void customForwardedHandlerForMultipleHost() {
 		testClientRequest(
 				clientRequestHeaders ->
--- a/reactor-netty-http/src/test/java/reactor/netty/http/server/ConnectionInfoTests.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/server/ConnectionInfoTests.java
@@ -382,6 +382,8 @@ class ConnectionInfoTests extends BaseHttpTest {
 				},
 				serverRequest -> {
 					Assertions.assertThat(serverRequest.remoteAddress().getHostString()).isEqualTo("192.168.0.1");
+					Assertions.assertThat(serverRequest.hostAddress().getHostString())
+							.containsPattern("^0:0:0:0:0:0:0:1(%\\w*)?|127.0.0.1$");
 					Assertions.assertThat(serverRequest.hostPort()).isEqualTo(8443);
 					Assertions.assertThat(serverRequest.hostName()).isEqualTo("a.example.com");
 					Assertions.assertThat(serverRequest.scheme()).isEqualTo("https");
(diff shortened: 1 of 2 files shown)


violetagg:
@pderop What do you think about this? We need to recreated the host information only if we are able to parse the port.

```
String hostHeader = request.headers().get(X_FORWARDED_HOST_HEADER);
if (hostHeader != null) {
	connectionInfo = connectionInfo.withHostAddress(
			AddressUtils.parseAddress(hostHeader.split(",", 2)[0].trim(),
					getDefaultHostPort(connectionInfo), DEFAULT_FORWARDED_HEADER_VALIDATION));
}

String portHeader = request.headers().get(X_FORWARDED_PORT_HEADER);
if (portHeader != null && !portHeader.isEmpty()) {
	String portStr = portHeader.split(",", 2)[0].trim();
	if (portStr.chars().allMatch(Character::isDigit)) {
		int port = Integer.parseInt(portStr);
		connectionInfo = new ConnectionInfo(
				AddressUtils.createUnresolved(connectionInfo.getHostAddress().getHostString(), port),
				connectionInfo.getHostName(), port, connectionInfo.getRemoteAddress(), connectionInfo.getScheme());
	}
	else if (DEFAULT_FORWARDED_HEADER_VALIDATION) {
		throw new IllegalArgumentException("Failed to parse a port from " + portHeader);
	}
}
```

reactor-netty-http/src/test/java/reactor/netty/http/server/ConnectionInfoTests.java
```diff
@@ -367,6 +367,30 @@ void xForwardedForMultipleHostAndPortAndProto() {
...
+
+		testClientRequest(
+				clientRequestHeaders -> {
+					clientRequestHeaders.add("Host", "a.example.com");
+					clientRequestHeaders.add("X-Forwarded-For", "192.168.0.1");
+					clientRequestHeaders.add("X-Forwarded-Port", "8443");
+					clientRequestHeaders.add("X-Forwarded-Proto", "https");
+				},
+				serverRequest -> {
+					Assertions.assertThat(serverRequest.remoteAddress().getHostString()).isEqualTo("192.168.0.1");
```
violetagg, line 384: Let's check also the host address as we do in the rest of the tests


```suggestion
					Assertions.assertThat(serverRequest.remoteAddress().getHostString()).isEqualTo("192.168.0.1");
										Assertions.assertThat(serverRequest.hostAddress().getHostString())
							.containsPattern("^0:0:0:0:0:0:0:1(%\\w*)?|127.0.0.1$");
					Assertions.assertThat(serverRequest.hostAddress().getPort()).isEqualTo(8443);
```

violetagg (approved):

pderop:
I have Applied the two feedbacks. Can you recheck please ?

Indeed, recreating the host information only if we are able to parse X-Forwarded-Port has the following benefit: if `X-Forwarded-Host` is present and contains a port and if `X-Forwarded-Port` is present but is malformed, then now we do not ignore the `X-Forwarded-Host` port (if it was present).

pderop:

PR #2762, opened 2023-04-04
Depend on `Netty` `v4.1.91.Final`

violetagg:
In case `H2C` client tries to connect to a secured `HTTP` server, the connection is directly closed. This is a result of this `Netty` change.
https://github.com/netty/netty/pull/13314


pderop (approved):
LGTM

PR #2773, opened 2023-04-14
Wrong host port 80 when only X-Forwarded-Proto: https is present

### LINKED ISSUE #2771: /cloudfoundryapplication-Endpoints queried for port 80 instead of 443 after update from reactor-netty 1.0.28 to 1.0.30
Initially I created this issue [here ](https://github.com/spring-projects/spring-boot/issues/34851) as the behaviour originated after updating from spring-boot 2.7.9 to 2.7.10 (raising reactor netty from 1.0.28 to 1.0.30). After some further research I think I might pin it down to the related reactor-netty update. 

Here is the behaviour:

We're running spring-boot-starter with spring-cloud-gateway 3.1.16 on cloud foundry platform. When updating from spring-boot-starter 2.7.9 to 2.7.10 the /cloudfoundryapplication-Endpoints (which are accessed by cloudfoundry-Apps-Manager for example by using Google Chrome) stop working.This means that using actuator endpoints like configuring loggers on the fly or viewing build info via spring-boot-maven-plugin) no longer work with cloudfoundry-Apps-Manager.

This behaviour seems to result from a bad ssl-handshake. By observing the network protocol in Chrome Browser the following seems to happen in version 2.7.10:

1. OPTIONS-Request to https://[host]/cloudfoundryapplication (port 443)

2. GET-Request to https://[host]/cloudfoundryapplication happens (port 443). The response body contains json with urls of actuator-endpoints. However, these endpoints contain port 80, example:

`{"_links":{"caches":{"href":"https://[host]/cloudfoundryapplication:80/loggers","templated":false},`

3. On subsequent requests of cloudfoundry-apps-manager to https://[host]:80/cloudfoundryapplication/info (or /loggers etc) no https-connection can be established (connection refused)

However, manual requests to https://[host]:443/cloudfoundryapplication/info return status code 200 and the expected result indicating that the acutator-endpoints are still configured properly

Until version 2.7.9 the request in 2) returned endpoints with no port (indicating using default https-port 443).
`{"_links":{"caches":{"href":"https://[host]/cloudfoundryapplication/loggers","templated":false},`

Checking the logs of a sample request reveals that only x_forwarded_proto and x_forwarded_for headers are present:

`2023-04-12T19:04:57.323+02:00   [2023-04-12T17:04:57.272549114Z] "GET /cloudfoundryapplication HTTP/1.1" 200 x_forwarded_for:"10.236.177.224, 10.236.177.224" x_forwarded_proto:"https" `

I may add that setting
```
server.forward-headers-strategy: none or
server.forward-headers-strategy: framework
```
restores the correct behaviour while
`server.forward-headers-strategy: native`

does not. However, native is the default setting in cloud foundry according to [Spring Docs ](https://docs.spring.io/spring-boot/docs/2.7.10/reference/htmlsingle/#howto.webserver.use-behind-a-proxy-server) and I cannot judge all effects especially in a gateway with the main purpose of routing. 

This issue might be related due to recent changes in DefaultHttpForwardedHeaderHandler by @violetagg but further analysis is beyond my skills. Thanks for checking. 

## Expected Behavior
Usage of port 443 instead of port 80 when x_forwarded_proto indicates usage of https (as before). 

## Actual Behavior
Usage of port 80.

## Steps to Reproduce
see above and initial [Spring Boot Issue](https://github.com/spring-projects/spring-boot/issues/34851) with sample application attached as zip. 

## Your Environment
* Reactor version(s) used: 1.0.28 vs 1.0.30

pderop:
Motivation:
When resolving forwarded headers (X-Forwarded/Forwarded), the following scenario seems to happen in cloud foundry:

- you send an https request to a cloud foundry application, for example an actuator request: https://cloudfoundryapplication/actuator
- the front-end proxy forwards the request to a springboot application on non secure http 8080
- The request contains a `Host` header **without any port**, as well as a `X-Forwarded-Proto: https` header
- in this case, the `DefaultHttpForwardedHeaderHandler` class will wrongly resolve the host port as `80` instead of 443, because the request is received on a non-secure connection.
- And the json response then contains an unexpected 80 port appended in the actuator urls:

```
{"_links":{"self":{"href":"https://cloudfoundryapplication:80/actuator","templated":false},"health-path":{"href":"https://cloudfoundryapplication:80/actuator/health/{*path}","templated":true},"health": ...
````

Have also applied the similar patch for Forwarded headers (like `Forwarded: proto=https`).
Fixes #2771


pderop:
in previous commit, drastically simplified the logic of the patch, based on Violeta's suggestions given privately (thanks !)

pderop:

PR #2789, opened 2023-04-27
Default forwarded header handler should only using ConnectionInfo public API

pderop:
Motivations:

`reactor.netty.http.server.DefaultHttpForwardedHeaderHandler` should use only the `reactor.netty.http.server.ConnectionInfo` public API if we want the custom parsers to have the same information as us. Now, in #2761 PR , the fix done in the DefaultHttpForwardedHeaderHandler method is using some private constructors from the ConnectionInfo when parsing a X-Forwarded-Port header. This breaks the contract because people cannot then do the same using their own custom handlers.

Related to #2751

Modifications:

- Introduced a new withHostAddress(InetSocketAddress hostAddress, String hostName, int hostPort) method in ConnectionInfo, and the DefaultHttpForwardedHeaderHandler is now using it.
- DefaultHttpForwardedHeaderHandler is still using ConnectionInfo.getHostName() , which is not public but people can implement their own logic in order to obtain the host name.
- a new CustomXForwardedHeadersHandler test class with a package that is different than reactor.netty.http.server has been introduced  and is meant to check that people can implement the same logic for parsinx X-Forwarded headers. 
- All xForwarded* tests are now run with both DefaultHttpForwardedHeaderHandler as well as with new CustomXForwardedHeadersHandler.


reactor-netty-http/src/test/java/reactor/netty/http/server/ConnectionInfoTests.java
```diff
@@ -713,6 +758,13 @@ void forwardedForIpV6() {
 				});
 	}
 
+	private void testClientRequest(Consumer<HttpHeaders> clientRequestHeadersConsumer,
```
violetagg, line 761: This new method is introduced just to invoke an existing method with the proper forwarded handler. Why don't we invoke the existing method with the proper forwarded handler as it is done in some of the tests?

reactor-netty-http/src/test/java/reactor/netty/http/server/forwardheaderhandler/CustomXForwardedHeadersHandler.java
```diff
@@ -0,0 +1,99 @@
...
+import io.netty.handler.codec.http.HttpRequest;
+import reactor.netty.http.server.ConnectionInfo;
+import reactor.netty.transport.AddressUtils;
+
+/**
+ * Custom X-Forwarded-XX headers handler which implement the same logic of the Default handler.
+ * This class is meant to verify that people can implement at least the same logic provided
+ * by the DefaultHttpForwardedHeaderHandler class.
+ */
+public final class CustomXForwardedHeadersHandler {
```
violetagg, line 28: Can we reduce the visibility? If it is not needed we don't want to expose test classes.

violetagg (changes requested):

reactor-netty-http/src/test/java/reactor/netty/http/server/ConnectionInfoTests.java
pderop: This new method has been introduced in order to reduce the changes made for all other tests.

Example: let's take the xForwardedFor, line 241, which is invoking the new introduced method:
```java
	@ParameterizedTest(name = "{displayName}({arguments})")
	@ValueSource(booleans = {true, false})
	void xForwardedFor(boolean useCustomForwardedHandler) {
		testClientRequest(
				clientRequestHeaders -> clientRequestHeaders.add("X-Forwarded-For",
						"[1abc:2abc:3abc::5ABC:6abc]:8080, 192.168.0.1"),
				serverRequest -> {
					Assertions.assertThat(serverRequest.remoteAddress().getHostString()).isEqualTo("1abc:2abc:3abc:0:0:0:5abc:6abc");
					Assertions.assertThat(serverRequest.remoteAddress().getPort()).isEqualTo(8080);
				},
				useCustomForwardedHandler);
	}
```
without the new method introduced line 761, the above method would be:
```
	@ParameterizedTest(name = "{displayName}({arguments})")
	@ValueSource(booleans = {true, false})
	void xForwardedFor(boolean useCustomForwardedHandler) {
		testClientRequest(
				clientRequestHeaders -> clientRequestHeaders.add("X-Forwarded-For",
						"[1abc:2abc:3abc::5ABC:6abc]:8080, 192.168.0.1"),
				serverRequest -> {
					Assertions.assertThat(serverRequest.remoteAddress().getHostString()).isEqualTo("1abc:2abc:3abc:0:0:0:5abc:6abc");
					Assertions.assertThat(serverRequest.remoteAddress().getPort()).isEqualTo(8080);
				},
				getForwardedHandler(useCustomForwardedHandler),
				Function.identity(),
				Function.identity(),
				false);
	}
```

So, with the new introduced method, the modifications done in all other xForwarded* are shorted.
But Ok,  I will revert and remove the new introduced method.

violetagg: ok then I'll leave you to decide

reactor-netty-http/src/test/java/reactor/netty/http/server/forwardheaderhandler/CustomXForwardedHeadersHandler.java
pderop: The package for the new CustomXForwardedHeadersHandler (`reactor.netty.http.server.forwardheaderhandler`) is different than the one from `DefaultHttpForwardedHeaderHandler` and `ConnectionsInfoTest`.
I did that on purpose in order to make sure the custom handler is only using public ConnectionInfo API. 
So, if we reduce the visibility of CustomXForwardedHeadersHandler, then the ConnectionsInfoTest won't be able to use it ?  

Is it ok to keep the existing public visibility ? Or else, we can use some reflections tricks from ConnectionInfoTests when getting access to this class ?

violetagg: Ok let's keep this visibility, but add notice that this is not for general purpose and can be changed at any time. Is this ok?

pderop: ok, I have updated the javadoc from CustomXForwardedHeadersHandler in order to indicate that the class is not a public API. see https://github.com/reactor/reactor-netty/pull/2789/commits/f8d975666cb2e7ddeb7b0fd747571e7455d4fa2c, can you recheck ?

reactor-netty-http/src/test/java/reactor/netty/http/server/ConnectionInfoTests.java
pderop: I would prefer to keep the existing method, in order to reduce the code diff.

pderop:
@violetagg ,

can you retake a look ?

So, have added public visibility to ConnectionInfo getHostName/getHostPort/getDefaultHostPort: it will simplify life of people if they want to implement their own forwarded handler logic.

pderop:
@violetagg , thanks for the review !

PR #2794, opened 2023-05-04
Adapt Http2SettingsSpecTests to netty PR #13369

pderop:
We need to adapt the Http2SettingsSpecTests, because since Netty [PR#13369](https://github.com/netty/netty/pull/13369),
the messages of the IllegalArgumentExceptions thrown when invalid Http2Settings are used have changed:

The following reactor-netty5 tests in the Http2SettingsSpecTests test are not passing anymore:

```
headerTableSizeBadValues
Expecting message to be:
  "Setting HEADER_TABLE_SIZE is invalid: -1"
but was:
  "Setting HEADER_TABLE_SIZE is invalid: -1, expected [0, 4294967295]"


maxFrameSizeBadValues
Expecting message to be:
  "Setting MAX_FRAME_SIZE is invalid: -1"
but was:
  "Setting MAX_FRAME_SIZE is invalid: -1, expected [16384, 16777215]"

maxHeaderListSizeBadValues
Expecting message to be:
  "Setting MAX_HEADER_LIST_SIZE is invalid: -1"
but was:
  "Setting MAX_HEADER_LIST_SIZE is invalid: -1, expected [0, 4294967295]"

initialWindowSizeBadValues
Expecting message to be:
  "Setting INITIAL_WINDOW_SIZE is invalid: -1"
but was:
  "Setting INITIAL_WINDOW_SIZE is invalid: -1, expected [0, 2147483647]"

maxConcurrentStreamsBadValues
Expecting message to be:
  "Setting MAX_CONCURRENT_STREAMS is invalid: -1"
but was:
  "Setting MAX_CONCURRENT_STREAMS is invalid: -1, expected [0, 4294967295]"
```

### DIFF
--- a/.github/workflows/check_netty_snapshots.yml
+++ b/.github/workflows/check_netty_snapshots.yml
@@ -29,4 +29,4 @@ jobs:
           distribution: 'temurin'
           java-version: 17
       - name: Build with Gradle
-        run: ./gradlew clean check --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.90.Final-SNAPSHOT'
+        run: ./gradlew clean check --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.91.Final-SNAPSHOT'
--- a/.github/workflows/check_netty_snapshots.yml
+++ b/.github/workflows/check_netty_snapshots.yml
@@ -29,4 +29,4 @@ jobs:
           distribution: 'temurin'
           java-version: 17
       - name: Build with Gradle
-        run: ./gradlew clean check --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.91.Final-SNAPSHOT'
+        run: ./gradlew clean check --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.92.Final-SNAPSHOT'
--- a/.github/workflows/check_netty_snapshots.yml
+++ b/.github/workflows/check_netty_snapshots.yml
@@ -29,4 +29,4 @@ jobs:
           distribution: 'temurin'
           java-version: 17
       - name: Build with Gradle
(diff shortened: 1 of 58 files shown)


violetagg:
@pderop Can you fix this for 1.0.x and 1.1.x as we have there the same issue
CI fails: https://github.com/reactor/reactor-netty/actions/runs/4872834651

pderop:
If I'm correct, the Netty PR #13369 is only part of 4.1.93.Final-SNAPSHOT, so shouldn't we wait for the 4.1.93.Final to be released before applying this patch into our reactor-netty 1.0.x and 1.1.x ? 

(the 1.0.x and 1.1.x are currently depending on 4.1.92.Final, which do not have the PR#13369)

pderop:
maybe what could be done in 1.0.x/1.1.x is to manage to expect **both** old and new error messages, so the tests would then pass fine with both 4.1.92.Final as well as with 4.1.93.Final-SNAPSHOT ? And when the 4.1.93.Final will be released, then we'll update the test with only the correct error messages ?
 WDYT ?

violetagg:
> If I'm correct, the Netty PR #13369 is only part of 4.1.93.Final-SNAPSHOT, so shouldn't we wait for the 4.1.93.Final to be released before applying this patch into our reactor-netty 1.0.x and 1.1.x ?
> 
> (the 1.0.x and 1.1.x are currently depending on 4.1.92.Final, which do not have the PR#13369)

yeah :(

either we use something like this `withMessageContaining`

```
	@Test
	void headerTableSizeBadValues() {
		assertThatExceptionOfType(IllegalArgumentException.class)
				.isThrownBy(() -> builder.headerTableSize(-1))
				.withMessageContaining("Setting HEADER_TABLE_SIZE is invalid: -1");
	}
```

or we need to check regularly that the CI fails only with this issue and nothing else, otherwise we might miss some change that might break Reactor Netty integration.

Wdyt?

pderop:
all right, I will do something like `withMessageContaining("Setting HEADER_TABLE_SIZE is invalid: -1")`, good suggestion, thanks.

pderop:
ok, I'm closing this current PR and will recreate a new one on top of 1.0.x branch ...

pderop:
Superseded by #2795

PR #2802, opened 2023-05-12
Ensure if a custom resolveCache is provided that cache TTL's are ignored

samueldlightfoot:
Currently if a custom resolve cache is provided in NameResolverProvider, calling DnsNameResolverBuilder::build will fail as both cache ttls (which have non-null defaults) and the resolveCache are populated. It is currently impossible to set cache ttls to null values in the NameResolverProvider

TTL fields:

- cacheMinTimeToLive
- cacheMaxTimeToLive
- cacheNegativeTimeToLive

See https://github.com/netty/netty/blob/4.1/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java#L494

Feel free to append any changes as necessary for the purpose of speed.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/NameResolverProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/NameResolverProvider.java
@@ -192,7 +192,9 @@ public final class NameResolverProvider {
 		NameResolverSpec resolvedAddressTypes(ResolvedAddressTypes resolvedAddressTypes);
 
 		/**
-		 * Sets the resolve cache to use for DNS resolution
+		 * Sets the resolve cache to use for DNS resolution. {@link NameResolverSpec#cacheMinTimeToLive},
+		 * {@link NameResolverSpec#cacheMaxTimeToLive} and {@link NameResolverSpec#cacheNegativeTimeToLive} will be
+		 * ignored if this is set.
 		 *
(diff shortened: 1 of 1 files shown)


violetagg:
@samueldlightfoot Why we need this change on the level of Reactor Netty?

Netty already takes care for this. Please review this:
- https://github.com/netty/netty/blob/af132ffc483f68d06eed746797fff2af652bcd35/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java#L440
- https://github.com/netty/netty/blob/af132ffc483f68d06eed746797fff2af652bcd35/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java#L444
- https://github.com/netty/netty/blob/af132ffc483f68d06eed746797fff2af652bcd35/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java#L452
- https://github.com/netty/netty/blob/af132ffc483f68d06eed746797fff2af652bcd35/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java#L501-L504

samueldlightfoot:
Thanks for looking into this.

I've raised this issue with Netty folks: https://github.com/netty/netty/issues/13384

PR #2820, opened 2023-06-02
Ensure any custom `CL`/`TE` header is removed when `304` response code

### LINKED ISSUE #2818: There are some problems with the response to the client request for 304
After we upgraded the version of reactor-netty-http from 1.0.26 to 1.0.27, we found that some clients had the 'unexpected end of stream' issue for 304 response requests. When we verified, we found that the lower version of postman would have the problem that the request connection was not disconnected.

## Expected Behavior
The request responds normally to 304

## Actual Behavior
some clients had the 'unexpected end of stream' issue.
With version 8.0.7 of Postman will have a 304 already responding, but the connection is not disconnected

## Steps to Reproduce

```java
    public Mono<ConfigsResponseBody> pollConfigsRelease(String instanceKey) {
        return Mono.<ConfigsResponseBody>create(request -> {
            // ConcurrentMap<String, MonoSink<ConfigsResponseBody>> requestPool
            requestPool.put(instanceKey, request);
            executorService.execute(this::handlerFullConfig);

        })
            .timeout(Duration.ofSeconds(5L),
                Mono.defer(() -> Mono.error(new ResponseStatusException(HttpStatus.NOT_MODIFIED))))
            .doFinally(signalType -> {
                requestPool.remove(instanceKey);
                log.info("request from {} is return, signalType is {}", instanceKey, signalType);
            });
    }
```

## Possible Solution
I found a patch that could cause this issue：
https://github.com/reactor/reactor-netty/commit/fcda69af6a85265aee457614ea01b74faea581d1?diff=unified
## Your Environment

* Reactor version(s) used:reactor-netty-http 1.0.27+
* JVM version (`java -version`): 8
* OS and version (eg. `uname -a`): Ubuntu4.15.0-91-generic x86_64 x86_64 x86_64 GNU/Linux  or window10

violetagg:
Fixes #2818

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/HttpOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/HttpOperations.java
@@ -118,19 +118,7 @@ public abstract class HttpOperations<INBOUND extends NettyInbound, OUTBOUND exte
 			return new PostHeadersNettyOutbound(((Mono<ByteBuf>) source)
 					.flatMap(b -> {
 						if (markSentHeaderAndBody(b)) {
-							HttpMessage msg;
-							if (HttpUtil.getContentLength(outboundHttpMessage(), -1) == 0 ||
-									isContentAlwaysEmpty()) {
-								if (log.isDebugEnabled()) {
-									log.debug(format(channel(), "Dropped HTTP content, since response has " +
-											"1. [Content-Length: 0] or 2. there must be no content: {}"), b);
-								}
-								b.release();
-								msg = newFullBodyMessage(Unpooled.EMPTY_BUFFER);
-							}
-							else {
-								msg = newFullBodyMessage(b);
-							}
+							HttpMessage msg = prepareHttpMessage(b);
 
 							try {
 								afterMarkSentHeaders();
@@ -161,18 +149,7 @@ public abstract class HttpOperations<INBOUND extends NettyInbound, OUTBOUND exte
 		ByteBuf b = (ByteBuf) message;
 		return new PostHeadersNettyOutbound(FutureMono.deferFuture(() -> {
 			if (markSentHeaderAndBody(b)) {
-				HttpMessage msg;
-				if (HttpUtil.getContentLength(outboundHttpMessage(), -1) == 0) {
-					if (log.isDebugEnabled()) {
-						log.debug(format(channel(), "Dropped HTTP content, " +
-								"since response has Content-Length: 0 {}"), b);
-					}
-					b.release();
-					msg = newFullBodyMessage(Unpooled.EMPTY_BUFFER);
-				}
-				else {
-					msg = newFullBodyMessage(b);
-				}
+				HttpMessage msg = prepareHttpMessage(b);
 
 				try {
 					afterMarkSentHeaders();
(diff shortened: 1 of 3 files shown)


violetagg:
The failed test on Windows OS is not related

pderop:
It seems that there might be some remaining use cases where a custom content length could still remain if a 304 response is returned

I am thinking about these two following use cases:

1) use case 1

For example, let's first add this test in the `dispostableServer` from `HttpResponseStatusCodesHandlingTests.noContentStatusCodes`test:

```
.get("/304-7", (req, res) -> res.status(HttpResponseStatus.NOT_MODIFIED)
    .header(HttpHeaderNames.CONTENT_LENGTH, "999")
    .send())
```

then add a `checkResponse` method call in order to invoke the above "/304-7" test:

```
		checkResponse("/304-7", client);
```

so, the /304-7 test will fail because it adds a custom invalid "Content-Length: 999" header (I know it's a corner case here, because no payload is sent while an invalid content-length is provided , but this CL is not removed, because I think `send()` does not invoke `HttpServerOperations.isContentAlwaysEmpty()`; so only the TE header will be removed by  `HttpServerOperations.newFullBodyMessage()`, but the CL header will remain. am I making sense here ?

2) use case 2:

another use case, let's use sendObject, and let's add a new /304-8 test in the `dispostableServer` from `HttpResponseStatusCodesHandlingTests.noContentStatusCodes`test:

```
.get("/304-8", (req, res) -> res.status(HttpResponseStatus.NOT_MODIFIED)
    .header(HttpHeaderNames.CONTENT_LENGTH, "6")
    .sendObject(Unpooled.wrappedBuffer("/304-8".getBytes(Charset.defaultCharset()))))
    
...
checkResponse("/304-8", client);
```

So,  here, the test fails because [HttpOperations.sendObject](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/main/java/reactor/netty/http/HttpOperations.java#L165) does not invoke `isContentAlwaysEmpty()`, while [HttpOperations.send(Publisher<? extends ByteBuf> source)](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/main/java/reactor/netty/http/HttpOperations.java#L123) does invoke isContentAlwaysEmpty().

based on the two above use cases, it sounds like two modifications could be done in the patch:

first, the  patch done in `HttpServerOperations.isContentAlwaysEmpty()` could maybe be moved to the [HttpServerOpeartions.newFullBodyMessage(ByteBuf body)](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java#L222) method ? something like:

```
	@Override
	protected HttpMessage newFullBodyMessage(ByteBuf body) {
		HttpResponse res =
				new DefaultFullHttpResponse(version(), status(), body);

		if (HttpResponseStatus.NOT_MODIFIED.code() == status().code()) {
			responseHeaders.remove(HttpHeaderNames.TRANSFER_ENCODING)
					.remove(HttpHeaderNames.CONTENT_LENGTH);
		}
                ...
```

and finally, shouldn't the [HttpOperations.sendObject](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/main/java/reactor/netty/http/HttpOperations.java#L165) be modified in order to invoke `isContentAlwaysEmpty()`, like it is done in [HttpOperations.send(Publisher<? extends ByteBuf> source)](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/main/java/reactor/netty/http/HttpOperations.java#L123) ? 
something like:

```
	public NettyOutbound sendObject(Object message) {
		if (!channel().isActive()) {
			ReactorNetty.safeRelease(message);
			return then(Mono.error(AbortedException.beforeSend()));
		}
		if (!(message instanceof ByteBuf)) {
			return super.sendObject(message);
		}
		ByteBuf b = (ByteBuf) message;
		return new PostHeadersNettyOutbound(FutureMono.deferFuture(() -> {
			if (markSentHeaderAndBody(b)) {
				HttpMessage msg;
				if (HttpUtil.getContentLength(outboundHttpMessage(), -1) == 0 ||
						isContentAlwaysEmpty()) {
					if (log.isDebugEnabled()) {
						log.debug(format(channel(), "Dropped HTTP content, " +
								"since response has Content-Length: 0 {}"), b);
					}
					b.release();
					msg = newFullBodyMessage(Unpooled.EMPTY_BUFFER);
				}
				else {
					msg = newFullBodyMessage(b);
				}

				try {
					afterMarkSentHeaders();
				}
				catch (RuntimeException e) {
					b.release();
					throw e;
				}

				return channel().writeAndFlush(msg);
			}
			return channel().writeAndFlush(b);
		}), this, b);
	}
```

doing so seems to address the two use case.
what do you think ?

violetagg:
@pderop I tend to agree for use case 2, can you explain more use case 1?

pderop:
The use case #1 is indeed about an **application error** that is adding a custom unexpected content-length header.

violetagg:
@pderop I address the use case 2, PTAL

violetagg:
The failed tests on Windows OS are not related

pderop (approved):
looking good ! thanks.

violetagg:
@pderop Thanks

PR #2840, opened 2023-06-21
Ensure handshake timeout is applied in case of SNI

violetagg:


reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -2209,6 +2211,51 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
...
+		createClient(disposableServer::address)
+		        .secure(spec -> spec.sslContext(clientSslContextBuilder)
+		                            .serverNames(new SNIHostName("test.com")))
+		        .get()
+		        .uri("/")
+		        .responseContent()
+		        .aggregate()
+		        .as(StepVerifier::create)
+		        .expectError()
+		        .verify(Duration.ofSeconds(5));
```
pderop, line 2254: what do you think about increasing the verify duration to 30 seconds, in order to avoid test failures if the CI is very slow ?

pderop (approved):
looking good.

just one minor thing: maybe it would be worth increasing the test verify duration, just in case the CI is very slow).

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
violetagg: but the timeout is 1ms?

pderop: ohhh yes ! sorry, then ignore my comment. 
Already approved this PR.

violetagg:
@pderop Thanks

PR #2842, opened 2023-06-23
Enhance LoopResources API in order to disable colocation

### LINKED ISSUE #2781: Sharing default `LoopResources` without colocation and without creating additional event loop groups
## Motivation

Out of the discussion at https://github.com/r2dbc/r2dbc-pool/issues/190#issuecomment-1520270200 and https://github.com/mariadb-corporation/mariadb-connector-r2dbc/issues/65, we found that colocation for long-lived connection can negatively impact performance because new long-lived database connections are created on the same event loop.

While it is now possible to create `LoopResources` without colocation, it would be good to be able to reuse the underlying `EventLoopGroup`s of the default instance without colocation to avoid creation of additional threads. Right now, in an arrangement with Reactor Netty and two R2DBC drivers (Postgres and MariaDB), an application would have:

- The default `LoopResources` instance used by default
- A customized `LoopResources` instance for MariaDB
- A customized `LoopResources` instance for Postgres

These would effectively create three `EventLoopGroup`s with three times the number of Threads imposing additional resource usage.

## Desired solution

Controlling colocation on the `runOn(…)` level would be neat. `LoopResources` bear quite some complexity regarding event loop group selection in addition to native transport selection, so having a simple way to reuse default event loop groups without colocation would be great.

## Considered alternatives

As outlined before, `TcpClient.runOn(…)` would work for disabling colocation at the cost of additional event loops. 

A system property switch is not a good alternative because other components, that benefit from colocation would suffer from disabling colocation.

Another alternative would be a wrapper in each R2DBC driver that wraps the default `EventLoopGroup` and unwraps it to obtain the non-colocated instance at the cost of complexity in each project.

pderop:
As discussed in #2781 GH issue, this PR proposes an enhancement in the `LoopResources` API in order to get an existing LoopResource, and mutate it to disable colocation.
(this enhancement is done on top of the main branch, so for 1.1.x reactor-netty versions).

So, to disable colocation on an existing LoopResources, using this PR, you can now call the new `LoopResources.disableColocation()`method that will return a LoopResources that will reuse the EventLoopGroup of the original LoopResources, but without colocation.

Here is an example of an R2DBC customizer which allows to reuse the default TcpResources without coloc:

```
@Component
public class R2DBCOptionsCustomizer implements ConnectionFactoryOptionsBuilderCustomizer {

    @Override
    public void customize(ConnectionFactoryOptions.Builder builder) {
        LoopResources loopResources = TcpResources.get().disableColocation();
        builder.option(MariadbConnectionFactoryProvider.LOOP_RESOURCES, loopResources);
    }
}
```

This could also be done in R2DBC driver implementations, for example, in the [MariaDBConnectionConfiguration](https://github.com/mariadb-corporation/mariadb-connector-r2dbc/blob/master/src/main/java/org/mariadb/r2dbc/MariadbConnectionConfiguration.java#L142), it could be possible to change:

```
    this.loopResources = loopResources != null ? loopResources : TcpResources.get();
```

by:

```
    this.loopResources = loopResources != null ? loopResources : TcpResources.get().disableColocation();
```

PS: this allows to overcome a performance problem described in 2781 issue, however, it might be useful to also consider the reactor-pool warmup optimization proposed in https://github.com/reactor/reactor-pool/pull/171, which might resolve the performance problem without having to do anything.

Fixes #2781


pderop:
@violetagg , please do not review for the moment, let's wait to see what Mark thinks about ?

reactor-netty-core/src/main/java/reactor/netty/resources/LoopResources.java
```diff
@@ -175,6 +175,18 @@ static LoopResources create(String prefix, int selectCount, int workerCount, boo
...
 
+	/**
+	 * Return a {@link LoopResources} that will share the same {@link EventLoopGroup} from this {@link LoopResources}, but with
+	 * colocation support disabled (if it was enabled).<p>
+	 * Disposing the returned {@link LoopResources} will also dispose the {@link EventLoopGroup} from the origin {@link LoopResources}.
+	 *
+	 * @return a LoopResources that will share the same {@link EventLoopGroup} from this {@link LoopResources}
+	 * @since 1.1.9
+	 */
+	default LoopResources disableColocation() {
```
mp911de, line 186: Does it make sense to have `disableColocation` on `TcpResources` as well?

pderop: The LoopResources.disableColocation is generic and can be applied to any default LoopResources, including the TcpResources (which implements the LoopResources interface), or the default HttpResources, or even the default UdpResources.

For example, in Mariadb driver, if users don't have provided any custom LoopResources, you can then configure the driver with a mutated version of the default TcpResources with colocation disabled like this in [MariadbConnectionConfiguration](https://github.com/mariadb-corporation/mariadb-connector-r2dbc/blob/master/src/main/java/org/mariadb/r2dbc/MariadbConnectionConfiguration.java#L142):

```
    this.loopResources = loopResources != null ? loopResources : TcpResources.get().disableColocation();

```

The TcpResources inherits from the LoopResources.
it's important to note that you need to use the mutated LoopResources that is **returned** by the disableColocation() method (which does not alter the original LoopResources that remains unchanged and  stay colocated mode, so other components which are happy with this default mode won't be impacted, like the HttpClient).

If you want to share the default HttpResources with the R2DBC driver, you could also do it (in this case R2DBC will reuse the same event loop group from Http applications:

```
    this.loopResources = loopResources != null ? loopResources : HttpResources.get().disableColocation();

```

all this to say that the `LoopResources.disableColocation()` is generic and applies to either one of the existing default loop resources (`HttpResources.get()`, `TcpResources.get()`, or `UdpResources.get())`, or to even any custom colocated LoopResources created using one of the `LoopResources.create(...)` methods.

pderop:
@violetagg , can you take a look please ?

reactor-netty-core/src/main/java/reactor/netty/resources/LoopResources.java
violetagg, line 186: @pderop This interface `LoopResources` can have different implementations. The proposed change `LoopResources.disableColocation()` is specific to the implementation in Reactor Netty only (`DefaultLoopResources`, `TcpResources` etc). In other implementations (for example https://github.com/turms-im/turms/blob/b2a817058b45b5e8bbb8c0a6e5808e92bfbbb884/turms-server-common/src/main/java/im/turms/server/common/access/common/LoopResourcesFactory.java#L30) this may not be the case and colocation might not exist as a configuration at all.
I suggest that we provide a change that can fit also with other implementations if the change is on the interface level.

violetagg (changes requested):

reactor-netty-core/src/main/java/reactor/netty/resources/LoopResources.java
pderop: What do you think if I  do not do the API change at the interface (instance) level, and instead of that, since we already have some `create(...)` methods which are using the ReactorNetty specific implementations (DefaultLoopResources, TcpResources, ...), then just add a new static create method like this in the LoopResources:

```
	/**
	 * Return a {@link LoopResources} that will share the same {@link EventLoopGroup} from an existing parent {@link LoopResources}.
	 * Disposing the returned {@link LoopResources} will also dispose the {@link EventLoopGroup} from the parent as well.
	 *
	 * @param parent the parent {@link LoopResources} whose underlying {@link EventLoopGroup} will be reused by the returned {@link LoopResources}
	 * @param colocate <code>false</code> will disable colocation from the parent {@link EventLoopGroup} if it was enabled in the parent.
	 *                 <code>true</code> means colocation will be enabled if it was disabled in the parent. The parent is not affected whether <code>true</code>
	 *                 or <code>false</code> is used.
	 * @return a LoopResources that will share the same {@link EventLoopGroup} from the specified parent, with or without colocation mode
	 * @since 1.1.9
	 */
	static LoopResources create(LoopResources parent, boolean colocate) {
		return new DefaultLoopResources(parent, colocate);
	}

```

This new static method is consistent with the other `create` methods that are depending on our ReactorNetty specfic implementations, and also, now we have a new
`DefaultLoopResources(LoopResources parent, boolean colocate)` constructor that is clearer than having a single constructor parameter `DefaultLoopResources(LoopResources parent)` that was implicitly disabling colocation.

would this be OK, and am I missing your point ?

violetagg: If we have in the future a request for changing other configurations how the API will look like?

pderop: You can ignore my last comment, indeed doing so will lead to a kind of bloated API, because in the future, if we have more new re-configurable configs, then we would end up
with many create method signatures like this, which would ruin the original cohesive feel of the API:

```
create(parent, boolean /* colocate */)
create(parent, boolean /* colocate */, boolean /* enable/disable new configuration1 */)
create(parent, boolean /* colocate */, boolean /* enable/disable new configuration1 */, boolean /* enable/disable new configuration2 */)
etc ...
```

in this case, a Builder API would be more appropriate.

But, I got back to your initial suggestion from https://github.com/reactor/reactor-netty/pull/2842#discussion_r1241802262:

> I suggest that we provide a change that can fit also with other implementations if the change is on the interface level.

and in the last commit, I revisited things, and it might be more aligned with what you have in mind, can you recheck ?

- first, I have reverted all changes from DefaultLoopResources, because in fact, it is not necessary to add complexity to this class.
- the LoopResources.disableColocation method is now using a helper class which is independant of our current Reactor-Netty LoopResources implementations (like DefaultLoopResources, TcpResources, ...)
- it should fit to other implementations like the one from `turms`; 
for example, if the turms `LoopResourcesFactory` has support for colocation and implements a `onClient(boolean useNative)` method that returns an EventLoopGroup decorated using `LoopResources.colocate(EventLoopGroup group)`, then the disableColocation() method shoud fit ?

let me know ?

pderop: ok, so I'll continue to think about this.

pderop:
I am closing this PR (see last comments from #2181 issue.)

Issue #2825, opened 2023-06-08
Misleading reactor.netty.http.client.PrematureCloseException: Connection prematurely closed BEFORE response


himanshunp:
When we use spring boot's web client with netty connection provider or http client and call http POST with very large JSON payload, it throws error reactor.netty.http.client.PrematureCloseException: Connection prematurely closed BEFORE response.
But actually this is misleading instead of propagating real http 400 BAD_REQUEST. I went into rabbit hole of this guide: https://projectreactor.io/docs/netty/release/reference/index.html#faq.connection-closed and still could not figure out until I kept increasing load on the REST api.

I was able to reproduce this locally running spring boot app without any load balancer or proxy in between.

## Expected Behavior

client should propagate actual http 400 BAD REQUEST error instead of masking with PrematureCloseException.
## Actual Behavior

throws error reactor.netty.http.client.PrematureCloseException: Connection prematurely closed BEFORE response
## Steps to Reproduce

Run the unit test:
https://github.com/himanshunp/spring-web-client/blob/master/src/test/java/com/parmar/himanshu/spring/webclient/ControllerTest.java#L37
If you look at the three parameters, the last one tests with very large json payload and reproduces above error. There should be a better way to capture http 400 and response body.

## Possible Solution

I was able to track down this by using X-Request-ID header. The documentation should add this as a debugging step.
## Your Environment

* Reactor version(s) used: 3.5.6
* Other relevant libraries versions (eg. `netty`, ...): `reactor-netty-core`: 1.1.7
* JVM version (`java -version`): 20
* OS and version (eg. `uname -a`): macOS Monterey 12.6.6 Darwin himanshus-mbp.lan 21.6.0 Darwin Kernel Version 21.6.0: Mon Apr 24 21:10:53 PDT 2023; root:xnu-8020.240.18.701.5~1/RELEASE_X86_64 x86_64

pderop:
Hi @himanshunp ,

thank you for reporting and for the nice reproducer project.

The PR #2864 may alleviate the issue. I have targeted it to 1.0.x branch.
please, expect some times before it can be validated and merged into the main 1.1.x branch.

Now, as a work around, in your reproducer, I think you can use Flux instead of Mono when sending the body, that might avoid the issue.
For example, in SvnClient.callWithWebclientWithConnectionPool method, you could do this change:

```
  public Integer callWithWebclientWithConnectionPool(String payload) {
    log.info("calling webclient with connection pool");
[...]

PR #2903, opened 2023-09-19
Fix HttpClientWithTomcatTest flaky test.

pderop:
This is a new attempt to fix the flaky test from HttpClientWithTomcatTest.testIssue2825, which was still sometimes getting some Premature Close Exception.

To summarize:

- The fix from #2864 significantly reduces the probability to get premature close exceptions when Tomcat sends an early final response 400 bad request while the client is still writing a large request body. Did a benchmark using a real network, it works well; no premature close exceptions.
- However, on localhost, and especially on the Github CI, the fix only alleviate the problem but we can't avoid it all the time:   sometimes, once the server has sent the 400 bad request, and once it has read up to 2 MB of extra request body, it may then close the connection using a TCP/RST, so in this case, if the client has not yet consumed the 400 bad request, it may miss it (when receiving the TCP/RST, the client OS may drop any unread data).
- The test is now doing his best in order to let the client have time to consume the 400 bad request, and if you disable the patch, the test will reproduce the issue described in #2825.

Did the following fixes:

- Fixed  the`PayloadSizeServlet` which was not counting correctly the number of bytes received.
- Fixed test http request PAYLOAD size: when the max payload size is exceeded, the test should at least write some extra bytes >= 2 MB, which is the value for maxSwallowSize on Tomcat. Else, the test would pass even without applying the fix from #2864 
- Reduce the SO_RCVBUF on the Tomcat server, to reduce the possibility that the client is writing while the server is closing the connection (with a small Tomcat SO_RCVBUF, the client will likely be blocked on socket writing while the server is closing the connection.
-  To avoid racy situation when the client can't read the response before the socket is closed with a TCP/RST or when the client is getting a broken pipe while writing, the testIssue2825 is now using a RetryWhen strategy in order to retry the request a couple of times before failing. This is needed, else the test may remain flaky when running on the CI.

### DIFF
--- a/reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientWithTomcatTest.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientWithTomcatTest.java
@@ -43,6 +43,7 @@ import reactor.netty.resources.ConnectionProvider;
 import reactor.test.StepVerifier;
 import reactor.util.function.Tuple2;
 import reactor.util.function.Tuples;
+import reactor.util.retry.Retry;
 
 import java.io.InputStream;
 import java.lang.reflect.Field;
@@ -71,7 +72,8 @@ import static reactor.netty.http.client.HttpClientOperations.SendForm.DEFAULT_FA
  */
 class HttpClientWithTomcatTest {
 	private static TomcatServer tomcat;
-	private static final byte[] PAYLOAD = String.join("", Collections.nCopies(TomcatServer.PAYLOAD_MAX + (1024 * 1024), "X"))
+	private static final String TOMCAT_RCVBUF_SIZE = String.valueOf(32 * 1024);
+	private static final byte[] PAYLOAD = String.join("", Collections.nCopies(TomcatServer.PAYLOAD_MAX + (10 * 1024 * 1024), "X"))
 			.getBytes(Charset.defaultCharset());
 
 	@BeforeAll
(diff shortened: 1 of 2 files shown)


reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientWithTomcatTest.java
```diff
@@ -333,31 +327,25 @@ void contentHeader() {
...
-	@MethodSource("testIssue2825Args")
-	void testIssue2825(Supplier<Publisher<ByteBuf>> payload, long bytesToSend) {
 		AtomicReference<SocketAddress> serverAddress = new AtomicReference<>();
 		HttpClient client = HttpClient.create()
 				.port(getPort())
 				.wiretap(false)
 				.metrics(true, ClientMetricsRecorder::reset)
-				.doOnConnected(conn -> serverAddress.set(conn.address()));
+				.doOnConnected(conn -> {
+					conn.channel().config().setOption(ChannelOption.SO_SNDBUF, 4 * 1024);
```
violetagg, line 340: This can be done as a direct configuration on the level of `HttpClient` -> `HttpClient.option(...)`

violetagg (changes requested):

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientWithTomcatTest.java
```diff
@@ -333,31 +327,25 @@ void contentHeader() {
...
-		Supplier<Publisher<ByteBuf>> postFlux = () -> Flux.just(Unpooled.wrappedBuffer(PAYLOAD));
-
-		return Stream.of(
-				Arguments.of(Named.of("postMono", postMono), Named.of("bytes", PAYLOAD.length)),
-				Arguments.of(Named.of("postFlux", postFlux), Named.of("bytes", PAYLOAD.length))
-		);
-	}
+	@Test
+	void testIssue2825() {
+		tomcat.setMaxSwallowSize(MAX_SWALLOW_SIZE);
```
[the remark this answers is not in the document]
pderop: good catch, thanks.
applied suggestion in https://github.com/reactor/reactor-netty/pull/2903/commits/80ec3af5854741b8dacbdacf6144ddd607a58da4

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientWithTomcatTest.java
pderop: applied suggestion in https://github.com/reactor/reactor-netty/pull/2903/commits/27457f55c1b01baeccb52fdcdfff6268020570d8

pderop:
just applied the suggested changes.

violetagg (approved):

pderop:
@violetagg , thanks for your review.

PR #2906, opened 2023-09-21
Depend on `Netty` `v4.1.98.Final`

violetagg:


violetagg:
With Netty 4.1.99.Final the build is green again

PR #2919, opened 2023-10-03
Ensure connection closed when non `200` status for request with `Expect` header and incoming data not completed

### LINKED ISSUE #2910: Webflux RestController unexpected 400 response with reused connection and incomplete expect-100 activity
We discovered a difference in behavior while using reactor-netty as a server compared to other servers such as tomcat or go http in regards to a specific scenario. The scenario is:
 - client makes POST to server 
 - The request contains an Expect: 100-continue header and a content-length > 0
 - The server responds with a status code before reading the body or closing the connection
 - The next request on the same connection is immediately seen as malformed and response code 400 is returned to client.

The expect-100 header in an http request allows the server to validate headers before asking the client to proceed sending the request body. Once the server validates the headers it gives a 100 continue response back to client and the client continues sending the body. 

What we observe is a client is reusing a keep-alive connection to the server. First request is a POST request and the response is unauthorized and given a 401 response. Since the 401 was returned to client and not a 100 continue then, the POST body is never sent from client to server. Due to the content-length specified in the POST request, the next bytes on the wire the server is looking for is the POST body - however its not the POST body but a completely new request and results in an immediate 400.

To summarize it seems this issue is very specific to connection reuse and the expect-100 header in a request that contains payload and the server returns a non 100/200 status code prior to reading the rest of the body or closing the connection. 

Per rfc https://www.rfc-editor.org/rfc/rfc9110#name-expect
"A server that responds with a final status code before reading the
      entire message body SHOULD indicate in that response whether it
      intends to close the connection or continue reading and discarding
      the request message (see Section 6.6 of [RFC7230])."

## Expected Behavior
The server should close the connection on a request that results in a non 200 status code that had expectations (such as expect 100)

## Actual Behavior
The server returns a status code prior to sending a 100 response to the client and due to the content-length - the next bytes the server wants to read from the wire is the POST body payload but its a new request and results in immediate 400.

## Steps to Reproduce

Repro steps are as follows:

1 - The code snippits/data relevant to app
2 - The commands to use over netcat while running the app on localhost:8080

In the following Controller, if you uncomment either of those lines - the app works as expected and no 400s will be observed.

#Controller 
```java
public final class Controller {
    @RequestMapping("/**")
    public Mono<Void> service( ServerHttpRequest incoming, ServerHttpResponse response) {
        System.out.println("expect100test Received request");

        Mono<Void> x= Mono.empty();
        if(incoming.getHeaders().get(HttpHeaders.AUTHORIZATION) == null) {
            //  response.getHeaders().add(HttpHeaders.CONNECTION, "close");
            //  Flux<DataBuffer> myBody = incoming.getBody();
            response.setStatusCode(HttpStatus.UNAUTHORIZED);
            System.out.println("expect100test sending 401");}
        else {
            response.setStatusCode(HttpStatus.OK);
            x= response.writeWith(incoming.getBody());
            System.out.println("expect100test  sending 200");
        }
        System.out.println("expect100test request complete");
        return x;
    }
}
```

#SecurityConfig 
```java
@EnableWebFluxSecurity
class SecurityConfig {
    @Bean
    public SecurityWebFilterChain springSecurityFilterChain(final ServerHttpSecurity http) {
        return http.authorizeExchange()
                .pathMatchers(HttpMethod.OPTIONS).denyAll()
                .pathMatchers(HttpMethod.TRACE).denyAll()
                .anyExchange().permitAll()
                .and()
                .csrf().disable()
                .build();
    }
}
```

#pom 
```java
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <parent>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-starter-parent</artifactId>
        <version>2.7.17-SNAPSHOT</version>
        <relativePath/> <!-- lookup parent from repository -->
    </parent>
    <groupId>com.example</groupId>
    <artifactId>expect100test</artifactId>
    <version>0.0.1-SNAPSHOT</version>
    <name>expect100test</name>
    <description>expect100test</description>
    <properties>
        <java.version>1.8</java.version>
    </properties>
    <dependencies>
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-webflux</artifactId>
            <scope>compile</scope>
        </dependency>
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-security</artifactId>
        </dependency>
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-actuator</artifactId>
        </dependency>

        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-test</artifactId>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>io.projectreactor</groupId>
            <artifactId>reactor-test</artifactId>
            <scope>test</scope>
        </dependency>
    </dependencies>
    <build>
        <plugins>
            <plugin>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-maven-plugin</artifactId>
            </plugin>
        </plugins>
    </build>
</project>
```

Steps to repro on localhost netcat:

1 - Run the app
2 - Start netcat to localhost at the port the app is running on `nc localhost 8080`
3 - Copy/Paste in the request without the POST body
```java
POST / HTTP/1.1
Host: localhost
User-Agent: curl/7.81.0
Accept: */*
Expect: 100-continue
Content-Length: 10
Content-Type: application/x-www-form-urlencoded
connection: keep-alive
```
4 - Press enter twice to send the request
5 - Copy/Paste in the second request to send on that same connection
```java
GET / HTTP/1.1
Host: localhost
User-Agent: curl/7.81.0
Accept: */*
Content-Type: application/x-www-form-urlencoded
Authorization: Yes
connection: keep-alive
```
6 - Its an immediate 400 for second request and connection is closed.

For testing purposes - here is how to send the POST body payload in the netcat request Copy/Paste:
```java
POST / HTTP/1.1
Host: localhost
User-Agent: curl/7.81.0
Accept: */*
Expect: 100-continue
Content-Length: 10
Content-Type: application/x-www-form-urlencoded
connection: keep-alive

[000, 000]
```

## Possible Solution
This is where we are seeking input. We are unsure if this issue is considered a bug in netty,  reactor-netty, Spring or if this is not a bug at all and its expected that the implementing application should handle edge cases like this programatically vs the underlying abstractions.

If implementing in the server it appears the following is a common approach:
Close the connection on any non-200 status codes that have expectations such as the expect-100 header present but no POST body was read.

## Your Environment
This has been observed in a variety of environments such as cloud foundry and localhost.

violetagg:
Fixes #2910

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
@@ -386,12 +386,25 @@ public class ChannelOperations<INBOUND extends NettyInbound, OUTBOUND extends Ne
 	}
 
 	/**
-	 * Return true if inbound traffic is not incoming or expected anymore
+	 * Return true if inbound traffic is not incoming or expected anymore.
+	 * The buffered data is consumed.
 	 *
-	 * @return true if inbound traffic is not incoming or expected anymore
+	 * @return true if inbound traffic is not incoming or expected anymore.
+	 * The buffered data is consumed
 	 */
 	public final boolean isInboundDisposed() {
 		return inbound.isDisposed();
+	}
+
+	/**
+	 * Return true if inbound traffic is not incoming or expected anymore.
+	 * The buffered data might still not be consumed.
+	 *
+	 * @return true if inbound traffic is not incoming or expected anymore.
+	 * The buffered data might still not be consumed.
+	 */
+	protected final boolean isInboundComplete() {
+		return inbound.inboundDone;
 	}
 
 	/**
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
@@ -113,6 +113,7 @@ class HttpServerOperations extends HttpOperations<HttpServerRequest, HttpServerR
 	final ServerCookieEncoder cookieEncoder;
 	final ServerCookies cookieHolder;
 	final HttpServerFormDecoderProvider formDecoderProvider;
+	final boolean is100ContinueExpected;
 	final boolean isHttp2;
 	final BiFunction<? super Mono<Void>, ? super Connection, ? extends Mono<Void>> mapHandle;
 	final HttpRequest nettyRequest;
@@ -138,6 +139,7 @@ class HttpServerOperations extends HttpOperations<HttpServerRequest, HttpServerR
 		this.cookieHolder = replaced.cookieHolder;
 		this.currentContext = replaced.currentContext;
 		this.formDecoderProvider = replaced.formDecoderProvider;
+		this.is100ContinueExpected = replaced.is100ContinueExpected;
 		this.isHttp2 = replaced.isHttp2;
 		this.mapHandle = replaced.mapHandle;
 		this.nettyRequest = replaced.nettyRequest;
@@ -186,6 +188,7 @@ class HttpServerOperations extends HttpOperations<HttpServerRequest, HttpServerR
 		this.cookieHolder = ServerCookies.newServerRequestHolder(nettyRequest.headers(), decoder);
 		this.currentContext = Context.empty();
 		this.formDecoderProvider = formDecoderProvider;
+		this.is100ContinueExpected = HttpUtil.is100ContinueExpected(nettyRequest);
 		this.isHttp2 = isHttp2;
 		this.mapHandle = mapHandle;
 		this.nettyRequest = nettyRequest;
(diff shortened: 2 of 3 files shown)


pderop (approved):
LGTM, nice test !

violetagg:
@pderop Thanks

Issue #2959, opened 2023-11-03
isConnectionReset returns false for SocketException(Connection Reset)


eternachen:
reactor.netty.channel.AbortedException
    
public static boolean isConnectionReset(Throwable err) {
        return err instanceof AbortedException && "Connection has been closed BEFORE send operation".equals(err.getMessage()) || err instanceof IOException && (err.getMessage() == null || err.getMessage().contains("Broken pipe") || err.getMessage().contains("Connection reset by peer")) || err instanceof SocketException && err.getMessage() != null && err.getMessage().contains("Connection reset by peer");
    }

Currently for Connection Reset issue, the error message is "Connection reset" instead of "Connection reset by peer".

![image](https://github.com/reactor/reactor-netty/assets/3196459/fecafb86-441d-491b-8d8f-0dc1ea88ec68)

* Reactor version(s) used: latest reactor
* Other relevant libraries versions (eg. `netty`, ...):
* JVM version (`java -version`): latest jdk17
* OS and version (eg. `uname -a`):  Darwin Kernel Version 22.6.0: Wed Jul  5 22:21:56 PDT 2023; root:xnu-8796.141.3~6/RELEASE_X86_64 x86_64

In https://github.com/corretto/corretto-17/blob/de588920c83977ba54d425805220dada5084d4b1/src/hotspot/share/runtime/os.cpp#L1466

It looks like     X(ECONNRESET, "Connection reset") \

violetagg:
@eternachen There are two different exceptions the one is `Connection reset`, the other is `Connection reset by peer`

violetagg:
@eternachen I'm closing this. The intent for `AbortedException#isConnectionReset` is to handle `Connection reset by peer` which is a different exception than `Connection reset`.

PR #2960, opened 2023-11-03
Fix for https://github.com/reactor/reactor-netty/issues/2959

eternachen:
reactor.netty.channel.AbortedException

public static boolean isConnectionReset(Throwable err) {
return err instanceof AbortedException && "Connection has been closed BEFORE send operation".equals(err.getMessage()) || err instanceof IOException && (err.getMessage() == null || err.getMessage().contains("Broken pipe") || err.getMessage().contains("Connection reset by peer")) || err instanceof SocketException && err.getMessage() != null && err.getMessage().contains("Connection reset by peer");
}

Currently for Connection Reset issue, the error message is "Connection reset" instead of "Connection reset by peer".


violetagg:
@eternachen I'm closing this. The intent for `AbortedException#isConnectionReset` is to handle `Connection reset by peer` which is a different exception than `Connection reset`.

PR #2969, opened 2023-11-12
Introducing HTTP3 support 

fokion:
Started reviewing and migrating parts

### DIFF
--- a/codequality/checkstyle.xml
+++ b/codequality/checkstyle.xml
@@ -92,6 +92,17 @@
 		<module name="RedundantImport"/>
 		<module name="UnusedImports"/>
 
+		<!-- Javadoc Comments -->
+		<module name="NonEmptyAtclauseDescription" />
+		<module name="AtclauseOrder">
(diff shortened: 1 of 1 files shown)


violetagg:
@fokion May be It is good if you can share what are the overall plans/architecture that you have in mind.

We can use this project https://github.com/netty/netty-incubator-codec-http3

fokion:
The reason why I added this draft pr is to share my thoughts. 

Are we going to add the dependency as it is or migrate it to this project? Moreover there are some updates since it was last implemented in that incubator.

violetagg:
We would like to use `netty-incubator-codec-http3` as a dependency, yes. If something is missing from SPEC point of view, we prefer to contribute it to `netty-incubator-codec-http3`.

fokion:
Ok in that case will add the dependency directly later today and raise any PRs there. Thanks for the clarification

violetagg:
This is superseded by #2972

PR #2972, opened 2023-11-15
WIP - HTTP3 migration

fokion:
### ⚠️ Still in progress

- creating config for http3, made an abstract class of the configuration and set the default min values in the streams based on the incubator project
- plan is to create a structure similar to the abstract class `QuicServer` and `QuicServerBind` for Http3 ( let me know if you have any other suggetions).

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,5 +3,6 @@ version=1.1.14-SNAPSHOT
 reactorNettyQuicVersion=0.1.14-SNAPSHOT
 reactorCoreVersion=3.5.13-SNAPSHOT
 reactorAddonsVersion=3.5.2-SNAPSHOT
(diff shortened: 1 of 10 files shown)


violetagg:
@fokion I need to take a look first at Netty HTTP3 codec. I haven't time to think about a possible implementation.
I'll try to find time next week to have at least rough idea. 
Things that we need to think in advance like pros/cons: the current HTTP server supports configuring the protocol version so that it is easy for users to switch between different HTTP versions. Do we want to think in that direction for HTTP/3? Or it is more likely that we want a new HTTP/3 server abstraction?

violetagg:
@fokion  @pderop Some very rough POC. Tell me what you think.
https://github.com/reactor/reactor-netty/tree/http3-server

violetagg:
This is superseded by #3176

PR #2998, opened 2023-12-13
Depend on `Netty` `v4.1.104.Final`

violetagg:


violetagg:
The build is currently failing https://github.com/netty/netty/issues/13728

PR #3000, opened 2023-12-14
Do not forward the last inbound data frame when it is empty

violetagg:
There is no need to forward an empty buffer to the response consumers, directly invoke complete.

Related to https://github.com/spring-projects/spring-framework/issues/31810


pderop (approved):

violetagg:

PR #3063, opened 2024-02-16
Enable `DNS` `retryOnTimeout` with `TCP`

### LINKED ISSUE #3059: Enabling DNS retryOnTimeout with TCP
Netty sometimes get a 5000ms timeout error when querying DNS. e.g.
`Caused by: io.netty.resolver.dns.DnsNameResolverTimeoutException: [19509: /172.33.0.2:53] DefaultDnsQuestion(<hidden> IN A) query '19509' via UDP timed out after 5000 milliseconds (no stack trace available)`

The team has now merged this branch that allows for a retry over TCP instead of UDP, in case of timeout.
I do not think it is possible to set this boolean from reactor-netty?
I would expect it to be in the `NameResolverProvider.NameResolverSpec` 

The merged branch:
https://github.com/netty/netty/pull/13757

violetagg:
Add an API for configuring the resolver to fallback to `TCP` if a timeout is detected. By default, the resolver will only try to use `TCP` if the response is marked as truncated.

Fixes #3059

### DIFF
--- a/docs/asciidoc/http-client.adoc
+++ b/docs/asciidoc/http-client.adoc
@@ -564,6 +564,8 @@ Default: -1 (to determine the value from the OS on Unix or use a value of 1 othe
 | `queryTimeout` | Sets the timeout of each DNS query performed by this resolver (resolution: milliseconds). Default: 5000.
 | `resolveCache` | The cache to use to store resolved DNS entries.
 | `resolvedAddressTypes` | The list of the protocol families of the resolved address.
+| `retryTcpOnTimeout` | Specifies whether this resolver will also fallback to TCP if a timeout is detected.
+By default, the resolver will only try to use TCP if the response is marked as truncated.
 | `roundRobinSelection` | Enables an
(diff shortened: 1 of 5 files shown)


pderop (approved):

AkashB23:
@violetagg  this configuration change that can be done with small change while creating an httpclient `  httpClient.resolver(
            nameResolverSpec -> nameResolverSpec.retryTcpOnTimeout(true)
        );` will be overridden by https://github.com/reactor/reactor-netty/blob/2cc87e2d148ce67e75627bfdb5d6226553b5a894/reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java#L426 and used here for a request https://github.com/reactor/reactor-netty/blob/2cc87e2d148ce67e75627bfdb5d6226553b5a894/reactor-netty-core/src/main/java/reactor/netty/tcp/TcpResources.java#L315 so whats the way to set tcpFallback for timeout ?

PR #3083, opened 2024-03-07
Provide better exception message for invalid `PORT` env variable

### LINKED ISSUE #3070: default port based on PORT environment variable
In the file https://github.com/reactor/reactor-netty/blob/559b5efaa538d3d1157b832cae271cd13129fb45/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L704, the default port is found using
```
Integer.parseInt(System.getenv("PORT"))
```
1) I could not find this use of the PORT environment variable in the documentation
2) If the PORT environment variable is already defined, its definition will be used when this class is loaded. In particular, if the PORT environment variable content is not parsable as an int, like a string for instance, it will result in an exception.

I suggest you document that use of the PORT environment variable, add extra care when parsing it as it could be not parseable, and even add a log indicating you are using it for debug purpose.

violetagg:
- Update javadoc
- Update reference documentation

Fixes #3070

### DIFF
--- a/docs/asciidoc/http-client.adoc
+++ b/docs/asciidoc/http-client.adoc
@@ -16,6 +16,7 @@ and adds Reactive Streams backpressure.
 
 To connect the `HTTP` client to a given `HTTP` endpoint, you must create and configure a
 {javadoc}/reactor/netty/http/client/HttpClient.html[`HttpClient`] instance.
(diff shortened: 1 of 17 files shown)


pderop (approved):

violetagg:

PR #3106, opened 2024-03-17
add netty epoll ARM 64 dependency

sullis:
Netty native transport epoll has 2 artifacts:

- x86 64
- aarch 64

This PR adds aarch 64 (ARM 64).


sullis:
GitHub Action build failure 
```
Error: Exception in thread "main" java.io.IOException: Downloading from https://services.gradle.org/distributions/gradle-7.6.4-bin.zip failed: timeout (10000ms)
```

violetagg:
For the moment we don't want to add more natives as transitive dependencies. This dependency can be added by users, similar to KQueue natives.

PR #3120, opened 2024-03-26
Migrate 1.2.x doc to Antora

pderop:
This PR is first attempt to migrate Reactor Netty 1.2.x documentation to Antora 

There are three logical commits in this PR:

- bfb304d6e01e76a5765efcc5271e3ca8600ffbef: it contains all modifs done in .adoc files, and also contains antora yaml files
-  cf9365f166b596fee3d4fefdcaec9c02b7267bcf: contains all adaptation done in various gradle scripts
- d3bf58953487aa196054461eb7cab3ad738aac19: contains adaptations done in the GH worfklows.

To test this PR, you can proceed like the following:

### Testing without PDF:

- clone this PR
- switch to java 17 (the docs sub project requires java17)
- build the doc and generate it to docs/build/site/index.html:
```
./gradlew antora
```
- you can now browse docs/build/site/index.html
- switch to java 8
- deploy reactor-netty docs to your local M2 
```
./gradlew publishToMavenLocal
```
- now, you can verify the content of ~/.m2/repository/io/projectreactor/netty/reactor-netty/1.2.0-SNAPSHOT/reactor-netty-1.2.0-SNAPSHOT-docs.zip or  ~/.m2/repository/io/projectreactor/netty/reactor-netty/1.2.0-SNAPSHOT/reactor-netty-1.2.0-SNAPSHOT-javadoc.jar

### Testing with PDF:

As before, PDF are not generated by default in snapshot mode. If you want to also include a PDF in the generated docs zip, then do the following:

- first, you will need to install **asciidoctor-pdf** command, because the antora pdf assembler requires it (this command is provided when doing "brew install asciidoctor"):
```
brew install asciidoctor
```
- switch to java17
- build antora and force PDF generation (in release mode, no need to use -PforcePdf, it's like before):
```
./gradlew antora -PforcePdf
this will generate  ./docs/build/assembler/reactor-netty/1.2.0/reactor-netty-reference-guide.pdf
```
- switch to java 8
- deploy reactor-netty docs to your local M2:
```
./gradlew publishToMavenLocal -PforcePdf
```
- verify the content of ~/.m2/repository/io/projectreactor/netty/reactor-netty/1.2.0-SNAPSHOT/reactor-netty-1.2.0-SNAPSHOT-docs.zip, which should also include a pdf in the root dir of the zip.


pderop:
force-pushed for centralizing antora versions in top level build.gradle

docs/antora-playbook.yml
```diff
@@ -0,0 +1,52 @@
...
+    - '@antora/collector-extension'
+    - '@antora/atlas-extension'
+    - id: pdf-extension # pdf-extension is enabled from antora-docs.gradle
+      require: '@antora/pdf-extension'
+      enabled: false
+    - require: '@springio/antora-extensions/root-component-extension'
+      root_component_name: 'reactor-netty'
+site:
+  title: Reactor Netty Reference Guide
+  url: https://pderop.github.io/test-antora-pages/docs/netty/reference/
```
violetagg, line 15: Is this URL correct?

docs/modules/ROOT/partials/wire-logger.adoc
```diff
@@ -7,58 +7,54 @@ By default, wire logging is disabled.
...
 and apply the following configuration:
 
-====
 [source,java,indent=0]
-.{examplesdir}/wiretap/Application.java
+.{examples-dir}/wiretap/Application.java
+[%unbreakable]
 ----
-include::{examplesdir}/wiretap/Application.java[lines=18..35]
+include::{examples-dir}/wiretap/Application.java[lines=18..35]
```
violetagg, line 14: Next time when you are preparing a PR and you want to do a renaming, I would love to see this renaming as a separate PR. Such renaming is not helping the reviewer with checking the real changes.

pderop: I should probably have given descriptions of all the changes in the PR, but there are so many modifications ...

So, please, let me explain here the reason why I changed the variable name:

unfortunately, `examplesdir` is deprecated in Antora, please check [this](https://docs.antora.org/antora/3.0/whats-new/#deprecation-summary), more specifically check the deprecation summary:

> The partialsdir and examplesdir attributes are deprecated. Use the resource ID of the [partial](https://docs.antora.org/antora/3.0/page/include-a-partial/) or [example](https://docs.antora.org/antora/3.0/page/include-an-example/) to reference the resource instead.

now, if the `examplesdir` variable name is kept, here is what we would get (see the title where the full pass is missing):

Screenshot 2024-03-27 at 09 31 15

so, this is why I had to change the `examplesdir` variable to another one (I used `examples-dir` instead), and then we now get the correct path in the title of the included source code (except that there is still the `examples$` prefix, but for the moment, I could not fix it):

Screenshot 2024-03-27 at 09 33 51

the renamed variable `examples-dir` is defined in the adoc files that are including the partial wire-logger.adoc, for example, see the heading in the tcp-client.adoc:

```
:sourcedir: {core}/src/main/java
:examples-dir: {examples}/src/main/java/reactor/netty/examples/documentation/tcp/client

```

docs/antora-playbook.yml
pderop: no, I have corrected it in https://github.com/reactor/reactor-netty/pull/3120/commits/3e12ac625eb3fac7c849a80f04b5c22ae594a362

violetagg:
@pderop Nice! Let's merge it!

violetagg (approved):

pderop:
@violetagg , thanks you for your review !

PR #3137, opened 2024-04-04
Remove wrong description in docsZip

pderop:
There is wrong description in docsZip gradle task from reactor-netty/build.gradle, it should be removed.


violetagg (approved):

pderop:
@violetagg , thanks.

PR #3138, opened 2024-04-04
Re enable full build with jdk17

pderop:
Since the migration to Antora, it's not possible anymore to do a full build using jdk17:

```
sdk use java 17.0.10-zulu
./gradlew publishToMavenLocal

> Configure project :
1.2.0-SNAPSHOT is a snapshot, will use 1.2.0.BUILD-202404040843 for bnd
[Spotless] Local run detected, ratchet from origin/main

> Task :docs:generatePomFileForMavenJavaPublication FAILED

FAILURE: Build failed with an exception.

* Where:
Script '/Users/***/Documents/reactor-netty/gradle/setup.gradle' line: 153

* What went wrong:
Execution failed for task ':docs:generatePomFileForMavenJavaPublication'.
> Could not apply withXml() to generated POM
   > Cannot invoke method children() on null object

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org/
```

This PR re enable full build using jdk17:

- in build.gradle: toolchain sets JDK17 for docs sub project
- in gradle/settings.gradle, the docs project must be excluded, it does not contain dependencies, that was causing the `Cannot invoke method children() on null object` error
- in reactor-netty/build.gradle: if JDK17 is used, then a dependency to `:docs:antora' is dynamically added for the docsZip task.

### DIFF
--- a/reactor-netty/build.gradle
+++ b/reactor-netty/build.gradle
@@ -46,10 +46,14 @@ task docsZip(type: Zip) {
 	archiveClassifier = 'docs'
 	duplicatesStrategy "fail"
 
+	if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_17)) {
+		logger.lifecycle("Adding dependency on :docs:antora (JDK version is >= 17)")
+		dependsOn(':docs:antora')
+	}
+
 	def docsDir = file('../docs/build/site')
 	def isSnapshot = project.version.endsWith('-SNAPSHOT')
 	def version = isSnapshot ? project.version.takeWhile { it != '-' } : project.version
-	def pdfFile = file("../docs/build/assembler/reactor-netty/${version}/reactor-netty-reference-guide.pdf")
 	boolean forcePdf = project.hasProperty('forcePdf')
 
 	doFirst {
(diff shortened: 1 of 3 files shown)


reactor-netty/build.gradle
```diff
@@ -63,14 +67,12 @@ task docsZip(type: Zip) {
...
 
-	if (pdfFile.exists()) {
-		if (!isSnapshot || forcePdf) {
-			logger.lifecycle("pdf ${pdfFile} included in docs zip")
-
-			from(pdfFile) {
-				rename { fileName ->
-					"docs/reactor-netty-reference-guide-${project.version}.pdf"
-				}
+	if (!isSnapshot || forcePdf) {
```
violetagg, line 70: Is this change related to java version?

pderop: yes, when building everything **in one shot** using JDK17, the removal of the `if (pdfFile.exists())` condition allows to include any pdf generated by the :docs:antora task. If we keep the `if (pdfFile.exists())` as before, I have observed that the PDF (if generated) is not actually included.

let me explain:

If we retain the `if (pdfFile.exists())` condition as before, this condition is evaluated during the configuration phase, inaccurately assuming the PDF's absence since it hasn't been generated at that stage (the configuration phase precedes the execution phase, as far as my understanding goes). it's the :docs:antora task which generates the PDF, but during the subsequent execution phase after the configuration phase, so the condition has already been assessed as false, preventing its inclusion in the zip.

Eliminating the `if (pdfFile.exists())` condition fixes the problem. It's worth noting that this solution reflects my current understanding, and there might be other alternative approaches.

violetagg: thanks for the explanation

violetagg (approved):

pderop:
Rebased on top of latest main branch in order to manually merge #3140 which was conflicting with this PR.

PR #3147, opened 2024-04-11
Remove deprecated `SslProvider$DefaultConfigurationSpec` and `SslProvider$DefaultConfigurationType`

violetagg:
This is a problematic deprecated API that we would like to remove sooner than later.

This is in preparation for HTTP/3 support.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017-2023 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2017-2024 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
```diff
@@ -597,7 +412,6 @@ else if (pipeline.get(NettyPipeline.NonSslRedirectDetector) != null) {
 	@Override
 	public String toString() {
 		return "SslProvider {" +
-				"type=" + type +
 				", handshakeTimeoutMillis=" + handshakeTimeoutMillis +
```
chemicL, line 415:
```suggestion
				"handshakeTimeoutMillis=" + handshakeTimeoutMillis +
```

violetagg: oh nice, thanks

chemicL (changes requested):
One small toString issue noticed, but otherwise looks good. Was there no test coverage for these deprecated mechanisms?

violetagg:
@chemicL Thanks

Issue #1531, opened 2021-03-05
HTTP/3 support


violetagg:
https://netty.io/news/2021/03/04/http3-0-0-1-Final.html
https://github.com/netty/netty-incubator-codec-http3

fokion:
Can I have a look ?

violetagg:
@fokion Yes sure. Thanks!
May be start experimenting with `main` branch (1.1.x version) as a base, because it will be hard with `netty5` branch (2.0.x)

fokion:
Noticed that we are compiling with jdk 8 but in mac os with apple silicon I got the following error

```
> Failed to calculate the value of task ':reactor-netty-core:compileJava' property 'javaCompiler'.
   > Unable to download toolchain matching the requirements ({languageVersion=8, vendor=any, implementation=vendor-specific}) from 'https://api.adoptium.net/v3/binary/latest/8/ga/mac/aarch64/jdk/hotspot/normal/eclipse'.
      > Could not read 'https://api.adoptium.net/v3/binary/latest/8/ga/mac/aarch64/jdk/hotspot/normal/eclipse' as it does not exist.


```

So started creating a branch migrating to JDK 17 https://github.com/fokion/reactor-netty/commit/bba09f5ebe981f63a409091ef0983be3296cb18f

violetagg:
> Noticed that we are compiling with jdk 8 but in mac os with apple silicon I got the following error
> 
> ```
> > Failed to calculate the value of task ':reactor-netty-core:compileJava' property 'javaCompiler'.
>    > Unable to download toolchain matching the requirements ({languageVersion=8, vendor=any, implementation=vendor-specific}) from 'https://api.adoptium.net/v3/binary/latest/8/ga/mac/aarch64/jdk/hotspot/normal/eclipse'.
>       > Could not read 'https://api.adoptium.net/v3/binary/latest/8/ga/mac/aarch64/jdk/hotspot/normal/eclipse' as it does not exist.
> ```
> 
> So I tried migrating to JDK 17 which had its own challenges [fokion@bba09f5](https://github.com/fokion/reactor-netty/commit/bba09f5ebe981f63a409091ef0983be3296cb18f)

Reactor Netty 2.0.x (which is still on milestone phases) is based on Java 17 and Netty 5, unfortunately we still don't have Netty HTTP3 ported for Netty 5

[...]

PR #3150, opened 2024-04-11
Make `ProtocolSslContextSpec` generic

violetagg:
- `QuicSslContextBuilder` does not extend `SslContextBuilder`.
- Add `Http3SslContextSpec`, similar to `Http2SslContextSpec` and `Http11SslContextSpec`.

Related to #1531

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
@@ -231,6 +231,17 @@ public final class SslProvider {
 		 *
 		 * @param spec SslContext builder that provides, specific for the protocol, default configuration
 		 * @return {@literal this}
+		 * @since 1.2.0
+		 */
+		Builder sslContext(GenericSslContextSpec<?> spec);
+
+		/**
+		 * SslContext builder that provides, specific for the protocol, default configuration
+		 * e.g. {@link DefaultSslContextSpec}, {@link TcpSslContextSpec} etc.
+		 * The default configuration is applied before any other custom configuration.
+		 *
+		 * @param spec SslContext builder that provides, specific for the protocol, default configuration
+		 * @return {@literal this}
 		 * @since 1.0.6
 		 */
 		Builder sslContext(ProtocolSslContextSpec spec);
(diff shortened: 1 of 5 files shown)


reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
```diff
@@ -224,6 +224,17 @@ public interface Builder {
...
+		 * SslContext builder that provides, specific for the protocol, default configuration
+		 * e.g. {@link DefaultSslContextSpec}, {@link TcpSslContextSpec} etc.
+		 * The default configuration is applied before any other custom configuration.
+		 *
+		 * @param spec SslContext builder that provides, specific for the protocol, default configuration
+		 * @return {@literal this}
+		 * @since 1.2.0
+		 */
+		Builder sslContext(GenericSslContextSpec<?> spec);
+
```
pderop, line 237: I just wonder why we keep the old signature for `sslContext` method that takes as argument a ProtocolSslContextSpec ?

```
		Builder sslContext(ProtocolSslContextSpec spec);
```

The `ProtocolSslContextSpec` class is now extending `GenericSslContextSpec<SslContextBuilder>`, so passing a ProtocolSslContextSpec to the new signature would work.

I may miss something obvious here, but why the old `Builder sslContext(ProtocolSslContextSpec spec)` signature is not removed ? maybe it's just because the old signature was not deprecated so far ?

violetagg: We can deprecated it only :(

pderop (approved):
+1

as discussed, maybe we can deprecate the old `sslContext(ProtocolSslContextSpec spec)`signature in SslProvider.SslContextSpec.
Something like:

```
		/**
		 * SslContext builder that provides, specific for the protocol, default configuration
		 * e.g. {@link DefaultSslContextSpec}, {@link TcpSslContextSpec} etc.
		 * The default configuration is applied before any other custom configuration.
		 *
		 * @param spec SslContext builder that provides, specific for the protocol, default configuration
		 * @return {@literal this}
		 * @since 1.0.6
		 * @deprecated as of 1.2.0. Prefer using {@link #sslContext(SslProvider.GenericSslContextSpec)}  
		 * This method will be removed in version 1.3.0.
		 */
		@Deprecated
		Builder sslContext(ProtocolSslContextSpec spec);
```

PR #3158, opened 2024-04-15
Limit permission in publish cleanup job

pderop:
in the publish CI, the `cleanup` job is using a too broad permission (`permissions: write-all`) when deleting the docs-build artifact.

According to https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#delete-an-artifact, we can limit the permission to only the necessary level: `actions: write`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -225,7 +225,8 @@ jobs:
     needs: [ reactor-netty-core, tagRelease, tagMilestone, deploySnapshot ]
     if: always() # cleanup always run after all needed jobs, regardless of whether they were successful
     runs-on: ubuntu-20.04
-    permissions: write-all
+    permissions:
+        contents: write
     steps:
       - name: delete antora docs-build artifact
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -225,7 +225,8 @@ jobs:
     needs: [ reactor-netty-core, tagRelease, tagMilestone, deploySnapshot ]
     if: always() # cleanup always run after all needed jobs, regardless of whether they were successful
     runs-on: ubuntu-20.04
-    permissions: write-all
+    permissions:
+        contents: write
```
violetagg, line 229: Is it `contents: write` or `actions: write`?

pderop: it's `actions: write` ... mistake from me :slightly_frowning_face: 

thanks for catching, I will correct this.

pderop: fixed in https://github.com/reactor/reactor-netty/pull/3158/commits/82249e4787cb3ea5ac440ba41defbe41b3b99f11

pderop:
@violetagg , thanks!

PR #3160, opened 2024-04-16
Deprecate `SslContextSpec#sslContext(reactor.netty.tcp.SslProvider.ProtocolSslContextSpec)`

violetagg:
`SslProvider.SslContextSpec#sslContext(reactor.netty.tcp.SslProvider.ProtocolSslContextSpec)` is deprecated in favour of  `SslProvider.SslContextSpec#sslContext(reactor.netty.tcp.SslProvider.GenericSslContextSpec<?>)`

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
@@ -243,7 +243,10 @@ public final class SslProvider {
 		 * @param spec SslContext builder that provides, specific for the protocol, default configuration
 		 * @return {@literal this}
 		 * @since 1.0.6
+		 * @deprecated as of 1.2.0. Prefer using {@link #sslContext(GenericSslContextSpec)}.
+		 * This method will be removed in version 1.3.0.
 		 */
+		@Deprecated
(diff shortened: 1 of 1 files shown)


reactor-netty-core/src/main/java/reactor/netty/tcp/SslProvider.java
```diff
@@ -243,7 +243,10 @@ public interface SslContextSpec {
 		 * @param spec SslContext builder that provides, specific for the protocol, default configuration
 		 * @return {@literal this}
 		 * @since 1.0.6
+		 * @deprecated as of 1.2.0. Prefer using {@link #sslContext(GenericSslContextSpec)}.
```
pderop, line 246: when generating the javadoc, I noticed that the suggested link is not clickable, see:

Screenshot 2024-04-16 at 12 02 54

The following fixes the link:

```suggestion
		 * @deprecated as of 1.2.0. Prefer using {@link #sslContext(SslProvider.GenericSslContextSpec)}.
```

violetagg: nice catch, fixed

pderop (approved):

PR #3164, opened 2024-04-17
Depend on `Micrometer` `v1.13.0-SNAPSHOT`

violetagg:
Depend on `Micrometer Tracing` `v1.3.0-SNAPSHOT`
Depend on `Micrometer Doc Generator` `v1.0.2`
Depend on `Context Propagation` `v1.1.0`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,11 +85,11 @@ ext {
 	}
 
 	//Metrics
-	micrometerVersion = '1.10.0' //optional baseline
-	micrometerTracingVersion = '1.0.0' //optional baseline
-	micrometerDocsVersion = '1.0.0' //optional baseline
+	micrometerVersion = '1.13.0-SNAPSHOT' //optional baseline
+	micrometerTracingVersion = '1.3.0-SNAPSHOT' //optional baseline
+	micrometerDocsVersion = '1.0.2' //optional baseline
 
-	contextPropagationDefaultVersion = '1.0.0' //optional baseline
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -85,11 +85,11 @@ ext {
 	}
 
 	//Metrics
-	micrometerVersion = '1.10.0' //optional baseline
-	micrometerTracingVersion = '1.0.0' //optional baseline
-	micrometerDocsVersion = '1.0.0' //optional baseline
+	micrometerVersion = '1.13.0-SNAPSHOT' //optional baseline
+	micrometerTracingVersion = '1.3.0-SNAPSHOT' //optional baseline
+	micrometerDocsVersion = '1.0.2' //optional baseline
```
pderop, line 90: I see that there is a newer 1.0.3-SNAPSHOT version available for `micrometer-docs-generator`: https://repo.spring.io/ui/native/snapshot/io/micrometer/micrometer-docs-generator/1.0.3-SNAPSHOT/

Should we keep the 1.0.2 version and not use the 1.0.3-SNAPSHOT ?

violetagg: version `1.0.2` is there since a long time and there is no due date neither for `1.0.3` nor for 1.1.0-M1, so imo we should stay on `1.0.2`

pderop: ok, thanks.

pderop (approved):
seems good to me

violetagg:
The failed CI is not relevant

PR #3197, opened 2024-04-24
Add dependency to `Netty HTTP3 Codec` for aggregated javadoc

violetagg:
Without this dependency the javadoc cannot be built correctly.

Related to #1531


pderop (approved):
looking good

violetagg:
@pderop Thanks!

PR #2755, opened 2023-03-30
Avoid null metric recorder remoteAddress parameters

pderop:
Motivation:

When a reactor-netty http server receives a request with an invalid request URI which can't be parsed, any registered metric recorders are invoked with some null `remoteAddress`parameters. Basically, when the URL can’t be parsed, a 400 bad request is sent using [HttpServerOperations.sendDecodingFailures](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java#L850)  which is using a [FailedHttpServerRequest](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java#L1023). And the constructor of FailedHttpServerRequest calls super(...) with null as the fifth parameter which is the connectionInfo.

That’s why [AbstractHttpServerMetricsHandler.recordWrite](https://github.com/reactor/reactor-netty/blob/1.0.x/reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java#L254) passes a null remoteAddress parameter because ops.remoteAddress() returns null (connectionInfo is null).

Modification:

The ConnectionInfo object is now used when sending a decoding error. Now, if the ConnectionInfo could not be created in some rare conditions like when the Host or Forwarded/X-Forwarded-For headers are malformed, then we fallback on established connection info.

Fixes #2748

PR #3210, opened 2024-04-29
Always use remote socket address for the metrics

violetagg:
This is a change additional to the change made with #2755


pderop (approved):

violetagg:

PR #3211, opened 2024-04-29
Cache the request info when collecting metrics

violetagg:
Do not search for `HttpServerOperations` every time when a request info is needed.

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
@@ -212,14 +212,12 @@ abstract class AbstractHttpServerMetricsHandler extends ChannelDuplexHandler {
 				ChannelOperations<?, ?> channelOps = ChannelOperations.get(ctx.channel());
 				if (channelOps instanceof HttpServerOperations) {
 					ops = (HttpServerOperations) channelOps;
-					if (!initialized) {
-						method = methodTagValue.apply(ops.method().name());
-						path = uriTagValue == null ? ops.path : uriTagValue.apply(ops.path);
-						// Always take the remote address from the operations in order to consider proxy information
-						// Use remoteSocketAddress() in order to obtain UDS info
-						remoteSocketAddress = ops.remoteSocketAddress();
-						initialized = true;
-					}
+					method = methodTagValue.apply(ops.method().name());
+					path = uriTagValue == null ? ops.path : uriTagValue.apply(ops.path);
+					// Always take the remote address from the operations in order to consider proxy information
+					// Use remoteSocketAddress() in order to obtain UDS info
+					remoteSocketAddress = ops.remoteSocketAddress();
+					initialized = true;
 					startRead(ops);
 				}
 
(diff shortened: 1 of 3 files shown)


violetagg:
The failed test on CI is not relevant

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
```diff
@@ -182,12 +205,22 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+		HttpServerOperations ops = null;
 		try {
 			if (msg instanceof HttpRequest) {
+				reset();
 				ChannelOperations<?, ?> channelOps = ChannelOperations.get(ctx.channel());
 				if (channelOps instanceof HttpServerOperations) {
-					HttpServerOperations ops = (HttpServerOperations) channelOps;
-					startRead(ops, uriTagValue == null ? ops.path : uriTagValue.apply(ops.path), methodTagValue.apply(ops.method().name()));
+					ops = (HttpServerOperations) channelOps;
+					if (!initialized) {
```
pderop, line 215: is it necessary to do this `if (!initialized) {` ?
The reset() method that has been invoked at line 211 has reset the `initialized` field to `false`, so maybe we can remove the `if` and always initialize the fields , or am I missing something else ?

violetagg: yeah true, let me fix it

reactor-netty-http/src/main/java/reactor/netty/http/server/AbstractHttpServerMetricsHandler.java
```diff
@@ -138,8 +154,19 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+						method = methodTagValue.apply(ops.method().name());
+						path = uriTagValue == null ? ops.path : uriTagValue.apply(ops.path);
+						// Always take the remote address from the operations in order to consider proxy information
+						// Use remoteSocketAddress() in order to obtain UDS info
+						remoteSocketAddress = ops.remoteSocketAddress();
+						initialized = true;
+					}
+					if (contextView == null) {
+						contextView = ops.currentContext();
```
pderop, line 166: if I'm correct, `ops.currentContext()` corresponds `HttpServerOperations.currentContext()` which returns the `HttpServerOperations.currentContext` attribute; but this attribute is `volatile`. 

Can we manage to avoid always accessing such volatile attribute when the AbstractHttpServerMetricsHandler is not a ContextAwareHttpServerMetricsHandler ?

For example, the following method could be added in the AbstractHttpServerMetricsHandler:

```
	protected ContextView currentContext(HttpServerOperations ops) {
		return Context.empty();
	}
```

and this method could be overriden only by the ContextAwareHttpServerMetricsHandler class:

```
	@Override
	protected final ContextView currentContext(HttpServerOperations ops) {
		return ops.currentContext();
	}
```

and instead of calling `contextView = ops.currentContext();`, the AbstractHttpServerMetricsHandler could then invoke `contextView = currentContext(ops);` ?

violetagg: Agree, please check the new commit

pderop (approved):
looking good.

violetagg:

PR #3230, opened 2024-05-07
Always add `proxy.address` tag for the client metrics

### LINKED ISSUE #3218: Micrometer Prometheus metrics export fails due to conditional `proxy.address` tag
With the introduction of the `proxy.address` metrics tag in #3081, we started to observe that Micrometer Prometheus metrics registration started to fail, due to the [requirement](https://github.com/micrometer-metrics/micrometer/blob/f1471c573d2067ca508387f67815b6b5858d53fc/implementations/micrometer-registry-prometheus-simpleclient/src/main/java/io/micrometer/prometheus/PrometheusMeterRegistry.java#L594-L600) that metrics with the same name have the same set of tags:
```
2024-05-01 14:03:54.728 ERROR 1 --- [or-http-epoll-4] PrometheusMetricsConfig () : Failed to register meter_id='MeterId{name='reactor.netty.http.client.data.received', tags=[tag(country=REDACTED),tag(proxy.address=REDACTED),tag(remote.address=REDACTED),tag(uri=http)]}': Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter named 'reactor.netty.http.client.data.received' containing tag keys [country, remote_address, uri]. The meter you are attempting to register has keys [country, proxy_address, remote_address, uri].
```

Besides reduced observability, this causes a stream of error logs. This happens because in one of our Spring Boot applications we use Netty to connect to multiple other systems, only one of which involves a proxy.

Would be possible to e.g. either have the `proxy.address` tag be conditionally omitted, or to always include it, with a default value (the empty string?) where relevant? (Maybe there are other approaches; just proposing the two "obvious" ones :smile:.)

NB: In the log line shown above, `tag(uri=http)` looks rather uninformative :eyes:. Didn't investigate that aspect.

violetagg:
When there is no proxy add `na` as a value tag.

Fixes #3218

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/Metrics.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/Metrics.java
@@ -293,6 +293,8 @@ public class Metrics {
 
 	public static final String UNKNOWN = "UNKNOWN";
 
+	public static final String NA = "na";
+
 	@Nullable
 	public static Observation currentObservation(ContextView contextView) {
 		if (contextView.hasKey(OBSERVATION_KEY)) {
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/MicrometerChannelMetricsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/MicrometerChannelMetricsHandler.java
@@ -35,6 +35,7 @@ import java.util.function.Supplier;
 
 import static reactor.netty.Metrics.CONNECT_TIME;
 import static reactor.netty.Metrics.ERROR;
+import static reactor.netty.Metrics.NA;
 import static reactor.netty.Metrics.OBSERVATION_REGISTRY;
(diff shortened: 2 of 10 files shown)


pderop:
in MicrometerChannelMetricsRecorder.java, line 176, is the following method still used ? if not, can we remove it ?

```
	@Nullable
	public final Timer getTlsHandshakeTimer(String name, @Nullable String address, String status) {
		MeterKey meterKey = new MeterKey(null, address, null, null, status);
		return MapUtils.computeIfAbsent(tlsHandshakeTimeCache, meterKey,
				key -> filter(Timer.builder(name)
				                   .tags(REMOTE_ADDRESS, address, STATUS, status)
				                   .register(REGISTRY)));
	}
```

pderop:
yes, indeed, it can't be removed, in this case, we can deprecate it.

pderop (approved):

PR #3231, opened 2024-05-07
Obtain the channel from `ChannelHandlerContext` instead of `HttpServerOperations`

violetagg:
This is in addition to #3211


pderop (approved):

violetagg:

PR #3235, opened 2024-05-08
Ensure `ConnectionProvider` metrics are disposed unconditionally when graceful shutdown

### LINKED ISSUE #3226: Metrics not disposed for `GracefulShutdownInstrumentedPool`
When working with various ConnectionProvider/MetricRegistrar arrangements, I am able to verify deRegistration of Metrics when I dispose the provider, but only if I do not set anything for `disposeTimeout`. Once I add that, it loses the ability to deregister its Metrics.

## Expected Behavior
`ConnectionProvider` Metrics should be deRegistered on dispose, regardless of `disposeTimeout` use.

## Actual Behavior
When adding `disposeTimeout` to a `ConnectionProvider`, Metrics are no longer deRegistered when the `ConnectionProvider` is disposed

## Steps to Reproduce
Can reproduce this problem with a tweak to an existing `reactor-netty-core` test:
[Add to `ConnectionProvider` setup in `PooledConnectionProviderDefaultMetricsTest.testConnectionPoolPendingAcquireSize`](https://github.com/rs017991/reactor-netty/commit/6125294e327ecabe579603456134a9f6d093bd39):
```java
.disposeTimeout(Duration.ofSeconds(1))
```

## Possible Solution
In `PooledConnectionProvider`, I see that Metric cleanup is only performed in `onErrorResume` for the `GracefulShutdownInstrumentedPool` block:
https://github.com/reactor/reactor-netty/blob/dd0c42d725e93f8da6cd2d053af8d9fde083382b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java#L201-L208
whereas in the de facto `else` it is done unconditionally:
https://github.com/reactor/reactor-netty/blob/dd0c42d725e93f8da6cd2d053af8d9fde083382b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java#L220-L223

Perhaps GracefulShutdownInstrumentedPool is supposed to perform the deRegistration on its own happy path, but I see nothing like that in its implementation. Without knowing how it was intended to work, not sure else I can glean...

## Your Environment

* Reactor version(s) used: encountered in `1.1.18`; reproduction was just done against `main`
* JVM version: `21`
* OS and version: `Windows 11`

violetagg:
Dispose metrics not only when `onErrorResume`, but also on a happy path.

Fixes #3226

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -186,28 +186,14 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 			                        if (pool instanceof GracefulShutdownInstrumentedPool) {
 			                            return ((GracefulShutdownInstrumentedPool<T>) pool)
 			                                    .disposeGracefully(disposeTimeout)
+			                                    .then(deRegisterDefaultMetrics(id, poolFactory.registrar, remoteAddress))
 			                                    .onErrorResume(t -> {
 			                                        log.error("Connection pool for [{}] didn't shut down gracefully", e.getKey(), t);
-			                                        return Mono.fromRunnable(() -> {
-			                                            if (poolFactory.registrar != null) {
-			                                                poolFactory.registrar.get().deRegisterMetrics(name, id, remoteAddress);
-			                                            }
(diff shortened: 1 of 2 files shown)


pderop (approved):

violetagg:

PR #3243, opened 2024-05-14
Allow using SocketAddress proxies instead of just InetSocketAddress

### LINKED ISSUE #3242: Support SocketAddress for proxies
Currently proxies are required to be InetSocketAddress, but netty allows any SocketAddress to be used as a proxy.
This can be seen in in the ProxyProvider class:

https://github.com/reactor/reactor-netty/blob/ed7c9644b040aff54cecccdcdcd554193c27c97a/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java#L429

## Motivation

I'm interested in using UnixDomainSocket for proxies instead of InetSocketAddress.

## Desired solution

Extend the functionality of the proxy provider class to allow passing a SocketAddress instead of a InetSocketAddress to a proxy.
I've made a rough example how that could be made in a fork of mine: https://github.com/AlexProgrammerDE/reactor-netty/commit/a76c01d563afbad4112153fce48b04afbafe49bd
It tries to keep backwards support until the next major release is made.
I'd recommend the maintainers use my commit as a reference.

## Considered alternatives

Support UnixDomainSocket instead, but it's probably easier to just support SocketAddress directly to allow other future implementations of that class.

AlexProgrammerDE:
This commit adds new socketAddress methods and tries to keep backwards compat for users using the old address methods, which are deprecated. I'd recommend removing socketAddress and renaming it back to address on the next major release.

Fixes #3242


AlexProgrammerDE:
I guess the error lies in that new methods to the interface were added?

AlexProgrammerDE:
Maybe this can be fixed by keeping address unimplemented and the socketAddress method defaulting to delegating to address. And then the implementing builder class overrides socketAddress and implements address as the one delegating to socketAddress. Seems a bit unnecessary though since this class is only supposed to be implemented by the builder class.

violetagg:
what do you think about this?

```
		default Builder socketAddress(SocketAddress address) {
			throw new UnsupportedOperationException();
		}

		default Builder socketAddress(Supplier<? extends SocketAddress> addressSupplier) {
			throw new UnsupportedOperationException();
		}
```

AlexProgrammerDE:
Should pass checks now and is ready for review.

AlexProgrammerDE:
It seems like the build failed because of unrelated reasons with ssl stuff on windows. I've merged now upstream into my branch and I hope the build passes now.

AlexProgrammerDE:
Seems like a request failure to maven central broke the build this time:
image

AlexProgrammerDE:
image
Seems like there are still errors that aren't network issues. I doubt they are related to the content of this PR though.

violetagg:
Failed tests on Mac OS CI are not related

violetagg (approved):
@AlexProgrammerDE Thanks for the PR!

@reactor/netty-team PTAL

pderop (approved):
@AlexProgrammerDE

Nice, thanks for the PR!

+1

Issue #3265, opened 2024-05-27
HttpClient.reactorNettyVersion() throws NullPointException when loaded through -javaagent, under jdk 1.8


KyoUK4n:
HttpClient.reactorNettyVersion() throws NullPointException when loaded through -javaagent, under jdk 1.8

It seems all versions are using the same way to get reactorNettyVersion.

## Possible Solution

I've created a pull request, which might fix it.  [#3266](https://github.com/reactor/reactor-netty/pull/3266)

## Environment

* Reactor version(s) used: 1.1.19
* JVM version (`java -version`): oracle jdk 1.8
* OS and version (eg. `uname -a`):  window 12 amd64

PR #3266, opened 2024-05-27
fix: HttpClient.reactorNettyVersion() throws NullPointException

KyoUK4n:
fix: [#3265](https://github.com/reactor/reactor-netty/issues/3265)


violetagg (approved):
@KyoUK4n Thanks for the PR!

PR #3271, opened 2024-05-31
Add possibility to delay the last flush operation

violetagg:
When read is in process, delay the last flush until the read is done.


pderop (approved):
nice improvements!
seems good to me

violetagg:
@pderop Thanks

PR #3276, opened 2024-06-04
Make `MonoSend.MAX_SIZE` configurable by system property

### LINKED ISSUE #3255: MonoSend MAX_SIZE and REFILL_SIZE are not configurable
While working on S3 file download server (can't use pre-signed URLs) it was found that S3 AWS CRT client is using big part size(8MB) of the transferred chunks of data by default. [AWS SDK issue](https://github.com/aws/aws-sdk-java-v2/issues/5158). 

It seems that reactor-netty by default requests `MonoSend.MAX_SIZE` and `MonoSend.REFILL_SIZE` in the `MonoSendMany` when processing `Flux<ByteBuffer>` for streaming body response.

As a result, reactor requests 128*8MB=1024MB from the S3 Client. If the clients are slow enough, and multiple files are downloaded in parallel, this causes OutOfDirectMemory due to S3 client pushing 1024Mb for each file into netty buffers.

There are two issues here:

1. One is on the AWS S3 SDK side as CRT client doesn't support rateLimit on the Flux.
2. Another is reactor-netty having hardcoded sizes for initial and refill requests.

We fixed the issue by reducing default size of the part/chunk size in S3, however, this causes a chatty traffic and additional requests between application and S3. It would be better, if we can control initial request/refill on the reactor-netty side.

## Basic code snippet:
```
@GetMapping(path="/{filekey}")
Mono<ResponseEntity<Flux<ByteBuffer>>> downloadFile(@PathVariable("filekey") String filekey) {    
    GetObjectRequest request = GetObjectRequest.builder()
      .bucket(s3config.getBucket())
      .key(filekey)
      .build();
    
    return Mono.fromFuture(s3client.getObject(request, AsyncResponseTransformer.toPublisher()))
      .map(response -> {
        checkResult(response.response());
        String filename = getMetadataItem(response.response(),"filename",filekey);            
        return ResponseEntity.ok()
          .header(HttpHeaders.CONTENT_TYPE, response.response().contentType())
          .header(HttpHeaders.CONTENT_LENGTH, Long.toString(response.response().contentLength()))
          .header(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=\"" + filename + "\"")
          .body(Flux.from(response));
      });
}
```

## Motivation
Motivation is to preserve big chunk size on the S3 client side, but keep a better backpressure between reactor and S3 Client. Default request of 128 makes sense for S3 NettyNioAsyncHttpClient which uses 8KB chunks by default, but for CRT client which uses 8MB chunks it is too much.

## Desired solution

MonoSend MAX_SIZE and REFILL_SIZE should be made configurable and not hardcoded. 

## Considered alternatives

I don't see any other alternatives, but maybe you can suggest something else.

### AWS Java SDK version used
2.25.10

### CRT Client Version
0.29.14

### Reactor netty and netty version
reactor-netty 1.1.3
netty 4.1.108.Final

### JDK version used
openjdk version "17.0.3" 2022-04-19
oracle jdk 17

### Operating System and version
windows 11 23H2 and some linux image in EKS

VadimKirilchuk:
MonoSend.MAX_SIZE is a constant used by MonoSendMany and its SendManyInner class to request initial items from a publisher to send to an output Channel. This PR introduces a way to configure MonoSend.MAX_SIZE by using `reactor.netty.channel.send.prefetch.maxSize`(Upd. `reactor.netty.send.maxPrefetchSize`) system property. A custom value for the property may be necessary if requested items from publisher are large and requesting default number of items (128) requires too much memory. 

Fixes #3255

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/MonoSend.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/MonoSend.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2023 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2019-2024 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -58,7 +58,8 @@ abstract class MonoSend<I, O> extends Mono<Void> {
 		return (ToIntFunction) SIZE_OF;
 	}
 
-	static final int                    MAX_SIZE    = 128;
+	static final int                    MAX_SIZE =
+			Integer.parseInt(System.getProperty("reactor.netty.channel.send.prefetch.maxSize", "128"));
 
 	static final int                    REFILL_SIZE = MAX_SIZE / 2;
 
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/MonoSend.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/MonoSend.java
@@ -59,7 +59,7 @@ abstract class MonoSend<I, O> extends Mono<Void> {
 	}
 
 	static final int                    MAX_SIZE =
-			Integer.parseInt(System.getProperty("reactor.netty.channel.send.prefetch.maxSize", "128"));
+			Integer.parseInt(System.getProperty(ReactorNetty.REACTOR_NETTY_SEND_MAX_PREFETCH_SIZE, "128"));
 
 	static final int                    REFILL_SIZE = MAX_SIZE / 2;
 
--- a/reactor-netty-core/src/main/java/reactor/netty/ReactorNetty.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/ReactorNetty.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2023 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2011-2024 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -197,6 +197,11 @@ public final class ReactorNetty {
 	 *  Specifies the zone id used by the access log.
 	 */
 	public static final ZoneId ZONE_ID_SYSTEM = ZoneId.systemDefault();
+
+	/**
+	 * Default prefetch size ({@link Subscription#request(long)}) for data stream Publisher, fallback to 128.
+	 */
+	public static final String REACTOR_NETTY_SEND_MAX_PREFETCH_SIZE = "reactor.netty.send.maxPrefetchSize";
 
 	/**
 	 * Try to call {@link ReferenceCounted#release()} if the specified message implements {@link ReferenceCounted}.


reactor-netty-core/src/main/java/reactor/netty/channel/MonoSend.java
```diff
@@ -58,7 +58,8 @@ static <O> ToIntFunction<O> defaultSizeOf() {
 		return (ToIntFunction) SIZE_OF;
 	}
 
-	static final int                    MAX_SIZE    = 128;
+	static final int                    MAX_SIZE =
+			Integer.parseInt(System.getProperty("reactor.netty.channel.send.prefetch.maxSize", "128"));
```
violetagg, line 62: I think we can drop `channel`
```suggestion
			Integer.parseInt(System.getProperty("reactor.netty.send.prefetch.maxSize", "128"));
```

violetagg (changes requested):

reactor-netty-core/src/main/java/reactor/netty/channel/MonoSend.java
violetagg: @pderop What do you think about the name of the system property?

pderop: I am first wondering if , for consistency reasons, should this new property be declared (or not ?) in the ReactorNetty class ? (most of reactor netty system properties are defined in the `reactor.netty.ReactorNetty`).

To shorten the property name, we indeed may drop `channel`.

now, based on our existing system properties, what about `reactor.netty.send.maxPrefetch` or  `reactor.netty.send.maxPrefetchSize` ? this is more consistent with our existing system properties like "reactor.netty.logChannelInfo" which are generally ending with a camel case string ?

violetagg: @pderop I agree

@VadimKirilchuk Can you move the system property to `ReactorNetty` and rename it to `reactor.netty.send.maxPrefetchSize`?

VadimKirilchuk: @violetagg Sure, just busy at the moment, will do asap.
I like new property name.

VadimKirilchuk:
@violetagg I did the change, some tests are failing locally, but seems to be unrelated.
I'll try to run my sample app with updated code and property to make sure it works fine.

VadimKirilchuk:
I have also checked running that locally with some stub Flux Publisher:
```
        return Mono.just(ResponseEntity.ok()
                .body(Flux.<ByteBuffer>generate(sink -> sink.next(ByteBuffer.wrap(SAMPLE)))
                        .take(1024)
//                        .delayElements(Duration.ofMillis(100))
                        .doOnRequest(value -> log.info("Requested {}", value))));
```

```
2024-06-07T14:58:03.980-04:00  INFO 14288 --- [ctor-http-nio-2] o.e.s.controller.DownloadController      : New request
2024-06-07T14:58:04.031-04:00  INFO 14288 --- [ctor-http-nio-2] o.e.s.controller.DownloadController      : Requested 1
2024-06-07T14:58:04.079-04:00  INFO 14288 --- [ctor-http-nio-2] o.e.s.controller.DownloadController      : Requested 63
2024-06-07T14:58:04.121-04:00  INFO 14288 --- [ctor-http-nio-2] o.e.s.controller.DownloadController      : Requested 32
...
2024-06-07T14:58:04.331-04:00  INFO 14288 --- [ctor-http-nio-2] reactor.netty.http.server.AccessLog      : 0:0:0:0:0:0:0:1 - - [07/июн./2024:14:58:03 -0400] "GET /api/download HTTP/1.1" 200 3072 473
```

Seems to work as expected.

violetagg (approved):

violetagg:
@VadimKirilchuk 

> Also, do you want me to create a similar PR for 1.2.x? I am not sure if you are simply doing merges from 1.1.x -> 1.2.x.

There is no need. We use `forward merge` strategy, so the change is already available in the `dev` branches.

PR #3280, opened 2024-06-04
Adapt access log to the delayed last flush operation

violetagg:
This is in addition to #3271


violetagg:

PR #3290, opened 2024-06-10
Close the connection when decoding exception happens

### LINKED ISSUE #3285: Internal state of HttpObjectDecoder is not reset when connection pooling is used
I have a simple Spring Cloud Gateway application setup, which reroutes all of the calls to a specific URL and appends
some specific Authorization headers to the outgoing call. I've stumbled upon a weird HTTP client behavior when connection pooling is used:

## Expected Behavior
I expect that the same remote sever response will be handled the same way for every request when connection pooling is used.

## Actual Behavior
1. Call the gateway application through Postman client
2. Request gets rerouted to the configured server by the gateway application
3. HTTP 204 No Content is returned by the remote server
4. Gateway application successfully returns the response to the Postman client
5. Call the gateway application through Postman for the second time with the exact same request
6. Request is rerouted
7. HTTP 204 No Content is returned by the remote server
8. Gateway application is stuck until remote server sends RST/FIN flag and connection is closed (~4 minutes)
9. Gateway application raises `Connection prematurely closed BEFORE response` error.

In other words, every 2nd request is stuck and connection gets reset by the remote server after some timeout.
Upon further investigation why is this happening I inspected the network traffic using Wireshark and managed to decrypt the TLS traffic. It looks like some additional HTTP "Continuation" packets are being sent by the server after serving the HTTP 204 response.
The packets themselves do not contain HTTP version, therefore message creation fails in the `decode(..)` method, since `HttpRequestDecoder` raises an exception when `HttpVersion.valueOf(initialLine[2])` is called. This falls into an `catch` block which invokes the `invalidMessage` method and sets the `currentState` of `HttpDecoder` to `State.BAD_MESSAGE`. Regardless of that the first request and response is somehow still handled by the gateway application and response is successfully returned to the Postman client. Upon second request, `HttpDecoder` still has the previous `currentState` as `State.BAD_MESSAGE` and fails to decode the very first HTTP packet with the 204 response since it fall through the `switch` case in the `decode` method.

All of this can be avoided if connection pooling is disabled, and for every request a new connection is made, this results in `HttpDecoder` always starting in `currentState = State.SKIP_CONTROL_CHARS` and being able to decode the response.

## Steps to Reproduce
Unfortunately I'm unable to provide a reproducible example since this issue is present only when these HTTP "Continuation" packets are present.
wireshark_sc

Version(s) used:
* Spring Cloud Gateway WebFlux `4.1.0`
* Reactor Netty `1.1.16` (`1.1.17` also tested)
* openjdk version `"17.0.5" 2022-10-18 LTS`
* OpenJDK Runtime Environment Corretto-`17.0.5.8.1 (build 17.0.5+8-LTS)`
* OpenJDK 64-Bit Server VM Corretto-`17.0.5.8.1 (build 17.0.5+8-LTS, mixed mode, sharing)`
* Local - Windows 10 Enterprise `21H2 19044.4412` as well `Red Hat Universal Base Image 9` in Openshift

violetagg:
Fixes #3285

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperationsHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2023 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2011-2024 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -109,21 +109,29 @@ final class ChannelOperationsHandler extends ChannelInboundHandlerAdapter {
 			return;
 		}
 		try {
-			ChannelOperations<?, ?> ops = ChannelOperations.get(ctx.channel());
+			Connection connection = Connection.from(ctx.channel());
+			ChannelOperations<?, ?> ops = connection.as(ChannelOperations.class);
 			if (ops != null) {
 				ops.onInboundNext(ctx, msg);
 			}
 			else {
-				if (log.isDebugEnabled()) {
-					if (msg instanceof DecoderResultProvider) {
-						DecoderResult decoderResult = ((DecoderResultProvider) msg).decoderResult();
-						if (decoderResult.isFailure()) {
+				if (msg instanceof DecoderResultProvider) {
+					DecoderResult decoderResult = ((DecoderResultProvider) msg).decoderResult();
+					if (decoderResult.isFailure()) {
+						if (log.isDebugEnabled()) {
 							log.debug(format(ctx.channel(), "Decoding failed."), decoderResult.cause());
 						}
-					}
 
+						//"FutureReturnValueIgnored" this is deliberate
+						ctx.close();
+						listener.onUncaughtException(connection, decoderResult.cause());
+					}
+				}
+
+				if (log.isDebugEnabled()) {
 					log.debug(format(ctx.channel(), "No ChannelOperation attached."));
 				}
+
 				ReferenceCountUtil.release(msg);
 			}
 		}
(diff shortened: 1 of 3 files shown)


pderop:
it seems some other tests are not passing anymore, like HttpServerTest.keepAlive()

violetagg:
> it seems some other tests are not passing anymore, like HttpServerTest.keepAlive()

yeah the change should be in the `if (decoderResult.isFailure()) {`

pderop:
I see that the ctx is closed on decoding failure, but only if the logger is enabled in DEBUG mode ?

I mean, instead of

```
				if (log.isDebugEnabled()) {
					if (msg instanceof DecoderResultProvider) {
						DecoderResult decoderResult = ((DecoderResultProvider) msg).decoderResult();
						if (decoderResult.isFailure()) {
							log.debug(format(ctx.channel(), "Decoding failed."), decoderResult.cause());
							ctx.close();
							listener.onUncaughtException(connection, decoderResult.cause());
						}
					}

					log.debug(format(ctx.channel(), "No ChannelOperation attached."));
				}
```

shouldn't it be something like this ?

```
				if (msg instanceof DecoderResultProvider) {
					DecoderResult decoderResult = ((DecoderResultProvider) msg).decoderResult();
					if (decoderResult.isFailure()) {
						if (log.isDebugEnabled()) {
							log.debug(format(ctx.channel(), "Decoding failed."), decoderResult.cause());
						}
						ctx.close();
						listener.onUncaughtException(connection, decoderResult.cause());
					}

					if (log.isDebugEnabled()) {
						log.debug(format(ctx.channel(), "No ChannelOperation attached."));
					}
				}
```

violetagg:
@pderop Thanks, check the new commit

pderop (approved):
LGTM :rocket:

violetagg:
@pderop I found one additional issue, commit is commit in the next minutes.

violetagg:
I will move the fix for the next release. I want to play more with various scenarios.

pderop:
ok Violeta.

violetagg:
@pderop The PR is ready for review

pderop (approved):

violetagg:
@pderop Thanks!

Issue #221, opened 2017-11-15
Connection closed prematurely following HTTP 500 responses


sergiords:
### Actual behavior

Following some requests with HTTP 500 status, other requests do not complete and raise `java.io.IOException: Connection closed prematurely` exceptions.

```sh
2017-11-15 20:58:04.205 ERROR 1166 --- [ctor-http-nio-6] o.s.web.reactive.function.client         : onError(java.io.IOException: Connection closed prematurely)
2017-11-15 20:58:04.205 ERROR 1166 --- [ctor-http-nio-6] o.s.web.reactive.function.client         : 

java.io.IOException: Connection closed prematurely
	at reactor.ipc.netty.http.client.HttpClientOperations.onInboundComplete(HttpClientOperations.java:262)
	at reactor.ipc.netty.channel.ChannelOperations.onHandlerTerminate(ChannelOperations.java:425)
	at reactor.ipc.netty.channel.ChannelOperationsHandler.channelInactive(ChannelOperationsHandler.java:105)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:245)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:231)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelInactive(AbstractChannelHandlerContext.java:224)
	at io.netty.channel.ChannelInboundHandlerAdapter.channelInactive(ChannelInboundHandlerAdapter.java:75)
	at io.netty.handler.codec.http.HttpContentDecoder.channelInactive(HttpContentDecoder.java:205)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:245)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:231)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelInactive(AbstractChannelHandlerContext.java:224)
	at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.fireChannelInactive(CombinedChannelDuplexHandler.java:420)
	at io.netty.handler.codec.ByteToMessageDecoder.channelInputClosed(ByteToMessageDecoder.java:377)
	at io.netty.handler.codec.ByteToMessageDecoder.channelInactive(ByteToMessageDecoder.java:342)
[...]

PR #3300, opened 2024-06-14
`Http2Pool`: when applying `acquireTimeout`, check for current pending `Borrowers`

violetagg:


reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -624,13 +624,23 @@ void pendingAcquireLimitReached(Borrower borrower, int maxPending) {
...
+		long estimateStreamsCount = totalMaxConcurrentStreams - acquired;
+		int permits = poolConfig.allocationStrategy().estimatePermitCount();
+		if (permits + estimateStreamsCount < postOffer) {
+			borrower.pendingAcquireStart = clock.millis();
+			if (!borrower.acquireTimeout.isZero()) {
+				borrower.timeoutTask = poolConfig.pendingAcquireTimer().apply(borrower, borrower.acquireTimeout);
+			}
+		}
+
+		int maxPending = poolConfig.maxPending();
```
pderop, line 643: overall, it seems good to me.

I just wonder why the `maxPending` variable has been moved to line 643, it's probably for readability reasons.
then since the `maxPending` variable is only used if `WIP.getAndIncrement(this) == 0`, then why not declaring this variable in the if block ? or am I missing the point ? something like:

```
		if (WIP.getAndIncrement(this) == 0) {
			int maxPending = poolConfig.maxPending();
            ...
```

violetagg: I moved it to the `if`, thanks

pderop (approved):
There is a pending discussion on the [reactor-pool PR #221](https://github.com/reactor/reactor-pool/pull/221#discussion_r1639402993) which may be related to this PR, so maybe more commits will come here, I don't know.

at this stage, this PR is looking good to me.
approving.

violetagg:
@pderop I think I applied everything from this discussion, no?

pderop:
yes, I think so.

pderop (approved):
to clarify, re-approving this PR.

violetagg:
@pderop Thanks

PR #3311, opened 2024-06-24
Add `HTTP/3` client initial settings

violetagg:
Related to #1531


pderop:
In `Http3ChannelInitializer`, I see that the `idleTimeout`attribute is not used anymore, because the quick server codec builder maxIdleTimeout is now initialized from `http3Settings.idleTimeout()`. 

So, is it necessary to keep this attribute ?

violetagg:

PR #3321, opened 2024-06-27
When an explicit `EventLoopGroup` is configured ensure only one connection pool is created

### LINKED ISSUE #3316: Client sockets are not re-used when explicit EventLoopGroup is passed
- TcpClient provides 2 methods to setup connections on eventloops: runOn(EvenLoopGroup) & runOn(LoopResources). [Code](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/tcp/TcpClient.java#L307)
 - LoopResources is a FunctionalInterface and it is created with a lambda: [Code](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/transport/Transport.java#L235)
 - If we pass eventLoopGroup to TcpClient, a new instance of LoopResouce is created with each call from the lambda. 
 - Reactor ClientTransport uses a HashMap with LoopResources in key to cache DNSResolver which creates a unique client socket: [Code](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransportConfig.java#L248)
 - Since LoopResources changes with every call, the code creates a new dnsResolver leading to a new client socket with each call.

## Expected Behavior

Client sockets should limit to EventLoopGroup size. 
## Actual Behavior

Client socket count keeps increasing with each connectNow() call.
## Steps to Reproduce

```java
@Test
public void repoCase() {
	LoopResources loopResources = LoopResources.create("test", -1, 80, true, false);
	for(int i=0;i<=400;i++) {
        TcpClient.create()
            .remoteAddress(
                () -> new DomainSocketAddress("/tmp/test.sock"))
            .runOn(loopResources)                     // Works
            // .runOn(loopResources.onClient(true))       // Does not work
            .connectNow();
            Thread.sleep(1000);
    }
}
```
Count sockets using: `ss -lunp | grep java | wc -l `. This number increases when eventLoopGroup is used.
## Possible Solution

 - Hash function works with eventLoopGroup(instead of loopResources) in cases eventloopGroup is passed to TcpClient
## Your Environment

* Reactor version(s) used: v1.1.17

violetagg:
Fixes #3316

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/Transport.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/Transport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2020-2024 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


pderop (approved):

violetagg:
@pderop Thanks

PR #3328, opened 2024-06-28
Add apidocs classifier for publishing to docs.spring.io

pderop:
in docs.spring.io, there is an `autorepo` script which automates the deployment to `docs.spring.io` of any docs zip deployed to artifactory.

the convention is the following:

1) any docs zip published to artifactory should contain `zip.type=docs`, and `zip.deployed=false` special artifactory properties. see [spring framework](https://github.com/spring-projects/spring-framework/blob/8ef74dfdad1f7c55757ed52599358815e2756c98/.github/workflows/build-and-deploy-snapshot.yml#L52) example, or [spring pulsar](https://github.com/spring-projects/spring-pulsar/blob/main/buildSrc/src/main/groovy/io/spring/gradle/convention/ArtifactoryPlugin.groovy#L51) example.

2) the structure of the docs zip should be like this:

- api/* -> contains javadoc
- reference/html/ -> contains reference flat html files
- reference/pdf/ -> contains reference pdf file (if any)

Now, our current reactor-netty-docs zip contains only the html files and the structure is different (`docs/*`).
We can't change the layout of the reactor-netty-docs zip, else it could break any other components, like the projectreactor.io site, which expects reference documentation to be located in docs/* sub directory of the docs zip.

So, this PR just adds a new reactor-netty `apidocs` classifier which contains:

- api/* -> javadoc
- reference/html/* -> reference html doc files
- reference/pdf/ -> pdf reference doc, if any

the new apidocs artifact will be published with the special properties `zip.type=docs` and `zip.deployed=false`, so the content of the apidocs artifact will also be copied to:

- `https://docs.spring.io/projectreactor/netty/docs/VERSION/api/` (javadoc)
- `https://docs.spring.io/projectreactor/netty/docs/VERSION/reference/html/` (reference doc)
- `https://docs.spring.io/projectreactor/netty/docs/VERSION/reference/pdf/` (pdf reference doc)

not that the autorepo script also uses [autoln](https://github.com/spring-io/autoln), which will maintain symbolic links like:

- `https://docs.spring.io/projectreactor/netty/docs/current ->` (the latest stable docs)
- `https://docs.spring.io/projectreactor/netty/docs/current-SNAPSHOT ->` (the most recent snapshot docs)
- `https://docs.spring.io/projectreactor/netty/docs/<major>.<minor>.x ->` (the most recently released version starting with `<major>.<minor>`)
- `https://docs.spring.io/projectreactor/netty/docs/<major>.<minor>.x-SNAPSHOT ->` (the most recent snapshot docs starting with `<major>.<minor>`)

### DIFF
--- a/reactor-netty/build.gradle
+++ b/reactor-netty/build.gradle
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2023 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2020-2024 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


reactor-netty/build.gradle
```diff
@@ -42,4 +42,32 @@ javadoc.dependsOn(aggregateJavadoc)
...
+		exclude "**/index.pdf"
+		into("reference/html/")
+	}
+	from(aggregateJavadoc) {
+		into "api"
+	}
+}
+
+//add apidocs.zip to the publication
+publishing.publications.mavenJava.artifact(apidocsZip)
```
violetagg, line 71: Does this means that we will publish this artifact to Maven Central?

pderop: yes, if this is an obstructive issue, is there a way to publish this new artifact to artifactory only ?
or alternatively, maybe there is another way: use rsync in order to push the javadocs/reference/pdf directly to docs.spring.io (then I'll need to explore that approach ...)

pderop: mmm ... this is interesting: in spring-security, it seems that the docs are deployed in docs.spring.io by doing some ssh commands in groovy code, see this [deployDocs](https://github.com/spring-projects/spring-security/blob/main/buildSrc/src/main/groovy/io/spring/gradle/convention/DeployDocsPlugin.groovy#L54-L80) groovy task in spring-security. maybe we should do the same.

I need to explore this; putting this PR in draft mode for the moment.

pderop: I could experiment this spring-security ssh commands, I see how it works and could test it using a local ssh server container.
I will close this current PR (because I don't want to loose it), and will create a new one based on the same approach used by spring-security.

pderop:
Closing this PR, and will create another PR which is based on the spring-security [ssh commands](https://github.com/spring-projects/spring-security/blob/main/buildSrc/src/main/groovy/io/spring/gradle/convention/DeployDocsPlugin.groovy#L54-L80) approach.

PR #3349, opened 2024-07-15
Polish Javadoc for HttpClient.compress()

izeye:
This PR polishes Javadoc for the `HttpClient.compress()`.

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -520,6 +520,8 @@ public abstract class HttpClient extends ClientTransport<HttpClient, HttpClientC
 	 *
 	 * <p>Note: Brotli and zstd compressions require additional dependencies.
 	 *
+	 * <p>Note: For zstd compression, {@literal Accept-Encoding: zstd} header needs to be added explicitly.
+	 *
 	 * @param compressionEnabled if true, compression (gzip, Brotli, and zstd) is enabled, otherwise disabled (default: false)
 	 * @return a new {@link HttpClient}
 	 */
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
@@ -516,9 +516,11 @@ public abstract class HttpClient extends ClientTransport<HttpClient, HttpClientC
 	}
 
 	/**
-	 * Specifies whether GZip compression is enabled.
+	 * Specifies whether compression (gzip, Brotli, and zstd) is enabled.
 	 *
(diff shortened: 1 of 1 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -516,9 +516,11 @@ public final HttpClient baseUrl(String baseUrl) {
 	}
 
 	/**
-	 * Specifies whether GZip compression is enabled.
+	 * Specifies whether compression (gzip, Brotli, and zstd) is enabled.
```
violetagg, line 519: Reactor Netty does not support `zstd` out of the box. One needs to add the header by oneself.

```suggestion
	 * Specifies whether compression (gzip and Brotli) is enabled.
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -516,9 +516,11 @@ public final HttpClient baseUrl(String baseUrl) {
...
 	 *
-	 * @param compressionEnabled if true GZip compression is enabled otherwise disabled (default: false)
+	 * <p>Note: Brotli and zstd compressions require additional dependencies.
```
violetagg, line 521:
```suggestion
	 * <p>Note: Brotli compression requires additional dependency.
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -516,9 +516,11 @@ public final HttpClient baseUrl(String baseUrl) {
...
+	 *
+	 * @param compressionEnabled if true, compression (gzip, Brotli, and zstd) is enabled, otherwise disabled (default: false)
```
violetagg, line 523:
```suggestion
	 * @param compressionEnabled if true, compression (gzip and Brotli) is enabled, otherwise disabled (default: false)
```

violetagg (changes requested):
@izeye Thanks for the PR!

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
izeye: @violetagg Thanks for the feedback!

I wanted to let users who want to use zstd encoding aware of this method as its invocation is required to use zstd encoding.

Is it okay if I add another note like the following?

```
<p>Note: For zstd compression, {@literal Accept-Encoding: zstd} header needs to be added explicitly.
```

violetagg: ok

izeye: @violetagg Thanks! I added the note in b91b30bb9adfe4c2557818a2bc13a886cd20ef23.

violetagg (approved):

Issue #1433, opened 2020-12-15
Event loop task duration tracking


osi:
Similar to #1398, it would be nice to be able to track the duration of event loop tasks. 

Reactor is able to do this for its schedulers via `Schedulers.enableMetrics`.

Additionally, it would be nice to set a threshold for the duration of any given task, and be able to log an event with details about that task (for identifying specific instances in the metrics)

kevinten10:
Full monitoring may cause redundant information and excessive performance loss.

Is it more reasonable to perform sampling of this part of the function?

like: `sample(() -> Schedulers.enableMetrics,
       (Predicate) () -> { /* Trigger threshold */ });`

osi:
Sampling may be fine. My goal is to identify work taking place on an event loop thread that is effectively blocking the loop due to how long it takes. (even if the work is fully non-blocking).

mateusz-nalepa:
Hey @violetagg !
I was thinking about adding something like this to the class `SingleThreadEventExecutor`

```java
private void execute(Runnable task, boolean immediate) {
     MeteredRunnable meteredTask = new MeteredRunnable(task, threadProperties.name());

...
class MeteredRunnable implements Runnable {

    private final Runnable task;
    private final String executorName;
    private final Long start;

    MeteredRunnable(Runnable task, String executorName) {
        this.task = task;
        this.executorName = executorName;
        this.start = System.nanoTime();
    }

    @Override
    public void run() {
        Duration duration = Duration.ofNanos(System.nanoTime() - start);
        System.out.println("Runnable took: " + duration + " for executor name: " + executorName);
        task.run();
    }
}
```

However, it turns out that class `SingleThreadEventExecutor` is located in https://github.com/netty/netty that has no access to Micrometer.
So I thought about some lambda in class `DefaultLoopResources` which could be passed down to `SingleThreadEventExecutor`, but I'm not sure about this approach :/ 

I'm open to your recommendations on how to add this metric! Right now, I'm out of ideas :D

violetagg:
@mateusz-nalepa I haven't spent time for researching this and how to be implemented so that we have both metrics and the performance is still intact.
There are some experiments, check linked issues here, also https://github.com/netty/netty/pull/11293, https://github.com/netty/netty/pull/9080

mateusz-nalepa:
Hi @violetagg 

I've failed with preparing this metric, but I was able to prepare workaround for `part of it`, I mean:

[...]

PR #3425, opened 2024-09-10
Enhance Task Monitoring by Printing Stack Traces for Blocking Tasks

qnnn:
Hello, I’m suggesting an enhancement to add blocking task detection. When the number of blocking tasks exceeds a specified threshold, it would trigger the printing of worker thread stack traces. This would enable faster identification and troubleshooting of the issues.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/LoopResources.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/LoopResources.java
@@ -35,6 +35,13 @@ import reactor.netty.ReactorNetty;
 public interface LoopResources extends Disposable {
 
 	/**
+	 * Default pending task log threshold, fallback to -1 (no logging).
+	 */
+	int DEFAULT_PENDING_TASK_LOG_THRESHOLD = Integer.parseInt(System.getProperty(
+			ReactorNetty.PENDING_TASK_LOG_THRESHOLD,
+			"" + -1));
+
+	/**
 	 * Default worker thread count, fallback to available processor
 	 * (but with a minimum value of 4).
 	 */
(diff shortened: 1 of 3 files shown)


reactor-netty-core/src/main/java/reactor/netty/resources/LoopResources.java
```diff
@@ -43,11 +50,12 @@ public interface LoopResources extends Disposable {
...
 
 	/**
-	 * Default selector thread count, fallback to -1 (no selector thread).
+	 * Default selector thread count, fallback to -1 (no selector thread),
+	 * but fallback to 1 if pending task logging is enabled.
 	 */
 	int DEFAULT_IO_SELECT_COUNT = Integer.parseInt(System.getProperty(
 			ReactorNetty.IO_SELECT_COUNT,
-			"" + -1));
+			DEFAULT_PENDING_TASK_LOG_THRESHOLD > -1 ? "" + 1 : "" + -1));
```
qnnn, line 58: If SelectLoop is not enabled, the number of pending tasks may be distorted (e.g., when sleeping in a code block).

violetagg:
@qnnn So you are interested in tasks that are blocking the event loop? Did you consider using [Blockhound](https://github.com/reactor/blockhound)?

qnnn:
> @qnnn So you are interested in tasks that are blocking the event loop? Did you consider using [Blockhound](https://github.com/reactor/blockhound)?

@violetagg I'm very sorry I just now noticed the Blockhound project; it looks really impressive. Thank you for your response, I will try to use it right away!

qnnn:
@violetagg Hello, I recently tried using the Blockhound project, which covers almost all possible blocking scenarios. It is very flexible and has better compatibility. I am really looking forward to a version that is compatible with Synchronized, but I have some concerns. Currently, many concurrent collections use Synchronized or ReentrantLock, and in concurrent situations, they can also park, triggering Blocking Detect. However, using them seems reasonable?

I reviewed the PR I submitted at that time; it looks rough and is very limited in use (it needs to trigger the detection based on step length (StepMeterRegistry Impl)). However, in a sense, when there are many pending tasks, does it also indicate a high load? It’s somewhat similar to CPU load; it seems like an effective monitoring implementation. If this makes sense, I can try to resubmit this PR to allow it to be detected by SelectLoop instead of being based on StepMeterRegistry Impl.

violetagg:
> However, in a sense, when there are many pending tasks, does it also indicate a high load?

yes it may indicate a high load

Are you looking for something like this 
https://github.com/reactor/reactor-netty/issues/1433
https://github.com/netty/netty/issues/12305

qnnn:
@violetagg Yes, this is exactly what I was looking for! There are a lot of useful points and cases mentioned here. I will take a closer look and try them out soon. Thank you for your guidance!

In my initial thoughts, pendingTasks is a very lightweight yet important metric. When its value is relatively high, I pay close attention to the current execution stack (as there might be some blocking or long-running tasks). However, monitoring it requires fetching data based on StepMeterRegistry implementation, since under high load, it's difficult to retrieve these metrics through the interface. Also, checking its stack manually at this point may not be timely enough.

PR #3436, opened 2024-09-18
Parse X-Forwarded-Prefix request header

chemicL:
The X-Forwarded-Prefix can be obtained via `HttpServerRequest#forwardedPrefix()`.

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/ConnectionInfo.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/ConnectionInfo.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2023 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2018-2024 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -57,6 +57,8 @@ public final class ConnectionInfo {
 
 	final boolean isInetAddress;
 
+	final String forwardedPrefix;
+
 	static ConnectionInfo from(Channel channel, HttpRequest request, boolean secured, SocketAddress remoteAddress,
 			@Nullable BiFunction<ConnectionInfo, HttpRequest, ConnectionInfo> forwardedHeaderHandler) {
 		String hostName = DEFAULT_HOST_NAME;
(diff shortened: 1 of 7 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/server/ConnectionInfo.java
```diff
@@ -94,12 +96,18 @@ static ConnectionInfo from(Channel channel, HttpRequest request, boolean secured
 
 	ConnectionInfo(SocketAddress hostAddress, String hostName, int hostPort,
 			SocketAddress remoteAddress, String scheme, boolean isInetAddress) {
+		this(hostAddress, hostName, hostPort, remoteAddress, scheme, isInetAddress, "");
```
chemicL, line 99: I wonder if we'd actually like to provide the means to tell the difference between a missing header and the content being empty. If so, this could be actually `null` instead of empty by default. WDYT @violetagg ?

reactor-netty-http/src/main/java/reactor/netty/http/server/ConnectionInfo.java
```diff
@@ -94,12 +96,18 @@ static ConnectionInfo from(Channel channel, HttpRequest request, boolean secured
...
+	}
+
+	ConnectionInfo(SocketAddress hostAddress, String hostName, int hostPort,
```
chemicL, line 102: I added a new constructor, but I can absolutely go ahead and modify all the usages of the above constructor to add the last parameter and remove the above.

violetagg: I like it with a new constructor

reactor-netty-http/src/main/java/reactor/netty/http/server/ConnectionInfo.java
violetagg: yeah may be we should return `null` if there is no header or `forwarded=false`

violetagg (approved):
nice work

chemicL:
@violetagg I addressed the aspect of distinguishing missing header vs empty content and also added the handling of comma-separated components with removal of the trailing forward slashes in each component.

violetagg (approved):

chemicL:
@violetagg as discussed, I updated the PR to include validation whether the header value starts with a slash.

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpForwardedHeaderHandler.java
```diff
@@ -138,7 +138,11 @@ private static String parseForwardedPrefix(String prefixHeader) {
 			}
 			prefix.append((endIndex != rawPrefix.length() ? rawPrefix.substring(0, endIndex) : rawPrefix));
 		}
-		return prefix.toString();
+		String parsedPrefix = prefix.toString();
+		if (!parsedPrefix.isEmpty() && DEFAULT_FORWARDED_HEADER_VALIDATION && !parsedPrefix.startsWith("/")) {
```
violetagg, line 142: instead of `startsWith("/")` what about `tempUri.charAt(0) != '/'`

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/DefaultHttpForwardedHeaderHandler.java
chemicL: Yep, that's better. Applied the change.

violetagg (approved):

Issue #3416, opened 2024-08-31
The EpollSocketChannel object is too large to be reclaimed by the jvm


userlaojie:
Hello, we are revamping our system with spring-webflux. After the service was started in a Linux environment, it was found that the memory kept increasing, and the memory was never reclaimed by the jvm. After pulling the service dump file, we suspected that the webclient connection pool was cross-referenced, resulting in the EpollSocketChannel object not being reclaimed.
Please help to check whether there is any problem with webclient configuration, or you can check it from other aspects. Thank you.

This is MAT after analyzing a single object over 80M.
![image](https://github.com/user-attachments/assets/42b8c39f-055c-4898-a8fe-ed9fac3c185f)
![image](https://github.com/user-attachments/assets/ae85ece0-c5de-4dcb-9407-a6e67b3c3721)

This is the jvm memory monitoring usage.
![image](https://github.com/user-attachments/assets/d4a196b9-f217-48fe-bb69-c79e69c836c4)

## Steps to Reproduce

The webclient configuration parameters are as follows:
```java
@Slf4j
@Configuration
public class WebClientConfig {
    @Data
    @ConfigurationProperties(prefix = "business.webclient")
    @Configuration
    static class WebClientConnectionConfig {
        private int pendingAcquireTimeout = 50;
        private int maxConnections = 32;
        private int pendingAcquireMaxCount = 1000;
        private long maxIdleTime = 10000;
        private long maxLifeTime = -1;
        private int connectTimeout = 2000;
        private long responseTimeout = 3000;
        private long writeTimeout = 10000;
        private long evictionIntervalTime = 120000;

        @Override
        public String toString() {
            return "WebClientConnectionConfig{" +
                    "pendingAcquireTimeout=" + pendingAcquireTimeout +
                    ", maxConnections=" + maxConnections +
                    ", pendingAcquireMaxCount=" + pendingAcquireMaxCount +
                    ", maxIdleTime=" + maxIdleTime +
                    ", maxLifeTime=" + maxLifeTime +
                    ", connectTimeout=" + connectTimeout +
                    ", responseTimeout=" + responseTimeout +
                    ", writeTimeout=" + writeTimeout +
                    ", evictionIntervalTime=" + evictionIntervalTime +
                    '}';
        }
    }

    @Bean
    public HttpClient httpClient(@Qualifier("webClientConfig.WebClientConnectionConfig")final WebClientConnectionConfig config) throws SSLException {
[...]

Issue #3367, opened 2024-07-24
Possible memory leak when using pooled connections


vitjouda:
I am facing periodic OOM issues with my service, which uses `reactor-netty` based `WebClient` for HTTP communication with other REST services. I am not sure this is 100% reactor-netty related, but the only thing that managed to mitigate this issue was switching to another `ClientHttpConnector`. Therefore I assume this is the correct place to start asking for help :).

When I inspect the memory dump of my service, I can see that most of the growing memory is retained by instances of `io.netty.channel.epoll.EpollSocketChannel`. When I inspect the path to GC root of some random `EpollSocketChannel` instance, there is a long chain of Reactor operators retained. I can see that there are "blocks" of the processing pipeline that correspond to different and unrelated server requests chained together, presumably by the `HttpClientOperations` instance holding reference to `PooledConnection` / `EpollSocketChannel` that is now assigned to another request. I tried to visualize it in the attached MAT screenshot, but I know its hard to explain.

<img width="1185" alt="image" src="https://github.com/user-attachments/assets/6703cada-6092-4e31-a80d-3a7e0234e583">

In the image, each color block represents a reference chain of `Channel ` to `HttpClientOperations`. Each block belongs to distinct server request processing pipeline, as shown in this diagram. There are hundreds of concurrent requests at any given time.

![Diagram](https://github.com/user-attachments/assets/70f7a788-f9ae-4fe4-ad1c-0e616f774a7c)

During the lifetime of the service, these chains get longer and retain more and more memory until OOM.

## Steps to Reproduce
Unfortunately I can´t provide reproducer because I never managed to simulate this behavior locally (and I tried for days) and I cannot share the actual memory dump. If I can provide any information that would help, please tell me.

## Your Environment

* Reactor version(s) used: reactor-core:3.6.7, reactor-netty:1.1.20, reactor-pool:1.0.6
* JVM version: 21
* OS and version: Ubuntu 22.04.3

violetagg:
@vitjouda with so little information we can't start any investigation.
- Is it HTTP/1.1 or HTTP/2?
- Can you share your HttpClient/WebClient configuration and the usage?

vitjouda:
I understand, I will provide everything I can.

- HTTP 1.1

Usage (adjusted so that properties are hardcoded with real world values)

```java
    @Bean
    @Scope("prototype")
    public WebClient.Builder webClientBuilder(ObjectProvider<WebClientCustomizer> customizerProvider) {
[...]

PR #3459, opened 2024-10-07
When terminating detach the connection from request/response objects

violetagg:
Related to #3416, #3367

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
@@ -28,6 +28,7 @@ import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.AbstractChannel;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelConfig;
+import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelMetadata;
@@ -703,6 +704,11 @@ public class ChannelOperations<INBOUND extends NettyInbound, OUTBOUND extends Ne
 			this.config = new DefaultChannelConfig(this);
 			this.localAddress = copy.localAddress();
 			this.remoteAddress = copy.remoteAddress();
+		}
+
+		@Override
+		public ChannelFuture closeFuture() {
+			return newSucceededFuture();
 		}
 
 		@Override
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
@@ -25,9 +25,16 @@ import java.util.function.Predicate;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
+import io.netty.channel.AbstractChannel;
 import io.netty.channel.Channel;
+import io.netty.channel.ChannelConfig;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelMetadata;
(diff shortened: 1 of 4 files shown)


violetagg:
@chemicL wait with the review - some of the tests are failing ...

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
```diff
@@ -681,4 +693,111 @@ static OnSetup empty() {
...
+				promise.setFailure(new UnsupportedOperationException());
+			}
+		}
+	}
+
+	static final class DisposedConnection implements Connection {
+
+		final Channel channel;
+
+		DisposedConnection(Channel copy) {
```
chemicL, line 794: I struggle with the name a bit. It's not really a copy as the name would imply. The behaviour you expect is to copy so you're treating a name as a verb, which is confusing. Am I getting it wrong? Perhaps `Channel actual` would be indicative that we're working with something that wraps the actual `Channel`. Then the same comment would apply to the argument of `DisposedChannel`'s constructor.

reactor-netty-http-brave/src/main/java/reactor/netty/http/brave/TracingHttpServerDecorator.java
```diff
@@ -266,27 +266,25 @@ static final class TracingMapHandle implements BiFunction<Mono<Void>, Connection
...
-			                               else {
-			                                   eventLoop.execute(() -> handler.handleSend(braveResponse, localSpan));
-			                               }
-			                       });
+			                       if (eventLoop.inEventLoop()) {
+			                           handler.handleSend(braveResponse, localSpan);
+			                       }
+			                       else {
+			                           eventLoop.execute(() -> handler.handleSend(braveResponse, localSpan));
+			                       }
```
chemicL, line 287: Took me a while to understand what's going on here, but I understand that at this point the `response.withConnection `lambda can potentially have the `DisposedConnection` which means there's no event loop associated any more so you are essentially capturing it in the closure through the above reference, correct?

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -3401,4 +3403,46 @@ static void testIssue3285SendRequest(HttpClient client, @Nullable Class<? extend
...
+		                            boolean result = connWeakRef.get().get() == null;
+		                            if (!result) {
+		                                System.gc();
+		                            }
+		                            return result;
+		                        })
+		                        .switchIfEmpty(Mono.error(new RuntimeException("failed"))
+		                        .flatMap(l -> Mono.empty())))
+		                .then()
+		                .contextWrite(Context.of(res.getClass(), res)))
```
chemicL, line 3443: I would love to see a comment that this serves the purpose of addressing another concern. While the main test flow validates that the connection is freed once response is retrieved as `res` reference, the `contextWrite` store ensures also that the res in `Context` also doesn't preserve the `Connection` reference. Am I understanding it right?

chemicL (approved):
Looks good, I just left some questions and requests for comments :)

reactor-netty-http-brave/src/main/java/reactor/netty/http/brave/TracingHttpServerDecorator.java
violetagg: correct

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
violetagg: correct

reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java
violetagg: change applied

chemicL (approved):
Thanks for the update! 🚢

violetagg:

PR #3481, opened 2024-10-23
Ensure body is sent when `DELETE` and `Transfer-Encoding`

violetagg:
According to the spec https://www.rfc-editor.org/rfc/rfc9110.html#name-delete

```
A client SHOULD NOT generate content in a DELETE request unless it is made
directly to an origin server that has previously indicated, in or out of band, that
such a request has a purpose and will be adequately supported.
```

When a user sends `DELETE` with content this means there is an indication that the server supports this type of requests.


violetagg:
@bachah Please open a new issue in order to track this regression. Thanks

PR #3514, opened 2024-11-18
Add timeout for `DefaultLoopResources#disposeLater`

### LINKED ISSUE #3509: CPU intesive work prevents LoopResources.disposeLater from running
Please see https://github.com/spring-projects/spring-boot/issues/21426 for background.

The following sample from @wilkinsona reproduces the issue:

```java
@Test
public void testGracefulShutdown() throws Exception {
	CountDownLatch latch1 = new CountDownLatch(2);
	CountDownLatch latch2 = new CountDownLatch(2);
	LoopResources loop = LoopResources.create("testGracefulShutdown");
	this.disposableServer = HttpServer.create().port(0).runOn(loop).doOnConnection(c -> {
		c.onDispose().subscribe(null, null, latch2::countDown);
		latch1.countDown();
	})
		// Register a channel group, when invoking disposeNow()
		// the implementation will wait for the active requests to finish
		.channelGroup(new DefaultChannelGroup(new DefaultEventExecutor()))
		.route(r -> r
			.get("/delay500",
					(req, res) -> res.sendString(Mono.just("delay500").delayElement(Duration.ofMillis(500))))
			.get("/cpuIntensive", (req, res) -> {
				// Simulate some long-running CPU-intensive work
				boolean stop = false;
				while (!stop) {
				}
				return res.sendString(Mono.just("cpuIntensive"));
			}))
		.wiretap(true)
		.bindNow(Duration.ofSeconds(30));

	HttpClient client = HttpClient.create().remoteAddress(this.disposableServer::address).wiretap(true);

	MonoProcessor<String> result = MonoProcessor.create();
	Flux.just("/delay500", "/cpuIntensive")
		.flatMap(s -> client.get().uri(s).responseContent().aggregate().asString())
		.collect(Collectors.joining())
		.subscribe(result);

	assertThat(latch1.await(30, TimeUnit.SECONDS)).isTrue();

	// Stop accepting incoming requests, wait at most 3s for the active requests to
	// finish
	try {
		this.disposableServer.disposeNow();
	}
	catch (IllegalStateException ex) {
		System.out.println(ex.getMessage());
		// The socket couldn't be stopped, continue with shutdown
	}

	System.out.println("Disposing of the loop resources");
	loop.disposeLater().block();
	System.out.println("Disposal complete");

	assertThat(latch2.await(30, TimeUnit.SECONDS)).isTrue();
}
```

violetagg:
When shutting down `EventLoopGroup`, `Netty` tries to add for an execution a task that will stop `EventLoops`. However if there is a CPU intensive task running on `EventLoop`, the execution of the shutting down task might never happen.

This change adds a timeout for `DefaultLoopResources#disposeLater` so that the user will be notified that some `EventLoops` cannot be stopped.

Note: Having blocking and CPU intensive tasks on `EventLoops` is strongly discouraged and can lead to unpredictable failures.

Fixes #3509

### DIFF
--- a/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
@@ -1894,6 +1894,67 @@ class HttpServerTests extends BaseHttpTest {
 	}
 
 	@Test
+	public void testGracefulShutdownIssue3509() throws Exception {
+		CountDownLatch latch1 = new CountDownLatch(2);
+		CountDownLatch latch2 = new CountDownLatch(1);
+		CountDownLatch latch3 = new CountDownLatch(1);
+		LoopResources loop = LoopResources.create("testGracefulShutdownIssue3509");
+		this.disposableServer =
+				createServer().runOn(loop)
+				              .doOnConnection(c -> {
+				                  c.onDispose().subscribe(null, null, latch2::countDown);
+				                  latch1.countDown();
+				              })
+				              // Register a channel group, when invoking disposeNow()
+				              // the implementation will wait for the active requests to finish
+				              .channelGroup(new DefaultChannelGroup(new DefaultEventExecutor()))
+				              .route(r ->
+				                  r.get("/delay500", (req, res) -> res.sendString(Mono.just("delay500").delayElement(Duration.ofMillis(500))))
+				                   .get("/cpuIntensive", (req, res) -> {
+				                       // Simulate some long-running CPU-intensive work
+				                       boolean stop = false;
+				                       while (!stop) {
+				                       }
+				                       return res.sendString(Mono.just("cpuIntensive"));
+				                   }))
+				              .bindNow(Duration.ofSeconds(30));
+
+		HttpClient client = createClient(this.disposableServer::address);
+
+		AtomicReference<String> result = new AtomicReference<>();
+		Flux.just("/delay500", "/cpuIntensive")
+		    .flatMap(s -> client.get().uri(s).responseContent().aggregate().asString())
+		    .subscribe(s -> {
+		        result.set(s);
+		        latch3.countDown();
+		    });
+
+		assertThat(latch1.await(30, TimeUnit.SECONDS)).isTrue();
+
+		// Stop accepting incoming requests, wait at most 1s for the active requests to finish
+		try {
+			// Reduce the timeout for testing purposes, by default it is 3s
+			this.disposableServer.disposeNow(Duration.ofSeconds(1));
+			fail("Expectation is that the socket cannot be closed");
+		}
+		catch (IllegalStateException e) {
+			// The socket couldn't be stopped, continue with shutdown
+			this.disposableServer = null;
+		}
+
+		// Reduce the quiet period and the timeout for testing purposes
+		// by default the quiet period is 2s and the timeout is 15s
+		loop.disposeLater(Duration.ofMillis(100), Duration.ofSeconds(1))
+		    .as(StepVerifier::create)
+		    .expectError(IllegalStateException.class)
+		    .verify(Duration.ofSeconds(5));
+
+		assertThat(latch2.await(30, TimeUnit.SECONDS)).isTrue();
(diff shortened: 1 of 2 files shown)


reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -1893,6 +1893,68 @@ public void channelRead(@NotNull ChannelHandlerContext ctx, @NotNull Object msg)
...
+				createServer().runOn(loop)
+				              .doOnConnection(c -> {
+				                  c.onDispose().subscribe(null, null, latch2::countDown);
+				                  latch1.countDown();
+				              })
+				              // Register a channel group, when invoking disposeNow()
+				              // the implementation will wait for the active requests to finish
+				              .channelGroup(new DefaultChannelGroup(new DefaultEventExecutor()))
+				              .route(r ->
+				                  r.get("/delay500", (req, res) -> res.sendString(Mono.just("delay500").delayElement(Duration.ofMillis(500))))
```
chemicL, line 1912: Do you think using a `Mono.fromCallable` with a latch in the lambda or use `VirtualTimeScheduler` with `StepVerifier` to emulate the passage of time could speed up the runtime of this test? As I imagine, this test will execute always above 500ms due to this delay and if tests like this add up they can slow down the build significantly.

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -1893,6 +1893,68 @@ public void channelRead(@NotNull ChannelHandlerContext ctx, @NotNull Object msg)
 				.isEqualTo("delay500delay1000");
 	}
 
+	@Test
+	public void testGracefulShutdownIssue3509() throws Exception {
+		CountDownLatch latch1 = new CountDownLatch(2);
+		CountDownLatch latch2 = new CountDownLatch(1);
+		CountDownLatch latch3 = new CountDownLatch(1);
```
chemicL, line 1900: Consider naming the latches according to their function:
`latch1` -> `connectedLatch`
`latch2` -> `disposeSubscribedLatch`
`latch3` -> `responseReceivedLatch`

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -1893,6 +1893,68 @@ public void channelRead(@NotNull ChannelHandlerContext ctx, @NotNull Object msg)
...
+				              })
+				              // Register a channel group, when invoking disposeNow()
+				              // the implementation will wait for the active requests to finish
+				              .channelGroup(new DefaultChannelGroup(new DefaultEventExecutor()))
+				              .route(r ->
+				                  r.get("/delay500", (req, res) -> res.sendString(Mono.just("delay500").delayElement(Duration.ofMillis(500))))
+				                   .get("/cpuIntensive", (req, res) -> {
+				                       // Simulate some long-running CPU-intensive work
+				                       boolean stop = false;
+				                       while (!stop) {
```
chemicL, line 1916: Where is this executed? Shouldn't there be some cleanup at the end of this test to make the CPU-intensive task to actually stop? Otherwise it can lead to resource exhaustion. Or is there some global cleanup happening not visible in the changeset?

chemicL (changes requested):
I have some comments regarding the tests. The production code LGTM.

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
violetagg: fixed

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
violetagg: fixed

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
violetagg: fixed

chemicL (approved):

violetagg:

PR #3517, opened 2024-11-26
Initialise lazy the `ProxyProvider` configuration

### LINKED ISSUE #3501: Unexpected Behavior: httpHeaders Callback Invoked for Every Request Instead of Only During Initial CONNECT in Netty HTTP Client with Proxy
When configuring a Netty HTTP client to use a proxy, I would have expected the headers callback specified in the `ProxyProvider` to be called only during the initial CONNECT request to the proxy server and that after the tunnel was created there would be no need to call the headers callback.  Is my understanding on the proxy correct or am I completely off base here?

## Expected Behavior

The `httpHeaders` method to be called only when creating the tunnel to the Proxy.

## Actual Behavior

The `httpHeaders` method is ran for every request the httpClient makes.

## Steps to Reproduce

Configure HTTP Client with a Proxy like this. When making requests observe that the `httpHeaders` method is ran for each request.

```
final var httpClient = HttpClient.create().proxy(proxy -> proxy.type(ProxyProvider.Proxy.HTTP).host(proxyUrl).port(proxyPort).httpHeaders(headers -> {
            System.out.println("in headers");
           //headers.add("Proxy-Authorization", "Negotiate " + getKerberosToken());
        }).build());
```

* Reactor version(s) used:  Reactor Netty 1.1.22

violetagg:
Configurations like `address`, `headers`, `password` are calculated just before sending the request. If a configuration is changed a new `connection pool` will be created.

Fixes #3501

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ClientTransport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2023 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2020-2024 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


chemicL:
Late LGTM 👍

violetagg:
> Late LGTM 👍

thanks a lot

PR #3521, opened 2024-11-27
Ensure all ConnectionProvider built-in metrics are disabled when a custom recorder is provided

### LINKED ISSUE #3519: Better align `reactor.netty.connection.provider.pending.connections.time` with other `reactor.netty.connection.provider` metrics
We are trying to avoid creating the default `reactor.netty.connection.provider.*` metrics. In doing so we end up facing two issues.

First, we use a custom `metrics` recorder on the `HttpClient` builder to record certain events. However, we also disable the `metrics` on the `ConnectionProvider` to avoid creating metrics there. However, the code within `PooledConnectionProvider` uses:

```
boolean metricsEnabled = poolFactory.metricsEnabled || config.metricsRecorder() != null;
```

As a result, the metrics are enabled since the `config.metricsRecorder()` is non-null.  I'm assuming this was done for a specific reason and is not worth changing.

We got around this issue by using a custom metric registrar that is no-op and avoids the default registration of metrics.  This works well, except that there are still 2 timers created within `DefaultPooledConnectionProvider` via `new MicrometerPoolMetricsRecorder` which ends up creating a success and failure timer under `reactor.netty.connection.provider.pending.connections.time`.

It would be good if somehow that also used the registrar to allow things to be ignored.

Either way, our hope is to get to a point where we can have custom registrars but still disable all metrics from being produced within Reactor Netty internally.

## Expected Behavior

Metrics should not be emitted if the builder is disabling metrics and/or using a custom registrar.

## Actual Behavior

Most metrics are properly not emitted with the exception of the `reactor.netty.connection.provider.pending.connections.time` timers that are created on a separate, but related path.

## Steps to Reproduce

```java
        @Test
	public void testConnectionProviderDisablesAllMetrics() {
		ConnectionProvider connectionProvider = ConnectionProvider.builder("foobar")
				// disable metrics to avoid producing reactor.netty.connection.* metrics
				// NOTE: we must also use a custom registrar to no-op since we also create a recorder below
				.metrics(false, () -> (poolName, id, remoteAddress, metrics) -> {
					// ignore
				}).build();

		HttpClient client = HttpClient.create(connectionProvider)
				// create a custom recorder to capture events and handle accordingly
				// NOTE: these are no-ops for testing purposes (in reality they are used)
				.metrics(true, () -> new ChannelMetricsRecorder() {
                    @Override public void recordDataReceived(SocketAddress remoteAddress, long bytes) { }
                    @Override public void recordDataSent(SocketAddress remoteAddress, long bytes) { }
                    @Override public void incrementErrorsCount(SocketAddress remoteAddress) { }
                    @Override public void recordTlsHandshakeTime(SocketAddress remoteAddress, Duration time, String status) { }
                    @Override public void recordConnectTime(SocketAddress remoteAddress, Duration time, String status) { }
                    @Override public void recordResolveAddressTime(SocketAddress remoteAddress, Duration time, String status) { }
                });

		// execute a request to create the pools
		Assert.assertEquals(200, client.get().uri("https://www.google.com/").response().block().status().code());

		// verify no reactor.netty.connection metrics
		// NOTE: this actually fails since the timers reactor.netty.connection.provider.pending.connections.time are created
		// DefaultPooledConnectionProvider -> PooledConnectionAllocator -> MicrometerPoolMetricsRecorder
		// this kicks off because metricsEnabled is true since the httpClient.metricsRecorder exists even though metricsEnabled is false
		// and it does not use the pool.metricsRegistrar so no way to ignore
		AtomicInteger count = new AtomicInteger();
		Metrics.REGISTRY.forEachMeter(meter -> {
			if (meter.getId().getName().startsWith("reactor.netty.connection")) {
				count.incrementAndGet();
				System.out.println("Incorrect meter: " + meter.getId());
			}
		});

		Assert.assertEquals(0, count.get());
	}
```

## Possible Solution

Honestly, not sure what is the best option here.  Ideally `metricsEnabled` being `false` would avoid all metrics. However, it appears that there was logic added explicitly for testing `httpClient.metricsRecorder` and likely for a good reason.  Further, the registrar is not really tied to that specific portion of code so not sure it'd make sense to lump into the registrar, even though the registrar is available at that point via the pool factory.

## Your Environment

* Reactor version(s) used: `reactor-netty-core:1.1.22`, `reactor-core:3.6.12`
* Other relevant libraries versions: `netty:4.1.115.Final`
```
openjdk version "21.0.2" 2024-01-16 LTS
OpenJDK Runtime Environment Zulu21.32+17-CA (build 21.0.2+13-LTS)
OpenJDK 64-Bit Server VM Zulu21.32+17-CA (build 21.0.2+13-LTS, mixed mode, sharing)
```
* OS and version (eg. `uname -a`): `Darwin 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000 arm64`

violetagg:
Fixes #3519

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -135,7 +135,7 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 				boolean metricsEnabled = poolFactory.metricsEnabled || config.metricsRecorder() != null;
 				String id = metricsEnabled ? poolKey.hashCode() + "" : null;
 
-				InstrumentedPool<T> newPool = metricsEnabled && Metrics.isMicrometerAvailable() ?
+				InstrumentedPool<T> newPool = metricsEnabled && poolFactory.registrar == null && Metrics.isMicrometerAvailable() ?
 						createPool(id, config, poolFactory, remoteAddress, resolverGroup) :
 						createPool(config, poolFactory, remoteAddress, resolverGroup);
 
--- a/reactor-netty-http/src/test/java/reactor/netty/resources/PooledConnectionProviderDefaultMetricsTest.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/resources/PooledConnectionProviderDefaultMetricsTest.java
@@ -19,6 +19,8 @@ import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Metrics;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import io.netty.handler.codec.http.HttpHeaderNames;
+import io.netty.handler.codec.http.HttpHeaderValues;
 import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import io.netty.handler.ssl.util.SelfSignedCertificate;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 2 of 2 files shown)


violetagg:

PR #3528, opened 2024-11-30
A draft for support h2 client ping #3301

### LINKED ISSUE #3301: HTTP/2 PING frame handling
HTTP/2 has a ping/pong [protocol](https://datatracker.ietf.org/doc/html/rfc7540#section-6.7), but I do not see how it can be customized via the reactor-netty API.

I only use HTTP/2 as a client so I do not know about the server side.

A feature that would be interesting is sending a PING frame at regular intervals (set by user), and if no PONG (PING ack) is received in a timely fashion (timeout set by user), then disconnect, and remove that connection from the pool.

ctlove0523:
Fixes #3301


violetagg:
@ctlove0523 The description says `draft` but the PR is not in draft mode. Is it ready for a review?

ctlove0523:
你好，你的邮件我已经收到，每天会在晚上查看邮件并回复。急事请打电话

PR #3530, opened 2024-12-03
Handle the incoming `LastHttpContent` with no content as if it is `EMPTY_LAST_CONTENT`

violetagg:
When the incoming `LastHttpContent` is `EMPTY_LAST_CONTENT`, Reactor Netty directly completes the inbound. Handle the incoming `LastHttpContent` with no content in the same way instead of buffering it in `FluxReceive`.

Adapt the tests - instead of counting the received content on the server, directly count the outbound data on the client.


chemicL (approved):

violetagg:

Issue #3538, opened 2024-12-09
HttpServerRequest::receiveContent() never emits any value nor completes when HTTP/1.1 TLS Upgrade (RFC-2817) kicks in


reta:
We have run into the issue with latest Apache HttpClient5 5.4.1 / Apache HttpCore5 5.3.1 as a client (Reactor Netty as the HTTP server with HTTP/1.1 and H2C protocols) that now tries HTTP/1.1 TLS Upgrade (RFC-2817) by default, see please [1] for context. The `HttpServerRequest::receiveContent()` never emits any value nor completes  since no next / completion callbacks are  called. 

I have traced this problem to `HttpServerOperations`, specifically to this block of code https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java#L812:

```
                      if (isFullHttpRequest) {
				FullHttpRequest request = (FullHttpRequest) msg;
				if (request.content().readableBytes() > 0) {
					super.onInboundNext(ctx, msg);
				}
				else {
					request.release();
				}
				if (isHttp2()) {
					//force auto read to enable more accurate close selection now inbound is done
					channel().config().setAutoRead(true);
					onInboundComplete();
				}
			}
```

In case of upgrade flow (which happens only for GET/OPTIONS/HEAD), the channel pipeline involves `HttpObjectAggregator` and emits the `FullHttpRequest` instance (precisely, `HttpObjectAggregator.AggregatedFullHttpRequest`) with empty content, which deviates from non-upgrade flow that emits `DefaultHttpRequest`, followed by last content. As the result, `onInboundNext` is not called, and because the upgrade did not switch protocols and stayed on HTTP/1.1, `onInboundComplete` is also not called.

[1] https://github.com/apache/httpcomponents-client/pull/542/files

## Expected Behavior

The  `HttpServerRequest::receiveContent()` should complete with empty result.

## Actual Behavior
The  `HttpServerRequest::receiveContent()` hangs.

## Steps to Reproduce
I have published a full reproducer here [2], it is very small but has pom.xml + client + server (easier to start with).

[2] https://github.com/reta/httpclient5-upgrade-reproducer

## Possible Solution
I assume that calling `onInboundComplete()` would be the right thing to do here, since there is no content and inbound request is processed.

## Your Environment

* Reactor version(s) used: 1.2.24
[...]

PR #3540, opened 2024-12-10
HttpServerRequest::receiveContent() never emits any value nor completes when HTTP/1.1 TLS Upgrade (RFC-2817) kicks in

reta:
We have run into the issue with latest Apache HttpClient5 5.4.1 / Apache HttpCore5 5.3.1 as a client (Reactor Netty as the HTTP server with HTTP/1.1 and H2C protocols) that now tries HTTP/1.1 TLS Upgrade (RFC-2817) by default, see please [1] for context. The HttpServerRequest::receiveContent() never emits any value nor completes since no next / completion callbacks are called.

Closes https://github.com/reactor/reactor-netty/issues/3538

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
@@ -816,6 +816,8 @@ class HttpServerOperations extends HttpOperations<HttpServerRequest, HttpServerR
 					request.release();
 					// HTTP/1.1 TLS Upgrade (RFC-2817) on empty requests (GET/HEAD/OPTIONS)
 					if (!isHttp2() && request.headers().contains(HttpHeaderNames.UPGRADE)) {
+						//force auto read to enable more accurate close selection now inbound is done
+						channel().config().setAutoRead(true);
 						onInboundComplete();
 					}
 				}
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
@@ -820,7 +820,7 @@ class HttpServerOperations extends HttpOperations<HttpServerRequest, HttpServerR
 					channel().config().setAutoRead(true);
 					onInboundComplete();
 				}
-				else if (!isHttp2() && request.headers().contains(HttpHeaderNames.UPGRADE)) {
+				else if (request.headers().contains(HttpHeaderNames.UPGRADE)) {
 					// HTTP/1.1 TLS Upgrade (RFC-2817) requests (GET/HEAD/OPTIONS) with empty / non-empty payload
 					stopReadTimeout();
 					//force auto read to enable more accurate close selection now inbound is done
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
@@ -826,6 +826,12 @@ class HttpServerOperations extends HttpOperations<HttpServerRequest, HttpServerR
 					channel().config().setAutoRead(true);
 					onInboundComplete();
 				}
+				else if (!isHttp2() && request.headers().contains(HttpHeaderNames.UPGRADE)) {
+					// HTTP/1.1 TLS Upgrade (RFC-2817) requests (GET/HEAD/OPTIONS) with empty / non-empty payload
+					// No need to call onInboundComplete(), it will be triggered by onInboundNext(...)
+					// since this is the last HTTP content
+					stopReadTimeout();
+				}
 			}
 		}
 		else if (msg instanceof LastHttpContent) {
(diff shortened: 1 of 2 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
```diff
@@ -814,6 +814,10 @@ else if (msg instanceof HttpRequest) {
 				}
 				else {
 					request.release();
+					// HTTP/1.1 TLS Upgrade (RFC-2817) on empty requests (GET/HEAD/OPTIONS)
+					if (!isHttp2() && request.headers().contains(HttpHeaderNames.UPGRADE)) {
+						onInboundComplete();
```
reta, line 819: I think we don't need to call `channel().config().setAutoRead(true);` for HTTP/1.1

violetagg: We always switch to auto-read=true between the different requests so I would prefer to have it here.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
```diff
@@ -814,6 +814,10 @@ else if (msg instanceof HttpRequest) {
 				}
 				else {
 					request.release();
+					// HTTP/1.1 TLS Upgrade (RFC-2817) on empty requests (GET/HEAD/OPTIONS)
```
violetagg, line 817: It is true that GET should be used without payload, however there are use cases when GET is actually used with payload. (I do not refer to this HTTP/1.1 TLS Upgrade, but in general)

What will happen if we have this request:

```
		createHttpClientForContextWithPort()
		        .protocol(HttpProtocol.HTTP11)
		        .headers(h ->
		            h.add(HttpHeaderNames.CONNECTION, HttpHeaderValues.UPGRADE)
		             .add(HttpHeaderNames.UPGRADE, "TLS/1.2"))
		        .request(HttpMethod.GET)
		        .send((req, res) -> res.sendString(Mono.just("testIssue3538")))
                ...
```

reta: 👍 , I will add test case for it, I would expect `onInboundNext()` call for sure, but `onInboundComplete()` is under the question

reta: Turned out to be more nuanced than that: 
 - by default, the request is rejected with 413
 - the default could be overridden with `.httpRequestDecoder(spec -> spec.h2cMaxContentLength(...))`

I have a test case for the latter but it fails with 

```
reactor.core.Exceptions$ReactiveException: reactor.netty.http.client.PrematureCloseException: Connection prematurely closed DURING response
```

Trying to figure out why.

reta: Resolved, @violetagg we should be all set, thank you!

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
```diff
@@ -814,12 +814,24 @@ else if (msg instanceof HttpRequest) {
...
 				if (isHttp2()) {
 					//force auto read to enable more accurate close selection now inbound is done
 					channel().config().setAutoRead(true);
 					onInboundComplete();
 				}
+				else if (!isHttp2() && request.headers().contains(HttpHeaderNames.UPGRADE)) {
+					// HTTP/1.1 TLS Upgrade (RFC-2817) requests (GET/HEAD/OPTIONS) with empty / non-empty payload
+					// No need to call onInboundComplete(), it will be triggered by onInboundNext(...)
+					// since this is the last HTTP content
+					stopReadTimeout();
```
violetagg, line 833: The current logic doesn't invoke `onInboundComplete` if the request is not HTTP/2, because the only other expected use case was `websocket`. For `websocket` use case we should not invoke `onInboundComplete`.

Now with this commit, we want to change the logic to handle also aggregated HTTP/1.1 requests. For them we need to guarantee that `onInboundComplete`, which marks end of request.
With your current test case you don't see the issue, because the test itself invokes `complete`.
If you remove from `EchoAction`

```
if (message instanceof LastHttpContent) {
    emitter.complete();
}
```

You will see that completion is never invoked. Reactor Netty should guarantee that completion is triggered and not the user's logic.

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
```diff
@@ -820,6 +820,13 @@ else if (msg instanceof HttpRequest) {
 					channel().config().setAutoRead(true);
 					onInboundComplete();
 				}
+				else if (!isHttp2() && request.headers().contains(HttpHeaderNames.UPGRADE)) {
```
violetagg, line 823:
```suggestion
				else if (request.headers().contains(HttpHeaderNames.UPGRADE)) {
```

violetagg (changes requested):
I think we are almost ready, just one small change.

violetagg (approved):
Thanks for the PR!

PR #3549, opened 2024-12-17
Do not add `Transfer-Encoding` for `DELETE` with no body

### LINKED ISSUE #3548: DELETE request without body uses transfer-encoding: chunked
With https://github.com/reactor/reactor-netty/pull/3481 (in v 1.1.24 and 1.2.0) the handling of DELETE requests in `HttpClientHandler` was changed. As mentioned by @bachah in https://github.com/reactor/reactor-netty/pull/3481#issuecomment-2547416357, the handler now always sets the HTTP request header `transfer-encoding: chunked` even if there is no payload (empty body). In previous versions that was not the case.

We are using SpringCloudGateway to route requests to a Payara web server. When upgrading the SpringBoot version of our gateway from 3.3.5 to 3.3.6, this pulled in reactor-netty-http v1.2.0 which added the changed behavior. Payara seems to treat incoming requests with `transfer-encoding: chunked` as having a body, and it forbids DELETE requests from sending a payload and therefore responds with status 400 Bad Request.

The Netty change broke our setup. We had to change the configuration of our Payara server as described [here](https://github.com/payara/Payara/issues/1545) to allow potential payloads for DELETE requests (even so in our use cases we never send a body).

We are fine now, but I guess this change may also affect others. Can the transfer-encoding header be omitted by default if a DELETE request does not contain a body?

violetagg:
This is a regression caused by #3481

Fixes #3548

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
@@ -576,8 +576,9 @@ class HttpClientConnect extends HttpClient {
 				ch.followRedirectPredicate(followRedirectPredicate);
 
 				if (!Objects.equals(method, HttpMethod.GET) &&
-							!Objects.equals(method, HttpMethod.HEAD) &&
-							!headers.contains(HttpHeaderNames.CONTENT_LENGTH)) {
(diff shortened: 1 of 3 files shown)


chemicL (approved):

violetagg:

Issue #3244, opened 2024-05-14
Support for changing the default compression level when enabling server response compression


juliojgd:
## Motivation

In the [server compression related documentation](https://projectreactor.io/docs/netty/release/reference/index.html#_compression) the only configurable parameters are _enabled_ (boolean) and minimum size for a payload to be compressed (int).

There is [no method in Reactor Netty HttpServer](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java) that allows to set the compression level (int) that is effectively available in [Netty's HttpContentCompressor class](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java) but it's not available in the default [SimpleCompressionHandler](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/server/SimpleCompressionHandler.java)

It cannot be set in the [Reactor Netty HttpServerResponse](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerResponse.java) either.

## Desired solution

To add a configurable property (maybe named "compression-level" or "gzip-compression-level") to allow changing the default compression level (6 - gzip) to any other compression level that the developer needs to set.

violetagg:
@juliojgd Are you interested in providing a PR? Which Reactor Netty version should enable this feature (1.1.x or 1.2.x)?

juliojgd:
@violetagg we are using Spring Boot 3.2.x so I guess we'd need 1.1.x. I'll try to provide a PR.

The source branch to checkout for the changes would be https://github.com/reactor/reactor-netty/tree/1.1.x , right?

Thanks in advance

violetagg:
> The source branch to checkout for the changes would be https://github.com/reactor/reactor-netty/tree/1.1.x , right?

yes

raccoonback:
@violetagg 
Hello,
I would like to ask if I can take ownership of this issue and work on resolving it.
Additionally, please let me know if there are any guidelines or considerations I should be aware of before starting.

violetagg:
@raccoonback Yes sure, go for it. Please follow the contribution guide https://github.com/reactor/.github/blob/main/CONTRIBUTING.md

blacelle:
I encountered an issue which can be adressed by this feature. Thanks!

[...]

PR #3565, opened 2024-12-28
Supports compression level

raccoonback:
## Motivation
Hello,

I have added a feature to support compression level configuration in Reactor Netty's `SimpleCompressionHandler`.

Feature Details:
- Compression methods such as gzip, snappy now allow specifying the compression level.
- For brotli, since it does not support compression level configuration, the existing default behavior remains unchanged.
- zstd does not support compression settings because the level that can be set may vary depending on native.

This enhancement is expected to enable users to better balance compression efficiency and performance.

Thanks.

## Issue
Fixs: https://github.com/reactor/reactor-netty/issues/3244


raccoonback:
It looks like there might be a better way, so I'll work on it again and ask for a review.

Issue #3561, opened 2024-12-23
Warning for log pattern used in PooledConnectionProvider


Doc94:
## Expected Behavior

The logger for pool status just print the log

## Actual Behavior

Currently exists cases where that log also print a warning for a bad patter.

## Steps to Reproduce

```java
@Test
public void repoCase() {

}
```

## Possible Solution

Fix the format in https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java

## Your Environment

* Reactor version(s) used: 2023.0.8
* Other relevant libraries versions (eg. `netty`, ...): Netty
* JVM version (`java -version`):
* OS and version (eg. `uname -a`):

Doc94:
Hi, the message was like.

```````
2024-12-22T12:51:26.846178Z reactor-http-nio-2 WARN found 4 argument placeholders, but provided 5 for pattern `[9eca61c3, L:/192.168.1.74:59917 - R:discord.com/162.159.138.232:443] {}, now: {} active connections, {} inactive connections and {} pending acquire requests.
```

raccoonback:
@violetagg 
Hello!
May I take charge of this issue and work on resolving it? 😊

violetagg:
> @violetagg Hello! May I take charge of this issue and work on resolving it? 😊

sure, thanks

PR #3571, opened 2024-12-30
Remove unnecessary conditional statements in log

raccoonback:
Fix: https://github.com/reactor/reactor-netty/issues/3561

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -365,12 +365,14 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 
 	protected static void logPoolState(Channel channel, InstrumentedPool<? extends Connection> pool, String msg, @Nullable Throwable t) {
 		InstrumentedPool.PoolMetrics metrics = pool.metrics();
-		log.debug(format(channel, "{}, now: {} active connections, {} inactive connections and {} pending acquire requests."),
+		log.debug(
+				format(channel, "{}, now: {} active connections, {} inactive connections and {} pending acquire requests."),
 				msg,
 				metrics.acquiredSize(),
 				metrics.idleSize(),
 				metrics.pendingAcquireSize(),
-				t == null ? "" : t);
+				t
+		);
 	}
 
 	final void scheduleInactivePoolsDisposal() {
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -55,6 +55,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.StringJoiner;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
@@ -364,15 +365,29 @@ public abstract class PooledConnectionProvider<T extends Connection> implements 
 	}
 
 	protected static void logPoolState(Channel channel, InstrumentedPool<? extends Connection> pool, String msg, @Nullable Throwable t) {
-		InstrumentedPool.PoolMetrics metrics = pool.metrics();
+		String logMsg = msg + ", " + stringifyPoolMetrics(pool);
+		if (t == null) {
+			log.debug(format(channel, logMsg));
+			return;
+		}
+
 		log.debug(
-				format(channel, "{}, now: {} active connections, {} inactive connections and {} pending acquire requests."),
-				msg,
-				metrics.acquiredSize(),
-				metrics.idleSize(),
-				metrics.pendingAcquireSize(),
-				t
+			format(channel, logMsg),
+			t
 		);
+	}
+
+	private static String stringifyPoolMetrics(InstrumentedPool<? extends Connection> pool) {
+		InstrumentedPool.PoolMetrics metrics = pool.metrics();
+		return new StringJoiner(" ")
+				.add("now:")
+				.add(String.valueOf(metrics.acquiredSize()))
+				.add("active connections,")
+				.add(String.valueOf(metrics.idleSize()))
+				.add("inactive connections")
+				.add(String.valueOf(metrics.pendingAcquireSize()))
+				.add("pending acquire requests.")
+				.toString();
 	}
 
 	final void scheduleInactivePoolsDisposal() {
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2024 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2018-2025 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


violetagg:
I think we need more to this solution.
Can you try the test below and ensure that with the fix we will be able to see the exception?

```java
@Test
void testLogFormat() {
	Channel channel = new EmbeddedChannel();
	useVerboseConsoleLoggers();
	try {
		InstrumentedPool<? extends Connection> pool = mock(InstrumentedPool.class);
		InstrumentedPool.PoolMetrics metrics = mock(InstrumentedPool.PoolMetrics.class);
		when(pool.metrics()).thenReturn(metrics);
		logPoolState(channel, pool, "log message");
		logPoolState(channel, pool, "log message", new Exception("test log format"));
	}
	finally {
		resetLoggerFactory();
		Connection.from(channel).dispose();
	}
}
```

raccoonback:
@violetagg Yes, I'll check it out!

raccoonback:
@violetagg 
I have tested the provided code locally, and it seems to work as expected without any issues!

violetagg:
@raccoonback That's interesting

When I'm using `useVerboseConsoleLoggers();` 

I'm seeing
 
```
[DEBUG] (Test worker) Using Verbose Console logging
[DEBUG] (Test worker) [embedded, L:embedded - R:embedded] log message, now: 0 active connections, 0 inactive connections and 0 pending acquire requests.
[DEBUG] (Test worker) [embedded, L:embedded - R:embedded] log message, now: 0 active connections, 0 inactive connections and 0 pending acquire requests.
```

When I'm using `useSl4jLoggers();`

I'm seeing also the exception

```
19:05:46.600 [Test worker] DEBUG r.n.r.PooledConnectionProvider - [embedded, L:embedded - R:embedded] log message, now: 0 active connections, 0 inactive connections and 0 pending acquire requests.
19:05:46.607 [Test worker] DEBUG r.n.r.PooledConnectionProvider - [embedded, L:embedded - R:embedded] log message, now: 0 active connections, 0 inactive connections and 0 pending acquire requests.
java.lang.Exception: test log format
	at reactor.netty.resources.DefaultPooledConnectionProviderTest.testLogFormat(DefaultPooledConnectionProviderTest.java:590)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:767)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
...
```

Can you share the output that you are seeing?

raccoonback:
@violetagg 
Same thing happened to me..
I'll look into this more and share it with you!

when useing `useVerboseConsoleLoggers()`;
```java
[DEBUG] (Test worker) Using Verbose Console logging
[DEBUG] (Test worker) [embedded, L:embedded - R:embedded] log message, now: 0 active connections, 0 inactive connections and 0 pending acquire requests.
[DEBUG] (Test worker) [embedded, L:embedded - R:embedded] log message, now: 0 active connections, 0 inactive connections and 0 pending acquire requests.
```

when useing `useSl4jLoggers()`;
```java
08:19:01.030 [Test worker] DEBUG r.n.r.PooledConnectionProvider - [embedded, L:embedded - R:embedded] log message, now: 0 active connections, 0 inactive connections and 0 pending acquire requests.
08:19:01.037 [Test worker] DEBUG r.n.r.PooledConnectionProvider - [embedded, L:embedded - R:embedded] log message, now: 0 active connections, 0 inactive connections and 0 pending acquire requests.
java.lang.Exception: test log format
	at reactor.netty.resources.DefaultPooledConnectionProviderTest.testLogFormat(DefaultPooledConnectionProviderTest.java:590)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:767)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
```

raccoonback:
Here is what I’ve found so far:
It seems there are two issues at play.

1. Even when an `Exception` is thrown, [`debug(String format, Object... arguments)`](https://github.com/reactor/reactor-core/blob/96e6824fc9ff96fd493385a36207ae8a4f500795/reactor-core/src/main/java/reactor/util/Loggers.java#L258-L260) is invoked instead of `debug(String msg, Throwable t)`.

2. While the **SLF4J** implementation (**logback**) handles cases where [`debug(String format, Object... arguments)` is called with only a `Throwable`](
https://github.com/qos-ch/logback/blob/1da2f171dc108a2576eb33918198db7f755b1224/logback-classic/src/main/java/ch/qos/logback/classic/spi/LoggingEvent.java#L140-L142), the console logger does not handle this scenario. 
In console logger, it only prints in `PrintStream`.
```java
public PrintStream format(String format, Object ... args) {
        try {
            synchronized (this) {
                ensureOpen();
                if ((formatter == null)
                    || (formatter.locale() != Locale.getDefault()))
                    formatter = new Formatter((Appendable) this);
                formatter.format(Locale.getDefault(), format, args);
            }
        } catch (InterruptedIOException x) {
            Thread.currentThread().interrupt();
        } catch (IOException x) {
            trouble = true;
        }
        return this;
    }
```

As a result, when `useSl4jLoggers()` is used, the stack trace is printed correctly, but this is not the case for the console logger.

raccoonback:
@violetagg The above case appears to be a `reactor-core` issue, so I will also send a PR to `reactor-core`!

violetagg:
@raccoonback Please wait, there is no issue with reactor-core, I just wanted to point that we provide 5 arguments for 4 placeholders. Depending on the different implementations this behaves differently - some may omit the Throwable, some may try to handle the Throwable and as in the reported issue, the logging implementation may even print bad pattern.
IMO we need to distinguish between two use cases - with/without Throwable and use the proper method - debug(String format, Object... arguments)/debug(String msg, Throwable t)

raccoonback:
@violetagg 
Thanks for the great advice! 😃
I believe we have three possible solutions for this issue. Could you share your thoughts on which approach might be the best?

1. Add an additional `log.debug("", t)` call.
```java
log.debug(
    format(channel, "{}, now: {} active connections, {} inactive connections and {} pending acquire requests."),
	msg,
	metrics.acquiredSize(),
	metrics.idleSize(),
	metrics.pendingAcquireSize()
);
if(t != null) {
     log.debug("", t);
}
```

2.  Set the log message in advance before calling `log.debug()`.
(Since most of them are primitive types, it would be a good idea to create a string log message in advance.)
```java
String message = new StringJoiner(", ")
			.add(msg)
			.add("now: " + metrics.acquiredSize() + " active connections")
			.add(metrics.idleSize() + " inactive connections and " + metrics.pendingAcquireSize() + " pending acquire requests.")
			.toString();
if(t != null) {
    log.debug(format(channel, message), t);
}
```

3. Create a **logging utility** to handle such cases separately.
```java

class LogUtils {

  // ...

}
```

Do you think any of these approaches would work best? Or do you have any other suggestions we could consider?

raccoonback:
@violetagg 
First of all, I modified it using method 2.
(If it is a `slf4j` environment, there is a **tradeoff of always trying string operations regardless of the log level**.)

ref. https://www.slf4j.org/faq.html#logging_performance

violetagg (approved):
Thanks for the PR!

violetagg:
I back ported the fix to `1.1.x` branch.

Issue #2746, opened 2023-03-27
HttpClient proxy configuration per request


jordillachmrf:
As explained here https://stackoverflow.com/questions/74893474/is-there-any-way-to-imlement-a-rotating-proxy-based-httpclient-in-reactor-netty it would be nice if reactor-netty HttpClient could be configured in a per request basis, instead of being "statically" configured as explained in the docs https://projectreactor.io/docs/netty/1.1.5/reference/index.html#_proxy_support_2

## Motivation
Some of our clients are GDRP constrained and based on the location of their servers and our servers(eeuu vs eu) we get 451 status codes(legal reasons)

## Desired solution
`HttpClient client = HttpClient.create()
				          .proxy((spec, req) -> spec.type(ProxyProvider.Proxy.HTTP)
				                             .host("proxy")
				                             .port(8080)
				                             .nonProxyHosts("localhost")
				                             .connectTimeoutMillis(20_000));
`
req -> HttpClientRequest, if HttpClientRequest would be available then we could route the request to our eeuu proxy or eu proxy based on the destination uri.

## Considered alternatives
Have various HttpClients's, handle 451 in our code and fallback to the appropriate HttpClient at runtime

pderop:
I will take a look at this issue soon.

violetagg:
@jordillachmrf Can you elaborate more about the desired solution? You know the destination URI as this is something the the  one that uses the `HttpClient` provides. Based on this you can then choose which `HttpClient` to use  i.e. the one configured with proxy for `x` or the one configured with proxy for `y`.

jordillachmrf:
And that's exactly how we are tackling this, but keeping the routing solution close the proxy configuration seems to us more appropriate than moving this logic one layer above. 
We have various "services" that use these various proxied webClients, we have had to wrap these into an artificial layer to hide this complexity. 
We were wondering that adding this flexibility when configuring the proxy would allow us to get rid of this

violetagg:
@jordillachmrf I'm gonna add this to the general backlog. Please use the described solution in my previous comment for the time being.
The team will discuss when and how this can be implemented. If you can work on this feature and provide a PR, this will be great.

longkunbetter:
@violetagg Is somebody working on this problem?

violetagg:
@longkunbetter We are not working on this one. I marked this with `help wanted` if somebody wants to work on this.

punitdarira:
Hi @violetagg,
[...]

PR #3593, opened 2025-01-19
Support for dynamic proxy configuration at the HTTP protocol level

raccoonback:
## Motivation
The existing `proxy(...)` and `noProxy()` methods are static, applying the same configuration to all requests. In scenarios where proxy behavior needs to change dynamically based on the request (e.g., host-specific proxies, conditional proxy usage), this static approach is insufficient.

The `proxyWhen(...)` method fills this gap by allowing developers to provide a function that resolves proxy settings dynamically during request execution.

## Description
- Added support for the `proxyWhen(BiConsumer<HttpClientConfig, ? super ProxyProvider.TypeSpec>)` method to `HttpClient`.
  - This method enables users to dynamically determine the proxy configuration on a per-request basis using a provided [`HttpClientConfig`](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java) function.
  - `proxyWhen(...)` evaluates the configuration function for each HTTP request, allowing fine-grained control over proxy settings.
- If `proxyWhen(...)` is used, any static proxy configurations applied via `proxy(...)` or `noProxy()` will be ignored.
- Internally, it uses `deferredConfig` to set the `proxyProvider` at the [`HttpClientConnect.connect()`](https://github.com/reactor/reactor-netty/blob/ec49b3dfb30ede20425a12d8d01fb3cbdd2e3496/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L106-L133) stage.

```java
    .proxyWhen((HttpClientConfig config, ProxyProvider.TypeSpec spec) -> {
        if (config.uri().equals("example.com")) {
            spec.type(ProxyProvider.Proxy.HTTP)
                .host("proxy.example.com")
                .port(8080);
        }
    });

client.get()
    .uri("http://example.com")
    .response()
    .block();
```

## Backward Compatibility
- The new `proxyWhen(...)` method is additive and does not interfere with existing static configurations unless explicitly used.
- Static configurations (`proxy(...)` and `noProxy()`) remain unchanged and fully functional.

## Related Issue
- https://github.com/reactor/reactor-netty/issues/2746

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017-2024 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2017-2025 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
@@ -389,7 +389,7 @@ public final class ProxyProvider {
 		return Integer.parseInt(port);
 	}
 
-	static final class Build implements TypeSpec, AddressSpec, Builder {
+	static final class Build implements TypeSpec, AddressSpec, Builder, Validator {
 
 		@SuppressWarnings("UnnecessaryLambda")
 		static final Supplier<? extends HttpHeaders> NO_HTTP_HEADERS = () -> null;
@@ -512,6 +512,11 @@ public final class ProxyProvider {
 		@Override
 		public ProxyProvider build() {
 			return new ProxyProvider(this);
+		}
+
+		@Override
+		public final boolean isConfiguredCorrectly() {
+			return type != null && host != null;
 		}
 	}
 
@@ -732,5 +737,15 @@ public final class ProxyProvider {
 		 * @return builds new ProxyProvider
 		 */
 		ProxyProvider build();
+	}
+
+	public interface Validator {
+
+		/**
+		 * Checks whether the proxy configuration is correctly set.
+		 *
+		 * @return true if the required settings are completed, false if any settings are missing
+		 */
+		boolean isConfiguredCorrectly();
 	}
 }
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
@@ -389,7 +389,7 @@ public final class ProxyProvider {
 		return Integer.parseInt(port);
 	}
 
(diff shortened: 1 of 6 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1634,6 +1635,35 @@ public Mono<Void> warmup() {
...
+							if (((ProxyProvider.Validator) spec).isConfiguredCorrectly()) {
+								ProxyProvider proxyProvider = ((ProxyProvider.Builder) spec).build();
+								config.proxyProvider(proxyProvider);
+							}
+
+							return Mono.just(config);
+						}
+				);
+		return dup;
+	}
```
violetagg, line 1665: What do you think if we use `Mono` for the proxy configuration? Imagine that you need to resolve something else before providing the proxy configuration.
Something similar to what we have for the headers e.g. this test case:
https://github.com/reactor/reactor-netty/blob/c5e4a96a297a2577fbc1a55f8255ac43f205b002/reactor-netty-http/src/test/java/reactor/netty/http/client/WebsocketClientOperationsTest.java#L85

```suggestion
	public final HttpClient proxyWhen(BiFunction<HttpClientConfig, ? super ProxyProvider.TypeSpec,
			Mono<? extends ProxyProvider.Builder>> proxyBuilder) {
		Objects.requireNonNull(proxyBuilder, "proxyBuilder");
		HttpClient dup = duplicate();
		dup.configuration().deferredConf(config -> {
			Mono<? extends ProxyProvider.Builder> mono = proxyBuilder.apply(config, ProxyProvider.builder());
			return mono == Mono.<ProxyProvider.Builder>empty() ?
					Mono.just(config) :
					mono.map(builder -> {
						config.proxyProvider(builder.build());
						return config;
					});
		});
		return dup;
	}
```

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
raccoonback: @violetagg 
Great point!
I’ll modify it to use Mono as suggested, ensuring it operates without blocking.
Thank you for your valuable feedback! 😊

raccoonback: @violetagg 
Hello, I’ve updated the code to reflect your feedback!
Please review the changes, and let me know if there are any additional improvements needed.
Thank you!

reactor-netty-core/src/main/java/reactor/netty/transport/ProxyProvider.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017-2024 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2017-2025 VMware, Inc. or its affiliates, All Rights Reserved.
```
violetagg, line 2: No changes in the file 

```suggestion
 * Copyright (c) 2017-2024 VMware, Inc. or its affiliates, All Rights Reserved.
```

violetagg (approved):
Nice!

PR #3612, opened 2025-02-02
Support a handler for checking connection status using Ping frame in HTTP/2

raccoonback:
## Related Issue
Fix https://github.com/reactor/reactor-netty/issues/3301

### DIFF
--- a/docs/modules/ROOT/pages/http-client.adoc
+++ b/docs/modules/ROOT/pages/http-client.adoc
@@ -318,6 +318,7 @@ include::{examples-dir}/http2/H2Application.java[lines=18..42]
 ----
 <1> Configures the client to support only `HTTP/2`
 <2> Configures `SSL`
+<3> You can configure the interval for checking `Ping` frames
 
 The following listing presents a simple `H2C` example:
 
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
@@ -1,0 +1,245 @@
+/*
+ * Copyright (c) 2025 VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty.http.client;
+
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelFutureListener;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPromise;
+import io.netty.handler.codec.http2.DefaultHttp2PingFrame;
+import io.netty.handler.codec.http2.Http2ChannelDuplexHandler;
+import io.netty.handler.codec.http2.Http2ConnectionEncoder;
+import reactor.util.Logger;
+import reactor.util.Loggers;
+import reactor.util.annotation.Nullable;
+
+import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ThreadLocalRandom;
+
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+/**
+ * Handler that supports connection health checks using HTTP/2 Ping Frames.
+ *
+ * <p>This Handler sends a ping frame at the specified interval when no frame is being read or written,
+ * ensuring the connection health is monitored. If a ping ACK frame is not received within the configured interval,
+ * the connection will be closed.</p>
+ *
+ * <p>Ping frame checking will not be performed while a read or write operation is in progress.</p>
+ *
+ * <p>Be cautious when setting a very short interval, as it may cause the connection to be closed,
+ * even if the keep-alive setting is enabled.</p>
+ *
+ * <p>If no interval is specified, no ping frame checking will be performed.</p>
+ *
+ * @author raccoonback
+ * @since 1.2.3
+ */
+public class Http2ConnectionLivenessHandler extends Http2ChannelDuplexHandler {
+
+	private static final Logger log = Loggers.getLogger(Http2ConnectionLivenessHandler.class);
+
+	private ScheduledFuture<?> pingScheduler;
+	private final ChannelFutureListener pingWriteListener = new PingWriteListener();
+	private final Http2ConnectionEncoder encoder;
+	private final long pingIntervalNanos;
+	private long lastSentPingData;
+	private long lastReceivedPingTime;
+	private long lastIoTime;
+	private boolean isPingAckPending;
+
+	public Http2ConnectionLivenessHandler(Http2ConnectionEncoder encoder, @Nullable Duration pingInterval) {
+		Objects.requireNonNull(encoder, "encoder");
+		this.encoder = encoder;
+
+		if (pingInterval != null) {
+			this.pingIntervalNanos = pingInterval.toNanos();
+		}
+		else {
+			this.pingIntervalNanos = 0L;
+		}
+	}
+
+	@Override
+	public void channelActive(ChannelHandlerContext ctx) throws Exception {
+		if (isPingIntervalConfigured()) {
+			isPingAckPending = false;
+			pingScheduler = ctx.executor()
+					.schedule(
+							new PingChecker(ctx),
+							pingIntervalNanos,
+							NANOSECONDS
+					);
+		}
+
+		ctx.fireChannelActive();
+	}
+
+	@Override
+	public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+		if (msg instanceof DefaultHttp2PingFrame) {
+			DefaultHttp2PingFrame frame = (DefaultHttp2PingFrame) msg;
+			if (frame.ack() && frame.content() == lastSentPingData) {
+				lastReceivedPingTime = System.nanoTime();
+			}
+		}
+		else {
+			lastIoTime = System.nanoTime();
+		}
+
+		ctx.fireChannelRead(msg);
+	}
+
+	@Override
+	@SuppressWarnings("FutureReturnValueIgnored")
+	public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
+		lastIoTime = System.nanoTime();
+
+		ctx.write(msg, promise);
+	}
+
+	@Override
+	public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+		cancel();
+		ctx.fireChannelInactive();
+	}
+
+	@Override
+	public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+		cancel();
+		ctx.fireExceptionCaught(cause);
+	}
+
+	private boolean isPingIntervalConfigured() {
+		return pingIntervalNanos > 0;
+	}
+
+	private void cancel() {
+		if (pingScheduler != null) {
+			pingScheduler.cancel(false);
+		}
+	}
+
+	private class PingChecker implements Runnable {
+
+		private final ChannelHandlerContext ctx;
+
+		PingChecker(ChannelHandlerContext ctx) {
+			this.ctx = ctx;
+		}
+
+		@Override
+		public void run() {
+			Channel channel = ctx.channel();
+			if (channel == null || !channel.isOpen()) {
+
+				return;
+			}
+
+			if (lastIoTime == 0 || isIoInProgress()) {
+				if (log.isDebugEnabled()) {
+					log.debug("{} channel is currently reading or writing data.", channel);
+				}
+
+				isPingAckPending = false;
+				pingScheduler = invokeNextSchedule();
+				return;
+			}
+
+			if (!isPingAckPending) {
+				if (log.isDebugEnabled()) {
+					log.debug("Attempting to send a ping frame to {} channel.", channel);
+				}
+
+				writePing(ctx);
+				pingScheduler = invokeNextSchedule();
+				return;
+			}
+
+			if (isOutOfTimeRange()) {
+				if (log.isInfoEnabled()) {
+					log.info("Closing {} channel due to delayed ping frame response (timeout: {} ns).", channel, pingIntervalNanos);
+				}
+
+				close(channel);
+				return;
+			}
+
+			isPingAckPending = false;
+			pingScheduler = invokeNextSchedule();
+		}
+
+		private void writePing(ChannelHandlerContext ctx) {
+			lastSentPingData = ThreadLocalRandom.current().nextLong();
+
+			encoder.frameWriter()
+					.writePing(ctx, false, lastSentPingData, ctx.newPromise())
+					.addListener(pingWriteListener);
+			ctx.flush();
+		}
+
+		private boolean isIoInProgress() {
+			return pingIntervalNanos > (System.nanoTime() - lastIoTime);
+		}
+
+		private boolean isOutOfTimeRange() {
+			return pingIntervalNanos < (System.nanoTime() - lastReceivedPingTime);
+		}
+
+		private ScheduledFuture<?> invokeNextSchedule() {
+			return ctx.executor()
+					.schedule(
+							new PingChecker(ctx),
+							pingIntervalNanos,
+							NANOSECONDS
+					);
+		}
+
+		private void close(Channel channel) {
+			channel.close()
+					.addListener(future -> {
+						if (future.isSuccess()) {
+							if (log.isDebugEnabled()) {
+								log.debug("{} channel closed an channel", channel);
+							}
+						}
+						else if (log.isDebugEnabled()) {
+							log.debug("{} channel failed to close an channel", channel, future.cause());
+						}
+					});
+		}
+	}
+
+	private class PingWriteListener implements ChannelFutureListener {
+
+		@Override
+		public void operationComplete(ChannelFuture future) throws Exception {
+			if (future.isSuccess()) {
+				if (log.isDebugEnabled()) {
+					log.debug("Wrote PING frame to {} channel.", future.channel());
+				}
+
+				isPingAckPending = true;
+			}
+			else if (log.isDebugEnabled()) {
+				log.debug("Failed to wrote PING frame to {} channel.", future.channel());
+			}
+		}
+	}
+}
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2024 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2020-2025 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -677,7 +677,7 @@ public final class HttpClientConfig extends ClientTransportConfig<HttpClientConf
 	}
 
 	static void configureHttp2Pipeline(ChannelPipeline p, HttpResponseDecoderSpec decoder,
-			Http2Settings http2Settings, ConnectionObserver observer) {
+			Http2Settings http2Settings, Duration pingInterval, ConnectionObserver observer) {
 		Http2FrameCodecBuilder http2FrameCodecBuilder =
 				Http2FrameCodecBuilder.forClient()
 				                      .validateHeaders(decoder.validateHeaders())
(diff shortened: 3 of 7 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+	private final long pingAckTimeoutNanos;
+	private final long pingScheduleIntervalNanos;
+	private final int pingAckDropThreshold;
+
+	private int pingAckDropCount;
+	private long lastSentPingData;
+	private long lastReceivedPingTime;
+	private long lastSendingPingTime;
+	private long lastIoTime;
+	private boolean isPingAckPending;
```
raccoonback, line 71: It is a flag indicating whether a `Ping Frame` has been sent and the system is waiting for the `ACK`.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+		else {
+			this.pingScheduleIntervalNanos = 0L;
+		}
+
+		if (pingAckDropThreshold != null) {
+			this.pingAckDropThreshold = pingAckDropThreshold;
+		}
+		else {
+			this.pingAckDropThreshold = 0;
+		}
```
raccoonback, line 98: It initializes the `pingAckTimeoutNanos`, `pingScheduleIntervalNanos`, and `pingAckDropThreshold` values provided by the user.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+		if (isPingIntervalConfigured()) {
+			isPingAckPending = false;
+			pingAckDropCount = 0;
+			pingScheduler = ctx.executor()
+					.schedule(
+							new PingChecker(ctx),
+							pingAckTimeoutNanos,
+							NANOSECONDS
+					);
+		}
```
raccoonback, line 112: The `scheduler` is executed only when the user has configured Ping settings and the Channel is active.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	@Override
+	public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+		if (msg instanceof Http2PingFrame) {
+			Http2PingFrame frame = (Http2PingFrame) msg;
+			if (frame.ack() && frame.content() == lastSentPingData) {
+				lastReceivedPingTime = System.nanoTime();
+			}
+		}
```
raccoonback, line 124: When the `ACK` for the **`last sent Ping Frame` is received**, it sets the `lastReceivedPingTime` to the current time.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+		else {
+			lastIoTime = System.nanoTime();
+		}
```
raccoonback, line 127: If a frame other than the Ping ACK is received, it sets the `lastIoTime` to the current time.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+		ctx.fireChannelRead(msg);
+	}
+
+	@Override
+	@SuppressWarnings("FutureReturnValueIgnored")
+	public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
+		lastIoTime = System.nanoTime();
+
+		ctx.write(msg, promise);
+	}
```
raccoonback, line 138: When a frame is written, it sets the `lastIoTime` to the current time.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+
+	@Override
+	public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+		cancel();
+		ctx.fireChannelInactive();
+	}
```
raccoonback, line 144: When the channel is closed, the scheduler is also stopped.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+		PingChecker(ChannelHandlerContext ctx) {
+			this.ctx = ctx;
+		}
+
+		@Override
+		public void run() {
+			Channel channel = ctx.channel();
+			if (channel == null || !channel.isOpen()) {
+				return;
+			}
```
raccoonback, line 170: If the channel is already closed, the scheduler stops without restarting.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+			if (lastIoTime == 0 || isIoInProgress()) {
+				if (log.isDebugEnabled()) {
+					log.debug("{} channel is currently reading or writing data.", channel);
+				}
+
+				isPingAckPending = false;
+				pingAckDropCount = 0;
+				pingScheduler = invokeNextSchedule();
+				return;
+			}
```
raccoonback, line 181: If other messages, besides the Ping Frame, are being read or written, the scheduler execution is scheduled.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+
+			if (!isPingAckPending) {
+				if (log.isDebugEnabled()) {
+					log.debug("Attempting to send a ping frame to {} channel.", channel);
+				}
+
+				writePing(ctx);
+				pingScheduler = invokeNextSchedule();
+				return;
+			}
```
raccoonback, line 191: If not waiting for a Ping ACK, a Ping frame is sent. 
Specifically, when the listener receives an event indicating a successful write, `isPingAckPending` is set to true, and `lastSendingPingTime` is set to the current time.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+				}
+
+				if (log.isInfoEnabled()) {
+					log.info("Drop ping ack frame in {} channel. (ping: {})", channel, lastSentPingData);
+				}
+
+				writePing(ctx);
+				pingScheduler = invokeNextSchedule();
+				return;
+			}
```
raccoonback, line 212: If the **Ping ACK is not received** within `pingAckTimeoutNanos`, but the retry count has not yet reached the `pingAckDropThreshold`, a **retry is attempted**.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+
+			isPingAckPending = false;
+			pingAckDropCount = 0;
+			pingScheduler = invokeNextSchedule();
```
raccoonback, line 216: If the Ping ACK is not received within `pingAckTimeoutNanos` and retries have reached the `pingAckDropThreshold`, 
the connection is considered invalid and **the channel is closed**.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+		}
+
+		private void writePing(ChannelHandlerContext ctx) {
+			lastSentPingData = ThreadLocalRandom.current().nextLong();
+
+			encoder.frameWriter()
+					.writePing(ctx, false, lastSentPingData, ctx.newPromise())
+					.addListener(pingWriteListener);
+			ctx.flush();
+		}
```
raccoonback, line 226: New data is generated, and a Ping frame is sent.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+
+		private boolean isIoInProgress() {
+			return pingAckTimeoutNanos >= (System.nanoTime() - lastIoTime);
+		}
+
+		private boolean isOutOfTimeRange() {
+			return pingAckTimeoutNanos < Math.abs(lastReceivedPingTime - lastSendingPingTime);
+		}
```
raccoonback, line 234: It determines whether the Ping ACK was received within `pingAckTimeoutNanos`.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionLivenessHandler.java
```diff
@@ -0,0 +1,285 @@
...
+				}
+
+				isPingAckPending = true;
+				lastSendingPingTime = System.nanoTime();
+			}
+			else if (log.isDebugEnabled()) {
+				log.debug("Failed to wrote PING frame to {} channel.", future.channel());
+			}
+		}
+	}
```
raccoonback, line 284: If the new Ping frame is successfully sent, the status is changed to 'ping awaiting', and `lastSendingPingTime` is updated to the current time.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java
```diff
@@ -688,8 +689,12 @@ static void configureHttp2Pipeline(ChannelPipeline p, HttpResponseDecoderSpec de
 					"reactor.netty.http.client.h2"));
 		}
 
+		Http2FrameCodec codec = http2FrameCodecBuilder.build();
+
 		p.addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.H2Flush, new FlushConsolidationHandler(1024, true))
-		 .addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.HttpCodec, http2FrameCodecBuilder.build())
+		 .addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.HttpCodec, codec)
+		 .addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.H2LivenessHandler,
+				 new Http2ConnectionLivenessHandler(codec.encoder(), pingAckTimeout, pingScheduleInterval, pingAckDropThreshold))
```
raccoonback, line 697: configure `Http2ConnectionLivenessHandler`.

reactor-netty-http/src/test/java/reactor/netty/resources/PooledConnectionProviderDefaultMetricsTest.java
```diff
@@ -388,6 +388,8 @@ private void testIssue3060(ConnectionProvider provider) throws Exception {
 	/* https://github.com/reactor/reactor-netty/issues/3519 */
 	@Test
 	public void testConnectionProviderDisableAllBuiltInMetrics() throws Exception {
+		REGISTRY.clear();
+
```
raccoonback, line 392: Although not related to the purpose of this PR, handling has been implemented for failing tests. 
To prevent interference with other tests, the `metrics` are initialized.

reactor-netty-http/src/test/java/reactor/netty/http/client/HttpClientTest.java
```diff
@@ -653,17 +650,9 @@ void testMaxConnectionPools(boolean withMaxConnectionPools) throws SSLException
...
-						.isEqualTo("Connection pool creation limit exceeded: {} pools created, maximum expected is {}");
-			}
-			else {
-				Mockito.verify(spyLogger, times(0))
-				       .warn(Mockito.eq("Connection pool creation limit exceeded: {} pools created, maximum expected is {}"),
-				               Mockito.eq(2), Mockito.eq(1));
-			}
+			Mockito.verify(spyLogger, times(0))
+					.warn(Mockito.eq("Connection pool creation limit exceeded: {} pools created, maximum expected is {}"),
+							Mockito.eq(2), Mockito.eq(1));
```
raccoonback, line 655: Although not related to the purpose of this PR, handling has been implemented for failing tests.
In `PooledConnectionProvider`, the `connectionPoolCount` is handled atomically to ensure that the log warning `log.warn("Connection pool creation limit exceeded: {} pools created, maximum expected is {}")` is not called multiple times. 
As a result, the test outcomes have been adjusted.

raccoonback:
@violetagg 
Hello!
I have rebased the branch onto `1.2.x` and changed the target branch of the PR to `1.2.x`.

violetagg:
@raccoonback I'll review this PR later this week

violetagg:
@raccoonback Nice idea! I think that we can reuse our current `reactor.netty.http.server.IdleTimeoutHandler` (we need to check whether the current position in the pipeline will be ok or whether we need to move it) and extend it with the HTTP/2 functionality ACK requirement. At the moment it handles idle timeout for HTTP/2 and H2C. The more ChannelHandler you have in the pipeline the worse is the performance so we always try to keep the ChannelHandlers number to the minimum if possible. Also all methods in `reactor.netty.http.server.IdleTimeoutHandler` are invoked on the event loop which guarantees that the functionality is thread-safe.

raccoonback:
@violetagg 
Hello, and thank you for your suggestion!
I have a few questions for clarification.

1. You suggested reusing `reactor.netty.http.server.IdleTimeoutHandler`.
Currently, the implementation works by **scheduling a PING frame transmission when there is no read/write activity**, then checking whether an ACK is received within the allowed time. (It also supports retries up to a certain threshold.)
'
My understanding is that you’re suggesting leveraging `IdleTimeoutHandler` to detect **when both read and write operations are idle** and incorporate it into this logic. Is that correct?
Specifically, do you mean modifying the handler so that when both read and write operations are idle, a PING frame is sent, and if no ACK is received within the retry threshold, the connection is closed?
'
If I’ve misunderstood, could you clarify what exactly you mean by "reuse"?

2. This Issue was intended to introduce **PING frame support only for the client-side**.
  - Are you suggesting extending this functionality to the **server-side** as well, for HTTP/2 and H2C environments?
  
3. Are you proposing making `IdleTimeoutHandler` usable for both client and server?
  - Since the current name (`server`) implies it is server-specific, would it be appropriate to modify it for client-side usage as well?

4. Would it be reasonable to move `IdleTimeoutHandler` to the `reactor.netty.http package`?
- If we make it usable for both client and server, relocating it to a more general package might be a better approach. I’d love to hear your thoughts on this.

Looking forward to your feedback!

violetagg:
> @violetagg Hello, and thanks for your suggestion! 😃 I have a few questions for clarification.
> 
> 1. You suggested reusing `reactor.netty.http.server.IdleTimeoutHandler`.
>    Currently, the implementation works by **scheduling a PING frame transmission when there is no read/write activity**, then checking whether an **ACK** is received within the allowed time. (It also supports retries up to a certain **threshold**.)
>    '
>    I understand that you’re suggesting leveraging `IdleTimeoutHandler` **to detect when both read and write operations are idle and incorporating it into this logic**. Is that correct?

Yes 
Currently we use it only for checking that there is no read, because we add this handler between the requests and we are sure there is no write operation. However this handler extends `io.netty.handler.timeout.IdleStateHandler` and for HTTP/2 we can configure it to check for both read and write.

>    Specifically, do you mean modifying the handler so that when both read and write operations are idle, a PING frame is sent, and if no ACK is received within the retry threshold, the connection is closed?

Yes

>    '
>    If I’ve misunderstood, could you clarify what exactly you mean by "reuse"? 😮
> 2. Can I understand that you are proposing to extend this feature not only on the **client side**, but also on the **server side** by `IdleTimeoutHandler`?

Yes

>    
>    * I thought that this Issue was intended to introduce **PING frame support only for the client-side** at first.
> 
> Looking forward to your feedback! 😊

We may need to move `reactor.netty.http.server.IdleTimeoutHandler` to the `reactor.netty.http` package so that it can be used by both server and client.

raccoonback:
@violetagg 

Thank you for the detailed explanation!
I will update `reactor.netty.http.server.IdleTimeoutHandler` to **check the Ping ACK internally** when using`HTTP/2` and `H2C`.

raccoonback:
@violetagg 
Hello!
I will reflect your PR review soon.
Thanks!

raccoonback:
@violetagg 
Updated to work based on `IdleTimeoutHandler`!

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -998,6 +1033,20 @@ public void handlerAdded(ChannelHandlerContext ctx) {
...
+						pipeline,
+						idleTimeout,
+						new Http2ConnectionLiveness(
+								upgrader.http2FrameCodec,
+								http2SettingsSpec != null ? http2SettingsSpec.pingAckTimeout() : null,
+								http2SettingsSpec != null ? http2SettingsSpec.pingScheduleInterval() : null,
+								http2SettingsSpec != null ? http2SettingsSpec.pingAckDropThreshold() : null
+						)
+				);
+			}
```
raccoonback, line 1049: In `H2CleartextCodec`, when upgrading from `HTTP/1.1` to the `H2C` protocol, the `IdleTimeoutHandler` is changed to be based on `Http2ConnectionLiveness`.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -1287,11 +1341,6 @@ protected void configurePipeline(ChannelHandlerContext ctx, String protocol) {
 				configureHttp11Pipeline(p, accessLogEnabled, accessLog, compressionOptions, compressPredicate, cookieDecoder, cookieEncoder, true,
 						decoder, formDecoderProvider, forwardedHeaderHandler, httpMessageLogFactory, idleTimeout, listener,
 						mapHandle, maxKeepAliveRequests, methodTagValue, metricsRecorder, minCompressionSize, readTimeout, requestTimeout, uriTagValue);
-
-				// When the server is configured with HTTP/1.1 and H2 and HTTP/1.1 is negotiated,
-				// when channelActive event happens, this HttpTrafficHandler is still not in the pipeline,
-				// and will not be able to add IdleTimeoutHandler. So in this use case add IdleTimeoutHandler here.
-				IdleTimeoutHandler.addIdleTimeoutHandler(ctx.pipeline(), idleTimeout);
```
raccoonback, line 1294: Move to be performed in `configureHttp11OrH2CleartextPipeline()`.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpTrafficHandler.java
```diff
@@ -546,7 +539,6 @@ void handleLastHttpContent(Object msg, ChannelPromise promise) {
 			ctx.executor().execute(this);
 		}
 		else {
-			IdleTimeoutHandler.addIdleTimeoutHandler(ctx.pipeline(), idleTimeout);
```
raccoonback, line 549: Modified to add the `IdleTimeoutHandler` for each protocol in **HttpServerConfig** instead of handling it in **HttpTrafficHandler**.

violetagg:
@raccoonback I plan to review this PR next week.

docs/modules/ROOT/pages/http-client.adoc
```diff
@@ -680,6 +680,50 @@ include::{examples-dir}/channeloptions/Application.java[lines=18..54]
 <4> Configures the time between individual `keepalive` probes to 1 minute.
 <5> Configures the maximum number of TCP `keepalive` probes to 8.
 
+
+==== IdleTimeout
+The maximum time (resolution: ms) that this connection stays opened and waits for HTTP request. Once the timeout is reached, the connection is closed.
```
violetagg, line 685: > The maximum time (resolution: ms) that this connection stays opened and waits for HTTP request

Can you please elaborate more on this?

This is HttpClient so it initiates the HTTP request. Also as part of the connection pool we already have a configuration for dropping the idle connections so this new configuration seems to be not needed?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpTrafficHandler.java
violetagg: The idea of the idle timeout on the server is that we check whether the connection is idle only between the requests, because of that the handler is removed when we receive a request and added when we finish the response.
`Specifies an idle timeout on the connection when it is waiting for an HTTP request`, this is similar with what we have on the connection pool, we apply idle timeout only for the connections that are in the pool and do not process request.
Currently the new implementation changes this and even if we process a request it will check for idle. Is that correct?

docs/modules/ROOT/pages/http-client.adoc
raccoonback: @violetagg 
This describes configuring `IdleTimeout` on the HttpClient side using `IdleTimeoutHandler`, independently of the connection pool settings.

From my understanding, since the connection pool already manages idle connections, applying an additional `IdleTimeoutHandler` at the `HttpClient` level may be redundant. 
Could you please confirm if this understanding is accurate?

violetagg: yes I think it is redundant

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpTrafficHandler.java
raccoonback: @violetagg 
That’s a great point.
As you mentioned, for `HTTP/1.1`, I also believe it's more appropriate to add the `IdleTimeoutHandler` **only when there is no ongoing message exchange**, especially to support connection reuse in connection pools.

Additionally, for `HTTP/2`, what do you think about applying a similar approach — adding the `IdleTimeoutHandler` only when idle, and then performing a health check using a `HTTP/2 ping frame` **only when an idle timeout actually occurs?**

In summary, regardless of the HTTP protocol, I think it makes sense to keep the current logic in `HttpTrafficHandler` — dynamically adding/removing the `IdleTimeoutHandler` based on whether the channel is actively exchanging messages.

What are your thoughts on this?

raccoonback:
@violetagg 
Hello!
I’ve applied the review feedback! 😀

raccoonback:
@violetagg 
Hello!
Please check this PR~ 😊

violetagg:
> @violetagg Hello! Please check this PR~ 😊

yep I'll check it next week

raccoonback:
@violetagg 
Hello. 
I would appreciate it if you could review this PR. 
Thanks!

violetagg:
> @violetagg Hello. I would appreciate it if you could review this PR. Thanks!

yeah I started reviewing it and later was distracted ... let me do the release tomorrow and we will include it in the next release ...

raccoonback:
@violetagg 
Thank you! 😀

violetagg:
@raccoonback I'm working on this one

raccoonback:
@violetagg 
Hello!
Does it mean you are conducting a review? 😀

raccoonback:
@violetagg
Hello!
Please check this PR~~~ 😊

violetagg:
@raccoonback Would you be able to review #3980. I based it on this PR, reduced the scope and added HttpClient support.

raccoonback:
@violetagg Good!
I will check it

PR #3647, opened 2025-02-20
Make `Subscription` non nullable and suppress `NullAway`

violetagg:


reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
```diff
@@ -127,7 +127,10 @@ static final class DisposableAcquire
 		// Deliberately suppress "NullAway"
 		// This is a lazy initialization
 		PooledRef<PooledConnection> pooledRef;
-		@Nullable Subscription subscription;
+		@SuppressWarnings("NullAway")
+		// Deliberately suppress "NullAway"
+		// This is a lazy initialization
+		Subscription subscription;
```
chemicL, line 133: I think it's easier to follow if the comment is above the relevant piece of code, so here's a suggestion of inverting the existing pattern. Also, a suggestion to explain why we don't expect an NPE instead of just mentioning lazy initialization. The deliberate intention is probably appearent from the context and the use of `SuppressWarnings("NullAway")` so I suggested removing that line.
```suggestion
		// Never null when accessed - only via dispose()
		// which is registered into sink.onCancel() callback.
		// See onSubscribe(Subscription).
		@SuppressWarnings("NullAway")
		Subscription subscription;
```

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
```diff
@@ -164,6 +167,7 @@ public Context currentContext() {
 
 		@Override
 		public void dispose() {
+			// sink.onCancel() happens after initializing the subscription
```
chemicL, line 170:
```suggestion
			// sink.onCancel() registration happens in onSubscribe()
```

chemicL (changes requested):
As discussed, I see how the subscription can not be accessed when it's null. However, to make it easier to connect the dots, I made a suggestion, which if you accept can be applied to all the similar cases in this PR.

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
violetagg: fixed

reactor-netty-core/src/main/java/reactor/netty/resources/DefaultPooledConnectionProvider.java
violetagg: fixed

chemicL (approved):

violetagg:

PR #3694, opened 2025-03-25
Add `Nullability` to `NameResolverProvider.NameResolverSpec#bindAddressSupplier`

violetagg:
When a bind operation is not expected the bind address supplier may return null.


chemicL (approved):

violetagg:

Issue #3257, opened 2024-05-22
Netty does not provide errorLog just accesLog


paulbors:
As a system administrator, I would like to see errorLogs in Netty web server similar to those of Apache HTTPD when the client socket is closed and we receive no accessLog.

## Motivation
Netty web server does not log web server errors for requests that do not complete (ie: connection terminated is not captured in accessLog which should really be following Apache HTTPD errorLog).

## Desired solution
A new ChannelDuplexHandler is added similar to BaseAccessLogHandler and its counterparts but for logging the web server access errors only.

## Considered alternatives
One could increase the logs to trace to follow the channel termination exception log events, but that is too verbose for a production environment.

## Additional context
Implement a similar errorLogs as those of Apache HTTPD:
https://httpd.apache.org/docs/2.4/logs.html#page-header

violetagg:
@paulbors Are you interested in providing a PR? Which Reactor Netty version should enable this feature (1.1.x or 1.2.x)?

raccoonback:
@violetagg 

Hello!
I would like to work on this issue.
Before proceeding, I have a few questions for clarification.

1. Should I approach this by adding an `ErrorLogHandler` similar to `BaseAccessLogHandler`?
2. Should an error log be recorded for all cases where a connection is `disconnected` or `closed`, and should `all exceptions` also be logged as errors?

If I’ve misunderstood anything or if there’s any additional information you can share, I’d appreciate your guidance!
Thank you!

violetagg:
> Should I approach this by adding an ErrorLogHandler similar to BaseAccessLogHandler?

@raccoonback I haven't thought about any particular implementation, so we can try adding an ErrorLogHandler similar to BaseAccessLogHandler.

They mention `Implement a similar errorLogs as those of Apache HTTPD`, can you give me more information about this?

raccoonback:
@paulbors 
Hello!
Please explain in more detail.

> Implement a similar errorLogs as those of Apache HTTPD:
https://httpd.apache.org/docs/2.4/logs.html#page-header

raccoonback:
@violetagg 
> As a system administrator, I would like to see errorLogs in Netty web server similar to those of Apache HTTPD when the client socket is closed and we receive no accessLog.

Maybe.. It seems they want to access server-side error logs related to client socket closed and exceptions.

In Apache HTTPD, it appears that a lot of information besides network is logged in ErrorLog...

[...]

Issue #781, opened 2019-07-10
Netty with accessEvents with logback-access


bhochhi:
Why netty has no accessEvents but just accessLog implementation? Is there any plan supporting that?

violetagg:
Such integration should be possible via https://projectreactor.io/docs/netty/release/api/reactor/netty/http/server/HttpServer.html#accessLog-boolean-reactor.netty.http.server.logging.AccessLogFactory-

Marking this as `help wanted` if somebody wants to work on this.

childewuque:
Thanks
It would be better if we could configure instead of coding

JooHyukKim:
has anyone worked on this yet?

violetagg:
@JooHyukKim Currently nobody is working on this one.

JooHyukKim:
@violetagg I see. Would you allow me to work on it? seems quite interesting

JooHyukKim:
@violetagg @bhochhi I'd like to hear more about your thoughts on what AccessEvents are and how they should behave.

violetagg:
> @violetagg I see. Would you allow me to work on it? seems quite interesting

@JooHyukKim Yes sure

violetagg:
> @violetagg @bhochhi I'd like to hear more about your thoughts on what AccessEvents are and how they should behave.

I believe that @bhochhi meant integration with `logback-access` https://logback.qos.ch/access.html

JooHyukKim:
trueeee. @bhochhi is that what you are suggesting? integration with `logback-access` like how Tomcat and Jetty are integrated with `logback-access` described in [https://logback.qos.ch/access.html](https://logback.qos.ch/access.html) ?

JooHyukKim:
`logback-access` is officailly described to [integrate with Serverlet Containers such as Jetty or Tomcat](https://logback.qos.ch/access.html#:~:text=integrates%20with%20Servlet%20containers%20such%20as%20Jetty%20or%20Tomcat). 
As far as I know, we also need **netty-specific** `Adaptor` and `ValveBase` implementation in `Logback-access`.
looking at [codebase of Tomcat and Jetty integration with logback-access](https://github.com/qos-ch/logback/tree/master/logback-access/src/main/java/ch/qos/logback/access).

Lemme know how you think :)
cheers

bhochhi:
[...]

PR #3700, opened 2025-03-28
Support error log handler for Http server

raccoonback:
## Description
This PR introduces an **error logging mechanism** similar to Apache HTTPD's errorLog, addressing [#3257](https://github.com/reactor/reactor-netty/issues/3257). 
Currently, Reactor Netty does not log errors when exceptions are fired, so this update ensures that such errors are captured separately from access logs.

## Key Changes
- Add `DefaultErrorLogHandler` handler to log error when exceptions are fired.
- The log format can be customized just like AccessLog using `ErrorLogFactory`. 
Log format customization, similar to `AccessLog`, is available using `ErrorLogFactory`.
- Filtering can also be done using `ErrorLogFactory`.
Filtering information includes `RemoteAddress`, `HttpServerInfos,` and `Throwable`.

## Related Issue
- https://github.com/reactor/reactor-netty/issues/3257

## Additional Context
For reference, this implementation follows the Apache HTTPD error log format:
🔗 [Apache HTTPD Error Log Documentation](https://httpd.apache.org/docs/2.4/logs.html#page-header)

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
@@ -296,6 +296,7 @@ public abstract class HttpServer extends ServerTransport<HttpServer, HttpServerC
 	 *
 	 * @param enable enable or disable the error log
 	 * @return a new {@link HttpServer}
+	 * @since 1.2.5
 	 */
 	public final HttpServer errorLog(boolean enable) {
 		HttpServer dup = duplicate();
@@ -316,8 +317,8 @@ public abstract class HttpServer extends ServerTransport<HttpServer, HttpServerC
 	 *                   (req, res) -> res.header(CONTENT_TYPE, TEXT_PLAIN)
 	 *                                    .sendString(Mono.just("Hello World!"))))
 	 *           .errorLog(true, ErrorLogFactory.createFilter(
-	 * 								args -> args.cause() instanceof RuntimeException,
-	 * 								args -> ErrorLog.create("host-name={}", args.httpServerInfos().hostName())))
+	 *                              args -> args.cause() instanceof RuntimeException,
+	 *                              args -> ErrorLog.create("host-name={}", args.httpServerInfos().hostName())))
 	 *           .bindNow()
 	 *           .onDispose()
 	 *           .block();
@@ -332,6 +333,7 @@ public abstract class HttpServer extends ServerTransport<HttpServer, HttpServerC
 	 * @param enable enable or disable the error log
 	 * @param errorLogFactory the {@link ErrorLogFactory} that creates an {@link ErrorLog} given an {@link ErrorLogArgProvider}
 	 * @return a new {@link HttpServer}
+	 * @since 1.2.5
 	 */
 	public final HttpServer errorLog(boolean enable, ErrorLogFactory errorLogFactory) {
 		Objects.requireNonNull(errorLogFactory, "errorLogFactory");
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
@@ -52,6 +52,7 @@ import reactor.netty.http.server.logging.AccessLogArgProvider;
 import reactor.netty.http.server.logging.AccessLogFactory;
 import reactor.netty.http.server.logging.error.ErrorLog;
 import reactor.netty.http.server.logging.error.ErrorLogArgProvider;
+import reactor.netty.http.server.logging.error.DefaultErrorLoggingEvent;
 import reactor.netty.http.server.logging.error.ErrorLogFactory;
 import reactor.netty.internal.util.Metrics;
 import reactor.netty.tcp.SslProvider;
@@ -1324,6 +1325,8 @@ public abstract class HttpServer extends ServerTransport<HttpServer, HttpServerC
 				}
 				catch (Throwable t) {
 					log.error(format(connection.channel(), ""), t);
+					connection.channel().pipeline()
+							.fireUserEventTriggered(new DefaultErrorLoggingEvent(t));
 					//"FutureReturnValueIgnored" this is deliberate
 					connection.channel()
 					          .close();
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
@@ -50,6 +50,9 @@ import reactor.netty.http.server.compression.HttpCompressionOptionsSpec;
 import reactor.netty.http.server.logging.AccessLog;
 import reactor.netty.http.server.logging.AccessLogArgProvider;
 import reactor.netty.http.server.logging.AccessLogFactory;
+import reactor.netty.http.server.logging.error.ErrorLog;
+import reactor.netty.http.server.logging.error.ErrorLogArgProvider;
+import reactor.netty.http.server.logging.error.ErrorLogFactory;
 import reactor.netty.internal.util.Metrics;
 import reactor.netty.tcp.SslProvider;
 import reactor.netty.tcp.TcpServer;
@@ -267,6 +270,74 @@ public abstract class HttpServer extends ServerTransport<HttpServer, HttpServerC
 		Objects.requireNonNull(accessLogFactory, "accessLogFactory");
 		HttpServer dup = duplicate();
 		dup.configuration().accessLog = accessLogFactory;
+		return dup;
+	}
+
+	/**
+	 * Enable or disable the error log. If enabled, the default log system will be used.
+	 * <p>
+	 * Example:
+	 * <pre>
+	 * {@code
+	 * HttpServer.create()
+	 *           .port(8080)
+	 *           .route(r -> r.get("/hello",
+	 *                   (req, res) -> res.header(CONTENT_TYPE, TEXT_PLAIN)
+	 *                                    .sendString(Mono.just("Hello World!"))))
+	 *           .errorLog(true)
+	 *           .bindNow()
+	 *           .onDispose()
+	 *           .block();
+	 * }
+	 * </pre>
+	 * <p>
+	 *
+	 * Note that this method takes precedence over the {@value reactor.netty.ReactorNetty#ERROR_LOG_ENABLED} system property.
+	 *
+	 * @param enable enable or disable the error log
+	 * @return a new {@link HttpServer}
+	 */
+	public final HttpServer errorLog(boolean enable) {
+		HttpServer dup = duplicate();
+		dup.configuration().errorLog = null;
+		dup.configuration().errorLogEnabled = enable;
+		return dup;
+	}
+
+	/**
+	 * Enable or disable the error log and customize it through an {@link ErrorLogFactory}.
+	 * <p>
+	 * Example:
+	 * <pre>
+	 * {@code
+	 * HttpServer.create()
+	 *           .port(8080)
+	 *           .route(r -> r.get("/hello",
+	 *                   (req, res) -> res.header(CONTENT_TYPE, TEXT_PLAIN)
+	 *                                    .sendString(Mono.just("Hello World!"))))
+	 *           .errorLog(true, ErrorLogFactory.createFilter(
+	 * 								args -> args.cause() instanceof RuntimeException,
+	 * 								args -> ErrorLog.create("host-name={}", args.httpServerInfos().hostName())))
+	 *           .bindNow()
+	 *           .onDispose()
+	 *           .block();
+	 * }
+	 * </pre>
+	 * <p>
+	 * The {@link ErrorLogFactory} class offers several helper methods to generate such a function,
+	 * notably if one wants to {@link ErrorLogFactory#createFilter(Predicate) filter} some exceptions out of the error log.
+	 * <p>
+	 * Note that this method takes precedence over the {@value reactor.netty.ReactorNetty#ERROR_LOG_ENABLED} system property.
+	 *
+	 * @param enable enable or disable the error log
+	 * @param errorLogFactory the {@link ErrorLogFactory} that creates an {@link ErrorLog} given an {@link ErrorLogArgProvider}
+	 * @return a new {@link HttpServer}
+	 */
+	public final HttpServer errorLog(boolean enable, ErrorLogFactory errorLogFactory) {
+		Objects.requireNonNull(errorLogFactory, "errorLogFactory");
+		HttpServer dup = duplicate();
+		dup.configuration().errorLog = enable ? errorLogFactory : null;
+		dup.configuration().errorLogEnabled = enable;
 		return dup;
 	}
 
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
@@ -293,6 +293,7 @@ public abstract class HttpServer extends ServerTransport<HttpServer, HttpServerC
 	 * <p>
 	 *
 	 * Note that this method takes precedence over the {@value reactor.netty.ReactorNetty#ERROR_LOG_ENABLED} system property.
+	 * By default, error logs are formatted as "[{datetime}] [pid {pid}] [client {remote address}] {error message}".
 	 *
 	 * @param enable enable or disable the error log
 	 * @return a new {@link HttpServer}
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
@@ -75,6 +75,9 @@ import reactor.netty.http.server.compression.HttpCompressionOptionsSpec;
 import reactor.netty.http.server.logging.AccessLog;
 import reactor.netty.http.server.logging.AccessLogArgProvider;
 import reactor.netty.http.server.logging.AccessLogHandlerFactory;
+import reactor.netty.http.server.logging.error.DefaultErrorLogHandler;
+import reactor.netty.http.server.logging.error.ErrorLog;
+import reactor.netty.http.server.logging.error.ErrorLogArgProvider;
 import reactor.netty.resources.LoopResources;
 import reactor.netty.tcp.SslProvider;
 import reactor.netty.transport.ServerTransportConfig;
@@ -320,6 +323,8 @@ public final class HttpServerConfig extends ServerTransportConfig<HttpServerConf
 
 	boolean                                                 accessLogEnabled;
 	Function<AccessLogArgProvider, AccessLog>               accessLog;
+	boolean                                                 errorLogEnabled;
+	Function<ErrorLogArgProvider, ErrorLog>               errorLog;
 	HttpCompressionOptionsSpec                              compressionOptions;
 	BiPredicate<HttpServerRequest, HttpServerResponse>      compressPredicate;
 	ServerCookieDecoder                                     cookieDecoder;
@@ -497,6 +502,8 @@ public final class HttpServerConfig extends ServerTransportConfig<HttpServerConf
 	static void addStreamHandlers(Channel ch,
 			boolean accessLogEnabled,
 			@Nullable Function<AccessLogArgProvider, AccessLog> accessLog,
+            boolean errorLogEnabled,
+            @Nullable Function<ErrorLogArgProvider, ErrorLog> errorLog,
 			@Nullable HttpCompressionOptionsSpec compressionOptions,
 			@Nullable BiPredicate<HttpServerRequest, HttpServerResponse> compressPredicate,
 			@Nullable Boolean connectProtocolEnabled,
@@ -574,6 +581,10 @@ public final class HttpServerConfig extends ServerTransportConfig<HttpServerConf
 			}
 		}
 
+		if (errorLogEnabled) {
+			pipeline.addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.ErrorLogHandler, new DefaultErrorLogHandler(errorLog));
+		}
+
 		if (log.isDebugEnabled()) {
 			log.debug(format(ch, "Initialized HTTP/2 stream pipeline {}"), pipeline);
 		}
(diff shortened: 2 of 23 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -574,6 +581,10 @@ else if (metricsRecorder instanceof ContextAwareHttpServerMetricsRecorder) {
 			}
 		}
 
+		if (errorLogEnabled) {
+			pipeline.addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.ErrorLogHandler, new DefaultErrorLogHandler(errorLog));
+		}
```
raccoonback, line 586: Register the `ErrorLogHandler` before the `ReactiveBridge` in HTTP/2 protocol..

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -802,12 +819,18 @@ else if (metricsRecorder instanceof ContextAwareHttpServerMetricsRecorder) {
 				}
 			}
 		}
+
+		if (errorLogEnabled) {
+			p.addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.ErrorLogHandler, new DefaultErrorLogHandler(errorLog));
+		}
```
raccoonback, line 825: Register the `ErrorLogHandler` before the `ReactiveBridge` in HTTP/1.1 protocol..

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -877,6 +900,10 @@ else if (metricsRecorder instanceof ContextAwareHttpServerMetricsRecorder) {
...
```
raccoonback, line 906: Register the `ErrorLogHandler` before the `ReactiveBridge`.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -802,12 +819,18 @@ else if (metricsRecorder instanceof ContextAwareHttpServerMetricsRecorder) {
...
 	}
 
 	@SuppressWarnings("deprecation")
 	static void configureHttp11Pipeline(ChannelPipeline p,
 			boolean accessLogEnabled,
 			@Nullable Function<AccessLogArgProvider, AccessLog> accessLog,
+            boolean errorLogEnabled,
+            @Nullable Function<ErrorLogArgProvider, ErrorLog> errorLog,
```
violetagg, line 833: Formatting

```suggestion
			boolean errorLogEnabled,
			@Nullable Function<ErrorLogArgProvider, ErrorLog> errorLog,
```

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -1022,6 +1049,8 @@ static final class H2Codec extends ChannelInitializer<Channel> {
 
 		final boolean                                                 accessLogEnabled;
 		final Function<AccessLogArgProvider, AccessLog>               accessLog;
+		final boolean                                                 errorLogEnabled;
+		final Function<ErrorLogArgProvider, ErrorLog>               errorLog;
```
violetagg, line 1053: Formatting

```suggestion
		final Function<ErrorLogArgProvider, ErrorLog>                 errorLog;
```

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -1096,6 +1129,8 @@ static final class Http11OrH2CleartextCodec extends ChannelInitializer<Channel>
...
```
violetagg, line 1133: Formatting

```suggestion
		final Function<ErrorLogArgProvider, ErrorLog>                 errorLog;
```

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -1233,6 +1272,8 @@ static final class H2OrHttp11Codec extends ApplicationProtocolNegotiationHandler
...
```
violetagg, line 1276: Formatting

```suggestion
		final Function<ErrorLogArgProvider, ErrorLog>                 errorLog;
```

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -1325,6 +1369,8 @@ static final class HttpServerChannelInitializer implements ChannelPipelineConfig
...
```
violetagg, line 1373: Formatting

```suggestion
		final Function<ErrorLogArgProvider, ErrorLog>                 errorLog;
```

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/DefaultErrorLogArgProvider.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Provides default information for logging errors that occur on the HTTP Server.
+ *
+ * @author raccoonback
+ */
+final class DefaultErrorLogArgProvider extends AbstractErrorLogArgProvider<DefaultErrorLogArgProvider> {
+
+	private Throwable cause;
+
+	DefaultErrorLogArgProvider(@Nullable SocketAddress remoteAddress) {
```
violetagg, line 36: In `AbstractErrorLogArgProvider` the `remoteAddress` is not marked as `Nullable`, but here it is?

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/DefaultErrorLogArgProvider.java
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	@Override
+	public Throwable cause() {
+		return cause;
+	}
+
+	public void clear() {
+		cause = null;
+	}
```
violetagg, line 52:
```suggestion
	void clear() {
		cause = null;
		errorDateTime = null;
		httpServerInfos = null;
	}
```

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/DefaultErrorLogHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+import io.netty.channel.ChannelHandlerContext;
+import reactor.util.annotation.Nullable;
+
+import java.util.function.Function;
+
+/**
+ * Handler for logging errors that occur in the HTTP Server.
+ *
+ * @author raccoonback
```
violetagg, line 26:
```suggestion
 * @author raccoonback
 * @since 1.2.5
```

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/ErrorLog.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Log the http error information into a Logger named {@code reactor.netty.http.server.ErrorLog} at ERROR level.
+ * <p>
+ * See {@link ErrorLogFactory} for convenience methods to create an access log factory to be passed to
+ * {@link reactor.netty.http.server.HttpServer#errorLog(boolean, ErrorLogFactory)} during server configuration.
+ *
+ * @author raccoonback
+ */
+public final class ErrorLog {
```
violetagg, line 31: Recently there was comment that `AccessLog` is final class which prevents the extensibility. May be we can consider the comment here https://github.com/reactor/reactor-netty/issues/781#issuecomment-2776094005 and make the `ErrorLog` interface?

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/ErrorLogArgProvider.java
```diff
@@ -0,0 +1,64 @@
...
+import reactor.util.annotation.Nullable;
+
+import java.net.SocketAddress;
+import java.time.ZonedDateTime;
+import java.util.function.BiFunction;
+
+/**
+ * Provides information for logging errors that occur on the HTTP Server.
+ *
+ * @author raccoonback
```
violetagg, line 29:
```suggestion
 * @author raccoonback
 * @since 1.2.5
```

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/ErrorLogFactory.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+/**
+ * An interface to declare more concisely a {@link Function} that apply an {@link ErrorLog} by an
+ * {@link ErrorLogArgProvider}.
+ * <p>
+ * Can be used in {@link reactor.netty.http.server.HttpServer#errorLog(boolean, ErrorLogFactory) errorLog} method for example.
+ *
+ * @author raccoonback
```
violetagg, line 27:
```suggestion
 * @author raccoonback
 * @since 1.2.5
```

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/error/ErrorLogTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static reactor.netty.http.server.logging.error.ErrorLog.LOGGER;
+
+/**
+ * This test class verifies {@link DefaultErrorLogHandler}.
+ *
+ * @author raccoonback
+ */
+class ErrorLogTest extends BaseHttpTest {
```
violetagg, line 40: I would like to check all three HTTP versions: 1.1, 2 and 3
May be instead of separate test class we can make something similar to
reactor.netty.http.HttpProtocolsTests#testAccessLog
reactor.netty.http.HttpProtocolsTests#testAccessLogWithForwardedHeader
reactor.netty.http.Http3Tests#testAccessLog
reactor.netty.http.Http3Tests#testAccessLogWithForwardedHeader

violetagg (changes requested):

I have some minor comments and questions. Please check them.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -270,6 +273,74 @@ public final HttpServer accessLog(Function<AccessLogArgProvider, AccessLog> acce
...
+	 *
+	 * @param enable enable or disable the error log
+	 * @return a new {@link HttpServer}
+	 */
+	public final HttpServer errorLog(boolean enable) {
+		HttpServer dup = duplicate();
+		dup.configuration().errorLog = null;
+		dup.configuration().errorLogEnabled = enable;
+		return dup;
+	}
```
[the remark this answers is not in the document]
violetagg: May be we can add to the javadoc the information for the format?

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/BaseErrorLogHandler.java
```diff
@@ -0,0 +1,59 @@
...
+import io.netty.channel.ChannelDuplexHandler;
+import reactor.util.annotation.Nullable;
+
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import java.util.function.Function;
+
+class BaseErrorLogHandler extends ChannelDuplexHandler {
+
+	static final String DEFAULT_LOG_FORMAT = "[{}] {}";
```
violetagg, line 27: I think that the format that we need is a bit different?
https://httpd.apache.org/docs/2.4/logs.html
https://httpd.apache.org/docs/2.4/mod/core.html#errorlogformat
May be something like this

```
	static String PID;
	static {
		String jvmName = ManagementFactory.getRuntimeMXBean().getName();
		PID = jvmName.split("@")[0];
	}

	static final String DEFAULT_LOG_FORMAT = "[{}] [pid " + PID + "] [client {}] {}";
	static final DateTimeFormatter DATE_TIME_FORMATTER =
			DateTimeFormatter.ofPattern("dd/MMM/yyyy:HH:mm:ss Z");
	static final String MISSING = "-";

	static final Function<ErrorLogArgProvider, ErrorLog> DEFAULT_ERROR_LOG =
			args -> ErrorLog.create(
					DEFAULT_LOG_FORMAT,
					args.errorDateTime().format(DATE_TIME_FORMATTER),
					refinedRemoteAddress(args.remoteAddress()),
					refinedExceptionMessage(args.cause()));
```

```
	private static String refinedExceptionMessage(Throwable throwable) {
		String error = throwable.getClass().getName();
		String message = throwable.getLocalizedMessage();
		return message == null ? error : error + "." + message;
	}
```

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/error/ErrorLogTest.java
```diff
@@ -0,0 +1,197 @@
...
+		ROOT.addAppender(mockedAppender);
+	}
+
+	@AfterEach
+	void tearDown() {
+		ROOT.detachAppender(mockedAppender);
+	}
+
+	@Test
+	void errorLogDefaultFormat() {
```
violetagg, line 63: Let say that the error is not triggered by the Netty pipeline but because of the reactive pipeline.

```
	@Test
	void errorLogDefaultFormat() {
		disposableServer = createServer()
				.handle((req, resp) -> Mono.error(new RuntimeException()))
				.errorLog(true)
				.bindNow();

		getHttpClientResponse("/example/test");

		Mockito.verify(mockedAppender, Mockito.times(1))
				.doAppend(loggingEventArgumentCaptor.capture());
		assertThat(loggingEventArgumentCaptor.getAllValues()).hasSize(1);

		LoggingEvent relevantLog = loggingEventArgumentCaptor.getAllValues().get(0);
		assertThat(relevantLog.getMessage())
				.isEqualTo(BaseErrorLogHandler.DEFAULT_LOG_FORMAT);
		assertThat(relevantLog.getFormattedMessage())
				.matches("\\[([0-9a-fA-F:.]+)(:\\d)*] -");
	}
```

Then we will miss the error. May be instead of using custom `ChannelHandler` we can use `reactor.netty.ConnectionObserver` and `onUncaughtException`. It should handle all errors but this has to be tested.

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/error/ErrorLogTest.java
raccoonback: @violetagg 
That's a great point.
I will reflect it!

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/BaseErrorLogHandler.java
raccoonback: I will reflect it!

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/DefaultErrorLogArgProvider.java
raccoonback: @violetagg 
I missed `Nullable` in `remoteAddress` of `AbstractErrorLogArgProvider`. 
I'll add it.

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/ErrorLog.java
raccoonback: @violetagg 
Yes, that's a good idea. 
I'll consider using ErrorLog as an interface.

reactor-netty-http/src/test/java/reactor/netty/http/server/logging/error/ErrorLogTest.java
raccoonback: @violetagg 
Thanks for the suggestion!
I tried testing it with the test case you shared. 
From what I observed, it seems that `ConnectionObserver.onUncaughtException()` is only triggered for exceptions related to the connection itself.


Do you happen to know if there's a way to catch exceptions that occur within the reactive pipeline as well?

violetagg: let me check

violetagg: ok for this use case we need something around here `reactor.netty.http.server.HttpServerOperations#onOutboundError`, this will handle error with reactive pipeline.

violetagg: For the use case 

```
		disposableServer = createServer()
				.handle((req, resp) -> {throw new RuntimeException();})
				.errorLog(true)
				.bindNow();
```

We need something around here `reactor.netty.http.server.HttpServer.HttpServerHandle#onStateChange`

raccoonback: @violetagg 
I couldn’t fully understand all the details you mentioned, so I’d like to check if I got it right! 😊

Is your suggestion to explore whether we can catch errors that occur in the **reactive pipeline** using either `reactor.netty.http.server.HttpServer.HttpServerHandle#onStateChange` or `reactor.netty.http.server.HttpServerOperations#onOutboundError`?

violetagg: I mean Reactor Netty logs error only on very few places and I think we need to cover them. For example the two tests above that I shared need changes in both `reactor.netty.http.server.HttpServer.HttpServerHandle#onStateChange` and `reactor.netty.http.server.HttpServerOperations#onOutboundError`

raccoonback: @violetagg 
Thanks for the clarification!
I agree with the points you mentioned.
I’ll review each of the use cases you shared and consider how to handle error logging accordingly.

raccoonback: @violetagg 
Exceptions thrown in the reactive pipeline, including unhandled errors, are now caught in `reactor.netty.http.server.HttpServer.HttpServerHandle#onStateChange` and `reactor.netty.http.server.HttpServerOperations#onOutboundError`, and are forwarded as user events.
A custom handler receives the event via `userEventTriggered` and logs the error accordingly.

raccoonback:
@violetagg 
Thank you for the review!
I've reflected all of your comments!

reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/DefaultErrorLogHandler.java
```diff
@@ -36,17 +39,25 @@ public DefaultErrorLogHandler(@Nullable Function<ErrorLogArgProvider, ErrorLog>
 
 	@Override
 	public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
-		if (errorLogArgProvider == null) {
-			errorLogArgProvider = new DefaultErrorLogArgProvider(ctx.channel().remoteAddress());
+		ErrorLog log;
```
violetagg, line 42: Can you elaborate more on this change, why do we need to lock?

raccoonback: @violetagg 

Great point!

In our architecture, `exceptionCaught()` can be triggered from both **Netty and Reactive pipelines**. For the Reactive pipeline, we've routed exception signals through `userEventTriggered()`, which then delegates to `exceptionCaught()`.

This setup allows for the possibility of `exceptionCaught()` being called concurrently from different threads—especially under load or when exceptions happen on both sides simultaneously.

`errorLogArgProvider` is a lazily initialized and reused mutable object that is not thread-safe. Without proper synchronization, concurrent access can lead to race conditions or inconsistent internal state. To mitigate this, a `ReentrantLock` is introduced to guard access to the shared instance.

violetagg: Netty guarantees that we will run always on an event loop, so I think that it is thread-safe, wdyt?

raccoonback: @violetagg 
Just to confirm — are you saying that `exceptionCaught()` or `userEventTriggered()` would never be triggered concurrently because they're always invoked via the event loop?

The reason I asked is that I was concerned about a case where `exceptionCaught()` is being handled in the Netty pipeline, and at the same time, an exception is raised in the Reactive pipeline — which results in `userEventTriggered()` being triggered, and that in turn also calls `exceptionCaught()`.

https://github.com/reactor/reactor-netty/blob/3042ce0f3e1c727d5fd54033198c960121ebb441/reactor-netty-http/src/main/java/reactor/netty/http/server/logging/error/DefaultErrorLogHandler.java#L68-L75

violetagg: In the reactive pipeline we call `connection.channel().pipeline().fireUserEventTriggered(new DefaultErrorLoggingEvent(t));`
the `pipeline` is implemented in a a way to always check whether we are in an event loop or not, see here
https://github.com/netty/netty/blob/cea9a5f0c7bc2cf512c6d7e13e3d9ee33776cfc6/transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java#L861
https://github.com/netty/netty/blob/cea9a5f0c7bc2cf512c6d7e13e3d9ee33776cfc6/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java#L372

So I think we should be ok?

raccoonback: @violetagg 
I understand.
Thank you for the detailed explanation!
I'll go ahead and remove the `ReentrantLock`.

raccoonback:
@violetagg 
I’ve applied the review feedback!

violetagg (approved):
Thanks for the PR!

PR #3772, opened 2025-05-19
POC - Measure How Long we are waiting for a thread

mateusz-nalepa:
Proof of Concept for issue https://github.com/reactor/reactor-netty/issues/3773

# Run tests from class `WaitingTest`

### Scenario 1: waitingForThreadTest

WAITING TEST ### DisposableAcquire started after: PT0.175384375S
WAITING TEST ### Blocking operation with id: id3 took: PT3.220496667S

WAITING TEST ### DisposableAcquire started after: PT3.183820333S
WAITING TEST ### Blocking operation with id: id1 took: PT6.228813792S

WAITING TEST ### DisposableAcquire started after: PT6.189854083S
WAITING TEST ### Blocking operation with id: id2 took: PT9.237844209S

### Scenario 2: waitingForThreadTestButThereIsPublishOn

WAITING TEST ### DisposableAcquire started after: PT0.00924375S
WAITING TEST ### DisposableAcquire started after: PT0.011155042S
WAITING TEST ### DisposableAcquire started after: PT0.011364792S

WAITING TEST ### Blocking operation with id: id3 took: PT3.01875575S
WAITING TEST ### Blocking operation with id: id2 took: PT3.018755541S
WAITING TEST ### Blocking operation with id: id1 took: PT3.020641291S


violetagg:
This proposal is not quite accurate.The time between creation of `DisposableAquire` and `onNext`/`run` doesn't mean that we wait. `onNext`/`run` is invoked when we have the connection that's true, but we have two scenarios - one when we have a connection in the connection pool and one when we don't have connections in the pool. When we are in the second use case this means that before delivering the connection we need to resolve the remote IP, establish a connection and eventually perform TLS handshake, so it is not that we don't do anything.

violetagg:
I think this is not what #3773 targets

Issue #3079, opened 2024-03-04
Support For Spnego Auth scheme support for netty HttpClient similar to Apache's


luckysardana:
We need to connect to some component via proxy, and this proxy works by kerberos auth. We need to do something like below:
```

CredentialsProvider credsr = new BasicCredentialsProvider();
    credsProvider.setCredentials(new AuthScope(null, -1, null), use_jaas_creds);
    Registry<AuthSchemeProvider> authSchemeRegistry = RegistryBuilder.<AuthSchemeProvider>create().register(AuthSchemes.SPNEGO, new SPNegoSchemeFactory(true)).build();
    CloseableHttpClient httpclient = HttpClients.custom()
        // set our proxy - httpclient doesn't use ProxySelector
        .setRoutePlanner(new DefaultProxyRoutePlanner(proxy))
        .setDefaultAuthSchemeRegistry(authSchemeRegistry)
        .setDefaultCredentialsProvider(creds).build();


```

We are not seeing any solution to run something like above Please help

naveenyamsani:
Can I please get the confirmation that it is a new enhancement or feature is already existed with reactor-netty HttpClient

naveenyamsani:
@violetagg do we have support For Spnego Auth scheme support for netty HttpClient similar to Apache's?

violetagg:
> @violetagg is there a support For Spnego Auth scheme with netty HttpClient similar to Apache's?

No

raccoonback:
@violetagg 
Hello,

I'm wondering if there are any plans to support this feature in Reactor Netty.
If so, would it be alright for me to look into it?

Thank you!

PR #3813, opened 2025-06-23
Support SPNEGO (Kerberos) Authentication in HttpClient

raccoonback:
## Motivation
This PR adds support for SPNEGO (Kerberos) authentication to HttpClient, addressing https://github.com/reactor/reactor-netty/issues/3079.
SPNEGO is widely used for HTTP authentication in enterprise environments, particularly those based on Kerberos.

### SpnegoAuthProvider
Provides SPNEGO authentication by generating a Kerberos-based token and attaching it to the `Authorization` header of outgoing HTTP requests.

### JaasAuthenticator
Provides a pluggable way to perform JAAS-based Kerberos login, making it easy to integrate with various authentication backends.

### HttpClient.spnego(...) API
Adds a new API to configure SPNEGO authentication for HttpClient instances.

```java
    .spnego(SpnegoAuthProvider.create(new JaasAuthenticator("KerberosLogin")));

client.get()
      .uri("http://protected.example.com/")
      .responseSingle((res, content) -> content.asString())
      .block();
```

#### jaas.conf
A JAAS configuration file in Java for integrating with authentication backends such as Kerberos.

```
KerberosLogin {
    com.sun.security.auth.module.Krb5LoginModule required
    client=true
    useKeyTab=true
    keyTab="/path/to/test.keytab"
    principal="test@EXAMPLE.COM"
    doNotPrompt=true
    debug=true;
};
```

#### krb5.conf
krb5.conf is a Kerberos client configuration file used to define how the client locates and communicates with the Kerberos Key Distribution Center (KDC) for authentication.

```
[libdefaults]
    default_realm = EXAMPLE.COM
[realms]
    EXAMPLE.COM = {
        kdc = kdc.example.com
    }
[domain_realms]
    .example.com = EXAMPLE.COM
    example.com = EXAMPLE.COM
```

## How It Works
- When a server responds with `401 Unauthorized` and a `WWW-Authenticate: Negotiate` header,
the client automatically generates a SPNEGO token using the Kerberos ticket and resends the request with the appropriate `Authorization` header.
- The implementation is based on Java's GSS-API and is compatible with standard Kerberos environments.

## Environment Configuration
Requires proper JAAS (jaas.conf) and Kerberos (krb5.conf) configuration.
See the updated documentation for example configuration files and JVM options.

## Additional Notes
- The `SpnegoAuthProvider` allows for easy extension and testing by supporting custom authenticators and GSSManager injection.
- The feature is fully compatible with **Java 1.6+** and works on both Unix and Windows environments.


raccoonback:
I tested Kerberos authentication using the krb5 available at https://formulae.brew.sh/formula/krb5.

reactor-netty-http/src/main/java/reactor/netty/http/client/JaasAuthenticator.java
```diff
@@ -0,0 +1,56 @@
...
+	 * @return the authenticated JAAS Subject
+	 * @throws LoginException if login fails
+	 */
+	@Override
+	public Subject login() throws LoginException {
+		LoginContext context = new LoginContext(contextName);
+		context.login();
+		return context.getSubject();
+	}
+}
```
raccoonback: It logs in using the specified JAAS login context name and returns the authenticated `Subject` object.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -446,6 +446,15 @@ public Context currentContext() {
...
 			if (newState == HttpClientState.RESPONSE_RECEIVED) {
+				HttpClientOperations operations = connection.as(HttpClientOperations.class);
+				if (operations != null && handler.spnegoAuthProvider != null) {
+					int statusCode = operations.status().code();
+					HttpHeaders headers = operations.responseHeaders();
+					if (handler.spnegoAuthProvider.isUnauthorized(statusCode, headers)) {
+						handler.spnegoAuthProvider.invalidateCache();
+					}
+				}
+
```
raccoonback: If a SPNEGO authentication expiration response is received via the HTTP response, the authentication header value is cleared.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -540,6 +552,19 @@ public SocketAddress get() {
...
+							Mono.defer(
+									() -> Mono.from(requestWithBodyInternal(ch))
+							)
+					);
+			}
+
+			return requestWithBodyInternal(ch);
+		}
+
+		private Publisher<Void> requestWithBodyInternal(HttpClientOperations ch) {
```
raccoonback: If SPNEGO authentication is configured through the HttpClient, authentication is attempted before sending the request.

reactor-netty-http/src/main/java/reactor/netty/http/client/SpnegoAuthProvider.java
```diff
@@ -0,0 +1,195 @@
...
+	 * @param request the HTTP client request to authenticate
+	 * @param address the target server address (used for service principal)
+	 * @return a Mono that completes when the authentication is applied
+	 * @throws RuntimeException if login or token generation fails
+	 */
+	public Mono<Void> apply(HttpClientRequest request, InetSocketAddress address) {
+		if (verifiedAuthHeader != null) {
+			request.header(HttpHeaderNames.AUTHORIZATION, verifiedAuthHeader);
+			return Mono.empty();
+		}
```
raccoonback: If a valid authentication token already exists, it is reused.

reactor-netty-http/src/main/java/reactor/netty/http/client/SpnegoAuthProvider.java
```diff
@@ -0,0 +1,195 @@
...
+							}
+						}
+					);
+				}
+				catch (LoginException e) {
+					throw new RuntimeException("Failed to login with SPNEGO", e);
+				}
+			})
+			.subscribeOn(boundedElastic())
+			.then();
```
raccoonback: After obtaining Kerberos credentials via JAAS login, a SPNEGO token for the target server is generated using the GSS-API, and the `Authorization: Negotiate <token>` header is added to the request.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -446,6 +447,16 @@ public Context currentContext() {
 		@Override
 		public void onStateChange(Connection connection, State newState) {
 			if (newState == HttpClientState.RESPONSE_RECEIVED) {
+				HttpClientOperations operations = connection.as(HttpClientOperations.class);
+				if (operations != null && handler.spnegoAuthProvider != null) {
+					if (shouldRetryWithSpnego(operations)) {
```
raccoonback: Re-authenticates with SPNEGO if necessary.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -719,6 +782,9 @@ public boolean test(Throwable throwable) {
 				redirect(re.location);
 				return true;
 			}
+			if (throwable instanceof SpnegoRetryException) {
+				return true;
+			}
```
raccoonback: Configures it to retry even if a `SpnegoRetryException` occurs.

raccoonback:
@violetagg 
Hello!
Please check this PR when you have a chance. 😃

wendigo:
This is so great! Looking forward to get this in :)

wendigo:
I can provide some guidance around APIs and configuration. Not every kerberos-enabled client uses JAAS, therefore the direct Subject/SPNEGO token support should be provided

raccoonback:
@wendigo 
Hello!
Thank you for the great point. 😀

I was thinking of allowing users to implement the [`SpnegoAuthenticator`](https://github.com/reactor/reactor-netty/pull/3813/files#diff-3566db025cfb3362b553d31a6dc304b5676e20bd1410ef65652f00b98cf32413) interface, similar to [`JaasAuthenticator`](https://github.com/reactor/reactor-netty/pull/3813/files#diff-3380c0044075ef6ff38ccb48fe29c85b692f5d3ee8af1f16a73a21811c685218), to support custom authentication logic if needed.

If I understood you correctly, you're suggesting that we should provide a way for users to directly supply a Subject, as in the example below:

```java
public class DirectSubjectAuthenticator implements SpnegoAuthenticator {

    // ...
    private Subject subject;

    @Override
    public Subject login() throws LoginException {
        return subject;
    }

    // ...
}
````

Would you be able to share a more concrete example or use case?
It would help refine the design in the right direction.

**Comment by wendigo** [comment]
Sure @raccoonback.

I'd like to use reactor-netty in the [trino](https://github.com/trinodb/trino) CLI/JDBC/client libraries.

We support delegated/constrained/unconstrained kerberos authentication. Relevant code is here:

https://github.com/trinodb/trino/tree/master/client/trino-client/src/main/java/io/trino/client/auth/kerberos

This is how we add it to the okhttp: https://github.com/trinodb/trino/blob/master/client/trino-client/src/main/java/io/trino/client/auth/kerberos/SpnegoHandler.java

Configurability is important as we expose configuration that allows the user to pass remote service name, service principal name, whether to canonicalize hostname: https://github.com/trinodb/trino/blob/master/client/trino-client/src/main/java/io/trino/client/auth/kerberos/SpnegoHandler.java#L50C5-L54C48

**Comment by raccoonback** [comment]
@violetagg 
I think supporting not only JAAS-based authentication but also allowing the user to provide a GSSCredential directly could improve configurability and flexibility.
This would be especially useful in environments where JAAS is not preferred or where credentials need to be managed programmatically.
What do you think about this direction?

cc. @wendigo

**Comment by violetagg** [comment]
I'm currently on vacation. When I return I'll check it.

На нд, 27.07.2025 г. в 18:21 KOSEUNGBIN ***@***.***> написа:

**Comment by raccoonback** [comment]
@wendigo 
I've added `GSSCredential-based SPNEGO authentication` and support for `service name` and `canonical hostname` configuration. 
Thank you for the suggestion.

File: reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java [hunk 8]
```diff
@@ -459,6 +470,42 @@ public void onStateChange(Connection connection, State newState) {
...
+		 * @param operations the HTTP client operations containing the response status and headers
+		 * @return {@code true} if SPNEGO re-authentication should be attempted, {@code false} otherwise
+		 */
+		private boolean shouldRetryWithSpnego(HttpClientOperations operations) {
+			int statusCode = operations.status().code();
+			HttpHeaders headers = operations.responseHeaders();
+
+			return handler.spnegoAuthProvider.isUnauthorized(statusCode, headers)
+				&& handler.spnegoAuthProvider.canRetry();
+		}
```
raccoonback: If the resource server responds with a requirement for SPNEGO authentication, the client will retry the authentication.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -459,6 +470,42 @@ public void onStateChange(Connection connection, State newState) {
...
+		private void retryWithSpnego(HttpClientOperations operations) {
+			handler.spnegoAuthProvider.invalidateTokenHeader();
+			handler.spnegoAuthProvider.incrementRetryCount();
+
+			if (log.isDebugEnabled()) {
+				log.debug(format(operations.channel(), "Triggering SPNEGO re-authentication"));
+			}
+
+			sink.error(new SpnegoRetryException());
+		}
```
raccoonback: The previously issued `SPNEGO Negotiate Token` is cleared, and a reissue attempt is made.

reactor-netty-http/src/main/java/reactor/netty/http/client/JaasAuthenticator.java
```diff
@@ -0,0 +1,92 @@
...
+			GSSContext context = manager.createContext(
+					serverName,
+					new Oid("1.3.6.1.5.5.2"), // SPNEGO
+					null,
+					GSSContext.DEFAULT_LIFETIME
+			);
+			context.requestMutualAuth(true);
+			return context;
+		});
+	}
```
raccoonback: Authentication is performed based on JAAS.

reactor-netty-http/src/main/java/reactor/netty/http/client/SpnegoAuthProvider.java
```diff
@@ -0,0 +1,315 @@
...
+
+		/**
+		 * Builds the SpnegoAuthProvider instance.
+		 *
+		 * @return a new SpnegoAuthProvider
+		 */
+		public SpnegoAuthProvider build() {
+			return new SpnegoAuthProvider(authenticator, unauthorizedStatusCode, serviceName, resolveCanonicalHostname);
+		}
+	}
```
raccoonback: The HTTP client receives the options required for SPNEGO authentication:

1. HTTP status to catch the authentication required error response from the resource server
2. Service principal name
3. Whether to handle canonical hostname

reactor-netty-http/src/main/java/reactor/netty/http/client/SpnegoAuthProvider.java
```diff
@@ -0,0 +1,315 @@
...
+	 * @param address the socket address to resolve hostname from
+	 * @return the resolved hostname (canonical if configured, otherwise standard hostname)
+	 */
+	private String resolveHostName(InetSocketAddress address) {
+		String hostName = address.getHostName();
+		if (resolveCanonicalHostname) {
+			hostName = address.getAddress().getCanonicalHostName();
+		}
+		return hostName;
+	}
```
raccoonback: It is handled as the canonical hostname.

reactor-netty-http/src/main/java/reactor/netty/http/client/SpnegoAuthProvider.java
```diff
@@ -0,0 +1,315 @@
...
+				}
+				catch (GSSException e) {
+					// Log but don't propagate disposal errors
+					if (log.isDebugEnabled()) {
+						log.debug("Failed to dispose GSSContext", e);
+					}
+				}
+			}
+		}
+	}
```
raccoonback: Generates a SPNEGO Negotiate token for the given host name.

raccoonback:
@violetagg
Hello.
I would appreciate it if you could review this PR.
Thanks!

violetagg:
> @violetagg Hello. I would appreciate it if you could review this PR. Thanks!

I'm just returning fro vacation, will check it in the next days or so

raccoonback:
@violetagg
Hello.
I would appreciate it if you could review this PR.
Thanks!

violetagg:
I will check this one ... just need to finalise some other tasks.

violetagg:
@raccoonback We discussed to introduce this feature for 1.3.0 version but it is based on 1.2.x branch. Which version are we targeting?

raccoonback:
@violetagg 
I think I got confused earlier.
As you mentioned, I’ll make the changes based on version **1.3.x**.
ref. https://github.com/reactor/reactor-netty/issues/3079#issuecomment-2893713964

raccoonback:
@violetagg 
I couldn’t find the **1.3.x** branch — would it be okay if I open the PR based on the **main** branch instead?

violetagg:
> @violetagg I couldn’t find the **1.3.x** branch — would it be okay if I open the PR based on the **main** branch instead?

yes `main` is the branch that is for `1.3.x` version

raccoonback:
@violetagg 
I’ve changed the target branch to **1.3.x**.
Please review PR.

violetagg:
After following the discussion here and having in mind the Reactor Netty API, I think that the solution should be a bit more generic. Also IMO, Reactor Netty is responsible to provide the infrastructure while the user - to provide the concrete implementation (for example how the token will be generated, cached etc).
My proposal is the following (similar to how follow redirect is implemented)

```
HttpClient.httpAuthentication(
			BiPredicate<HttpClientRequest, HttpClientResponse> authenticationPredicate,
			Consumer<HttpClientRequest> authenticationConsumer)
```

And the usage will be 

```
HttpClient.create()
			.httpAuthentication(
					(request, response) ->
						return response.status().code() == 401 &&
								response.responseHeaders().contains("WWW-Authenticate", "Negotiate", true)
					,
					request -> {
							String token = getOrGenerateToken(request.resourceUrl());
							request.header("Authorization", "Negotiate " + token);
					}
			);
```

violetagg:
The API is a bit rough, if we have a delayed token generation we may need the authenticationConsumer to return Mono that we can execute before sending the request. What do you think?

raccoonback:
@violetagg 
Thank you for the review. 😀
I have a few questions.

1. Regarding the **token issuance** and **re-authentication behavior** you mentioned — should we let users define only those parts?
Or do you also think users should define the handling of components like `GssCredentialAuthenticator`?

2. Currently, the authentication header value is stored and reused for subsequent HTTP requests.
Do you also expect this kind of handling to be managed through user-defined logic?

raccoonback:
@violetagg 
I initially considered this as a feature addition specifically for SPNEGO authentication...
However, are you suggesting adding only the authentication flow itself, without including the detailed SPNEGO implementation or dependencies ~~?

violetagg:
> @violetagg Thank you for the review. 😀 I have a few questions.
> 
> 1. Regarding the **token issuance** and **re-authentication behavior** you mentioned — should we let users define only those parts?
>    Or do you also think users should define the handling of components like `GssCredentialAuthenticator`?
> 2. Currently, the authentication header value is stored and reused for subsequent HTTP requests.
>    Do you also expect this kind of handling to be managed through user-defined logic?

I would like the user to be responsible for the token generation (thus our API will work not only for SPNEGO) and also the decision how the token will be used/reused, what kind of cache that will be used, what will be the expiration time etc., it is not Reactor Netty decision but user's.

violetagg:
> @violetagg I initially considered this as a feature addition specifically for SPNEGO authentication... However, are you suggesting adding only the authentication flow itself, without including the detailed SPNEGO implementation or dependencies ~~?

yes, I understand
however I think that if we try to provide a generic API, later it will be able to be used not only with SPNEGO

We can show case in the examples various http authentications

raccoonback:
@violetagg 
I’ll update it according to the direction you suggested.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -456,6 +457,17 @@ public Context currentContext() {
...
+				if (operations != null && handler.authenticationPredicate != null) {
+					if (handler.authenticationPredicate.test(operations, operations)) {
+						if (log.isDebugEnabled()) {
+							log.debug(format(operations.channel(), "Authentication predicate matched, triggering retry"));
+						}
+						sink.error(new HttpClientAuthenticationException());
+						return;
+					}
+				}
+
```
raccoonback: It propagates an exception to trigger a retry, when the authentication requirements are met.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -550,6 +568,19 @@ public SocketAddress get() {
...
+				return authenticator.apply(ch, ch.address())
+					.then(
+							Mono.defer(
+									() -> Mono.from(requestWithBodyInternal(ch))
+							)
+					);
+			}
+
+			return requestWithBodyInternal(ch);
+		}
```
raccoonback: The authenticator is executed first and then the HTTP request is attempted, when a retry is triggered due to HTTP authentication.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -728,6 +759,11 @@ public boolean test(Throwable throwable) {
 				redirect(re.location);
 				return true;
 			}
+			if (throwable instanceof HttpClientAuthenticationException) {
+				// Set flag to trigger authenticator on retry
+				authenticationAttempted = true;
+				return true;
+			}
```
raccoonback, line 759: The HTTP request is retried, when the HTTP authentication requirements are met.

raccoonback:
@violetagg 
I’ve improved it in the guide you suggested.
Please review it!

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1696,6 +1696,50 @@ public final HttpClient wiretap(boolean enable) {
...
+	 * </pre>
+	 *
+	 * @param predicate determines when authentication should be applied, receives the request
+	 *                  and response to decide if authentication is needed (e.g., check for 401 status)
+	 * @param authenticator applies authentication to the request, receives the request and remote address,
+	 *                      returns a Mono that completes when authentication is applied
+	 * @return a new {@link HttpClient}
+	 * @since 1.3.0
+	 */
+	public final HttpClient httpAuthentication(
```
violetagg, line 1794: Let's have two APIs (similar to what we have already)
- httpAuthentication - this will provide just a BiConsumer
- httpAuthenticationWhen - this will use the current signature

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
violetagg: It is better to use the mechanism that we use for follow redirect. reactor.netty.http.client.HttpClientOperations#onInboundNext -> see how "redirecting" is used. The idea is that once we decide to retry we will need to take care of this connection, which means we drain the incoming messages, no message goes to the end user etc  (we don't know whether the server will send only the headers or also some message). Basically we need to return to the connection pool a clean connection. Also because we don't send any message to the end user we can switch to auto-read = true

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
violetagg: Isn't it too early here to provide the HttpClientOperations? HttpClientOperations is still empty and we need to configure it which happens in the new method requestWithBodyInternal.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientAuthenticationException.java
```diff
@@ -0,0 +1,36 @@
...
+ * <p>
+ * This exception is used internally by the generic HTTP authentication framework
+ * to signal that the current request requires authentication and should be retried.
+ * The framework will invoke the configured authenticator before retrying the request.
+ * </p>
+ *
+ * @author Oliver Ko
+ * @since 1.3.0
+ */
+final class HttpClientAuthenticationException extends RuntimeException {
```
violetagg, line 29: What do you think about omitting the stack trace similar to RedirectClientException?

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
raccoonback: @violetagg
Since the decision to apply the authenticator is made using a `BiPredicate<HttpClientRequest, HttpClientResponse> predicate`, I think that `httpAuthentication` requires both a `BiPredicate` and a `BiConsumer`.
What do you think?

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientAuthenticationException.java
raccoonback: @violetagg
Okay, I will omit the stack trace!
I have one question: is there a particular reason for omitting the stack trace?

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
raccoonback: @violetagg I'll check it.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
raccoonback: That makes sense.
I will update it so that the authenticator is applied after the configuration.
`Configuration → apply authenticator → request`.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
violetagg: yes

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientAuthenticationException.java
violetagg: Creating the stack trace costs, and here we would like just to use this exception as a signal to retry the request. Also there is no need of the stack trace i.e. it is not user facing exception.

raccoonback:
@violetagg 
Should this feature also be considered in `Http2WebsocketClientOperations` and `WebsocketClientOperations`?
Since **redirecting** is already handled in those Operations as well, I wanted to check.

violetagg:
> @violetagg Should this feature also be considered in `Http2WebsocketClientOperations` and `WebsocketClientOperations`? Since **redirecting** is already handled in those Operations as well, I wanted to check.

yes

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1696,6 +1696,84 @@ public final HttpClient wiretap(boolean enable) {
...
+	 * @param authenticator applies authentication to the request, receives the request and remote address,
+	 *                      returns a Mono that completes when authentication is applied
+	 * @return a new {@link HttpClient}
+	 * @since 1.3.0
+	 * @see #httpAuthenticationWhen(BiPredicate, BiFunction)
+	 */
+	public final HttpClient httpAuthentication(
+			BiFunction<? super HttpClientRequest, ? super SocketAddress, ? extends Mono<Void>> authenticator) {
+		return httpAuthenticationWhen(HttpClientConfig.AUTHENTICATION_PREDICATE, authenticator);
+	}
```
raccoonback, line 1807: support HTTP authentication that retries on 401 Unauthorized responses.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1696,6 +1696,84 @@ public final HttpClient wiretap(boolean enable) {
...
+	public final HttpClient httpAuthenticationWhen(
+			BiPredicate<HttpClientRequest, HttpClientResponse> predicate,
+			BiFunction<? super HttpClientRequest, ? super SocketAddress, ? extends Mono<Void>> authenticator) {
+		Objects.requireNonNull(predicate, "predicate");
+		Objects.requireNonNull(authenticator, "authenticator");
+		HttpClient dup = duplicate();
+		dup.configuration().authenticationPredicate = predicate;
+		dup.configuration().authenticator = authenticator;
+		return dup;
+	}
```
raccoonback, line 1922: support HTTP authentication for the client with custom authentication logic and retry predicate.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientAuthenticationException.java
raccoonback: @violetagg it reflected!

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConfig.java
```diff
@@ -841,6 +845,9 @@ else if (metricsRecorder instanceof ContextAwareHttpClientMetricsRecorder) {
 			                                               .codeAsText())
 			                                   .matches();
 
+	static final BiPredicate<HttpClientRequest, HttpClientResponse> AUTHENTICATION_PREDICATE =
+			(req, res) -> res.status().equals(HttpResponseStatus.UNAUTHORIZED);
+
```
raccoonback: provide default `authentication predicated`.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -604,48 +615,57 @@ Publisher<Void> requestWithBody(HttpClientOperations ch) {
...
 
-				ch.redirectRequestConsumer(consumer);
-				if (handler != null) {
-					Publisher<Void> publisher = handler.apply(ch, ch);
-					return ch.equals(publisher) ? ch.send() : publisher;
-				}
-				else {
-					return ch.send();
-				}
+				return result;
```
raccoonback, line 670: I improved it so that the `authenticator` is set after initializing `HttpClientOperations`.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -848,7 +861,7 @@ protected void onInboundNext(ChannelHandlerContext ctx, Object msg) {
 				}
 			}
 			else {
-				// when redirecting no need of manual reading
+				// when redirecting or authenticating no need of manual reading
```
raccoonback, line 881: It enables auto-read when a redirect or authentication is required.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -401,6 +411,9 @@ protected void afterInboundComplete() {
 		if (redirecting != null) {
 			listener().onUncaughtException(this, redirecting);
 		}
+		else if (authenticating != null) {
+			listener().onUncaughtException(this, authenticating);
+		}
```
raccoonback, line 428: It propagates a `HttpClientAuthenticationException` when authentication is required and the request needs to be retried.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -903,15 +916,15 @@ protected void onInboundNext(ChannelHandlerContext ctx, Object msg) {
 			if (lastHttpContent != LastHttpContent.EMPTY_LAST_CONTENT) {
 				// When there is HTTP/2 response with INBOUND HEADERS(endStream=false) followed by INBOUND DATA(endStream=true length=0),
 				// Netty sends LastHttpContent with empty buffer instead of EMPTY_LAST_CONTENT
-				if (redirecting != null || lastHttpContent.content().readableBytes() == 0) {
+				if (redirecting != null || authenticating != null || lastHttpContent.content().readableBytes() == 0) {
```
raccoonback, line 936: It drains all messages from the current connection, when a retry is required due to authentication.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -941,9 +954,9 @@ protected void onInboundNext(ChannelHandlerContext ctx, Object msg) {
 			return;
 		}
 
-		if (redirecting != null) {
+		if (redirecting != null || authenticating != null) {
 			ReferenceCountUtil.release(msg);
-			// when redirecting auto-read is set to true, no need of manual reading
+			// when redirecting or authenticating auto-read is set to true, no need of manual reading
```
raccoonback, line 976: It drains all messages from the current connection, when a retry is required due to authentication.

raccoonback:
@violetagg 
I’ve applied all the review comments.
Please take another look. 🙂

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
violetagg: I needed to be more concrete - I was thinking about these below. The idea is that if you don't need to delay the token generation (i.e. asking some other component/service for a token) then you can immediately calculate it and add it to the headers thus no need of `Mono` otherwise you use the variant with `Mono`. Does this make sense?

```
	public final HttpClient httpAuthentication(
			BiPredicate<HttpClientRequest, HttpClientResponse> predicate,
			BiConsumer<? super HttpClientRequest, ? super SocketAddress> authenticator) {
	}

	public final HttpClient httpAuthenticationWhen(
			BiPredicate<HttpClientRequest, HttpClientResponse> predicate,
			BiFunction<? super HttpClientRequest, ? super SocketAddress, ? extends Mono<Void>> authenticator) {
	}

```

raccoonback: @violetagg 
Feedback has been applied.
Please review it again. 🙏

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2WebsocketClientOperations.java
```diff
@@ -111,7 +111,7 @@ else if (msg instanceof HttpResponse) {
 
 			setNettyResponse(response);
 
-			if (notRedirected(response)) {
+			if (notRedirected(response) && notAuthenticated()) {
```
violetagg: May be we can use another name?

```suggestion
			if (notRedirected(response) && authenticationNotRequired()) {
```

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2WebsocketClientOperations.java
```diff
@@ -131,9 +131,16 @@ else if (msg instanceof HttpResponse) {
...
+				if (redirecting != null) {
+					// Deliberately suppress "NullAway"
+					// redirecting != null in this case
+					listener().onUncaughtException(this, redirecting);
+				}
+				else if (authenticating != null) {
+					// Deliberately suppress "NullAway"
+					// authenticating != null in this case
+					listener().onUncaughtException(this, authenticating);
+				}
```
violetagg, line 139: The comments can be removed as we check now for null, also `NullAway` can be removed from the suppression.

```suggestion
				if (redirecting != null) {
					listener().onUncaughtException(this, redirecting);
				}
				else if (authenticating != null) {
					listener().onUncaughtException(this, authenticating);
				}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1696,6 +1696,92 @@ public final HttpClient wiretap(boolean enable) {
...
+	 * }
+	 * </pre>
+	 *
+	 * @param predicate determines when authentication should be applied, receives the request
+	 *                  and response to decide if authentication is needed (e.g., check for 401 status)
+	 * @param authenticator applies authentication to the request, receives the request and remote address
+	 * @return a new {@link HttpClient}
+	 * @since 1.3.0
+	 * @see #httpAuthenticationWhen(BiPredicate, BiFunction)
+	 */
```
violetagg, line 1743: I have some suggestions to the javadoc

```suggestion
	/**
	 * Enables a mechanism for an automatic retry of the requests when HTTP authentication is expected:
	 * <p>
	 * <ol>
	 *     <li>The initial request is sent without authentication</li>
	 *     <li>If the response matches the {@code predicate} (e.g., 401 Unauthorized), the request is retried</li>
	 *     <li>Before retry, the {@code authenticator} adds credentials to the request</li>
	 * </ol>
	 * </p>
	 * <p>
	 * Use this method when authentication credentials can be computed synchronously (without I/O operations).
	 * For async credential retrieval (e.g., fetching tokens from external services), use
	 * {@link #httpAuthenticationWhen(BiPredicate, BiFunction)} instead.
	 * </p>
	 *
	 * <p>Example - Bearer token authentication with custom retry logic:</p>
	 * <pre>
	 * {@code
	 * HttpClient client = HttpClient.create()
	 *     .httpAuthentication(
	 *         // Retry on 401 Unauthorized
	 *         (req, res) -> res.status().code() == 401,
	 *         // Add authentication header before retry
	 *         (req, addr) -> {
	 *             String token = generateAuthToken(addr);
	 *             req.header(HttpHeaderNames.AUTHORIZATION, "Bearer " + token);
	 *         }
	 *     );
	 * }
	 * </pre>
	 *
	 * @param predicate determines whether to retry with authentication; receives the original request
	 *                  and the response from the failed attempt (e.g., check for 401 Unauthorized status)
	 * @param authenticator applies authentication credentials to the request before retry; receives
	 *                      the request and remote address
	 * @return a new {@link HttpClient}
	 * @since 1.3.1
	 * @see #httpAuthenticationWhen(BiPredicate, BiFunction)
	 */
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1696,6 +1696,92 @@ public final HttpClient wiretap(boolean enable) {
...
+	 * }
+	 * </pre>
+	 *
+	 * @param predicate determines when authentication should be applied, receives the request
+	 *                  and response to decide if authentication is needed (e.g., check for 401 status)
+	 * @param authenticator applies authentication to the request, receives the request and remote address,
+	 *                      returns a Mono that completes when authentication is applied
+	 * @return a new {@link HttpClient}
+	 * @since 1.3.0
+	 */
```
violetagg, line 1854: I have some suggestions to the javadoc

```suggestion
	/**
	 * Enables a mechanism for an automatic retry of the requests when HTTP authentication is expected,
	 * with support for async credential retrieval:
	 * <p>
	 * <ol>
	 *     <li>The initial request is sent without authentication</li>
	 *     <li>If the response matches the {@code predicate} (e.g., 401 Unauthorized), the request is retried</li>
	 *     <li>Before retry, the {@code authenticator} asynchronously adds credentials to the request</li>
	 * </ol>
	 * </p>
	 * <p>
	 * Use this method when authentication credentials require async operations (e.g., fetching tokens
	 * from external services, reading from databases, or performing I/O). For synchronous credential
	 * computation, {@link #httpAuthentication(BiPredicate, BiConsumer)} provides a simpler API.
	 * </p>
	 *
	 * <p>Example - Async token retrieval with custom retry logic:</p>
	 * <pre>
	 * {@code
	 * HttpClient client = HttpClient.create()
	 *     .httpAuthenticationWhen(
	 *         // Retry on 401 Unauthorized
	 *         (req, res) -> res.status().code() == 401,
	 *         // Fetch token asynchronously and add authentication header before retry
	 *         (req, addr) -> tokenService.fetchToken(addr)
	 *             .doOnNext(token ->
	 *                 req.header(HttpHeaderNames.AUTHORIZATION, "Bearer " + token))
	 *             .then()
	 *     );
	 * }
	 * </pre>
	 *
	 * @param predicate determines whether to retry with authentication; receives the original request
	 *                  and the response from the failed attempt (e.g., check for 401 Unauthorized status)
	 * @param authenticator applies authentication credentials to the request before retry; receives
	 *                      the request and remote address, returns a {@link Mono} that completes when
	 *                      authentication credentials have been applied
	 * @return a new {@link HttpClient}
	 * @since 1.3.1
	 * @see #httpAuthentication(BiPredicate, BiConsumer)
	 */
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1696,6 +1696,92 @@ public final HttpClient wiretap(boolean enable) {
...
+	 * @param predicate determines when authentication should be applied, receives the request
+	 *                  and response to decide if authentication is needed (e.g., check for 401 status)
+	 * @param authenticator applies authentication to the request, receives the request and remote address
+	 * @return a new {@link HttpClient}
+	 * @since 1.3.0
+	 * @see #httpAuthenticationWhen(BiPredicate, BiFunction)
+	 */
+	public final HttpClient httpAuthentication(
+			BiPredicate<HttpClientRequest, HttpClientResponse> predicate,
+			BiConsumer<? super HttpClientRequest, ? super SocketAddress> authenticator) {
```
violetagg: The first param is without generics while the second is with generics?

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
```diff
@@ -1696,6 +1696,92 @@ public final HttpClient wiretap(boolean enable) {
...
+	 * @param predicate determines when authentication should be applied, receives the request
+	 *                  and response to decide if authentication is needed (e.g., check for 401 status)
+	 * @param authenticator applies authentication to the request, receives the request and remote address,
+	 *                      returns a Mono that completes when authentication is applied
+	 * @return a new {@link HttpClient}
+	 * @since 1.3.0
+	 */
+	public final HttpClient httpAuthenticationWhen(
+			BiPredicate<HttpClientRequest, HttpClientResponse> predicate,
+			BiFunction<? super HttpClientRequest, ? super SocketAddress, ? extends Mono<Void>> authenticator) {
```
violetagg: The first param is without generics while the second is with generics?

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientAuthenticationException.java
```diff
@@ -0,0 +1,42 @@
...
+ * Exception thrown to trigger HTTP authentication retry.
+ * <p>
+ * This exception is used internally by the generic HTTP authentication framework
+ * to signal that the current request requires authentication and should be retried.
+ * The framework will invoke the configured authenticator before retrying the request.
+ * </p>
+ *
+ * @author Oliver Ko
+ * @since 1.3.0
+ */
```
violetagg, line 28: I have some suggestions to the javadoc

```suggestion
/**
 * This exception is used internally to signal that the current request requires HTTP authentication and should be retried.
 * The {@code authenticator} configured via {@link HttpClient#httpAuthentication(java.util.function.BiPredicate, java.util.function.BiConsumer)}
 * or {@link HttpClient#httpAuthenticationWhen(java.util.function.BiPredicate, java.util.function.BiFunction)}
 * will be invoked before retrying the request.
 *
 * @author Oliver Ko
 * @since 1.3.1
 * @see HttpClient#httpAuthentication(java.util.function.BiPredicate, java.util.function.BiConsumer)
 * @see HttpClient#httpAuthenticationWhen(java.util.function.BiPredicate, java.util.function.BiFunction)
 */
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -837,7 +850,7 @@ protected void onInboundNext(ChannelHandlerContext ctx, Object msg) {
 						httpMessageLogFactory().debug(HttpMessageArgProviderFactory.create(response)));
 			}
 
-			if (notRedirected(response)) {
+			if (notRedirected(response) && notAuthenticated()) {
```
violetagg: May be we can use another name?

```suggestion
			if (notRedirected(response) && authenticationNotRequired()) {
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -977,6 +990,20 @@ final boolean notRedirected(HttpResponse response) {
 		return true;
 	}
 
+	@SuppressWarnings("NullAway")
+	final boolean notAuthenticated() {
+		// Deliberately suppress "NullAway"
+		// authenticationPredicate is checked for null before calling this method
```
violetagg: We don't need a suppression as we check for null. Also may be we need another name?

```suggestion
	final boolean authenticationNotRequired() {
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
```diff
@@ -977,6 +990,20 @@ final boolean notRedirected(HttpResponse response) {
...
+		if (authenticationPredicate != null && authenticationPredicate.test(this, this)) {
```
violetagg: What will happen if for example the server constantly returns 401? When are we going to stop retrying? May be we need something similar to what we have with follow redirect?

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -144,7 +144,7 @@ public void onInboundNext(ChannelHandlerContext ctx, Object msg) {
 
 			setNettyResponse(response);
 
-			if (notRedirected(response)) {
+			if (notRedirected(response) && notAuthenticated()) {
```
violetagg: May be we need another name?

```suggestion
			if (notRedirected(response) && authenticationNotRequired()) {
```

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -165,9 +165,16 @@ public void onInboundNext(ChannelHandlerContext ctx, Object msg) {
...
+				if (redirecting != null) {
+					// Deliberately suppress "NullAway"
+					// redirecting is initialized in notRedirected(response)
+					listener().onUncaughtException(this, redirecting);
+				}
+				else if (authenticating != null) {
+					// Deliberately suppress "NullAway"
+					// authenticating is initialized in notAuthenticated()
+					listener().onUncaughtException(this, authenticating);
+				}
```
violetagg, line 174: The comments can be removed as we check now for null, also NullAway can be removed from the suppression.

```suggestion
				if (redirecting != null) {
					listener().onUncaughtException(this, redirecting);
				}
				else if (authenticating != null) {
					listener().onUncaughtException(this, authenticating);
				}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClient.java
raccoonback: @violetagg 
I reviewed the followRedirect implementation as a reference, and as you pointed out, it makes sense to add generics to the `BiPredicate<HttpClientRequest, HttpClientResponse>` as well. I will update the code accordingly.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientOperations.java
raccoonback: @violetagg 
Good point.
Should we let the user specify the maximum retry count?
Or would it be better for us to define a default retry limit ourselves?

violetagg: If we expect more than 1 retry then may be only the user can specify this.

raccoonback:
@violetagg 
I’ve applied all the review comments!
Please take another look.

raccoonback:
@violetagg 
Hello.
I'd appreciate it if you could review this PR.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -749,8 +753,12 @@ public boolean test(Throwable throwable) {
 				return true;
 			}
 			if (throwable instanceof HttpClientAuthenticationException) {
-				// Set flag to trigger authenticator on retry
-				authenticationAttempted = true;
+				// Check if we've exceeded the max retry limit
+				if (authenticationRetries >= maxAuthenticationRetries) {
```
violetagg: We should do the comparison in HttpClientOperations so that if we decide that there will be no retry, we can deliver  the incoming response to the user otherwise having the check here we will not be able to deliver the response to the user.

raccoonback: @violetagg good point.
I'll update it.

raccoonback:
> Let's also expose the number of authentication retries similar to reactor.netty.http.client.HttpClientInfos#redirectedFrom, it might be helpful when users need to decide whether want to continue the retry or not. What do you think?

@violetagg 
That’s an excellent suggestion. I’ll go ahead and add it.

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
raccoonback: @violetagg 
To ensure retry attempts are tracked even when a new connection is established, I’ll keep the authentication retry count within `HttpClientHandler`.

raccoonback:
@violetagg
Hello.
I've reflect it.
Please check this PR!

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -499,6 +501,11 @@ static final class HttpClientHandler extends SocketAddress
 		volatile boolean                        shouldRetry;
 		volatile @Nullable HttpHeaders          previousRequestHeaders;
 
+		@Nullable BiPredicate<? super HttpClientRequest, ? super HttpClientResponse> authenticationPredicate;
+		@Nullable BiFunction<? super HttpClientRequest, ? super SocketAddress, ? extends Mono<Void>> authenticator;
+		AtomicInteger authenticationRetries;
```
violetagg: We can use "volatile" here in this scenario because we need to guarantee that the user will see the last updated value. The user will need this information when they decide whether to retry or not. The increase of the value happens after the invocation of the predicate just before the retry and creation of the new request. Also we do not expect to have multithreaded attempt for increase. What do you think?

raccoonback: @violetagg 
I didn’t choose `AtomicInteger` specifically for multithreading concerns, but because using a `volatile int` triggered a compile-time warning. To avoid that warning, I initially switched to AtomicInteger.

However, your suggestion makes sense — I’ll update the implementation to use a `volatile int` without triggering the warning.

raccoonback: @violetagg
I've reflect it.
Please check this PR!

raccoonback:
@violetagg
Hello.
I've reflect it.
Please check this PR. 😀😃

violetagg:

PR #3861, opened 2025-07-29
Fix message for `handlerConfigurator`

klopfdreh:
Small fix for the message if the handlerConfigurator is null


violetagg:
I back ported this to `1.2.x` branch

PR #3972, opened 2025-10-17
DefaultLoopIOUring should not depend on netty incubator io_uring package

j-bahr:
Now that Reactor Netty depends on Netty 4.2.X we should no longer depend on the io_uring incubator package.

> In Netty 4.2.0.Final we have graduated the io_uring transport from incubator, to a fully supported first-class transport module. (https://netty.io/news/2025/04/03/4-2-0.html)

If a project includes reactor netty and uses a JDK lower than 17, we get the default `DefaultLoopIOUring` class. If the incubator packages are not on the class path `isIoUringAvailable` will end up evaluating to false (even when io.netty:netty-transport-native-io_uring is included). Given that the incubator is not being updated we should mirror the logic from `DefaultLoopIOUring` from the multi release (17) class.


violetagg:
This is superseded by #3974

PR #3974, opened 2025-10-18
`IO_Uring` loop resources should support `Java 11+`

j-bahr:
As of Netty 4.2 io_uring support has been promoted out of incubator `netty-incubator-transport-native-io_uring` and is now located at `netty-transport-native-io_uring`

> In Netty 4.2.0.Final we have graduated the io_uring transport from incubator, to a fully supported first-class transport module. (https://netty.io/news/2025/04/03/4-2-0.html)

Everything in the netty io_uring module is compiled for Java 9 and above excluding `io.netty.channel.uring.IoUring` to allow for calls to `isAvailable()` from Java 8 code.

Currently in reactor-netty we compile support for io_uring as multi release source with a release version 17, which means that only Java 17+ will be able to run the code. Older releases such as JDK 11 will get the default implementation which is compiled against the incubator release. This actually causes failures if you supply your own io_uring event loop in java 11 and the incubator is missing from the classpath.

This PR aims to fix two things (1) we should not compile against incubator code as it's now unsupported (2) we should allow io_uring for all releases above Java 9.

1) To remove io_uring incubator the default, Java 8, source must throw exceptions for all methods in `DefaultLoopIOUring ` and expressly mark `isIoUringAvailable = false`.

2) To enable use of io_uring in JDK 11 and above we need to move the implementation of `DefaultLoopIOUring` to a Java 11 multi release folder and update the gradle build file. I suggest targeting Java 11 here because Java 9 & 10 are not LTS release, not receiving updates and there is a low probability that they are being used in production.

This should allow users using Java 11+ to use the actual io_uring module from netty, using either reactor-netty's default loop resources or by allowing user to supply their own.

### DIFF
--- a/.github/workflows/check_netty_4_1_snapshots.yml
+++ b/.github/workflows/check_netty_4_1_snapshots.yml
@@ -30,21 +30,16 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '8'
-      - name: Set up JDK 11
-        uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165
-        with:
-            distribution: 'zulu'
-            java-version: '11'
       - name: Set up JDK 17
         uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165
         with:
             distribution: 'graalvm'
             java-version: '17.0.12'
       - name: Build with Gradle against Netty 4.1
-        run: ./gradlew clean check --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.129.Final-SNAPSHOT'
+        run: ./gradlew clean check -x :reactor-netty-core:java17Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.129.Final-SNAPSHOT'
       - name: GraalVM smoke tests against Netty 4.1
         run: ./gradlew :reactor-netty-graalvm-smoke-tests:nativeTest --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.129.Final-SNAPSHOT'
       - name: Build with Gradle against Netty 4.2
-        run: ./gradlew clean check --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.2.8.Final-SNAPSHOT'
+        run: ./gradlew clean check -x :reactor-netty-core:java17Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.2.8.Final-SNAPSHOT'
       - name: GraalVM smoke tests against Netty 4.2
         run: ./gradlew :reactor-netty-graalvm-smoke-tests:nativeTest --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.2.8.Final-SNAPSHOT'


violetagg:
@j-bahr We cannot remove the IO_Uring incubator because of compatibility reasons and because there is nothing that can be used when on JDK 8, so please keep it.

j-bahr:
@violetagg, thanks for the feedback, I am happy to add back the logic for the io_uring incubator so that we can merge the other fixes.

Can you help me understand the rationale behind using the incubator package for Java 8? Java 8 users still have the ability to use `nio` or `epoll` transports and it seems unlikely that many folks would include the io_uring incubator packages on the class path given that it's end of life and will not receive bug fixes or security patches.

violetagg:
> @violetagg, thanks for the feedback, I am happy to add back the logic for the io_uring incubator so that we can merge the other fixes.
> 
> Can you help me understand the rationale behind using the incubator package for Java 8? Java 8 users still have the ability to use `nio` or `epoll` transports and it seems unlikely that many folks would include the io_uring incubator packages on the class path given that it's end of life and will not receive bug fixes or security patches.

Unfortunately we don't know whether it is used or not. We can add a note that the incubator is not supported anymore and deprecate it then eventually in the next version we can remove the support. Wdyt?

.github/workflows/check_netty_4_1_snapshots.yml
```diff
@@ -36,10 +36,11 @@ jobs:
             distribution: 'graalvm'
             java-version: '17.0.12'
       - name: Build with Gradle against Netty 4.1
-        run: ./gradlew clean check -x :reactor-netty-core:java17Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.129.Final-SNAPSHOT'
+        run: ./gradlew clean check -x :reactor-netty-core:java17Test -x :reactor-netty-core:java11Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.129.Final-SNAPSHOT'
```
violetagg: Do we need this change? This workflow is against branch `1.2.x`, (see above in the checkout action)

.github/workflows/check_netty_4_1_snapshots.yml
```diff
@@ -36,10 +36,11 @@ jobs:
...
       - name: GraalVM smoke tests against Netty 4.1
         run: ./gradlew :reactor-netty-graalvm-smoke-tests:nativeTest --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.129.Final-SNAPSHOT'
       - name: Build with Gradle against Netty 4.2
-        run: ./gradlew clean check -x :reactor-netty-core:java17Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.2.8.Final-SNAPSHOT'
+        run: ./gradlew clean check -x :reactor-netty-core:java17Test -x :reactor-netty-core:java11Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.2.8.Final-SNAPSHOT'
```
violetagg: Do we need this change? This workflow is against branch `1.2.x`, (see above in the checkout action)

.github/workflows/check_netty_4_2_snapshots.yml
```diff
@@ -31,6 +31,6 @@ jobs:
             distribution: 'graalvm'
             java-version: '17.0.12'
       - name: Build with Gradle
-        run: ./gradlew clean check -x :reactor-netty-core:java17Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.2.8.Final-SNAPSHOT'
+        run: ./gradlew clean check -x :reactor-netty-core:java17Test -x :reactor-netty-core:java11Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.2.8.Final-SNAPSHOT'
```
violetagg: Do we need JDK 11 setup?

.github/workflows/publish.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         run: ./gradlew qualifyVersionGha
       - name: run reactor-netty-core checks
         id: reactor-netty-core-checks
-        run: ./gradlew reactor-netty-core:check -x :reactor-netty-core:java17Test
+        run: ./gradlew reactor-netty-core:check -x :reactor-netty-core:java17Test -x :reactor-netty-core:java11Test
```
violetagg: Do we need JDK 11 setup?

reactor-netty-core/build.gradle
```diff
@@ -102,6 +105,7 @@ dependencies {
 		}
 		else if (forceTransport == "io_uring" && osdetector.os == "linux") {
 			testImplementation "io.netty:netty-transport-native-io_uring:$nettyVersion$os_suffix"
+			testImplementation "io.netty.incubator:netty-incubator-transport-native-io_uring:$nettyIoUringVersion"
```
violetagg:
```suggestion
			testImplementation "io.netty.incubator:netty-incubator-transport-native-io_uring:$nettyIoUringVersion$os_suffix"
```

.github/workflows/check_graalvm.yml
```diff
@@ -28,6 +28,11 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '8'
+      - name: Set up JDK 11
+        uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165
+        with:
+            distribution: 'zulu'
+            java-version: '11'
```
violetagg: Below we test IO_Uring on JDK 17, do we want to test it on JDK 11?

.github/workflows/check_netty_4_1_snapshots.yml
```diff
@@ -30,16 +30,22 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '8'
+      - name: Set up JDK 11
+        uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165
+        with:
+            distribution: 'zulu'
+            java-version: '11'
```
violetagg:
```suggestion
```

.github/workflows/check_netty_4_1_snapshots.yml
```diff
@@ -30,16 +30,22 @@ jobs:
...
       - name: Set up JDK 17
         uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165
         with:
             distribution: 'graalvm'
             java-version: '17.0.12'
       - name: Build with Gradle against Netty 4.1
-        run: ./gradlew clean check -x :reactor-netty-core:java17Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.129.Final-SNAPSHOT'
+        run: ./gradlew clean check --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.129.Final-SNAPSHOT'
```
violetagg:
```suggestion
        run: ./gradlew clean check -x :reactor-netty-core:java17Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.129.Final-SNAPSHOT'
```

.github/workflows/check_netty_4_1_snapshots.yml
```diff
@@ -30,16 +30,22 @@ jobs:
...
       - name: GraalVM smoke tests against Netty 4.1
         run: ./gradlew :reactor-netty-graalvm-smoke-tests:nativeTest --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.1.129.Final-SNAPSHOT'
       - name: Build with Gradle against Netty 4.2
-        run: ./gradlew clean check -x :reactor-netty-core:java17Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.2.8.Final-SNAPSHOT'
+        run: ./gradlew clean check --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.2.8.Final-SNAPSHOT'
```
violetagg:
```suggestion
        run: ./gradlew clean check -x :reactor-netty-core:java17Test --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.2.8.Final-SNAPSHOT'
```

.github/workflows/check_netty_4_1_snapshots.yml
```diff
@@ -30,16 +30,22 @@ jobs:
...
       - name: GraalVM smoke tests against Netty 4.2
         run: ./gradlew :reactor-netty-graalvm-smoke-tests:nativeTest --no-daemon -PforceTransport=${{ matrix.transport }} -PforceNettyVersion='4.2.8.Final-SNAPSHOT'
+
```
violetagg:
```suggestion
```

violetagg:
Thanks for the PR!

j-bahr:
@violetagg thanks for merging and thanks for fixing a few of the issues I was working through on the workflow files.

PR #3980, opened 2025-10-24
Add `HTTP/2` `PING-based` connection health checks

violetagg:
Introduce `PING` frame support for proactive connection health monitoring on idle `HTTP/2` connections. The feature sends `PING` frames to detect unresponsive peers and close stale connections.

Configuration via `Http2SettingsSpec`:
- `pingAckTimeout`: Timeout for `ACK` response
- `pingAckDropThreshold`: Maximum `PING` attempts (default: 1)

Requires idle timeout configuration:
- Client: `ConnectionProvider.maxIdleTime`
- Server: `HttpServer.idleTimeout`

The feature is opt-in and integrates automatically with connection pool eviction (client) and idle timeout handling (server).

This implementation is based on PR #3612, reduces the scope and adds `HttpClient` support.

### DIFF
--- a/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpIdleTimeoutTest.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpIdleTimeoutTest.java
@@ -84,7 +84,7 @@ class HttpIdleTimeoutTest extends BaseHttpTest {
 		        .as(StepVerifier::create)
 		        .expectNext("closedAfterIdleTimeout")
 		        .expectComplete()
-		        .verify(Duration.ofSeconds(5));
+		        .verify(Duration.ofSeconds(10));
 
 		Mono.delay(Duration.ofMillis(200))
 		    .block();
@@ -119,7 +119,7 @@ class HttpIdleTimeoutTest extends BaseHttpTest {
 		        .as(StepVerifier::create)
 		        .expectNext("openedBeforeIdleTimeout")
 		        .expectComplete()
-		        .verify(Duration.ofSeconds(5));
+		        .verify(Duration.ofSeconds(10));
 
 		Mono.delay(Duration.ofMillis(100))
 		    .block();


raccoonback:
@violetagg 
I’ll review it by the end of this week! 😀

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
```diff
@@ -0,0 +1,225 @@
...
+		if (isPingAckPending) {
+			if (msg instanceof Http2PingFrame) {
+				Http2PingFrame frame = (Http2PingFrame) msg;
+				if (frame.ack() && frame.content() == lastSentPingData) {
+					cancel();
+				}
+			}
+			else if (msg instanceof Http2HeadersFrame || msg instanceof Http2DataFrame) {
+				cancel();
+			}
```
raccoonback, line 139: Oh, so it also handles the case when a header frame is received. 👍

raccoonback, line 139: Since `IdleTimeoutHandler` is triggered when no frame I/O occurs,
is it correct to understand that there’s no need to send another ping after receiving a frame?

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
```diff
@@ -0,0 +1,225 @@
...
+
+		private final ChannelHandlerContext ctx;
+
+		PingTimeoutTask(ChannelHandlerContext ctx) {
+			this.ctx = ctx;
+		}
+
+		@Override
+		public void operationComplete(ChannelFuture future) {
+			if (!future.channel().isActive()) {
```
raccoonback, line 156: Since the future will be newly created, would it be fine not to check the active state of the current context channel?
(ex. `!ctx.channel().isActive()`)

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
```diff
@@ -579,16 +621,28 @@ static final class PooledConnectionAllocator {
...
+						new Http11EvictionPredicate(poolFactory.maxIdleTime(), poolFactory.maxLifeTime());
+				Http2EvictionPredicate http2EvictionPredicate = new Http2EvictionPredicate(poolFactory.maxLifeTime());
+				this.parent = parent.mutate().evictionPredicate(http11EvictionPredicate).build();
+				poolConfigFunction = poolConfig -> new Http2Pool(poolConfig, poolFactory.allocationStrategy(),
+						http2EvictionPredicate, poolFactory.maxIdleTime());
+			}
+			else {
+				this.parent = parent;
+				poolConfigFunction = poolConfig -> new Http2Pool(poolConfig, poolFactory.allocationStrategy());
+			}
```
raccoonback, line 639: Is it intended to allow users to customize HTTP/2 connection eviction?

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
```diff
@@ -375,6 +399,17 @@ public void onStateChange(Connection connection, State newState) {
...
 			}
+			else if (newState == UPGRADE_SUCCESSFUL) {
+				Http2Pool http2Pool = http2Pool();
+				if (http2Pool.evictionPredicate != null) {
+					ChannelHandlerContext frameCodec = http2PooledRef(pooledRef).slot.http2FrameCodecCtx();
+					if (frameCodec != null) {
+						IdleTimeoutHandler.addIdleTimeoutHandler(connection.channel().pipeline(), Duration.ofMillis(http2Pool.maxIdleTime),
+								new Http2ConnectionLiveness(((Http2FrameCodec) frameCodec.handler()),
+										http2SettingsSpec.pingAckDropThreshold(), http2SettingsSpec.pingAckTimeout().toNanos()));
+					}
```
raccoonback, line 410: It seems we also need to check whether `pingAckTimeout` is null here.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
```diff
@@ -364,6 +378,16 @@ else if (p.state != null) {
...
 			}
 
+			Http2Pool http2Pool = http2Pool();
+			if (http2Pool.evictionPredicate != null) {
+				ChannelHandlerContext frameCodec = http2PooledRef(pooledRef).slot.http2FrameCodecCtx();
+				if (frameCodec != null) {
+					IdleTimeoutHandler.addIdleTimeoutHandler(channel.pipeline(), Duration.ofMillis(http2Pool.maxIdleTime),
+							new Http2ConnectionLiveness(((Http2FrameCodec) frameCodec.handler()),
+									http2SettingsSpec.pingAckDropThreshold(), http2SettingsSpec.pingAckTimeout().toNanos()));
+				}
```
raccoonback, line 388: It seems we also need to check whether `pingAckTimeout` is null here.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
```diff
@@ -600,5 +654,41 @@ Publisher<Connection> connectChannel() {
...
+				boolean isNotHttp2 = true;
+				if (owner instanceof DisposableAcquire) {
+					Http2Pool.Http2PooledRef http2PooledRef = http2PooledRef(((DisposableAcquire) owner).pooledRef);
+					ChannelHandlerContext frameCodec = http2PooledRef.slot.http2FrameCodecCtx();
+					isNotHttp2 = frameCodec == null;
+				}
+				return (isNotHttp2 && maxIdleTime != -1 && meta.idleTime() >= maxIdleTime)
+						|| (maxLifeTime != -1 && meta.lifeTime() >= maxLifeTime);
+			}
+		}
```
raccoonback, line 679: Is this intended to check for the `h2c` case?

raccoonback (commented):
@violetagg 
I added some comments with a few questions.
I’d appreciate it if you could review them.

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
violetagg: My understanding is that if we receive headers/data (which means a request/response) or ack to our ping frame then we can stop. The handler is still there so if the connection stays again idle for the configured timeout, we will start sending ping frames again.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
violetagg: we provide an API for a custom eviction predicate (reactor.netty.resources.ConnectionProvider.ConnectionPoolSpec#evictionPredicate), it is documented that if you provide a custom eviction predicate this will disable every Reactor Netty eviction predicate.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
violetagg: once we upgrade the connection the H2C has Http2FrameCodec

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
violetagg: it is intentional because `http2Pool.evictionPredicate != null` is true only if `http2SettingsSpec.pingAckTimeout() != null`, this is the artificial eviction predicate that we add for ping frame support.
I don't mind to check again for null although it is redundant or to add comment so that it is clear why we don't check for null

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
violetagg: it is intentional because `http2Pool.evictionPredicate != null` is true only if `http2SettingsSpec.pingAckTimeout() != null`, this is the artificial eviction predicate that we add for ping frame support.
I don't mind to check again for null although it is redundant or to add comment so that it is clear why we don't check for null

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
violetagg: yeah I think headers/data defines a request/response

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
violetagg: And indeed I think it will be dangerous if despite the custom eviction predicate we start closing the connections based on our new ping frame feature.

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
violetagg: Can you elaborate more?

violetagg:
@raccoonback Thanks for the review! Let me merge this one.

PR #3987, opened 2025-10-28
Allow setting `.pendingAcquireMaxCount()` to zero in `ConnectionProvider`

### LINKED ISSUE #3986: Setting `pendingAcquireMaxCount` and `pendingAcquireTimeout` to zero in `ConnectionProvider`
I'm wondering if reactor-netty can support setting `pendingAcquireMaxCount` and `pendingAcquireTimeout` in `ConnectionProvider`, in order to fail fast if no connection can be acquired.

## Motivation

My motivation is that **I can't estimate a proper `maxConnections`**, and **I don't want requests wait for connection**, so I implement a `HybridConnectionProvider` to composite `PooledConnectionProvider` and `NewConnectionProvider`:

* If the pending queue in `PooledConnectionProvider` is full (which I want set it to 0, and throw `PoolAcquirePendingLimitException` immediately), I can new a connection via `NewConnectionProvider`.
* If not, use `PooledConnectionProvider` as normal.

## Desired solution

My implement like:

```java
    public Mono<? extends Connection> acquire(@NonNull TransportConfig config, @NonNull ConnectionObserver connectionObserver, Supplier<? extends SocketAddress> remoteAddress, AddressResolverGroup<?> resolverGroup) {
        Mono<Connection> poolAcquire = (Mono<Connection>) pooledConnectionProvider.acquire(config, connectionObserver, remoteAddress, resolverGroup);

        return poolAcquire
                .doOnError(throwable -> log.info("Encounter exception in pooledConnectionProvider, create temp connection!", throwable))
                .onErrorResume(PoolAcquirePendingLimitException.class, __ -> newConnection(config, connectionObserver, remoteAddress, resolverGroup))
                .onErrorResume(PoolAcquireTimeoutException.class, __ -> newConnection(config, connectionObserver, remoteAddress, resolverGroup));
    }
```

I hope the `ConnectionProvider` can:

```java
        return ConnectionProvider.builder("webclient-pool")
                .pendingAcquireMaxCount(0)
                .pendingAcquireTimeout(Duration.ZERO)
                .build();
```

## Considered alternatives

No

## Additional context

No

Martin7-1:
Fixes #3986.

I'm new to `reactor-netty`. I created thisPR as described in CONTRIBUTING.md, please let me know if there is something wrong

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
@@ -605,18 +605,23 @@ public interface ConnectionProvider extends Disposable {
 
 		/**
 		 * Set the options to use for configuring {@link ConnectionProvider} the maximum number of registered
-		 * requests for acquire to keep in a pending queue
-		 * When invoked with -1 the pending queue will not have upper limit.
+		 * requests for acquire to keep in a pending queue.
+		 * <ul>
+		 * <li>When invoked with {@code -1} the pending queue will not have an upper limit.</li>
+		 * <li>When invoked with {@code 0} the pending queue will not be used and any acquire attempt
+		 * will fail immediately when no idle connection is available (fail-fast).</li>
+		 * <li>When invoked with a positive value, that many acquire requests can be queued.</li>
+		 * </ul>
 		 * Default to {@code 2 * max connections}.
 		 *
 		 * @param pendingAcquireMaxCount the maximum number of registered requests for acquire to keep
 		 * in a pending queue
 		 * @return {@literal this}
-		 * @throws IllegalArgumentException if pendingAcquireMaxCount is negative
+		 * @throws IllegalArgumentException if pendingAcquireMaxCount is negative (except {@code -1} for unbounded)
 		 */
 		public final SPEC pendingAcquireMaxCount(int pendingAcquireMaxCount) {
-			if (pendingAcquireMaxCount != -1 && pendingAcquireMaxCount <= 0) {
-				throw new IllegalArgumentException("Pending acquire max count must be strictly positive");
+			if (pendingAcquireMaxCount != -1 && pendingAcquireMaxCount < 0) {
+				throw new IllegalArgumentException("Pending acquire max count must be non-negative, or -1 for unbounded");
 			}
 			this.pendingAcquireMaxCount = pendingAcquireMaxCount;
 			return get();


reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -615,8 +616,8 @@ public final SPEC maxConnections(int maxConnections) {
 		 * @throws IllegalArgumentException if pendingAcquireMaxCount is negative
 		 */
 		public final SPEC pendingAcquireMaxCount(int pendingAcquireMaxCount) {
-			if (pendingAcquireMaxCount != -1 && pendingAcquireMaxCount <= 0) {
-				throw new IllegalArgumentException("Pending acquire max count must be strictly positive");
+			if (pendingAcquireMaxCount != -1 && pendingAcquireMaxCount < 0) {
+				throw new IllegalArgumentException("Pending acquire max count must be non-negative");
```
violetagg: What do you think about this clarification?

```suggestion
				throw new IllegalArgumentException("Pending acquire max count must be non-negative, or -1 for unbounded");
```

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -607,6 +607,7 @@ public final SPEC maxConnections(int maxConnections) {
 		 * Set the options to use for configuring {@link ConnectionProvider} the maximum number of registered
 		 * requests for acquire to keep in a pending queue
 		 * When invoked with -1 the pending queue will not have upper limit.
+		 * When invoked with 0 the pending queue has no capacity, any acquire attempt will fail immediately.
```
violetagg: What do you think about this clarification?

```suggestion
		 * requests for acquire to keep in a pending queue.
		 * <ul>
		 * <li>When invoked with {@code -1} the pending queue will not have an upper limit.</li>
		 * <li>When invoked with {@code 0} the pending queue will not be used and any acquire attempt
		 * will fail immediately when no idle connection is available (fail-fast).</li>
		 * <li>When invoked with a positive value, that many acquire requests can be queued.</li>
		 * </ul>
```

violetagg (changes requested):
Thanks for the PR!

I added some suggestions for more clarity.

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
```diff
@@ -615,8 +616,8 @@ public final SPEC maxConnections(int maxConnections) {
 		 * @throws IllegalArgumentException if pendingAcquireMaxCount is negative
```
violetagg: May be this one also needs adjustment.

```suggestion
		 * @throws IllegalArgumentException if pendingAcquireMaxCount is negative (except {@code -1} for unbounded)
```

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
Martin7-1: Looks good! I just change it for clarification.

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
Martin7-1: Looks good! I just change it for clarification.

reactor-netty-core/src/main/java/reactor/netty/resources/ConnectionProvider.java
Martin7-1: Done.

Issue #4016, opened 2025-11-25
HTTP Client is interpreting 201 CREATED response as a redirection


eze210:
The Location header is valid for 3xx and 201 responses. For 3xx responses, the Location header indicates the URL of the redirection request, but for 201, it indicates where is the new created resource.

Reproduced with 1.3.0

## Expected Behavior
For 201 responses, the followRedirects predicate shouldn't be invoked, and the redirection request should not happen.

## Actual Behavior
HTTP Client follows a "redirect" for 201 CREATED responses with Location header.

## Steps to Reproduce

Unit test:

```java
	@Test
	void testFollowRedirectPredicateDoesNotApplyToNonRedirectStatus() {
		final int serverPort = SocketUtils.findAvailableTcpPort();

		disposableServer =
				createServer(serverPort)
				          .host("localhost")
				          .route(r -> r.get("/not-a-redirect", (req, res) -> res.status(201).header(HttpHeaderNames.LOCATION, "/created-resource-location"))
				                       .get("/created-resource-location", (req, res) -> res.status(200)
				                                                           .sendString(Mono.just("Success"))))
				          .bindNow();

		HttpClientResponse response =
				createClient(disposableServer::address)
				          .followRedirect((req, res) -> {
							// 201 CREATED is not a redirect status code, so the predicate should not be applied
							return true;
				          })
				          .get()
				          .uri("/not-a-redirect")
				          .response()
				          .block(Duration.ofSeconds(30));

		// A 201 CREATED with a Location header should not be followed as a redirect,
		// even if the predicate returns true.
		assertThat(response).isNotNull();
		assertThat(response.status()).isEqualTo(HttpResponseStatus.CREATED);
		assertThat(response.responseHeaders().get(HttpHeaderNames.LOCATION)).isEqualTo("/created-resource-location");
	}
```

## Possible Solution
HttpClientOperations#notRedirected(HttpResponse) may also check the status code before calling the followRedirectPredicate. Created a Pull Request with this possible solution #4017 .

## Your Environment

* Reactor version(s) used: 1.3.0
* Other relevant libraries versions (eg. `netty`, ...): N/A
* JVM version (`java -version`): Temurin JDK 17
* OS and version (eg. `uname -a`): Irrelevant

violetagg:
[...]

PR #4017, opened 2025-11-25
Only follow redirects for 3xx status codes

eze210:
Fixes issue #4016


violetagg:
Close as the behaviour is expected, see the comment https://github.com/reactor/reactor-netty/issues/4016#issuecomment-3576282624

PR #4030, opened 2025-12-11
Add `maxConnections` configuration to `HttpServer`

### LINKED ISSUE #3978: A way to restrict no. of connections on reactor server
We are looking for a way to restrict the no. of connections that get established on our reactor netty server(standalone application not spring boot). We could see that there are some explicit ways to do it manually by adding inbound channelhandlers but that will go through the initial connection establishment and then discarding of the connection. 
Is there a way you could provide this as a feature which implicitly does not let the total no. of connections on reactor server go beyond a configured threshold. Would be great if the connection establishment itself could be avoided.

raccoonback:
## Summary
Adds `maxConnections` configuration to `HttpServer` to limit the maximum number of concurrent connections. When the limit is reached, new connections are immediately closed.

### API Changes
  - **HttpServer**: Added public `maxConnections(int)` method to configure connection limit
  - **ServerTransport**: Added protected `maxConnections(int)` base method for potential future extension

### Core Implementation
- **ServerTransportConfig**:
  - Added `maxConnections` field (default: `-1` for no limit) and `AtomicInteger activeConnections` for tracking
  - Enhanced `defaultChildObserver()` to include `maxConnections <= 0` check for optimization
  - Implemented connection limit enforcement in `ServerTransportDoOnConnection` with proper HTTP/2 stream filtering
- **Http3ChannelInitializer**:
  - Add `Http3ConnectionObserverBridge` to better reflect its role as a bridge between Netty channel events and ConnectionObserver

## Behavior
- **Default**: -1 (no limit)
- **HTTP/1.1**: Each connection is counted
- **HTTP/2**: Only TCP connections are counted, not individual streams
- **HTTP/3**: QUIC channels are counted, not individual streams
- **Rejection**: Connections exceeding the limit are immediately closed

## Example Usage
  ```java
  HttpServer.create()
            .port(8080)
            .maxConnections(100)  // Limit to 100 concurrent connections
            .handle((request, response) -> response.sendString(Mono.just("Hello World!")))
            .bindNow();
```

## Related Issue
- Fixes #3978

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
@@ -25,7 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
@@ -119,7 +119,7 @@ public abstract class ServerTransport<T extends ServerTransport<T, CONF>,
 			if (!config.channelType(isDomainSocket).equals(DatagramChannel.class)) {
 				Acceptor acceptor = new Acceptor(config.childEventLoopGroup(), channelInitializer,
 						config.childOptions, config.childAttrs, isDomainSocket,
-						config.maxConnections, config.activeConnections);
+						config.maxConnections);
 				channelInitializer = new AcceptorInitializer(acceptor);
 			}
 			TransportConnector.bind(config, channelInitializer, local, isDomainSocket)
@@ -383,26 +383,29 @@ public abstract class ServerTransport<T extends ServerTransport<T, CONF>,
 
 	static class Acceptor extends ChannelInboundHandlerAdapter {
 
+		static final AtomicIntegerFieldUpdater<Acceptor> ACTIVE_CONNECTIONS =
+				AtomicIntegerFieldUpdater.newUpdater(Acceptor.class, "activeConnections");
+
 		final EventLoopGroup childGroup;
 		final ChannelHandler childHandler;
 		final Map<ChannelOption<?>, ?> childOptions;
 		final Map<AttributeKey<?>, ?> childAttrs;
 		final boolean isDomainSocket;
 		final int maxConnections;
-		final AtomicInteger activeConnections;
+
+		volatile int activeConnections;
 
 		@Nullable Runnable enableAutoReadTask;
 
 		Acceptor(EventLoopGroup childGroup, ChannelHandler childHandler,
 				Map<ChannelOption<?>, ?> childOptions, Map<AttributeKey<?>, ?> childAttrs,
-				boolean isDomainSocket, int maxConnections, AtomicInteger activeConnections) {
+				boolean isDomainSocket, int maxConnections) {
 			this.childGroup = childGroup;
 			this.childHandler = childHandler;
 			this.childOptions = childOptions;
 			this.childAttrs = childAttrs;
 			this.isDomainSocket = isDomainSocket;
 			this.maxConnections = maxConnections;
-			this.activeConnections = activeConnections;
 		}
 
 		@Override
@@ -410,18 +413,16 @@ public abstract class ServerTransport<T extends ServerTransport<T, CONF>,
 		public void channelRead(ChannelHandlerContext ctx, Object msg) {
 			final Channel child = (Channel) msg;
 			if (maxConnections > 0) {
-				int current = activeConnections.incrementAndGet();
-				if (current > maxConnections) {
-					activeConnections.decrementAndGet();
+				int current = ACTIVE_CONNECTIONS.get(this);
+				if (current >= maxConnections) {
 					if (log.isDebugEnabled()) {
 						log.debug(format(child, "Connection rejected: max connections limit reached ({})"), maxConnections);
 					}
 					forceClose(child, new RuntimeException("Connection rejected: max connections limit reached"));
 					return;
 				}
-				else {
-					child.closeFuture().addListener(future -> activeConnections.decrementAndGet());
-				}
+				ACTIVE_CONNECTIONS.incrementAndGet(this);
+				child.closeFuture().addListener(future -> ACTIVE_CONNECTIONS.decrementAndGet(this));
 			}
 
 			child.pipeline().addLast(childHandler);

### CODE REVIEW

File: reactor-netty-http/src/main/java/reactor/netty/http/server/Http3ChannelInitializer.java [hunk 1]
```diff
@@ -60,9 +71,19 @@ final class Http3ChannelInitializer extends ChannelInitializer<Channel> {
...
 	protected void initChannel(Channel channel) {
+		ChannelInitializer<Channel> wrappedInitializer = new ChannelInitializer<Channel>() {
+			@Override
+			protected void initChannel(Channel ch) {
+				if (ch instanceof QuicChannel) {
+					ch.pipeline().addLast(new Http3ConnectionObserverBridge(connectionObserver));
+				}
+				ch.pipeline().addLast(quicChannelInitializer);
+			}
+		};
```
raccoonback: Wrapped existing `quicChannelInitializer` with `wrappedInitializer`

reactor-netty-http/src/main/java/reactor/netty/http/server/Http3ChannelInitializer.java
```diff
@@ -123,4 +144,25 @@ static void streamChannelOptions(QuicServerCodecBuilder quicServerCodecBuilder,
...
+			if (ctx.channel() instanceof QuicChannel) {
+				if (log.isDebugEnabled()) {
+					log.debug(format(ctx.channel(), "HTTP/3: Firing State.CONNECTED for QuicChannel"));
+				}
+				Connection connection = Connection.from(ctx.channel());
+				connectionObserver.onStateChange(connection, ConnectionObserver.State.CONNECTED);
+			}
+			super.channelActive(ctx);
+		}
+	}
```
raccoonback: `Http3ConnectionObserverBridge` bridges Netty's QUIC channel lifecycle events to Reactor-Netty's `ConnectionObserver` pattern by manually firing `State.CONNECTED` when a `QuicChannel` becomes active, 
  enabling HTTP/3 connections to be counted and limited by the maxConnections logic.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -1286,6 +1290,23 @@ protected void initChannel(Channel ch) {
...
+		}
+
+		@Override
+		public void channelActive(ChannelHandlerContext ctx) throws Exception {
+			// Trigger State.CONNECTED for the HTTP/2 connection so that maxConnections limiting works
+			Connection c = Connection.from(ctx.channel());
+			listener.onStateChange(c, ConnectionObserver.State.CONNECTED);
+			super.channelActive(ctx);
+		}
+	}
```
raccoonback: `H2ConnectionHandler` triggers `State.CONNECTED` event to the `ConnectionObserver` when an HTTP/2 connection becomes active, enabling HTTP/2 TCP connections to be counted and limited by the 
  maxConnections logic.

reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransportConfig.java
```diff
@@ -232,22 +246,63 @@ static final class ServerTransportDoOnConnection implements ConnectionObserver {
...
+					}
+				}
+
+				// Add to ChannelGroup if configured
+				if (channelGroup != null) {
+					channelGroup.add(channel);
+					if (isHttp2Stream) {
+						// HTTP/2 - add both the stream and the connection
+						channelGroup.add(parent);
+					}
```
raccoonback: `ServerTransportDoOnConnection.onStateChange(...)` was enhanced to distinguish HTTP/2 streams from actual connections (TCP, QUIC) and enforce maxConnections limit by atomically tracking active 
  connections, immediately closing new connections when the limit is exceeded while ensuring HTTP/2 streams are not counted separately.

raccoonback:
@violetagg
Hello!
Please check this PR when you have a chance. 😃

raccoonback:
@violetagg 
It looks like the test failed because Gradle couldn’t be downloaded for some reason.
Please rerun it.

docs/modules/ROOT/pages/http-server.adoc
```diff
@@ -366,6 +366,40 @@ include::{examples-dir}/lifecycle/Application.java[lines=18..39]
 <1> `Netty` pipeline is extended with `ReadTimeoutHandler` when a remote client is connected.
 <2> `Netty` pipeline is extended with `LoggingHandler` when initializing the channel.
 
+[[connection-limit]]
+== Connection Limit
+
+You can configure the `HTTP` server to limit the maximum number of concurrent connections.
```
violetagg:
```suggestion
You can configure the `HttpServer` to limit the maximum number of concurrent connections.
```

reactor-netty-http/src/http3Test/java/reactor/netty/http/Http3Tests.java
```diff
@@ -1229,6 +1230,57 @@ void httpClientSmokeTest() {
...
+									latch.countDown();
+								}
+						);
+			}
+
+			assertThat(latch.await(3, TimeUnit.SECONDS)).isTrue();
+			assertThat(connectionCount.get()).isEqualTo(2);
+			assertThat(successCount.get()).isEqualTo(2);
+			assertThat(failureCount.get()).isEqualTo(1);
+		} finally {
```
violetagg:
```suggestion
		}
		finally {
```

violetagg:
@raccoonback It seems there is some issue with GitHub Actions ...

violetagg:
@raccoonback Although there are issues with GitHub Actions, it seems that your PR has some checkstyle issues.

raccoonback:
@violetagg 
I’ve updated it with the fixes.
Please take another look. 😀

violetagg:
May be it would be better if we implement the feature here reactor.netty.transport.ServerTransport.Acceptor. What do you think?

raccoonback:
@violetagg 
Thank you for the suggestion! I think implementing this in `ServerTransport.Acceptor` is architecturally cleaner. 
However, I'd like to confirm one thing about `HTTP/3` support before proceeding.

Based on what I’ve confirmed so far, the following applies. Which approach would be more appropriate for handling `HTTP/3`?

### TCP-based protocols (HTTP/1.1, HTTP/2):
```  
ServerBootstrap
    └─> Acceptor.channelRead()  <- Can intercept here
         └─> child.pipeline().addLast(childHandler)
```
  The `Acceptor` is invoked when `!config.channelType(isDomainSocket).equals(DatagramChannel.class) `(ServerTransport.java:118), which covers TCP connections perfectly.

### QUIC-based HTTP/3:
```
DatagramChannel (UDP)
    └─> QuicServerCodecBuilder (Netty QUIC codec)
         └─> QuicChannel created internally by Netty
              └─> handler.initChannel()  <- Earliest point we can intercept
```

 For QUIC, there's no equivalent to `Acceptor` because:
  - `QuicServerCodecBuilde`r handles channel creation internally
  - We can only hook via handler(`ChannelInitializer`) ...

raccoonback:
@violetagg 
For `HTTP/3`, would it make sense to manage `maxConnections` separately at the time of `QuicChannel` initialization, instead of using the `Acceptor`?

That is:
- **HTTP/1.1** and **HTTP/2**
  - `Acceptor`: manages maxConnections for **HTTP/1.1** and **HTTP/2**.

- **HTTP/3**
  - `Http3ConnectionObserverBridge`: manages `maxConnections` for **HTTP/3**.
    - The name `Http3ConnectionObserverBridge` will be renamed to something more appropriate.

violetagg:
@raccoonback Let's do this feature only for TCP connections. If there is a requirement for HTTP/3 then we can extend this feature. What do you think?

raccoonback:
@violetagg 
That sounds good to me. I’ll update this to support only `HTTP/1.1`, `H2C`, and `HTTP/2` based **TCP**. and we can extend it later if `HTTP/3` support is needed.

raccoonback:
@violetagg 
I’ve updated the implementation so that `maxConnections` is enforced in the `Acceptor` only for **HTTP/1.1**, **H2C**, and **HTTP/2**.
Additionally, I added validation to throw an `UnsupportedOperationException` at bind time when `maxConnections` is configured with the **HTTP/3** protocol.

reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
```diff
@@ -367,22 +388,41 @@ static class Acceptor extends ChannelInboundHandlerAdapter {
...
+					if (log.isDebugEnabled()) {
+						log.debug(format(child, "Connection rejected: max connections limit reached ({})"), maxConnections);
+					}
+					forceClose(child, new RuntimeException("Connection rejected: max connections limit reached"));
+					return;
+				}
+				else {
+					child.closeFuture().addListener(future -> activeConnections.decrementAndGet());
+				}
+			}
```
violetagg, line 426: I think we can simplify a bit the operations as the increment happens only in our acceptor event loop, and the operation that is in another event loop is only decrement. Also I think that active connections variable belongs to the `Acceptor`

```
		volatile int activeConnections;
		static final AtomicIntegerFieldUpdater<Acceptor> ACTIVE_CONNECTIONS =
				AtomicIntegerFieldUpdater.newUpdater(Acceptor.class, "activeConnections");
```

```suggestion
			if (maxConnections > 0) {
				int current = ACTIVE_CONNECTIONS.get(this);
				if (current >= maxConnections) {
					if (log.isDebugEnabled()) {
						log.debug(format(child, "Connection rejected: max connections limit reached ({})"), maxConnections);
					}
					forceClose(child, new RuntimeException("Connection rejected: max connections limit reached"));
					return;
				}
				ACTIVE_CONNECTIONS.incrementAndGet(this);
				child.closeFuture().addListener(future -> ACTIVE_CONNECTIONS.decrementAndGet(this));
			}
```

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerBind.java
```diff
@@ -76,6 +77,13 @@ else if ((config._protocols & HttpServerConfig.h3) == HttpServerConfig.h3) {
 						"Configured HTTP/3 protocol without TLS. Configure TLS via HttpServer#secure"));
 			}
 		}
+
+		if ((config._protocols & HttpServerConfig.h3) == HttpServerConfig.h3 && config.maxConnections() >= 0) {
```
violetagg: We do not allow `0` at config time

```suggestion
		if ((config._protocols & HttpServerConfig.h3) == HttpServerConfig.h3 && config.maxConnections() > 0) {
```

reactor-netty-core/src/main/java/reactor/netty/transport/ServerTransport.java
raccoonback: @violetagg 
Good point. 
I’ll update the implementation to simplify this as suggested.

(Using `AtomicIntegerFieldUpdater` allows atomic operations without additional object allocation. 👍 )

raccoonback:
@violetagg
I’ve incorporated all the suggested review comments.
I’d appreciate it if you could take another look. 😀

PR #4039, opened 2025-12-16
Fix logging tests expected timezone

eze210:
My timezone is GMT-3, then these tests fail for me :)


violetagg:
@eze210 in which version (1.2.x or 1.3.x) do we want to have this

violetagg (approved):

eze210:
@violetagg it's just a fix in the test, I would add it just in main branch

violetagg:
Thanks for the PR!

Issue #4050, opened 2026-01-06
Http2Pool: concurrent acquires may allocate extra connections due to async deliver window


zimatars:
### Summary
When using HTTP/2, `Http2Pool` should preferentially multiplex concurrent requests on the same connection (until max concurrent streams is reached). However, under concurrent acquires, the pool can allocate a new connection even though an existing HTTP/2 connection still has available stream capacity.

This seems to be caused by a small timing window in `reactor.netty.http.client.Http2Pool#drainLoop`: the selected `slot` is removed from the internal queue, but it is only re-offered back to the pool asynchronously on the Channel's EventLoop. A second concurrent acquire that runs before the EventLoop task executes may not see the reusable connection and may trigger a new allocation.

### Affected code
`reactor.netty.http.client.Http2Pool#drainLoop` schedules delivery on the EventLoop:

```java
slot.connection.channel().eventLoop().execute(() -> {
    borrower.deliver(new Http2PooledRef(slot));
    drain();
});
```

`Borrower#deliver` is where concurrency is incremented and the slot is re-offered back to the pool. Because this happens asynchronously, there is a window where the slot is not visible to subsequent borrowers.

### Expected behavior
With HTTP/2 multiplexing enabled and the connection still below max concurrent streams:
- concurrent acquires should reuse the same HTTP/2 connection (open multiple streams),
- and should not allocate additional connections due to scheduling/timing.

### Actual behavior
Under concurrent acquires, the pool may allocate a new connection even though an existing HTTP/2 connection can still open a new stream.

### Reproduction (minimal)
This can be reproduced with a setup that:
1. Warm-ups the pool so one HTTP/2 connection is created and cached.
2. Triggers two acquires back-to-back / concurrently, while delaying the EventLoop execution of the deliver task.
3. Observes allocator invocation count (or allocated connections) > 1.

Pseudo steps:
- Create an HTTP/2 channel (e.g. `EmbeddedChannel` with `Http2FrameCodec` + `Http2MultiplexHandler`)
- Build `Http2Pool` with max connections > 1
- Acquire once and release to keep the connection cached
- Call `acquire().subscribe(...)` twice without running the channel pending tasks
- Check allocator is invoked only once (expected) but currently can be invoked twice (actual)
- Then run pending tasks to complete the delivers

### Reproducer (JUnit)
[...]

PR #4051, opened 2026-01-06
Improve `Http2Pool` connection reuse for concurrent acquires

zimatars:
This PR is for issue #4050 .

When using HTTP/2, concurrent requests should prefer multiplexing on an existing connection (until max concurrent streams is reached). There is a timing window in `Http2Pool#drainLoop` where a selected slot is temporarily not visible to other borrowers until the async deliver runs on the Channel EventLoop. During that window, concurrent acquires may allocate an extra connection.

Change: Reserve stream capacity and re-offer the selected slot before scheduling the async deliver, so concurrent borrowers can still see and reuse the connection.

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2AllocationStrategy.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2AllocationStrategy.java
@@ -68,15 +68,23 @@ public final class Http2AllocationStrategy implements ConnectionProvider.Allocat
 		Builder minConnections(int minConnections);
 
 		/**
-		 * Configures whether the HTTP/2 pool should strictly prefer reusing existing connections (multiplexing)
-		 * and avoid opening additional connections until all opened connections reach their max concurrent streams.
+		 * Configures whether the HTTP/2 pool should avoid opening additional connections as long as
+		 * existing connections have not reached their max concurrent streams limit.
+		 * When enabled, the pool may operate with fewer connections (even a single one) and will only
+		 * allocate a new connection when all existing connections have reached their max concurrent streams.
+		 * <p>
+		 * This behavior is automatically enabled when {@link #minConnections(int)} is configured with a value
+		 * greater than zero. Use this option to enable the same behavior without setting a minimum connections constraint.
 		 * <p>
 		 * Default to {@code false}.
 		 *
 		 * @param strictConnectionReuse whether strict connection reuse should be enabled
 		 * @return {@code this}
+		 * @since 1.3.3
 		 */
-		Builder strictConnectionReuse(boolean strictConnectionReuse);
+		default Builder strictConnectionReuse(boolean strictConnectionReuse) {
+			return this;
+		}
 	}
 
 	/**


violetagg:
@zimatars Can you please rebase in order to pickup #4061, it is relevant to the changes in this PR.

zimatars:
@violetagg I’ve rebased this PR on the latest main and included #4061. Could you please check it when you have a chance?

violetagg:
@zimatars I think we should apply this also to the branch when a connection is established. What do you think?

zimatars:
Hi @violetagg Thanks for the suggestion. I agree this should also cover the allocation path (when a connection is established).

On the “connection is established” path, changing `deliver(...)` alone doesn't seem sufficient. The allocation decision is made earlier in `drainLoop()`, so there is still an unavoidable gap between “permit granted / allocation in-flight” and “slot becomes visible in the queue”. A concurrent acquire that runs during this gap can observe `findConnection(...) == null` while `permitGranted > idleSize` (permit count includes in-flight allocations, while idle size is what is currently visible in the queue) and trigger another allocation.

### Proposed approach (drainLoop allocation decision)
Instead of adding more logic to `deliver(...)` in the “new connection established” branch, I was thinking to adjust the allocation decision in `drainLoop()`:

1. Capture `resourcesCount = idleSize` before scanning the queue.
2. Call `findConnection(resources, resourcesCount)` (or keep `findConnection(resources)` but use the captured `resourcesCount` for the decision).
3. If no slot is found and **`permitGranted > resourcesCount`**, treat it as “allocations / deliveries are already in-flight (a connection exists but is temporarily not visible)” and **do not allocate a new connection in this iteration**. The next `drain()` will be triggered once the in-flight deliver/allocation completes.
4. Otherwise, keep the existing behavior in the `slot == null` path unchanged (permit checks + allocate when needed).

### Scenarios
- **Slot temporarily not visible:** `resourcesCount == 0` while `permitGranted > 0` → a connection/slot exists, but it is temporarily not visible in the queue (selected by another borrower or allocation in-flight) → wait instead of allocating another connection.
- **No eligible slot in the queue:** `resourcesCount > 0` and `permitGranted == resourcesCount` but `findConnection == null` → all visible slots were filtered out by `findConnection` (e.g. reached max streams, GO_AWAY/eviction predicate/closed/inactive, etc.) and there is no in-flight allocation → proceed with allocating a new connection (subject to permits).
- **Connection is being established:** `permitGranted > resourcesCount` → permit was granted, but the new connection/slot is not visible in the queue yet → avoid starting another allocation in parallel.

If you think this approach makes sense, I'm happy to update the PR accordingly.

violetagg:
@zimatars What do you think about pushing this as is and you will create a new PR for the connection established scenario?

zimatars:
@violetagg  I was planning to use the same drainLoop() allocation decision adjustment to cover both the existing-connection visibility window and the “connection is established” window. With that, I think the extra deliver(...) changes in my commit can be reverted (keeping the main/#4061 deliver semantics unchanged) and we can keep the PR focused on the drainLoop() side. Would that be acceptable?

violetagg:
ok

zimatars:
@violetagg I’ve pushed an updated implementation in this PR. I **reverted the earlier `deliver(...)` changes** and instead adjusted the **allocation decision in `Http2Pool#drainLoop()`** so it covers both:

- the **selected-slot not visible** window (slot temporarily removed from the queue until the async `deliver` runs), and
- the **new-connection established** window (permit granted / allocation in-flight before the new slot becomes visible).

### Changes in `drainLoop()`
In the `slot == null` branch, we detect “there is already an in-flight allocation/delivery” by comparing:

- `permitGranted` (includes in-flight allocations), vs
- `resourcesCount` (what is currently visible in the queue, captured from `idleSize`)

If **`permitGranted > resourcesCount`**, we treat it as “a connection/slot exists but is temporarily not visible”, so we **don’t start another allocation in this drain iteration**. The next `drain()` is triggered once the in-flight deliver/allocation completes.

### Behavior differences
This makes HTTP/2 prefers multiplexing more consistently, but it changes a few edge behaviors:

1) **Pool from 0 → 1 (first connection):**
   - Concurrent acquires will **wait for the first connection to be established and delivered** instead of racing to create multiple connections in parallel.

2) **Existing connection temporarily not visible (selected by another borrower, before async deliver):**
   - Concurrent acquires will **wait for the selected slot to be re-offered / delivered**, instead of allocating an extra connection during the window.

3) **Max concurrent streams reached:**
   - If all visible connections are at max streams, we allocate **one new connection** (subject to permits), then concurrent acquires **wait for that new connection to become visible/delivered**.
   - We only allocate additional connections once the newly created one is also saturated (or otherwise not eligible).

### Test updates
Because the pool now multiplexes more aggressively, two tests needed to be updated:

- `HttpServerTests#doTestGracefulShutdown`: HTTP/2 GOAWAY is **connection-level**, and with multiplexing both concurrent requests can share a single connection, so the test should assert **at least one** GOAWAY rather than two.
- `Http2PooledConnectionProviderCustomMetricsTest#measureActiveStreamsSize`: the test previously used `doOnConnected` as a proxy for “5 connections”, but with HTTP/2 multiplexing the 5 requests can share **one connection with 5 streams**. The test now waits for metrics to observe `activeStreamSize == 5`.

violetagg:
I checked the changes and they look ok, however I think that the old change is still relevant, basically we return the connection immediately to the idle resources so that it can be reused. What do you think? Combining both implementations?
Do you use some performance tests for this implementation? If yes can you share some results?

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -412,10 +413,12 @@ void drainLoop() {
...
 				}
 				else {
-					int resourcesCount = idleSize;
 					if (minConnections > 0 &&
 							poolConfig.allocationStrategy().permitGranted() >= minConnections &&
-							resourcesCount == 0) {
+							poolConfig.allocationStrategy().permitGranted() > resourcesCount) {
+						// connections allocations were triggered
+					}
+					else if (minConnections == 0 && poolConfig.allocationStrategy().permitGranted() > resourcesCount) {
```
violetagg: With the current implementation, the meaning of `poolConfig.allocationStrategy().permitGranted() > resourcesCount` from one hand it is really `we gave the permit but the connection is in a process of establishment`, but from another hand it might be `we gave the permit, but the connection is taken in order to create a stream` because `resourcesCount` is the idle size only not all. What do you think?

zimatars:
@violetagg Thanks for the feedback — I agree the earlier `deliver(...)` change is still relevant (re-offering the connection immediately so it remains reusable). I’ll push an updated commit that combines the `deliver(...)` adjustment with the current `drainLoop()` allocation decision change.

I don’t have a full benchmark suite yet, but I did a quick (non-rigorous) local performance check to validate the impact.

### Setup (local)
- **CPU**: AMD Ryzen 7 8845HS (single host)
- **Topology**: JMeter -> **Server A** -> (h2c) -> **Server B**
- **Note**: Server A forwards requests to Server B, and Server B returns the response directly. Everything runs on the same machine, so these numbers are mainly directional rather than definitive.
- **Client pool config (A -> B)**: `protocol=HttpProtocol.H2C`, `maxConnections = Integer.MAX_VALUE`


### Results (200 JMeter threads)

**Before the optimization**
- **Throughput**: ~3551 req/s
- **Connections between A and B**: ~127

**After the optimization (prefer reuse / multiplexing)**
- **Throughput**: ~2208 req/s
- **Connections between A and B**: ~1

**After the optimization + setting `minConnections = 4`**
- **Throughput**: ~3721 req/s
- **Connections between A and B**: ~4


 With the optimization, the client side tends to **reuse a single HTTP/2 (h2c) connection** by default. On my setup this led to **only one connection between A and B**, which also means the channel is effectively handled by a single EventLoop thread and CPU utilization can be lower, limiting throughput.

 By setting `minConnections = 4`, this configuration allows a small number of parallel connections, which improves CPU utilization and throughput in this specific workload.
 
 Now that the pool more consistently reuses existing HTTP/2 connections, the default setup can converge to a single connection. Do you think it would make sense to revisit the default `minConnections` for HTTP/2 (e.g. defaulting to a small number, or something proportional to CPU cores) to improve out-of-the-box throughput while still benefiting from multiplexing?

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
zimatars: You’re right. permitGranted > resourcesCount can indeed indicate two different in-flight situations:
1) the new-connection established window (permit granted, connection/slot not yet visible in the queue)
2) the selected-slot not visible window (a slot was taken to open a stream and is temporarily not visible because delivery is async and resourcesCount tracks idle only)

That’s exactly what this guard is intended to cover to avoid triggering parallel allocations in both cases.

violetagg:
I'm wondering whether we should limit this optimisation to the min connections only or to add a configuration. The current proposal is such that until we reach max concurrent streams on all opened connections we will not open a new connection, from one hand this is nice but from other hand there are use cases when opening more that one connection scales better. What do you think?

zimatars:
@violetagg  I’d like to confirm my understanding of “limit this optimisation to the min connections only”: do you mean we should apply the optimisation only when minConnections is explicitly configured (i.e. keep the current behaviour unchanged when minConnections is not set / defaults to 0), and once the pool has reached minConnections we then prefer reuse and avoid opening additional connections until max concurrent streams is reached?

violetagg:
> @violetagg I’d like to confirm my understanding of “limit this optimisation to the min connections only”: do you mean we should apply the optimisation only when minConnections is explicitly configured (i.e. keep the current behaviour unchanged when minConnections is not set / defaults to 0), and once the pool has reached minConnections we then prefer reuse and avoid opening additional connections until max concurrent streams is reached?

yes correct or to introduce a configuration that enables this behaviour. What do you think?

zimatars:
@violetagg  I’d prefer adding a strictConnectionReuse flag configuration and keeping it disabled by default to preserve current scaling behaviour.

violetagg:
> @violetagg I’d prefer adding a strictConnectionReuse flag configuration and keeping it disabled by default to preserve current scaling behaviour.

ok

zimatars:
@violetagg I've pushed the changes. I added a strictConnectionReuse flag in Http2AllocationStrategy along with a setter in the Builder (which looks like the cause of the API compatibility failures).

Now, the connection reuse optimization is enabled if strictConnectionReuse is true or minConnections is explicitly set. Could you please review the code when you have a moment?

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2AllocationStrategy.java
```diff
@@ -66,6 +66,17 @@ public interface Builder {
...
+		/**
+		 * Configures whether the HTTP/2 pool should strictly prefer reusing existing connections (multiplexing)
+		 * and avoid opening additional connections until all opened connections reach their max concurrent streams.
+		 * <p>
+		 * Default to {@code false}.
+		 *
+		 * @param strictConnectionReuse whether strict connection reuse should be enabled
+		 * @return {@code this}
+		 */
+		Builder strictConnectionReuse(boolean strictConnectionReuse);
```
violetagg:
```suggestion

		/**
		 * Configures whether the HTTP/2 pool should avoid opening additional connections as long as
		 * existing connections have not reached their max concurrent streams limit.
		 * When enabled, the pool may operate with fewer connections (even a single one) and will only
		 * allocate a new connection when all existing connections have reached their max concurrent streams.
		 * <p>
		 * This behavior is automatically enabled when {@link #minConnections(int)} is configured with a value
		 * greater than zero. Use this option to enable the same behavior without setting a minimum connections constraint.
		 * <p>
		 * Default to {@code false}.
		 *
		 * @param strictConnectionReuse whether strict connection reuse should be enabled
		 * @return {@code this}
		 * @since 1.3.3
		 */
		default Builder strictConnectionReuse(boolean strictConnectionReuse) {
			return this;
		}
```

reactor-netty-http/src/test/java/reactor/netty/http/client/Http2AllocationStrategyTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2026 VMware, Inc. or its affiliates, All Rights Reserved.
```
violetagg:
```suggestion
 * Copyright (c) 2022-2026 VMware, Inc. or its affiliates, All Rights Reserved.
```

violetagg:
@zimatars Thanks for the PR!

PR #4118, opened 2026-02-28
Add WebSocket client metrics

### LINKED ISSUE #3820: Support for WebSocket client metrics
## Motivation

Currently, Reactor Netty provides metrics support for HTTP client/server interactions, but **WebSocket** client connections lack similar observability and metrics support. This makes it difficult to monitor and troubleshoot WebSocket connection lifecycles, message throughput, error rates, and latencies in production environments.

## Desired solution

Introduce built-in metrics collection for Reactor Netty WebSocket clients, similar to what is available for HTTP clients (`MicrometerHttpMetricsRecorder`). Like metrics for,  

* WebSocket session open/close count.
* Session duration.
* Send/receive message counts and rates
* Bytes sent/received per session / connection.

Metrics should be consistent with existing Micrometer integration and maybe follow OpenTelemetry semantic conventions. (if and when they come up)

## Considered alternatives

* Manually instrumenting each WebSocket client with `doOnConnect`, `doOnError`, `doOnNext`, etc., but this is error-prone and inconsistent.
* Wrapping `WebsocketClientSpec` and `WebSocketSession` to track metrics manually, which adds complexity and duplication.

LivingLikeKrillin:
## Summary

Add metrics support for WebSocket clients, including handshake time,
data received/sent, connection duration, and error counting.

- Add constants, meter/observation documentation, and recorder interfaces
  with Micrometer-based implementation
- Add channel handlers that replace the HTTP metrics handler on WebSocket
  upgrade and track connection duration via handler lifecycle
- Exclude control frames (Close, Ping, Pong) from data metrics

## Test plan

- [x] Handshake time timer
- [x] Data received time and distribution summary
- [x] Data sent time and distribution summary
- [x] Connection duration timer
- [x] Multiple connections aggregation
- [x] URI tag value normalization function
- [x] Handshake failure scenario
- [x] HTTP/1.1 and H2 protocol combinations

Related to #3820


violetagg:
@LivingLikeKrillin On most classes you have `@author raccoonback` is that the correct author? Also the copyright is 2025 while it should be 2026, also change the `@since 1.3.2` to `1.3.5`

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,262 @@
...
+			if (log.isWarnEnabled()) {
+				log.warn(format(ctx.channel(), "Exception caught while recording metrics."), e);
+			}
+		}
+
+		ctx.fireExceptionCaught(cause);
+	}
+
+	void initMetrics(ChannelHandlerContext ctx) {
+		ChannelOperations<?, ?> channelOps = ChannelOperations.get(ctx.channel());
```
violetagg: The constructor is always invoked within the WebsocketClientOperations/Http2WebsocketClientOperations. You can calculate immediately the `path` and the `contextView`

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,262 @@
...
+
+	private static long extractProcessedDataFromBuffer(Object msg) {
+		if (msg instanceof ByteBufHolder) {
+			return ((ByteBufHolder) msg).content().readableBytes();
+		}
+		else if (msg instanceof ByteBuf) {
+			return ((ByteBuf) msg).readableBytes();
+		}
+		return 0;
+	}
```
violetagg: You know that the object is `WebSocketFrame`

```suggestion
	private static long extractProcessedDataFromBuffer(WebSocketFrame msg) {
		return msg.content().readableBytes();
	}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,262 @@
...
+			recorder().incrementErrorsCount(remoteAddress, proxyAddress, path != null ? path : "unknown");
+		}
+	}
+
+	protected void recordRead(io.netty.channel.Channel channel, SocketAddress address) {
+		if (path == null) {
+			return;
+		}
+		if (proxyAddress == null) {
+			recorder().recordDataReceivedTime(address, path, "GET", "n/a",
```
violetagg: While it is true that for HTTP/1.1 the method is `GET`, for HTTP/2 is `CONNECT`

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,262 @@
...
+
+		//"FutureReturnValueIgnored" this is deliberate
+		ctx.write(msg, promise);
+	}
+
+	@Override
+	public void channelRead(ChannelHandlerContext ctx, Object msg) {
+		try {
+			if (msg instanceof WebSocketFrame && isDataFrame(msg)) {
+				dataReceivedTime = System.nanoTime();
```
violetagg: What will happen if the frame is `ContinuationWebSocketFrame`

violetagg (changes requested):
Thanks for the PR!
I started with the review. Can you please resolve the issues that I've identified till now?

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Moved the `path` and `contextView` computation into `swapMetricsHandler()` so they're resolved eagerly and passed directly through the constructor. Removed `initMetrics()` and the `uriTagValue` field.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Added `wsHttpMethod()` to `WebsocketClientOperations` (returns `"GET"`) with an override in `Http2WebsocketClientOperations` (returns `"CONNECT"` per RFC 8441). Replaced all hardcoded `"GET"` with the `method` field.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Applied the suggested change and also updated `isDataFrame()` to take `WebSocketFrame`. The `instanceof WebSocketFrame` check in `write()`/`channelRead()` now casts first, then delegates to both methods with the proper type.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: `ContinuationWebSocketFrame` extends `WebSocketFrame` and is not a control frame (Close/Ping/Pong), so it correctly passes the `isDataFrame()` check and its bytes are counted. The refactored code makes this more explicit by casting to `WebSocketFrame` first in `write()`/`channelRead()`.

violetagg: `dataReceiveTime` - is it the time for receiving/sending the whole frame, or just this particular part? Because if you have `ContinuationWebSocketFrame` this matters.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
violetagg: Can we have `path = null`?

reactor.netty.http.client.AbstractWebSocketClientMetricsHandler#resolvePath doesn't return null,
and `uriTagValue` has a definition `@Nullable Function<String, String>`, which means the function is not allowed to return null.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Currently it measures per-frame: `dataReceivedTime` is reset on every frame and `recordRead()`/`recordWrite()` is called immediately. With `ContinuationWebSocketFrame`, this means each fragment is recorded separately with near-zero duration.

I think per-message measurement would be more meaningful — accumulate bytes across fragments, set the start time only on the first frame, and record on `isFinalFragment()`. What do you think?

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: `resolvePath()` always returns non-null and `uriTagValue` returns `String`, so `path` can never be null after eager initialization. Will remove the `@Nullable` annotation and the null checks.

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
```diff
@@ -61,6 +61,14 @@ public class Metrics {
 	 */
 	public static final String HTTP_CLIENT_PREFIX = "reactor.netty.http.client";
 
+	/**
+	 * Name prefix that will be used for the WebSocket client's metrics
+	 * registered in Micrometer's global registry.
+	 *
+	 * @since 1.3.5
+	 */
+	public static final String WEBSOCKET_CLIENT_PREFIX = "reactor.netty.http.client.websocket";
```
violetagg: What do you think if we change this from `reactor.netty.http.client.websocket` to `reactor.netty.websocket.client`?

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,244 @@
...
+
+	private static final Logger log = Loggers.getLogger(AbstractWebSocketClientMetricsHandler.class);
+
+	final String method;
+	final @Nullable SocketAddress proxyAddress;
+	final SocketAddress remoteAddress;
+
+	@Nullable String path;
+
+	@Nullable ContextView contextView;
```
violetagg:
```suggestion
	String path;

	ContextView contextView;
```

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,244 @@
...
+	long dataSent;
+
+	long dataReceivedTime;
+
+	long dataSentTime;
+
+	long connectionStartTime;
+
+	protected AbstractWebSocketClientMetricsHandler(SocketAddress remoteAddress, @Nullable SocketAddress proxyAddress,
+			@Nullable String path, @Nullable ContextView contextView, String method) {
```
violetagg:
```suggestion
			String path, ContextView contextView, String method) {
```

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,244 @@
...
+		this.method = method;
+		this.path = path;
+		this.contextView = contextView;
+		this.proxyAddress = proxyAddress;
+		this.remoteAddress = remoteAddress;
+	}
+
+	protected AbstractWebSocketClientMetricsHandler(AbstractWebSocketClientMetricsHandler copy) {
```
violetagg: Do we need a copy constructor?

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,244 @@
...
+	@SuppressWarnings("FutureReturnValueIgnored")
+	public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
+		try {
+			if (msg instanceof WebSocketFrame) {
+				WebSocketFrame frame = (WebSocketFrame) msg;
+				if (isDataFrame(frame)) {
+					dataSentTime = System.nanoTime();
+					dataSent += extractProcessedDataFromBuffer(frame);
+
+					recordWrite(remoteAddress);
```
violetagg: What do you think about something like this?

```suggestion
					if (dataSentTime == 0) {
						dataSentTime = System.nanoTime();
					}
					dataSent += extractProcessedDataFromBuffer(frame);

					if (frame.isFinalFragment()) {
						promise.addListener(f -> {
							try {
								dataSent = 0;
								dataSentTime = 0;
								recordWrite(remoteAddress);
							}
							catch (RuntimeException e) {
								if (log.isWarnEnabled()) {
									log.warn(format(ctx.channel(), "Exception caught while recording metrics."), e);
								}
							}
						});
					}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,244 @@
...
+	public void channelRead(ChannelHandlerContext ctx, Object msg) {
+		try {
+			if (msg instanceof WebSocketFrame) {
+				WebSocketFrame frame = (WebSocketFrame) msg;
+				if (isDataFrame(frame)) {
+					dataReceivedTime = System.nanoTime();
+					long bytes = extractProcessedDataFromBuffer(frame);
+					dataReceived += bytes;
+
+					recordRead(ctx.channel(), remoteAddress);
```
violetagg: What do you think about something like this?

```suggestion
					if (dataReceivedTime == 0) {
						dataReceivedTime = System.nanoTime();
					}
					dataReceived += extractProcessedDataFromBuffer(frame);

					if (frame.isFinalFragment()) {
						dataReceived = 0;
						dataReceivedTime = 0;
						recordRead(ctx.channel(), remoteAddress);
					}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,244 @@
...
+		else {
+			recorder().recordDataSentTime(address, proxyAddress, path, method,
+					Duration.ofNanos(System.nanoTime() - dataSentTime));
+
+			recorder().recordDataSent(address, proxyAddress, path, dataSent);
+		}
+		dataSent = 0;
+	}
+
+	static String resolvePath(HttpClientOperations ops) {
```
violetagg: Can we move this to reactor.netty.http.client.WebsocketClientOperations as it belongs there?

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -85,7 +90,7 @@ void initHandshaker(URI currentURI, WebsocketClientSpec websocketClientSpec) {
 		// Returned value is deliberately ignored
 		addHandlerFirst(NettyPipeline.HttpAggregator, new HttpObjectAggregator(8192));
 
-		removeHandler(NettyPipeline.HttpMetricsHandler);
+		swapMetricsHandler(currentURI);
```
violetagg: Is the method parameter needed?

```suggestion
		swapMetricsHandler();
```

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -148,11 +153,18 @@ public void onInboundNext(ChannelHandlerContext ctx, Object msg) {
...
+						micrometerWsHandler.recordHandshakeComplete(channel(),
+								String.valueOf(response.status().code()));
+					}
 					// This change is needed after the Netty change https://github.com/netty/netty/pull/11966
 					ctx.read();
 					listener().onStateChange(this, HttpClientState.RESPONSE_RECEIVED);
 				}
 				catch (Exception e) {
+					if (micrometerWsHandler != null) {
+						micrometerWsHandler.recordHandshakeFailure(channel(), "ERROR");
```
violetagg: Do we have other status than `ERROR`?

```suggestion
						micrometerWsHandler.recordHandshakeFailure(channel());
```

violetagg (changes requested):

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
LivingLikeKrillin: `reactor.netty.websocket.client` is a cleaner separation. Will update.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Will apply.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Will apply.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: The copy constructor is never invoked — WebSocket handlers are always created fresh in `swapMetricsHandler()`. Will remove.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: One question on the ordering inside the listener — `recordWrite()` reads `dataSent` and `dataSentTime` directly, so resetting them to 0 before the call would record zero bytes. Should the reset happen after `recordWrite()`?

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Same question as the `write()` suggestion — `recordRead()` reads `dataReceived` and `dataReceivedTime` directly. Should the reset happen after `recordRead()`?

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Only called from `swapMetricsHandler()` and depends on `HttpClientOperations`. Will move it there.

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
LivingLikeKrillin: Will apply.

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
LivingLikeKrillin: No other status — it's always `"ERROR"`. Will remove the parameter.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
violetagg: sorry it's a mistake

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
violetagg: sorry it's a mistake

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,226 @@
...
+
+	private static final Logger log = Loggers.getLogger(AbstractWebSocketClientMetricsHandler.class);
+
+	final String method;
+	final @Nullable SocketAddress proxyAddress;
+	final SocketAddress remoteAddress;
+
+	String path;
+
+	ContextView contextView;
```
violetagg:
```suggestion
	final String path;

	final ContextView contextView;
```

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,226 @@
...
+	long dataReceived;
+
+	long dataSent;
+
+	long dataReceivedTime;
+
+	long dataSentTime;
+
+	long connectionStartTime;
+
```
violetagg, line 64: What do you think about moving the handshake timer so that it is available for all variants?

```suggestion

	long handshakeStartTime;

```

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,226 @@
...
+		this.proxyAddress = proxyAddress;
+		this.remoteAddress = remoteAddress;
+	}
+
+	@Override
+	public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
+		super.handlerAdded(ctx);
+		connectionStartTime = System.nanoTime();
+	}
+
```
violetagg, line 88: What do you think about moving the handshake timer so that it is available for all variants?

```suggestion

	void startHandshake(Channel channel) {
		handshakeStartTime = System.nanoTime();
	}

	void recordHandshakeComplete(Channel channel, String status) {
		Duration time = Duration.ofNanos(System.nanoTime() - handshakeStartTime);
		if (proxyAddress == null) {
			recorder().recordWebSocketHandshakeTime(remoteAddress, path, status, time);
		}
		else {
			recorder().recordWebSocketHandshakeTime(remoteAddress, proxyAddress, path, status, time);
		}
	}

	void recordHandshakeFailure(Channel channel) {
		Duration time = Duration.ofNanos(System.nanoTime() - handshakeStartTime);
		if (proxyAddress == null) {
			recorder().recordWebSocketHandshakeTime(remoteAddress, path, "ERROR", time);
		}
		else {
			recorder().recordWebSocketHandshakeTime(remoteAddress, proxyAddress, path, "ERROR", time);
		}
	}

```

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,226 @@
...
+						dataSentTime = 0;
+					}
+				}
+			}
+		}
+		catch (RuntimeException e) {
+			if (log.isWarnEnabled()) {
+				log.warn(format(ctx.channel(), "Exception caught while recording metrics."), e);
+			}
+		}
```
violetagg, line 175: What do you think if we do the recording once we know that the write was successful?

```suggestion
		if (msg instanceof WebSocketFrame) {
			WebSocketFrame frame = (WebSocketFrame) msg;
			if (isDataFrame(frame)) {
				if (dataSentTime == 0) {
					dataSentTime = System.nanoTime();
				}
				dataSent += extractProcessedDataFromBuffer(frame);

				if (frame.isFinalFragment()) {
					promise.addListener(f -> {
						try {
							recordWrite(remoteAddress);
							dataSentTime = 0;
						}
						catch (RuntimeException e) {
							if (log.isWarnEnabled()) {
								log.warn(format(ctx.channel(), "Exception caught while recording metrics."), e);
							}
						}
					});
				}
			}
		}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/ContextAwareWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,106 @@
...
+			String method) {
+		super(remoteAddress, proxyAddress, path, contextView, method);
+		this.recorder = recorder;
+	}
+
+	@Override
+	protected ContextAwareWebSocketClientMetricsRecorder recorder() {
+		return recorder;
+	}
+
```
violetagg, line 50: Bring here an implementation for handshake

```suggestion

	@Override
	void recordHandshakeComplete(io.netty.channel.Channel channel, String status) {
		Duration time = Duration.ofNanos(System.nanoTime() - handshakeStartTime);
		if (proxyAddress == null) {
			recorder.recordWebSocketHandshakeTime(contextView, remoteAddress, path, status, time);
		}
		else {
			recorder.recordWebSocketHandshakeTime(contextView, remoteAddress, proxyAddress, path, status, time);
		}
	}

	@Override
	void recordHandshakeFailure(io.netty.channel.Channel channel) {
		Duration time = Duration.ofNanos(System.nanoTime() - handshakeStartTime);
		if (proxyAddress == null) {
			recorder.recordWebSocketHandshakeTime(contextView, remoteAddress, path, "ERROR", time);
		}
		else {
			recorder.recordWebSocketHandshakeTime(contextView, remoteAddress, proxyAddress, path, "ERROR", time);
		}
	}

```

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,173 @@
...
+ * @since 1.3.5
+ */
+final class MicrometerWebSocketClientMetricsHandler extends AbstractWebSocketClientMetricsHandler {
+	final MicrometerWebSocketClientMetricsRecorder recorder;
+
+	@Nullable HandshakeTimeHandlerContext handshakeTimeHandlerContext;
+	@Nullable Observation handshakeTimeObservation;
+	@Nullable ContextView parentContextView;
+
+	long handshakeStartTime;
```
violetagg: This is moved to AbstractWebSocketClientMetricsHandler

```suggestion
```

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,173 @@
...
+		this.recorder = recorder;
+	}
+
+	@Override
+	protected WebSocketClientMetricsRecorder recorder() {
+		return recorder;
+	}
+
+	void startHandshake(Channel channel) {
+		handshakeStartTime = System.nanoTime();
```
violetagg: This is moved to AbstractWebSocketClientMetricsHandler

```suggestion
		super.startHandshake(channel);
```

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -68,6 +71,8 @@ class WebsocketClientOperations extends HttpClientOperations
 	final Sinks.One<WebSocketCloseStatus> onCloseState;
 	final boolean proxyPing;
 
+	@Nullable MicrometerWebSocketClientMetricsHandler micrometerWsHandler;
```
violetagg: We have handshake timer for all

```suggestion
	@Nullable AbstractWebSocketClientMetricsHandler micrometerWsHandler;
```

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -224,6 +236,59 @@ boolean isHandshakeComplete() {
...
+			String httpMethod = wsHttpMethod();
+
+			AbstractWebSocketClientMetricsHandler wsHandler;
+			if (httpHandler instanceof MicrometerHttpClientMetricsHandler) {
+				MicrometerWebSocketClientMetricsHandler micrometerHandler = new MicrometerWebSocketClientMetricsHandler(
+						MicrometerWebSocketClientMetricsRecorder.INSTANCE,
+						httpHandler.remoteAddress, httpHandler.proxyAddress, resolvedPath, ctxView, httpMethod);
+				micrometerHandler.startHandshake(channel);
+				this.micrometerWsHandler = micrometerHandler;
+				wsHandler = micrometerHandler;
```
violetagg: Let's here only construct the new handler

```suggestion
				wsHandler = new MicrometerWebSocketClientMetricsHandler(
						MicrometerWebSocketClientMetricsRecorder.INSTANCE,
						httpHandler.remoteAddress, httpHandler.proxyAddress, resolvedPath, ctxView, httpMethod);
```

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -224,6 +236,59 @@ boolean isHandshakeComplete() {
...
+			}
+			else if (httpHandler instanceof ContextAwareHttpClientMetricsHandler) {
+				ContextAwareHttpClientMetricsHandler ctxHandler = (ContextAwareHttpClientMetricsHandler) httpHandler;
+				wsHandler = new ContextAwareWebSocketClientMetricsHandler(
+						new DefaultContextAwareWebSocketClientMetricsRecorder(ctxHandler.recorder),
+						httpHandler.remoteAddress, httpHandler.proxyAddress, resolvedPath, ctxView, httpMethod);
```
violetagg, line 270: Wrap the recorder only if it is not ContextAwareWebSocketClientMetricsRecorder

```suggestion
				ContextAwareWebSocketClientMetricsRecorder wsRecorder;
				if (ctxHandler.recorder instanceof ContextAwareWebSocketClientMetricsRecorder) {
					wsRecorder = (ContextAwareWebSocketClientMetricsRecorder) ctxHandler.recorder;
				}
				else {
					wsRecorder = new DefaultContextAwareWebSocketClientMetricsRecorder(ctxHandler.recorder);
				}
				wsHandler = new ContextAwareWebSocketClientMetricsHandler(
						wsRecorder,
						httpHandler.remoteAddress, httpHandler.proxyAddress, resolvedPath, ctxView, httpMethod);
```

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -224,6 +236,59 @@ boolean isHandshakeComplete() {
...
+			}
+			else {
+				wsHandler = new WebSocketClientMetricsHandler(
+						MicrometerWebSocketClientMetricsRecorder.INSTANCE,
+						httpHandler.remoteAddress, httpHandler.proxyAddress, resolvedPath, ctxView, httpMethod);
+			}
```
violetagg: We will need a wrapper DefaultWebSocketClientMetricsRecorder similar to what we have to context aware. What do you think?

```suggestion
			else if (httpHandler instanceof HttpClientMetricsHandler) {
				HttpClientMetricsHandler plainHandler = (HttpClientMetricsHandler) httpHandler;
				WebSocketClientMetricsRecorder wsRecorder;
				if (plainHandler.recorder instanceof WebSocketClientMetricsRecorder) {
					wsRecorder = (WebSocketClientMetricsRecorder) plainHandler.recorder;
				}
				else {
					wsRecorder = new DefaultWebSocketClientMetricsRecorder(plainHandler.recorder);
				}
				wsHandler = new WebSocketClientMetricsHandler(
						wsRecorder,
						httpHandler.remoteAddress, httpHandler.proxyAddress, resolvedPath, ctxView, httpMethod);
			}
			else {
				return;
			}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
```diff
@@ -224,6 +236,59 @@ boolean isHandshakeComplete() {
...
+
+			channel.pipeline().addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.WsMetricsHandler, wsHandler);
```
violetagg, line 290:
```suggestion
			wsHandler.startHandshake(channel);
			this.micrometerWsHandler = wsHandler;
			channel.pipeline().addBefore(NettyPipeline.ReactiveBridge, NettyPipeline.WsMetricsHandler, wsHandler);
```

violetagg (changes requested):
I have proposal for moving handshake timer to the abstract class so that it is available for all handlers. What do you think?

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,226 @@
...
+				!(msg instanceof PongWebSocketFrame);
+	}
+
+	private static long extractProcessedDataFromBuffer(WebSocketFrame msg) {
+		return msg.content().readableBytes();
+	}
+
+	protected abstract WebSocketClientMetricsRecorder recorder();
+
+	protected void recordConnectionClosed(ChannelHandlerContext ctx) {
```
violetagg: Do we need the method param?

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,226 @@
...
+		Duration duration = Duration.ofNanos(System.nanoTime() - connectionStartTime);
+		if (proxyAddress == null) {
+			recorder().recordWebSocketConnectionDuration(remoteAddress, path, duration);
+		}
+		else {
+			recorder().recordWebSocketConnectionDuration(remoteAddress, proxyAddress, path, duration);
+		}
+	}
+
+	protected void recordException(ChannelHandlerContext ctx) {
```
violetagg: Do we need the method param?

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
```diff
@@ -0,0 +1,226 @@
...
+		if (proxyAddress == null) {
+			recorder().incrementErrorsCount(remoteAddress, path);
+		}
+		else {
+			recorder().incrementErrorsCount(remoteAddress, proxyAddress, path);
+		}
+	}
+
+	protected void recordRead(io.netty.channel.Channel channel, SocketAddress address) {
```
violetagg: Do we need the method param `channel`?

violetagg (changes requested):

violetagg:
Please also ensure you have `Signed-off-by` so that we have a happy DCO check

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Applied.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Applied.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Applied. The base implementation calls `recorder().recordWebSocketHandshakeTime()` with duration. `MicrometerWebSocketClientMetricsHandler` overrides `startHandshake()` to add observation logic, and `ContextAwareWebSocketClientMetricsHandler` overrides `recordHandshakeComplete()`/`recordHandshakeFailure()` to pass `contextView`.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Applied. One addition: used `promise.unvoid()` before `addListener()` to handle `VoidChannelPromise`, which does not support listeners. This pattern is already used in `TransportConnector` and `ProxyProvider` in the codebase.

reactor-netty-http/src/main/java/reactor/netty/http/client/ContextAwareWebSocketClientMetricsHandler.java
LivingLikeKrillin: Applied.

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
LivingLikeKrillin: Applied.

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
LivingLikeKrillin: Applied.

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
LivingLikeKrillin: Applied.

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
LivingLikeKrillin: Created. Delegates data-level methods (`recordDataSent`, `recordDataReceived`, etc.) to the wrapped `HttpClientMetricsRecorder` and no-ops WebSocket-specific methods (`recordWebSocketHandshakeTime`, `recordWebSocketConnectionDuration`).

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
LivingLikeKrillin: Applied. `wsHandler.startHandshake(channel)`, `this.micrometerWsHandler = wsHandler`, and the pipeline addition are now after the if-else block, shared by all three handler variants.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Removed.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Removed.

reactor-netty-http/src/main/java/reactor/netty/http/client/AbstractWebSocketClientMetricsHandler.java
LivingLikeKrillin: Removed.

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerWebSocketClientMetricsHandler.java
LivingLikeKrillin: Applied.

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerWebSocketClientMetricsHandler.java
LivingLikeKrillin: Applied.

LivingLikeKrillin:
Hi @violetagg, round 3 changes are pushed. Here's what was addressed:

- ✅ **Handshake timer moved to `AbstractWebSocketClientMetricsHandler`** —
  `handshakeStartTime`, `startHandshake`, `recordHandshakeComplete`, and
  `recordHandshakeFailure` now live in the abstract class.
- ✅ **`MicrometerWebSocketClientMetricsHandler`** now calls
  `super.startHandshake(channel)` and keeps only the Micrometer Observation
  setup on top. `recordHandshakeComplete`/`recordHandshakeFailure` are
  overridden to stop the Observation instead of invoking the recorder, as the
  Observation replaces recorder-based handshake time recording.
- ✅ **`ContextAwareWebSocketClientMetricsHandler`** overrides
  `recordHandshakeComplete` and `recordHandshakeFailure` to pass `contextView`
  to the recorder.
- ✅ **`micrometerWsHandler` field type** in `WebsocketClientOperations`
  changed to `AbstractWebSocketClientMetricsHandler`, since the handshake
  timer is now available for all variants.
- ✅ **Micrometer branch in `swapMetricsHandler` simplified** to only
  construct the new handler, matching your suggestion.
- ✅ **`DefaultWebSocketClientMetricsRecorder` wrapper added** to wrap a plain
  `HttpClientMetricsRecorder` when the user supplies a non-WebSocket-aware
  recorder, mirroring the context-aware pattern.
- ✅ **Unused parameters removed** from `AbstractWebSocketClientMetricsHandler`
  — `ctx` from `recordConnectionClosed` and `recordException`, and `channel`
  from `recordRead`, as suggested.
- ✅ **`path` and `contextView` fields** in
  `AbstractWebSocketClientMetricsHandler` are now `final`.

The DCO check is green. Could you take another look when you have a moment?

reactor-netty-core/src/main/java/reactor/netty/Metrics.java
violetagg:
```suggestion
 * Copyright (c) 2019-2026 VMware, Inc. or its affiliates, All Rights Reserved.
```

reactor-netty-http/src/main/java/reactor/netty/http/client/WebsocketClientOperations.java
violetagg:
```suggestion
 * Copyright (c) 2011-2026 VMware, Inc. or its affiliates, All Rights Reserved.
```

violetagg (changes requested):

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerWebSocketClientMetricsRecorder.java
```diff
@@ -0,0 +1,257 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty.http.client;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.Timer;
+import org.jspecify.annotations.Nullable;
+import reactor.netty.Metrics;
```
violetagg:
```suggestion
```

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerWebSocketClientMetricsRecorder.java
```diff
@@ -0,0 +1,257 @@
...
+
+import static reactor.netty.Metrics.CONNECTION_DURATION;
+import static reactor.netty.Metrics.DATA_RECEIVED;
+import static reactor.netty.Metrics.DATA_RECEIVED_TIME;
+import static reactor.netty.Metrics.DATA_SENT;
+import static reactor.netty.Metrics.DATA_SENT_TIME;
+import static reactor.netty.Metrics.ERRORS;
+import static reactor.netty.Metrics.HANDSHAKE_TIME;
+import static reactor.netty.Metrics.NA;
+import static reactor.netty.Metrics.PROXY_ADDRESS;
```
violetagg:
```suggestion
```

reactor-netty-http/src/main/java/reactor/netty/http/client/MicrometerWebSocketClientMetricsRecorder.java
```diff
@@ -0,0 +1,257 @@
...
+import static reactor.netty.Metrics.REGISTRY;
+import static reactor.netty.Metrics.REMOTE_ADDRESS;
+import static reactor.netty.Metrics.STATUS;
+import static reactor.netty.Metrics.URI;
```
violetagg:
```suggestion
```

violetagg (changes requested):

reactor-netty-http/src/test/java/reactor/netty/http/WebSocketClientMetricsHandlerTests.java
```diff
@@ -0,0 +1,496 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty.http;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import io.netty.buffer.Unpooled;
+import io.netty.handler.codec.http.websocketx.CloseWebSocketFrame;
```
violetagg:
```suggestion
```

violetagg (changes requested):

LivingLikeKrillin:
Thanks for picking up the copyright and unused-import cleanup directly.

The remaining CI failure was `NativeConfigTest.testChannelHandler()` — the new WebSocket `ChannelHandler` classes weren't registered in `reflect-config.json` for GraalVM native image. Pushed `b48178d8` to add them:

- `AbstractWebSocketClientMetricsHandler`
- `ContextAwareWebSocketClientMetricsHandler`
- `MicrometerWebSocketClientMetricsHandler`
- `WebSocketClientMetricsHandler`

Verified locally by running the same commands CI does:

- `./gradlew clean spotlessCheck -PspotlessFrom=origin/main` — PASS
- `./gradlew clean japicmp` — PASS
- `./gradlew :reactor-netty-http:check -PforceTransport=nio` (with the Java 8 flags) — PASS, no flakes

Ready for another CI run.

violetagg:
I'm gonna review this at the end of the week.

violetagg:
@LivingLikeKrillin I’m tied up with some other priorities for the next couple of weeks, so I likely won't get a chance to look at this until after that. I'll pick this up as soon as my schedule clears up—thanks for your patience!

LivingLikeKrillin:
Hi @violetagg — while you're away, I did a self-review of the WebSocket client metrics implementation and noticed that the handshake-time timer was only being recorded on successful handshake or when `finishHandshake` itself threw. Four other failure paths silently dropped the metric.

Pushed two commits:

**1801d4e7 — Record WebSocket handshake failure on missing termination paths**
- Adds a `handshakeFinalized` idempotency guard so `recordHandshakeComplete`/`recordHandshakeFailure` become no-ops on second invocation
- Wires `recordHandshakeFailure` into the four previously missing paths (premature close in H1, init-handshaker write listener in H1 and H2, non-OK FullHttpResponse in H2)
- Moves `recordHandshakeComplete` after `listener().onStateChange()` so exceptions thrown by downstream observers correctly route through the catch block and trigger `recordHandshakeFailure`
- Adds `testWebSocketHandshakeFailureOnHttp2NonOK` covering the H2 non-OK FullHttpResponse path
- Redirect and auth-required responses are intentionally not instrumented here — the retry attempt records its own timer sample, and tagging the original attempt as `STATUS=ERROR` would phantom-pair with the eventual `STATUS=101` success

**39a0b745 — Polish handshake-finalization plumbing**
- Aligns `handshakeFinalized` with the `closeSent` `AtomicIntegerFieldUpdater` pattern already used in `WebsocketClientOperations`
- Extracts a `recordHandshakeFailure(Channel)` helper to deduplicate the six call sites (four new wirings + the two existing `catch (Exception e)` blocks); wraps the recorder call in try/catch + WARN log matching the existing convention in `AbstractWebSocketClientMetricsHandler.handlerRemoved`, `write`, `channelRead`, and `exceptionCaught`

Looking forward to your feedback.

LivingLikeKrillin:
Quick self-review summary while the review is paused — capturing context for the recent commits in one place.

**Sent-data per-message attribution (new fix + regression test)**
`dataSent`/`dataSentTime` are shared handler fields that were read and reset inside the async write-completion listener. When several messages are written in a single burst, a later message mutates those fields before the previous message's listener fires, so the whole burst collapsed into a single `data.sent` record and `data.sent.time` got a `now - 0` (garbage) value. Fixed by snapshotting the values before registering the listener and recording from the snapshot. Added `testWebSocketMultipleMessagesSentMetrics`, which asserts per-message byte attribution and a sane send time. The inbound path records synchronously, so it is unaffected.

**Documentation (new)**
Added the `reactor.netty.websocket.client.*` meters table to the HTTP client reference docs, which were previously missing.

**Handshake-failure paths (pushed earlier)**
For context, the two earlier commits cover handshake-time being dropped on several termination paths.

**Investigated, no change needed**
Also checked the `followRedirect` and `httpAuthenticationWhen` flows (3xx/401 on the upgrade): the intermediate connection does not record a spurious handshake metric, so nothing changed there.

The commits can be squashed or reorganized as preferred.

violetagg:
@LivingLikeKrillin I plan to review and merge this PR this week.

violetagg (approved):

violetagg:
@LivingLikeKrillin Thanks for the PR!

Issue #940, opened 2019-12-23
reactor netty does not instant propagate exception upon an attempt to send data to closed connection (regression from spring-boot 2.1.4, reactor-netty 0.8.6)


bsavov:
When try to send message (SSE) to already closed connection (channel) I do not get instant exception, instead the exception is propagated on the next attempt to send another message.
Up to spring-boot 2.1.4 everything was fine and the exception was propagated instant upon an attempt to send something to the channel. After that (upgrade to any spring-boot.version >2.1.4) the exception is thrown always on second message.

## Expected Behavior

Instant exception upon an attempt to send (SSE) event to closed connection as it does up to spring-boot version 2.1.4 (including)
## Actual Behavior

Exception is not thrown on the first unsuccessful sent (SSE) event to closed connection, but on the second one, thus the first event is lost

## Steps to Reproduce

The issue explained in stackoverflow: https://stackoverflow.com/questions/59419542/detect-disconnect-on-sse-with-reactor-netty-spring-boot-2-2-x
simplified code for both working 2.1.4 and non-working 2.2.2 versions available in github: https://github.com/bsavov/bs-sse
```java
@RestController
@Slf4j
public class SseController {
    private int counter = 0;

    @Autowired
    private SseService sseService;

    @GetMapping(value = "/sse", produces = MediaType.TEXT_EVENT_STREAM_VALUE)
    public Flux<ServerSentEvent<String>> subsEvents() {
        return sseService.getEvents()
                .map(s -> ServerSentEvent.<String>builder().data(s).build());
    }

    public Flux<String> getEvents() {
        AtomicInteger lastProcessed = new AtomicInteger();
        return Flux
                .interval(Duration.ofSeconds(1))
                .map(i -> i + " -> " + counter)
                .doOnNext(v -> lastProcessed.set(counter++))
                .doOnEach(sig -> {
                    logger.info("{} vs {}", sig.get(), sig.getThrowable());
                })
                .doFinally(st -> {
                    logger.info("Completed with signal type: {}. lastProcessed: {}", st, lastProcessed);
                    counter = lastProcessed.get();
                });
    }
}
```
## Your Environment

* Reactor version(s) used: reactor-netty 0.9.2-RELEASE which comes with spring-boot 2.2.2
* Other relevant libraries versions (eg. `netty`, ...):
* JVM version (`javar -version`):
  * java version "1.8.0_231"
    Java(TM) SE Runtime Environment (build 1.8.0_231-b11)
[...]

PR #4189, opened 2026-04-10
Fix `keep-alive` race when creating `HttpServerOperations`

### LINKED ISSUE #4188: HTTP/1.1 keep-alive can hang when a new request inherits the previous request's `ChannelOperations` as parent   connection
## Summary

**Impact:** Requests can hang on affected keep-alive connections; in one failure mode the connection can become permanently unusable.

In HTTP/1.1 keep-alive scenarios, `HttpTrafficHandler` can incorrectly assign a previous, terminating `HttpServerOperations` as the parent connection for a new request. This creates a transitive dependency chain that can later resolve to `DisposedChannel`, which in our testing led to either silent hangs or `AbortedException`.

## Expected Behavior

On an HTTP/1.1 keep-alive connection, each new request should receive an `HttpServerOperations` whose parent `Connection` is the stable connection bound to the channel (for example `ReactorNetty.SimpleConnection`), not the previous request's `HttpServerOperations`.

Since `HttpServerOperations` extends `ChannelOperations` and can terminate concurrently, using a previous request's ops as the parent can expose the new request to the disposal state of an unrelated request. The connection should continue serving sequential requests on the same keep-alive socket.

## Actual Behavior

There is a TOCTOU race involving parent selection in `HttpTrafficHandler.channelRead()` and concurrent termination of the previous request's `HttpServerOperations`. `Connection.from(ctx.channel())` can return a previous request's `HttpServerOperations` that terminates between the `Connection.from()` call and the subsequent `new HttpServerOperations(...)` / `bind()` call. The result is a chain of transitive references that resolves to `DisposedChannel`, leaving the affected request — and in one of the two failure modes, the entire keep-alive connection — stuck until the client times out and abandons the connection.

### Root Cause

When a new HTTP request arrives on a keep-alive connection, `HttpTrafficHandler.channelRead()` creates a new `HttpServerOperations` using `Connection.from(ctx.channel())` as the parent connection:

```java
ops = new HttpServerOperations(Connection.from(ctx.channel()), listener, request, ...);
ops.bind();
listener.onStateChange(ops, ConnectionObserver.State.CONFIGURED);
ctx.fireChannelRead(msg);
```

`Connection.from(ctx.channel())` reads the channel's `CONNECTION` attribute, which at that moment may still hold the previous request's `HttpServerOperations` (e.g. `ops2`) if it has not terminated yet.
After this read, but before the newly created operations later dereference their parent in the `bind()` or send path, `ops2.terminate()` can run on another thread and complete successfully — setting `ops2.connection = new DisposedConnection(channel())` inside the `if (rebind(connection))` block.

At that point the newly created `ops3` has `ops2` as its parent, but `ops2.connection` has already transitioned to `DisposedConnection`. The chain becomes:

```text
ops3.channel()
 → ops3.connection.channel()      // ops3.connection = ops2
 → ops2.connection.channel()      // ops2.connection = DisposedConnection (after terminate)
 → DisposedChannel                // inert placeholder
```

Additionally, `ops2.terminate()` succeeded with `rebind(ops1)` — a `CAS(ops2, ops1)` on the real channel attribute — which means the real channel's `CONNECTION` attribute is now `ops1`. Meanwhile `ops3.bind()` updates only `DisposedChannel.attr(CONNECTION)`, so the real channel attribute remains unchanged. As a result:

- The real channel attribute is pinned to `ops1` (a zombie whose `OUTBOUND_CLOSE` is already `cancelledSubscription`).
- `ops3` is bound only to `DisposedChannel`, invisible to the rest of the pipeline.

### How ops1 becomes a zombie

Before the race above occurs, `ops1` (request 1's `HttpServerOperations`) enters a zombie state through this sequence:

1. `ops1` handles request 1, writes the response, and `handleLastHttpContent()` calls `ctx.read()` — enabling the next request.
2. Request 2 arrives. `Connection.from(ctx.channel())` returns `ops1` (still in the attribute). `ops2` is created with `ops1` as parent. `ops2.bind()` updates the channel attribute to `ops2`.
3. `ops1`'s reactive chain completes on another thread (e.g. `Mono.delay()` on `Schedulers.parallel()`).
   - `ops1.onComplete()` sets `OUTBOUND_CLOSE` to `cancelledSubscription`.
   - `ops1.terminate()` attempts `CAS(ops1, SimpleConnection)` on the channel attribute, but the attribute is now `ops2` — the CAS fails.
   - Because `connection = new DisposedConnection(channel())` is inside the `if (rebind(connection))` block, `ops1.connection` remains `SimpleConnection`. However, `ops1.OUTBOUND_CLOSE` is permanently set to `cancelledSubscription`.

### Case A: ops2.terminate() races between Connection.from and bind — silent hang

This is the primary failure mode. The race window is between `Connection.from(ctx.channel())` returning `ops2` and the `new HttpServerOperations(ops2)` / `ops3.bind()` call completing.

**Timeline:**

| Step | Event Loop Thread | Parallel Scheduler Thread |
|------|-------------------|---------------------------|
| 1 | `Connection.from(channel)` → `ops2` | |
| 2 | | `ops2.onComplete()` |
| 3 | | `OUTBOUND_CLOSE.set(ops2, cancelledSubscription)` |
| 4 | | `ops2.terminate()` |
| 5 | | `rebind(ops1)`: `CAS(ops2, ops1)` succeeds → attr = `ops1` |
| 6 | | `ops2.connection = new DisposedConnection(channel())` |
| 7 | `new HttpServerOperations(ops2)` → `ops3` | |
| 8 | `ops3.channel()` → `DisposedChannel` | |
| 9 | `ops3.bind()` → `DisposedChannel.attr.set(ops3)` — updates only the disposed channel attribute | |
| 10 | real channel attr = `ops1` (unchanged) | |

After `ops3.bind()`:

- `ops3` is bound to `DisposedChannel.attr` — invisible to the real pipeline.
- The real channel's `CONNECTION` attribute points to `ops1` (the zombie).

When `ctx.fireChannelRead(msg)` propagates to `ChannelOperationsHandler`:

```java
Connection connection = Connection.from(ctx.channel()); // → ops1 (real channel attr)
ChannelOperations<?, ?> ops = connection.as(ChannelOperations.class); // → ops1
if (ops != null) {
    ops.onInboundNext(ctx, msg); // delivers request 3 to ops1
}
```

`ops1.onInboundNext()` receives the `HttpRequest`, fires `listener().onStateChange(this, REQUEST_RECEIVED)`, and `HttpServerHandle.onStateChange()` runs:

```java
HttpServerOperations ops = (HttpServerOperations) connection; // ops1
Publisher<Void> publisher = handler.apply(ops, ops);
// ...
mono.subscribe(ops.disposeSubscriber()); // subscribes to ops1
```

`ops1.onSubscribe(s)` calls `Operators.setOnce(OUTBOUND_CLOSE, this, s)`. Since `ops1.OUTBOUND_CLOSE` is already `cancelledSubscription` (set by `ops1.onComplete()` earlier), `setOnce` returns `false`. The subscription is silently dropped: `s.request(Long.MAX_VALUE)` is never called, the handler chain never executes, and no response is ever written for request 3.

**Result:** No exception is logged. The TCP connection remains open. The client waits until its own timeout expires because no response is ever written.

### Case B: disposed parent observed at send time — AbortedException logged

In this case, `ops2.terminate()` has already completed by the time the new `HttpServerOperations` is created with `ops2` as parent. The handler chain runs, but when it attempts to write the response:

```java
// HttpServerOperations.then()
if (!channel().isActive()) {            // DisposedChannel.isActive() = false
    return Mono.error(AbortedException.beforeSend());
}
```

Representative stack trace from our environment:

```text
reactor.netty.channel.AbortedException: Connection has been closed BEFORE send operation
  at reactor.netty.channel.AbortedException.beforeSend(AbortedException.java:59)
  at reactor.netty.http.server.HttpServerOperations.then(HttpServerOperations.java:677)
  at reactor.netty.ReactorNetty$OutboundThen.<init>(ReactorNetty.java:744)
  at reactor.netty.ReactorNetty$OutboundThen.<init>(ReactorNetty.java:733)
  at reactor.netty.NettyOutbound.then(NettyOutbound.java:358)
  at reactor.netty.http.server.HttpServerOperations.send(HttpServerOperations.java:536)
  at org.springframework.http.server.reactive.ReactorServerHttpResponse.writeWithInternal(ReactorServerHttpResponse.java:96)
```

This case is easier to diagnose because it leaves a log trail. However, because the error-handling cleanup path also walks `ops.channel()` — which resolves to `DisposedChannel` — the real connection was not closed promptly in our environment. From the client's perspective, it still appeared as a hang until timeout.

### When the race window opens

The race window exists when:

1. **The response has already been written.** `handleLastHttpContent()` calls `ctx.read()`, so the next request can arrive on the same connection.
2. **The previous request's reactive chain has not completed yet.** For example, the handler returns `res.then().then(Mono.delay(...).then())`, where the response is sent immediately but completion is delayed.
3. **The delayed completion happens on another thread.** For example, `Mono.delay()` completes on `Schedulers.parallel()`, while `channelRead()` for the next request runs on the event loop.

This kind of post-response completion path is common in frameworks such as Spring Cloud Gateway, where filters may continue with logging, metrics, tracing, or coroutine continuation work after the response has already been sent.

### Impact

Both failure modes produce the same client-visible symptom: the affected request does not receive a response, while the underlying TCP connection remains open.

- **Case A** is silent and permanent for the keep-alive connection. The real channel's `CONNECTION` attribute is pinned to a zombie `ops1` whose `OUTBOUND_CLOSE` is `cancelledSubscription`. Every subsequent request on that connection is delivered to `ops1`, whose `onSubscribe` silently fails. No exception is logged, and thread dumps show no blocked threads.
- **Case B** logs `AbortedException("Connection has been closed BEFORE send operation")`, making it easier to diagnose, but from the client's point of view it is still a hang until timeout.
- Which case occurs depends on the exact timing of `ops2.terminate()` relative to the event loop's `channelRead` processing.

## Steps to Reproduce

> Note: this test does not try to force the memory-visibility race itself. In our environment, the actual timing-sensitive hang was only observed reliably under Linux. Instead, the test reproduces the structural precondition that makes the race possible: `HttpTrafficHandler.channelRead()` can observe the previous request's `ChannelOperations` as the result of `Connection.from(ctx.channel())` when creating the next request's ops. The fix removes that structural precondition entirely by no longer performing a per-request `Connection.from(ctx.channel())` lookup.

```java
@Test
@Timeout(30)
void testKeepAliveNewOpsUsesStableParentConnection() throws Exception {
    // ChannelOperations.connection is package-private; reflection is used only
    // to observe the per-request parent connection.
    Field connectionField = ChannelOperations.class.getDeclaredField("connection");
    connectionField.setAccessible(true);

    int requestCount = 5;
    List<Connection> parentConnections = new CopyOnWriteArrayList<>();
    CountDownLatch latch = new CountDownLatch(requestCount);

    disposableServer = createServer()
            .doOnConnection(conn -> {
                try {
                    parentConnections.add((Connection) connectionField.get(conn));
                    latch.countDown();
                }
                catch (IllegalAccessException e) {
                    throw new RuntimeException(e);
                }
            })
            .handle((req, res) -> {
                // Send a complete response immediately (Content-Length: 0),
                // then delay chain completion by 2 seconds. This keeps ops1
                // alive in the channel attribute when the next request arrives
                // via keep-alive, which is the structural precondition for the race.
                res.header(HttpHeaderNames.CONTENT_LENGTH, "0");
                return res.then().then(Mono.delay(Duration.ofSeconds(2)).then());
            })
            .bindNow();

    ConnectionProvider p = ConnectionProvider.create("testToctouRace", 1);
    try {
        Flux.range(0, requestCount)
            .concatMap(i ->
                createClient(p, disposableServer.port())
                    .get()
                    .uri("/")
                    .responseSingle((r, bytes) -> {
                        assertThat(r.status().code()).isEqualTo(200);
                        return bytes.asString().defaultIfEmpty("");
                    }))
            .blockLast(Duration.ofSeconds(15));
    }
    finally {
        p.disposeLater().block(Duration.ofSeconds(5));
    }

    assertThat(latch.await(15, TimeUnit.SECONDS)).as("all requests processed").isTrue();
    assertThat(parentConnections).hasSize(requestCount);

    // All ops must share the same parent connection.
    Connection firstParent = parentConnections.get(0);
    for (int i = 1; i < parentConnections.size(); i++) {
        assertThat(parentConnections.get(i))
                .as("ops[%d] parent connection should be the same stable initial connection", i)
                .isSameAs(firstParent);
    }

    // The parent must NOT be a ChannelOperations (which can be terminated/disposed).
    assertThat(firstParent)
            .as("parent connection should not be a terminable ChannelOperations")
            .isNotInstanceOf(ChannelOperations.class);
}
```

Without the fix, the test fails deterministically on the first keep-alive reuse with:

```text
[ops[1] parent connection should be the same stable initial connection]
Expecting actual:
  GET{uri=/, connection=SimpleConnection{channel=[...]}} // previous HttpServerOperations
and:
  AtomicLong(1) // stable SimpleConnection
to refer to the same object
```

With the fix applied, the test passes locally.

## Possible Solution

Instead of calling `Connection.from(ctx.channel())` for each new request, capture the initial stable connection once when the handler is added, and reuse it for all subsequent `HttpServerOperations` creation:

```java
Connection initialConnection;

@Override
public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
    super.handlerAdded(ctx);
    this.ctx = ctx;
    this.initialConnection = Connection.from(ctx.channel());
}

@Override
public void channelRead(ChannelHandlerContext ctx, Object msg) {
    // ...
    ops = new HttpServerOperations(initialConnection, listener, request, ...);
    ops.bind();
}
```

The same change applies to the pipelined request path in `HttpTrafficHandler.run()`.

This fix is safe because `initialConnection` (the bootstrap-time `SimpleConnection`) is stable for the entire lifecycle of the channel, unlike the request-scoped `ChannelOperations` instances. Reusing that stable connection avoids linking a new request to the lifecycle of an earlier request.

I plan to open a companion PR with this change and the regression test above.

## Affected Versions

Confirmed on `1.2.4`.

Based on local inspection / `git blame`:

- the `channelRead()` path appears to originate from `193539fa3f` (`fix #1001 Improve decoding failure handling`, 2020-02-18)
- the pipelined `run()` path appears to originate from `7ad753fc91` (`Ensure correct access log handling for HTTP/1.1 pipelining`, 2023-01-24)

I have not independently revalidated every historical release, so I would describe the broader impact as likely affecting releases that contain these code paths.

## Environment

- Reactor Netty version: `1.2.4`
- Reactor Core version: `3.6.x` (transitively resolved)
- Netty version: `4.1.x` (transitively resolved)
- Transport: NIO
- JVM: Eclipse Temurin OpenJDK 21 (`21.0.9+10-LTS`)
- `JAVA_TOOL_OPTIONS`: `-XX:ActiveProcessorCount=8`
- The production hang was observed under Linux with:

```text
openjdk version "21.0.9" 2025-10-21 LTS
OpenJDK Runtime Environment Temurin-21.0.9+10 (build 21.0.9+10-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.9+10 (build 21.0.9+10-LTS, mixed mode, sharing)
```

- The same application setup did not reproduce the hang on native macOS in my testing.
- The regression test above is OS-agnostic because it checks the structural precondition, not the race timing.

koisyu:
## Summary

  Fixes #4188

  `HttpTrafficHandler` created new `HttpServerOperations` from `Connection.from(ctx.channel())`. On HTTP/1.1 keep-alive
  connections, that lookup could still return the previous request's `HttpServerOperations` instead of the stable channel-
  bound connection.

  This change captures the stable connection once in `handlerAdded()` and reuses it for subsequent request-scoped
  `HttpServerOperations`, including the pipelined path. That prevents a new request from inheriting a terminating
  `ChannelOperations` as its parent.

  ## Changes

  - capture the initial stable connection in `HttpTrafficHandler.handlerAdded()`
  - use that connection when creating `HttpServerOperations` in `channelRead()`
  - use the same connection in the pipelined request path in `run()`
  - add a regression test covering keep-alive reuse

  ## Testing

  - added `testKeepAliveNewOpsUsesStableParentConnection`
  - verified the test fails without the fix
  - verified the test passes with the fix

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
@@ -798,6 +798,13 @@ class HttpServerOperations extends HttpOperations<HttpServerRequest, HttpServerR
 	}
 
 	@Override
+	protected void afterInboundComplete() {
+		if (!isHttp2()) {
+			channel().pipeline().fireUserEventTriggered(HttpServerOperationsTerminatedEvent.INSTANCE);
+		}
+	}
+
+	@Override
 	protected void onInboundNext(ChannelHandlerContext ctx, Object msg) {
 		Class<?> msgClass = msg.getClass();
 		if (msgClass == DefaultHttpRequest.class) {
--- a/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2025 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2011-2026 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -3926,4 +3926,86 @@ class HttpServerTests extends BaseHttpTest {
 		public void recordResponseTime(SocketAddress remoteAddress, String uri, String method, String status, Duration time) {
 		}
 	}
+
+	@Test
+	@Timeout(30)
+	void testIssue4188() throws Exception {
+		// https://github.com/reactor/reactor-netty/issues/4188
+		//
+		// Verifies the fix for a TOCTOU race condition in HttpTrafficHandler:
+		//
+		// Before the fix, when a keep-alive request arrived while the previous ops was
+		// still the current channel-bound ops, Connection.from(ctx.channel()) could
+		// return that request-scoped HttpServerOperations. If that ops later terminated,
+		// the new request could inherit an unstable parent chain.
+		//
+		// The fix buffers the next HTTP/1.1 keep-alive request while the previous ops
+		// is still channel-bound, and resumes read processing only after the previous
+		// ops reaches afterInboundComplete() inside terminate(), i.e. after the channel
+		// rebind point and before transition to DisposedConnection.
+
+		Field connectionField = ChannelOperations.class.getDeclaredField("connection");
+		connectionField.setAccessible(true);
+
+		int requestCount = 3;
+		List<Connection> parentConnections = new CopyOnWriteArrayList<>();
+		CountDownLatch latch = new CountDownLatch(requestCount);
+
+		disposableServer = createServer()
+				.doOnConnection(conn -> {
+					try {
+						Connection parent = (Connection) connectionField.get(conn);
+						parentConnections.add(parent);
+						latch.countDown();
+					}
+					catch (IllegalAccessException e) {
+						throw new RuntimeException(e);
+					}
+				})
+				.handle((req, res) -> {
+					// Send response immediately (Content-Length: 0), but delay chain
+					// completion. This keeps the current ops alive in the channel
+					// attribute when the next keep-alive request arrives via
+					// setAutoRead(true), which is the precondition for the TOCTOU race.
+					res.header(HttpHeaderNames.CONTENT_LENGTH, "0");
+					return res.then()
+					          .then(Mono.delay(Duration.ofMillis(200)).then());
+				})
+				.bindNow();
+
+		ConnectionProvider p = ConnectionProvider.create("testIssue4188", 1);
+
+		try {
+			Flux.range(0, requestCount)
+			    .concatMap(i ->
+			        createClient(p, disposableServer.port())
+			            .get()
+			            .uri("/")
+			            .responseSingle((r, bytes) -> {
+			                assertThat(r.status().code()).isEqualTo(200);
+			                return bytes.asString().defaultIfEmpty("");
+			            }))
+			    .blockLast(Duration.ofSeconds(20));
+		}
+		finally {
+			p.disposeLater().block(Duration.ofSeconds(5));
+		}
+
+		assertThat(latch.await(20, TimeUnit.SECONDS)).as("all requests processed").isTrue();
+		assertThat(parentConnections).hasSize(requestCount);
+
+		// All ops must share the same parent connection
+		Connection firstParent = parentConnections.get(0);
+		for (int i = 1; i < parentConnections.size(); i++) {
+			assertThat(parentConnections.get(i))
+					.as("ops[%d] parent connection should be the same stable initial connection", i)
+					.isSameAs(firstParent);
+		}
+
+		// The parent must NOT be a ChannelOperations (which can be terminated/disposed)
+		assertThat(firstParent)
+				.as("parent connection should not be a terminable ChannelOperations")
+				.isNotInstanceOf(ChannelOperations.class);
+	}
+
 }
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/Http3ServerOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/Http3ServerOperations.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024-2025 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2024-2026 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -83,6 +83,11 @@ final class Http3ServerOperations extends HttpServerOperations {
 			return H3;
 		}
 		throw new IllegalStateException("request not parsed");
+	}
+
+	@Override
+	protected void afterInboundComplete() {
+		// noop
 	}
 
 	static final HttpVersion H3 = HttpVersion.valueOf("HTTP/3.0");
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperationsTerminatedEvent.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperationsTerminatedEvent.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2026 VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty.http.server;
+
+/**
+ * A user event fired through the channel pipeline during
+ * {@code ChannelOperations#terminate()} via
+ * {@link HttpServerOperations#afterInboundComplete() afterInboundComplete()}.
+ * <p>
+ * This event signals that the previous {@link HttpServerOperations} has passed
+ * the channel rebind point and that keep-alive request handoff may resume.
+ *
+ * @since 1.2.18
+ */
+final class HttpServerOperationsTerminatedEvent {
+
+	static final HttpServerOperationsTerminatedEvent INSTANCE = new HttpServerOperationsTerminatedEvent();
+
+	private HttpServerOperationsTerminatedEvent() {
+	}
+
+	@Override
+	public String toString() {
+		return "HttpServerOperationsTerminatedEvent";
+	}
+}
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpTrafficHandler.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpTrafficHandler.java
@@ -233,6 +233,16 @@ final class HttpTrafficHandler extends ChannelDuplexHandler implements Runnable 
 					return;
 				}
 
+				ChannelOperations<?, ?> existingOps = ChannelOperations.get(ctx.channel());
+				if (existingOps instanceof HttpServerOperations) {
+					if (HttpServerOperations.log.isDebugEnabled()) {
+						HttpServerOperations.log.debug(format(ctx.channel(), "Buffering keep-alive request, previous ops not yet terminated"));
+					}
+					overflow = true;
+					doPipeline(ctx, new HttpRequestHolder(request));
+					return;
+				}
+
 				HttpServerOperations ops;
 				ZonedDateTime timestamp = ZonedDateTime.now(ReactorNetty.ZONE_ID_SYSTEM);
 				ConnectionInfo connectionInfo = null;
@@ -511,6 +521,16 @@ final class HttpTrafficHandler extends ChannelDuplexHandler implements Runnable 
 		ctx.write(response, promise);
 	}
 
+	@Override
+	public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+		if (evt == HttpServerOperationsTerminatedEvent.INSTANCE) {
+			if (persistentConnection && ctx.channel().isActive()) {
+				resumeRead();
+			}
+		}
+		super.userEventTriggered(ctx, evt);
+	}
+
 	@SuppressWarnings("FutureReturnValueIgnored")
 	void handleLastHttpContent(Object msg, ChannelPromise promise) {
 		finalizingResponse = true;
@@ -546,6 +566,9 @@ final class HttpTrafficHandler extends ChannelDuplexHandler implements Runnable 
 			}
 		}
 
+	}
+
+	void resumeRead() {
 		if (pipelined != null && !pipelined.isEmpty()) {
 			if (HttpServerOperations.log.isDebugEnabled()) {
 				HttpServerOperations.log.debug(format(ctx.channel(), "Draining next pipelined " +


violetagg:
@koisyu Thanks for the PR! However I think that the PR doesn't fix the root cause. We need to guarantee that the old ops will be properly terminated before new ops are created.

koisyu:
Thanks for the feedback.

  Based on your review comments, I reworked the fix to address the keep-alive handoff race directly instead of caching `initialConnection`.

  My understanding of the root cause is that there is a TOCTOU window between selecting the parent with
  `Connection.from(ctx.channel())` and binding the next `HttpServerOperations`, during which the previous ops may terminate and transition its internal connection state.

  The revised approach is:

  1. `HttpServerOperations.afterInboundComplete()` fires a `HttpServerOperationsTerminatedEvent` through the pipeline.
     This is reached from inside `ChannelOperations.terminate()` after `rebind(connection)` succeeds and before the ops transitions to `DisposedConnection`, so the real channel is still reachable at that point.

  2. In `HttpTrafficHandler.channelRead()`, if a new HTTP/1.1 request arrives while the previous `HttpServerOperations` is still the current channel-bound ops, the request is buffered via `doPipeline()` instead of immediately creating new ops.

  3. In `HttpTrafficHandler.userEventTriggered()`, receiving `HttpServerOperationsTerminatedEvent` resumes the keep-alive
  handoff by draining the buffered request or issuing `ctx.read()`.

  4. `setAutoRead(true)` in `HttpServerOperations.handleLastHttpContent()` is left in place so the behavior related to #940
  is preserved.

  Key design choices:
  - use a pipeline user event to avoid introducing a direct reverse dependency from `HttpServerOperations` to
  `HttpTrafficHandler`
  - resume keep-alive handoff only after the previous ops has passed the `rebind(connection)` point in `terminate()`
  - keep the `LAST_FLUSH_WHEN_NO_READ` path aligned with its current synchronous termination behavior

  Does this direction look closer to what you had in mind before I update the PR?

koisyu:
@violetagg I pushed an updated version of the branch based on the previous feedback.
Could you please take another look when you have a chance and let me know whether this approach looks reasonable now?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperationsResumeReadEvent.java
```diff
@@ -0,0 +1,39 @@
...
+ * A user event fired through the channel pipeline from
+ * {@link HttpServerOperations#afterInboundComplete()} while
+ * {@code ChannelOperations#terminate()} is still in progress.
+ * <p>
+ * This event signals that the previous {@link HttpServerOperations} has passed
+ * the channel rebind point and that keep-alive request handoff may resume.
+ *
+ * @since 1.2.x
+ */
+final class HttpServerOperationsResumeReadEvent {
```
violetagg: What do you think about this name `HttpServerOperationsTerminatedEvent`?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperationsResumeReadEvent.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A user event fired through the channel pipeline from
+ * {@link HttpServerOperations#afterInboundComplete()} while
+ * {@code ChannelOperations#terminate()} is still in progress.
+ * <p>
+ * This event signals that the previous {@link HttpServerOperations} has passed
+ * the channel rebind point and that keep-alive request handoff may resume.
+ *
+ * @since 1.2.x
```
violetagg:
```suggestion
 * @since 1.2.18
```

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
```diff
@@ -797,6 +797,13 @@ else if (channel().pipeline().get(NettyPipeline.CompressionHandler) == null) {
 		return this;
 	}
 
+	@Override
+	protected void afterInboundComplete() {
```
violetagg, line 801: We have to override this method in reactor.netty.http.server.Http3ServerOperations - we don't need this for HTTP/3

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -3926,4 +3926,85 @@ public void recordDataSentTime(SocketAddress remoteAddress, String uri, String m
...
+					}
+				})
+				.handle((req, res) -> {
+					// Send response immediately (Content-Length: 0), but delay chain
+					// completion. This keeps the current ops alive in the channel
+					// attribute when the next keep-alive request arrives via
+					// setAutoRead(true), which is the precondition for the TOCTOU race.
+					res.header(HttpHeaderNames.CONTENT_LENGTH, "0");
+					return res.then()
+					          .then(Mono.delay(Duration.ofSeconds(2)).then());
```
violetagg: Can we reduce the delay?

violetagg (changes requested):
Thanks for the rework!

violetagg:
Can you add "Signed-off-by:" to the commit so that DCO check can pass?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperationsResumeReadEvent.java
koisyu: Thanks for the review. Fixed.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerOperations.java
koisyu: Thanks for the review. Fixed.

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
koisyu: Thanks for the review. Fixed.

koisyu:
> Can you add "Signed-off-by:" to the commit so that DCO check can pass?

Sure, I've added the sign-off and updated the commit. Thanks.

koisyu:
@violetagg Thank you for the thoughtful review! I've addressed all the comments and updated the PR accordingly. Please let me know if there's anything else I should adjust.

koisyu:
@violetagg   Thanks for checking. The other environments completed successfully, so these failures may be unrelated to this change. Could you please re-run the failed job when you have a chance?

https://github.com/reactor/reactor-netty/actions/runs/24467135689/job/71500153052?pr=4189

image

violetagg (approved):

PR #4212, opened 2026-05-08
Throw `DecoderException` with hostname when `SNI` AsyncMapping resolves to null

### LINKED ISSUE #4210: NullPointerException in SniProvider when AsyncMapping returns null
## Expected Behavior
When an `AsyncMapping` configured via `setSniAsyncMappings` resolves with `null`, `SniProvider.SniHandler.onLookupComplete` should fail with a clear, diagnosable exception that includes the offending hostname.

## Actual Behavior
`onLookupComplete` checks `future.isSuccess()` but does not validate that `future.getNow()` is non-null. A null result causes a bare `NullPointerException` with no hostname information.

`reactor-netty-core/src/main/java/reactor/netty/tcp/SniProvider.java`:

```java
SslProvider sslProvider = future.getNow();    // can be null
SslHandler sslHandler = null;
try {
    sslHandler = sslProvider.getSslContext().newHandler(ctx.alloc()); // NPE
```

Stack trace:

```
java.lang.NullPointerException: Cannot invoke "reactor.netty.tcp.SslProvider.getSslContext()" because "sslProvider" is null
    at reactor.netty.tcp.SniProvider$SniHandler.onLookupComplete(SniProvider.java:131)
    at io.netty.handler.ssl.AbstractSniHandler.onLookupComplete(AbstractSniHandler.java:191)
```

Two things stand out:

- The javadoc of `Builder.setSniAsyncMappings` does not state that the `AsyncMapping` must return a non-null `SslProvider`, so callers have no signal that null-success is invalid usage.
- Error handling in the same method is inconsistent: the `!future.isSuccess()` path wraps the failure in `DecoderException` with the hostname, while the null-success path produces a bare NPE that loses both the wrapping and the hostname.

## Steps to Reproduce
This was discovered downstream in Spring Boot. See https://github.com/spring-projects/spring-boot/issues/50246

Even though the Spring Boot side can avoid the null, I think this should also be addressed in reactor-netty itself, since other consumers using `setSniAsyncMappings` could easily run into the same bare NPE.

Minimal reproducer:

```java
@Test
public void sniHandlerThrowsBareNpeOnNullSuccess() {
    HttpServer.create()
        .secure(spec -> spec
            .sslContext(defaultSslContext)
            .setSniAsyncMappings((serverName, promise) -> promise.setSuccess(null)))
        .bindNow();
    // Any TLS client that sends an SNI extension triggers the bare NPE.
}
```

## Possible Solution
Add a null check that mirrors the existing failure path in the same method:

```java
SslProvider sslProvider = future.getNow();
if (sslProvider == null) {
    throw new DecoderException("failed to get the SslContext for " + hostname
            + ": the SNI mapping returned null");
}
```

This preserves the hostname in the error message and uses the same `DecoderException` type already used by the failure branch.

Optional follow-up: document the non-null contract on `Builder.setSniAsyncMappings` javadoc.

I'm willing to submit a PR if this is accepted.
## Your Environment

* Reactor version(s) used: reactor-netty 1.2.17 (verified `main` branch is also affected)
* Other relevant libraries versions: netty 4.1.x, Spring Boot 3.5.14 (downstream)
* JVM version: Temurin 21
* OS and version: Linux x86_64

kwondh5217:
If the `AsyncMapping` configured via `SslProvider.Builder.setSniAsyncMappings` resolves successfully with a `null` `SslProvider`, `SniProvider.SniHandler.onLookupComplete` produces a bare `NullPointerException` without the offending hostname.

This change throws a `DecoderException` carrying the hostname instead.

Fixes #4210

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/tcp/SniProvider.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/tcp/SniProvider.java
@@ -26,6 +26,7 @@ import io.netty.util.DomainWildcardMappingBuilder;
 import io.netty.util.Mapping;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.FutureListener;
 import io.netty.util.concurrent.Promise;
 import io.netty.util.internal.PlatformDependent;
 import reactor.netty.NettyPipeline;
@@ -112,7 +113,22 @@ final class SniProvider {
 
 		@Override
 		protected Future<SslProvider> lookup(ChannelHandlerContext ctx, String hostname) {
-			return mappings.map(hostname, ctx.executor().newPromise());
+			Promise<SslProvider> promise = ctx.executor().newPromise();
+			mappings.map(hostname, ctx.executor().newPromise())
+			        .addListener((FutureListener<SslProvider>) future -> {
+			            Throwable cause = future.cause();
+			            if (cause != null) {
+			                promise.setFailure(cause);
+			            }
+			            else if (future.getNow() == null) {
+			                promise.setFailure(new DecoderException("failed to get the SslContext for " + hostname
+			                        + ": the SNI mapping returned null"));
+			            }
+			            else {
+			                promise.setSuccess(future.getNow());
+			            }
+			        });
+			return promise;
 		}
 
 		@Override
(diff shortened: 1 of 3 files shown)


violetagg (changes requested):
Thanks for the PR!

With the current fix we still have "USER_EVENT: SniCompletionEvent(SUCCESS='test.com'")" which I think is not correct, the `SniCompletionEvent` should fail as well so that the channel handlers that depend on it are not misleaded.

kwondh5217:
Updated, please take a look !

violetagg:
Try to trigger the build

violetagg (approved):

violetagg:
I back ported the fix to 1.2.x version

PR #4217, opened 2026-05-13
Fix missing metrics for `AdaptiveByteBufAllocator`

### LINKED ISSUE #4209: reactor.netty.bytebuf.allocator.* metrics missing starting with Netty 4.2
Since Netty 4.2 there is a new default allocator: AdaptiveByteBufAllocator
See release notes: https://netty.io/news/2025/04/03/4-2-0.html

That allocator does not produce metrics.

With the update to Spring Boot 4 we lost these metrics for our applications, since that includes the update to Netty 4.2.

We use reactor-netty-core and http in version 1.3.5
And netty in version 4.2.12

## Expected Behavior

We should still get these metrics. As far as I can tell this new allocator still allocates direct memory, and we'd like to see how much memory is being used by reactor-netty.
Unless I'm misunderstanding how the new allocator works, and it's not necessary any longer.

## Actual Behavior

Metrics are lost.

## Steps to Reproduce

Enable reactor netty metrics on a Spring Boot Webflux application, and check the reported metrics on actuator endpoint. Or debug the reactor.netty.transport.TransportConfig.TransportChannelInitializer#initChannel to see that the instanceof checks fail.

Use `-Dio.netty.allocator.type=pooled` then the pooled allocator, which is the previous default, is picked, and the metrics are there again.

## Possible Solution

Add support in reactor.netty.transport.TransportConfig.TransportChannelInitializer#initChannel for io.netty.buffer.AdaptiveByteBufAllocator

## Your Environment

* Reactor version(s) used: 1.3.5
* Other relevant libraries versions (eg. `netty`, ...): netty in version 4.2.12
* JVM version (`java -version`): Java 25
* OS and version (eg. `uname -a`): Locally for Ubuntu and in distroless container

lukas-riedler-dynatrace:
Newer Netty versions use the AdaptiveByteBufAllocator by default. Add test for the new allocator, and the existing support for the unpooled one.

Fixes #4209

### DIFF
--- a/reactor-netty-http/src/test/java/reactor/netty/transport/ByteBufAllocatorMetricsTest.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/transport/ByteBufAllocatorMetricsTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2024 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2019-2026 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


reactor-netty-http/src/test/java/reactor/netty/transport/ByteBufAllocatorMetricsTest.java
```diff
@@ -75,6 +78,43 @@ void tearDown() {
 
 	@Test
 	void test() throws Exception {
```
violetagg:
```suggestion
	void testPooled() throws Exception {
```

violetagg (changes requested):
Thanks for the PR!

violetagg (approved):

PR #4244, opened 2026-06-05
Add SSL support for HTTP proxy connections

### LINKED ISSUE #2348: Support for Proxy-over-SSL
## Motivation

I have a network which is reachable by a proxy only. The proxy itself is an apache2 configured as proxy. The proxy is exposed via https. The ressources within the network are reachable by http.

This setting works easily with curl:

```
curl --proxy https://proxy:secret@my-rpoxy-server:443 http://10.200.6.245/
```

* Proxy-Connection is established with SSL
* connection from proxy to backend is established without SSL

## Desired solution

This should also be possible with netty. I have seen some potentially related issues which are closed without obvious reason (https://github.com/reactor/reactor-netty/pull/2178, https://github.com/reactor/reactor-netty/issues/2165).

codingkiddo:
### Motivation

This adds support for connecting to an HTTP proxy over SSL/TLS.

The target scenario is:

```text
HttpClient --TLS--> HTTPS proxy --plain HTTP--> backend
```

This is related to #2348, where the proxy endpoint itself is HTTPS while the backend resource can still be plain HTTP.

### Changes

* Adds proxy-level SSL configuration to `ProxyProvider`
* Adds dedicated proxy SSL pipeline handlers
* Ensures proxy SSL is inserted before the proxy handler
* Adds unit coverage for proxy SSL pipeline ordering

### Testing

Ran:

```bash
./gradlew :reactor-netty-core:compileJava
./gradlew :reactor-netty-core:test --tests "reactor.netty.transport.ProxyProviderTest"
```

### TODO

* Add end-to-end HTTP client test for HTTPS proxy -> HTTP backend

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/NettyPipeline.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/NettyPipeline.java
@@ -114,6 +114,9 @@ public interface NettyPipeline {
 	String OnChannelReadIdle     = LEFT + "onChannelReadIdle";
 	String OnChannelWriteIdle    = LEFT + "onChannelWriteIdle";
 	String ProxyHandler          = LEFT + "proxyHandler";
+	String ProxySslHandler       = LEFT + "proxySslHandler";
+	String ProxySslLoggingHandler = LEFT + "proxySslLoggingHandler";
+	String ProxySslReader        = LEFT + "proxySslReader";
 	/**
 	 * Use to register a special handler which ensures that any {@link io.netty.channel.VoidChannelPromise}
 	 * will be converted to "unvoided" promises.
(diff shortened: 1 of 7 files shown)


violetagg:
Thanks for the PR!
I have a few initial comments before I dive into a more in-depth review.

- The change reuses `reactor.netty.tcp.SslProvider.SslReadHandler` however I'm not sure that the proxy SslReadHandler has to forward the events. Wdyt?

- Several places in the code use a construction like this `pipeline().get(SslHandler.class)`, till now this was returning the application SslHandler, while now it is returning the proxy SslHandler?

- It will be good to have E2E test

codingkiddo:
> Thanks for the PR! I have a few initial comments before I dive into a more in-depth review.
> 
> * The change reuses `reactor.netty.tcp.SslProvider.SslReadHandler` however I'm not sure that the proxy SslReadHandler has to forward the events. Wdyt?
> * Several places in the code use a construction like this `pipeline().get(SslHandler.class)`, till now this was returning the application SslHandler, while now it is returning the proxy SslHandler?
> * It will be good to have E2E test

Thanks for the review. I updated the proxy SSL path to use a proxy-specific read handler instead of reusing the application SslReadHandler, so the proxy TLS handshake is not forwarded as an application SSL handshake.

I also updated the client/metrics paths that expected the application SSL handler to look up the named NettyPipeline.SslHandler instead of using get(SslHandler.class), so they do not accidentally pick up NettyPipeline.ProxySslHandler.

I’ll work on the E2E test next.

PR #4255, opened 2026-06-12
Exclude `HTTP/2` connections from the pool during the `PING` liveness check

### LINKED ISSUE #4245: Pooled HTTP/2 connections not evicted from pool on IO errors
In the event of half-open TCP connections in the pool where the peer has crashed without orderly terminating the connection, subsequent attempts to utilize the pooled connection will of course result in an error (ie. ReadTimeoutException). With HTTP 1.1, those will be evicted from the pool and on retry a new connection is acquired/established. With HTTP/2, the IO error is observed on the stream channel leaving the physical TCP parent connection untouched in the pool. Retries will continue to use the broken connection inadvertently running into the same error by default.

## Expected Behavior
The connection pool should expose the same behaviour for HTTP/2 as for HTTP/1.1 and stop reusing the connection in case of IO errors. 

## Actual Behavior
HTTP/2 streams are closed but TCP connection is kept and utilized for subsequent requests.

## Steps to Reproduce
Please find minimal repro case here: 
https://github.com/chringwer/http2-pool-eviction/blob/master/src/test/java/com/example/http2_pool_eviction/Http2PoolEvictionTest.java

## Possible Solution
My current workaround is:

```
    @ChannelHandler.Sharable
    private static final class TimeoutInvalidationHandler extends ChannelInboundHandlerAdapter {
      static final TimeoutInvalidationHandler INSTANCE = new TimeoutInvalidationHandler();

      @Override
      public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
        if (cause instanceof ChannelException) {
          Channel parent = ctx.channel().parent();

          if (parent != null) {
            parent.attr(CONNECTION_IO_ERROR_OBSERVED).set(TRUE);
          }
        }

        super.exceptionCaught(ctx, cause);
      }
    }
  }
```

with

```
.doOnRequest(
    (request, channel) ->
        channel.addHandlerLast(
            "timeout-invalidator", TimeoutInvalidationHandler.INSTANCE))
```

and 

```
.evictionPredicate( (conn, meta) ->
    TRUE.equals(conn.channel().attr(CONNECTION_IO_ERROR_OBSERVED).get()))
```

LivingLikeKrillin:
## Summary

Fixes #4245

When an HTTP/2 connection goes half-open (the peer silently stops responding), a request on it
times out, and the pool keeps handing out that same dead connection — so the immediate retry
fails too.

As discussed, **a timeout is the wrong eviction trigger**: it means a request was slow, not that
the connection is broken. The right signal is the existing HTTP/2 **PING liveness** probe, which
already fires on idle (`pingAckTimeout` + `maxIdleTime`) and distinguishes a slow-but-alive
connection from a dead one. The missing piece was that while the probe is in flight, the pool
still hands the connection out.

## Fix

While a PING liveness probe is in flight, **exclude the connection from acquisition**; re-admit it
on ACK, or let it be evicted on ack-timeout:

- `Http2ConnectionLiveness` flips a `volatile boolean connectionLivenessCheckInProgress` on the
  `Http2Pool.Slot` when the probe starts (`writePing`) and clears it when the probe ends (`cancel`,
  which fires on ACK, inbound HEADERS/DATA, channel-inactive, or handler-removed).
- `findConnection` treats that flag as a guard: a slot under probe is re-offered (`offerSlot` +
  `continue`), not handed out — alongside the existing `isActive` / `goAwayReceived` /
  `evictionPredicate` / `canOpenStream` guards.
- On ack-timeout the channel is closed and removed lazily by the existing `isActive` guard, so a
  never-cleared flag cannot strand a slot.

This replaces the earlier `streamIoErrorObserved` (timeout-reactive) mechanism, which is removed.

**The cross-thread seam is kept internal:** the per-slot flag is written via an overloaded
`Http2ConnectionLiveness` constructor (`@since 1.3.7`); the public `HttpConnectionLiveness`
interface is unchanged. Happy to redirect the wiring in review if you would prefer a different shape.

## Test plan

- [x] `Http2PoolTest.connectionLivenessCheckInProgressExcludesConnectionFromAcquisition`: with the
      flag set, acquire allocates a fresh connection (the probed one is not reused); cleared, reuse
      resumes with no new allocation.
- [x] `Http2PoolPingLivenessEvictionTest.brokenConnectionIsEvictedByPingLivenessProbe`: reuses the
      half-open `TcpProxy` harness; with `pingAckTimeout` + `maxIdleTime`, a broken pooled connection
      is evicted by the probe (latched on the parent `closeFuture`). Parameterized over the H2 / H2C /
      HTTP11+H2C-upgrade wiring sites.
- [x] `:reactor-netty-http:test` (the two above), `:reactor-netty-core:japicmp` /
      `:reactor-netty-http:japicmp` (one additive `Http2ConnectionLiveness` ctor, `@since 1.3.7`,
      zero breakages) — green locally.

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
@@ -24,6 +24,7 @@ import io.netty.handler.codec.http2.Http2FrameCodec;
 import io.netty.handler.codec.http2.Http2FrameWriter;
 import io.netty.handler.codec.http2.Http2HeadersFrame;
 import io.netty.handler.codec.http2.Http2PingFrame;
+import io.netty.util.AttributeKey;
 import org.jspecify.annotations.Nullable;
 
 import java.util.concurrent.ScheduledFuture;
@@ -68,15 +69,28 @@ import static reactor.netty.ReactorNetty.format;
  */
 public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
 
+	/**
+	 * Channel attribute, set by the connection pool on a pooled HTTP/2 connection, that lets the pool
+	 * observe the PING liveness probe: the listener is notified with {@code true} when a probe starts
+	 * and with {@code false} when it ends. This is the pool's signal to exclude the connection from
+	 * acquisition while it is being probed and to re-admit it once the probe completes. It is internal
+	 * wiring between the pool and this handler; the probe behaviour itself is unconditional and not
+	 * configurable.
+	 *
+	 * @since 1.3.7
+	 */
+	public static final AttributeKey<@Nullable Consumer<Boolean>> CONNECTION_LIVENESS_CHECK_LISTENER =
+			AttributeKey.valueOf(Http2ConnectionLiveness.class, "CONNECTION_LIVENESS_CHECK_LISTENER");
+
 	private final Http2FrameWriter http2FrameWriter;
 	private final int pingAckDropThreshold;
 	private final long pingAckTimeoutNanos;
-	private final @Nullable Consumer<Boolean> connectionLivenessCheckListener;
 
 	private boolean isPingAckPending;
 	private long lastSentPingData;
 	private @Nullable ScheduledFuture<?> pingScheduler;
 	private int pingAttempts;
+	private @Nullable Channel channel;
 
 	/**
 	 * Constructs a new {@code Http2ConnectionLiveness} instance.
@@ -86,25 +100,9 @@ public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
 	 * @param pingAckTimeoutNanos the timeout in nanoseconds for receiving a PING ACK response
 	 */
 	public Http2ConnectionLiveness(Http2FrameCodec http2FrameCodec, int pingAckDropThreshold, long pingAckTimeoutNanos) {
-		this(http2FrameCodec, pingAckDropThreshold, pingAckTimeoutNanos, null);
-	}
-
-	/**
-	 * Constructs a new {@code Http2ConnectionLiveness} instance.
-	 *
-	 * @param http2FrameCodec the HTTP/2 frame codec
-	 * @param pingAckDropThreshold the maximum number of PING frame transmission attempts before closing the connection
-	 * @param pingAckTimeoutNanos the timeout in nanoseconds for receiving a PING ACK response
-	 * @param connectionLivenessCheckListener notified with {@code true} when a liveness check (PING probe) starts and
-	 * with {@code false} when it ends; lets the connection pool exclude the connection from acquisition while probing
-	 * @since 1.3.7
-	 */
-	public Http2ConnectionLiveness(Http2FrameCodec http2FrameCodec, int pingAckDropThreshold, long pingAckTimeoutNanos,
-			@Nullable Consumer<Boolean> connectionLivenessCheckListener) {
 		this.http2FrameWriter = http2FrameCodec.encoder().frameWriter();
 		this.pingAckDropThreshold = pingAckDropThreshold;
 		this.pingAckTimeoutNanos = pingAckTimeoutNanos;
-		this.connectionLivenessCheckListener = connectionLivenessCheckListener;
 	}
 
 	/**
@@ -122,8 +120,11 @@ public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
 	}
 
 	private void notifyConnectionLivenessCheck(boolean inProgress) {
-		if (connectionLivenessCheckListener != null) {
-			connectionLivenessCheckListener.accept(inProgress);
+		if (channel != null) {
+			Consumer<Boolean> listener = channel.attr(CONNECTION_LIVENESS_CHECK_LISTENER).get();
+			if (listener != null) {
+				listener.accept(inProgress);
+			}
 		}
 	}
 
@@ -239,6 +240,7 @@ public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
 		private void writePing() {
 			isPingAckPending = true;
 			pingAttempts++;
+			channel = ctx.channel();
 			notifyConnectionLivenessCheck(true);
 
 			lastSentPingData = ThreadLocalRandom.current().nextLong();
--- a/reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2025-2026 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -28,6 +28,7 @@ import org.jspecify.annotations.Nullable;
 
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Consumer;
 
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static reactor.netty.ReactorNetty.format;
@@ -70,6 +71,7 @@ public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
 	private final Http2FrameWriter http2FrameWriter;
 	private final int pingAckDropThreshold;
 	private final long pingAckTimeoutNanos;
+	private final @Nullable Consumer<Boolean> connectionLivenessCheckListener;
 
 	private boolean isPingAckPending;
 	private long lastSentPingData;
@@ -84,9 +86,25 @@ public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
 	 * @param pingAckTimeoutNanos the timeout in nanoseconds for receiving a PING ACK response
 	 */
 	public Http2ConnectionLiveness(Http2FrameCodec http2FrameCodec, int pingAckDropThreshold, long pingAckTimeoutNanos) {
+		this(http2FrameCodec, pingAckDropThreshold, pingAckTimeoutNanos, null);
+	}
+
+	/**
+	 * Constructs a new {@code Http2ConnectionLiveness} instance.
+	 *
+	 * @param http2FrameCodec the HTTP/2 frame codec
+	 * @param pingAckDropThreshold the maximum number of PING frame transmission attempts before closing the connection
+	 * @param pingAckTimeoutNanos the timeout in nanoseconds for receiving a PING ACK response
+	 * @param connectionLivenessCheckListener notified with {@code true} when a liveness check (PING probe) starts and
+	 * with {@code false} when it ends; lets the connection pool exclude the connection from acquisition while probing
+	 * @since 1.3.7
+	 */
+	public Http2ConnectionLiveness(Http2FrameCodec http2FrameCodec, int pingAckDropThreshold, long pingAckTimeoutNanos,
+			@Nullable Consumer<Boolean> connectionLivenessCheckListener) {
 		this.http2FrameWriter = http2FrameCodec.encoder().frameWriter();
 		this.pingAckDropThreshold = pingAckDropThreshold;
 		this.pingAckTimeoutNanos = pingAckTimeoutNanos;
+		this.connectionLivenessCheckListener = connectionLivenessCheckListener;
 	}
 
 	/**
@@ -96,9 +114,16 @@ public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
 	public void cancel() {
 		isPingAckPending = false;
 		pingAttempts = 0;
+		notifyConnectionLivenessCheck(false);
 		if (pingScheduler != null) {
 			pingScheduler.cancel(false);
 			pingScheduler = null;
+		}
+	}
+
+	private void notifyConnectionLivenessCheck(boolean inProgress) {
+		if (connectionLivenessCheckListener != null) {
+			connectionLivenessCheckListener.accept(inProgress);
 		}
 	}
 
@@ -214,6 +239,7 @@ public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
 		private void writePing() {
 			isPingAckPending = true;
 			pingAttempts++;
+			notifyConnectionLivenessCheck(true);
 
 			lastSentPingData = ThreadLocalRandom.current().nextLong();
 
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
@@ -16,6 +16,7 @@
 package reactor.netty.http.client;
 
 import io.netty.channel.Channel;
+import io.netty.channel.ChannelException;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.http2.Http2FrameCodec;
 import io.netty.handler.codec.http2.Http2StreamChannel;
@@ -468,6 +469,7 @@ final class Http2ConnectionProvider extends PooledConnectionProvider<Connection>
 				Http2StreamChannel ch = future.getNow();
 
 				if (!channel.isActive() || frameCodec == null ||
+						http2PooledRef.slot.streamIoErrorObserved ||
 						((Http2FrameCodec) frameCodec.handler()).connection().goAwayReceived() ||
 						!((Http2FrameCodec) frameCodec.handler()).connection().local().canOpenStream()) {
 					invalidate(this);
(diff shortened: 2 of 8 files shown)


LivingLikeKrillin:
The CI failures on the first run were caused by an interaction between this change and two existing tests, plus a few unrelated items — summary:

**Adjusted in 9e267e8.** `testServerConnectionsMicrometerConnectionClose` / `testServerConnectionsRecorderConnectionClose` dispatch a request that hits the response timeout — which is now exactly the stream IO error this PR reacts to, so the pooled HTTP/2 connection is evicted and closed mid-test. Their HTTP/2 branches asserted the connection survives the timeout and raced with the asynchronous close (that's why re-runs flip between green and red). The branches now await the close and assert the post-eviction state.

While adjusting them I noticed one thing worth flagging: when the server is configured with multiple protocols, the connection close is never reflected in `reactor.netty.http.server.connections.total` — the handler that recorded the opened connection is replaced when the pipeline is reconfigured for HTTP/2, and the stream-level handlers skip connection events. The test asserts the current behavior with a comment; happy to open a separate issue if you agree it's a gap.

**Also in 9e267e8.** The two CodeQL findings (deprecated `SslContextSpec#sslContext`) are replaced with the `GenericSslContextSpec` variant.

**Not related to this PR (no change).** `Http2ConnectionLivenessTest.clientCloseConnectionIfPingFrameDelayed` (exact ping-count assertion vs 100ms timer on a loaded runner; the request in that test succeeds, so the new code path is never reached), `HttpClientTest.testSelectedIps` / `staleDisposeAfterConnectionReuseDoesNotCloseActiveChannel` and `PooledConnectionProviderDefaultMetricsTest.testIssue3060PendingAcquireTimeoutReached` (HTTP/1.1-only or DNS, macOS runner slowness), and the 33s `ubuntu-22.04, native` failure (Gradle distribution download timeout).

violetagg:
Thanks for the PR!
My main concern is using the stream/response timeout as the eviction trigger. It can be set globally or per request, but either way it just means one request took too long — which isn't really a signal that the connection is broken. It could easily be intermittent in which case the connection is fine and the next request would succeed. Evicting it on a single timeout means we throw away a healthy connection and reconnect for no reason.
I think the right direction is the HTTP/2 PING feature, since that actually probes the connection end-to-end. The missing piece is that today HTTP/2 PING doesn't take the connection out of the pool while it's being checked — the pool keeps handing it out until the ping eventually closes the channel. So we'd want to: temporarily exclude the connection from the available ones, ping it, and then either put it back if it answers or evict it if it doesn't.
That way intermittent timeouts don't kill healthy connections, and genuinely dead ones still get cleaned up.
WDYT?

LivingLikeKrillin:
Right — a single timeout means one request was slow, not that the connection's broken, so I'm with you that it isn't the signal to evict on. And the pool-exclusion gap is real: the liveness check only closes the channel, it doesn't take the slot out while it's probing.

The one thing I wasn't sure about: PING liveness today only fires on an idle connection (after `maxIdleTime` of quiet). #4245's case is an active request timing out on an otherwise-silent half-open connection, where the retry needs a fresh connection right away — well before an idle probe would run. So I'm not sure idle-PING alone reaches that case. How were you picturing it covering the in-flight timeout — would the timeout itself kick off the probe, or did you have a different trigger in mind?

(Scope note for the issue: PING is opt-in via `pingAckTimeout`, so whatever we do only covers pools that enabled it — the default-config case may need tracking separately on #4245.)

violetagg:
I'd actually like to drop the `StreamIoErrorObserver` approach.
My main issue is that it only kicks in when a `response timeout` is set, and it ties a connection-level decision to a per-request timeout. I don't think we want to trigger PING off individual request timeouts. Liveness should be a connection-level thing, and we already have that: the idle handler on the parent connection fires the PING when the connection goes quiet.
I also don't think the "idle probe won't run during an active request" concern holds - a connection waiting on a response that never comes is idle (request was sent, nothing coming back), so the idle timer fires and PING goes out anyway. And that sorts the two cases out for free: a slow-but-alive peer ACKs the ping and we keep it, a dead one doesn't and we drop it. If one stream times out while others on the same connection are still flowing, we don't probe or evict - which is right, the connection's clearly fine.
So the only real gap is: PING closes the channel but doesn't pull the connection out of the pool while it's probing. If we add that - exclude it during the ping, then put it back if ACKed or evict if not.
WDYT?

LivingLikeKrillin:
You're right — I'd conflated an app-level in-flight request with I/O idle. A connection that's sent its request and is getting nothing back has no reads or writes, so the parent idle timer fires and the PING goes out on its own; the ACK then splits slow-but-alive from dead, and leaving a connection alone while its other streams are still flowing is the right call. So — no per-request-timeout trigger; liveness stays connection-level on the existing idle PING.

That leaves the gap you pointed to: the probe doesn't pull the connection out of the pool. I'll have the liveness check mark the slot "under probe" when it sends the PING (the slot's reachable from where the check is wired up), keep it out of acquisition while pending, and re-admit on ACK or evict on timeout.

I'll drop the `StreamIoErrorObserver` approach and rework this PR along those lines for your review.

LivingLikeKrillin:
Reworked as we agreed — dropped the timeout-reactive `streamIoErrorObserved` approach and built the fix on the existing idle-PING liveness instead:

- While a PING liveness probe is in flight, the connection is excluded from acquisition (a `volatile` flag on the `Http2Pool.Slot`, set in `writePing` / cleared in `cancel`), re-admitted on ACK and evicted on ack-timeout.
- `findConnection` re-offers a slot under probe rather than handing it out, next to the existing guards.

I kept the cross-thread seam internal — the flag is written via an overloaded `Http2ConnectionLiveness` constructor (`@since 1.3.7`), so the public `HttpConnectionLiveness` interface is untouched and you can redirect the wiring in review if you would prefer. (Pushed as `c66d005fa`; PR title/description updated off the old stream-IO framing.)

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
```diff
@@ -84,9 +86,25 @@ public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
...
+	 *
+	 * @param http2FrameCodec the HTTP/2 frame codec
+	 * @param pingAckDropThreshold the maximum number of PING frame transmission attempts before closing the connection
+	 * @param pingAckTimeoutNanos the timeout in nanoseconds for receiving a PING ACK response
+	 * @param connectionLivenessCheckListener notified with {@code true} when a liveness check (PING probe) starts and
+	 * with {@code false} when it ends; lets the connection pool exclude the connection from acquisition while probing
+	 * @since 1.3.7
+	 */
+	public Http2ConnectionLiveness(Http2FrameCodec http2FrameCodec, int pingAckDropThreshold, long pingAckTimeoutNanos,
+			@Nullable Consumer<Boolean> connectionLivenessCheckListener) {
```
violetagg: Do we need this extra constructor. `connectionLivenessCheckListener` should be something internal that we don't need from the user?

violetagg (changes requested):

LivingLikeKrillin:
Making the listener internal comes down to a trade-off between two options:

**A — keep the flag on the Slot (current approach).** The probe-in-progress flag stays a field on the pool's `Slot`, read directly on the acquire path — the faster of the two. But since the handler (`reactor.netty.http`) and the `Slot` (`reactor.netty.http.client`) are in different packages, wiring it this way needs a public seam on the handler — currently the constructor parameter — which is the exposure you flagged.

**B — channel attribute.** The handler sets a channel attribute when the probe starts and clears it when it ends, and the pool reads it in `findConnection`. This drops the extra constructor and the `Consumer` entirely. The costs: the end path (`cancel()`) has no `ChannelHandlerContext` today, so the handler would hold a `Channel` reference to clear the attr; the pool does an attribute lookup on the acquire path instead of the direct field read — a bit slower, though the `Slot` can cache the `Attribute` handle to bring it close to a plain field read; and it needs a shared `AttributeKey` reachable from both packages (a public key, or a `valueOf(String)` contract).

So it's really: keep the faster direct field read but with a public constructor (A) vs. drop the public constructor at the cost of a slightly slower attribute lookup (B).

If the non-public constructor is the priority for you, I'll go with B. Does that work?

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
violetagg, line 28:
```suggestion
import io.netty.util.AttributeKey;
import org.jspecify.annotations.Nullable;
```

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
```diff
@@ -70,6 +71,7 @@ public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
 	private final Http2FrameWriter http2FrameWriter;
```
violetagg, line 85:
```suggestion
	/**
	 * Channel attribute, set by the connection pool on a pooled HTTP/2 connection, that lets the pool
	 * observe the PING liveness probe: the listener is notified with {@code true} when a probe starts
	 * and with {@code false} when it ends. This is the pool's signal to exclude the connection from
	 * acquisition while it is being probed and to re-admit it once the probe completes. It is internal
	 * wiring between the pool and this handler; the probe behaviour itself is unconditional and not
	 * configurable.
	 *
	 * @since 1.3.7
	 */
	public static final AttributeKey<Consumer<Boolean>> CONNECTION_LIVENESS_CHECK_LISTENER =
			AttributeKey.valueOf(Http2ConnectionLiveness.class, "CONNECTION_LIVENESS_CHECK_LISTENER");

	private final Http2FrameWriter http2FrameWriter;
```

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
```diff
@@ -70,6 +71,7 @@ public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
...
 	private final int pingAckDropThreshold;
 	private final long pingAckTimeoutNanos;
+	private final @Nullable Consumer<Boolean> connectionLivenessCheckListener;
```
violetagg:
```suggestion
```

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
violetagg:
```suggestion
```

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
```diff
@@ -84,9 +86,25 @@ public final class Http2ConnectionLiveness implements HttpConnectionLiveness {
...
+	 * @param connectionLivenessCheckListener notified with {@code true} when a liveness check (PING probe) starts and
+	 * with {@code false} when it ends; lets the connection pool exclude the connection from acquisition while probing
+	 * @since 1.3.7
+	 */
+	public Http2ConnectionLiveness(Http2FrameCodec http2FrameCodec, int pingAckDropThreshold, long pingAckTimeoutNanos,
+			@Nullable Consumer<Boolean> connectionLivenessCheckListener) {
 		this.http2FrameWriter = http2FrameCodec.encoder().frameWriter();
 		this.pingAckDropThreshold = pingAckDropThreshold;
 		this.pingAckTimeoutNanos = pingAckTimeoutNanos;
+		this.connectionLivenessCheckListener = connectionLivenessCheckListener;
```
violetagg:
```suggestion
```

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
violetagg, line 92:
```suggestion
	private int pingAttempts;
	private @Nullable Channel channel;
```

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
```diff
@@ -96,12 +114,19 @@ public Http2ConnectionLiveness(Http2FrameCodec http2FrameCodec, int pingAckDropT
...
+		notifyConnectionLivenessCheck(false);
 		if (pingScheduler != null) {
 			pingScheduler.cancel(false);
 			pingScheduler = null;
 		}
 	}
 
+	private void notifyConnectionLivenessCheck(boolean inProgress) {
+		if (connectionLivenessCheckListener != null) {
+			connectionLivenessCheckListener.accept(inProgress);
```
violetagg:
```suggestion
		if (channel != null) {
			Consumer<Boolean> listener = channel.attr(CONNECTION_LIVENESS_CHECK_LISTENER).get();
			if (listener != null) {
				listener.accept(inProgress);
			}
```

reactor-netty-http/src/main/java/reactor/netty/http/Http2ConnectionLiveness.java
```diff
@@ -214,6 +239,7 @@ private boolean isExceedAckDropThreshold() {
 		private void writePing() {
 			isPingAckPending = true;
 			pingAttempts++;
+			notifyConnectionLivenessCheck(true);
```
violetagg, line 244:
```suggestion
			channel = ctx.channel();
			notifyConnectionLivenessCheck(true);
```

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
```diff
@@ -392,13 +392,15 @@ else if (p.state != null) {
...
+				Http2Pool.Slot slot = http2PooledRef(pooledRef).slot;
+				ChannelHandlerContext frameCodec = slot.http2FrameCodecCtx();
 				if (frameCodec != null) {
 					IdleTimeoutHandler.addIdleTimeoutHandler(channel.pipeline(), Duration.ofMillis(http2Pool.maxIdleTime),
 							new Http2ConnectionLiveness(((Http2FrameCodec) frameCodec.handler()),
 									// Deliberately suppress "NullAway"
 									// http2SettingsSpec != null && http2SettingsSpec.pingAckTimeout() != null in this case
-									http2SettingsSpec.pingAckDropThreshold(), http2SettingsSpec.pingAckTimeout().toNanos()));
+									http2SettingsSpec.pingAckDropThreshold(), http2SettingsSpec.pingAckTimeout().toNanos(),
+									inProgress -> slot.connectionLivenessCheckInProgress = inProgress));
```
violetagg: revert this

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
```diff
@@ -417,13 +419,15 @@ public void onStateChange(Connection connection, State newState) {
...
+					Http2Pool.Slot slot = http2PooledRef(pooledRef).slot;
+					ChannelHandlerContext frameCodec = slot.http2FrameCodecCtx();
 					if (frameCodec != null) {
 						IdleTimeoutHandler.addIdleTimeoutHandler(connection.channel().pipeline(), Duration.ofMillis(http2Pool.maxIdleTime),
 								new Http2ConnectionLiveness(((Http2FrameCodec) frameCodec.handler()),
 										// Deliberately suppress "NullAway"
 										// http2SettingsSpec != null && http2SettingsSpec.pingAckTimeout() != null in this case
-										http2SettingsSpec.pingAckDropThreshold(), http2SettingsSpec.pingAckTimeout().toNanos()));
+										http2SettingsSpec.pingAckDropThreshold(), http2SettingsSpec.pingAckTimeout().toNanos(),
+										inProgress -> slot.connectionLivenessCheckInProgress = inProgress));
```
violetagg: revert this

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
```diff
@@ -392,13 +392,15 @@ else if (p.state != null) {
 
 			Http2Pool http2Pool = http2Pool();
 			if (http2Pool.evictionPredicate != null) {
-				ChannelHandlerContext frameCodec = http2PooledRef(pooledRef).slot.http2FrameCodecCtx();
+				Http2Pool.Slot slot = http2PooledRef(pooledRef).slot;
+				ChannelHandlerContext frameCodec = slot.http2FrameCodecCtx();
```
violetagg, line 396:
```java
						installConnectionLivenessCheckListener(connection.channel(), slot, http2Pool);
```

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
```diff
@@ -417,13 +419,15 @@ public void onStateChange(Connection connection, State newState) {
 			else if (newState == UPGRADE_SUCCESSFUL) {
 				Http2Pool http2Pool = http2Pool();
 				if (http2Pool.evictionPredicate != null) {
-					ChannelHandlerContext frameCodec = http2PooledRef(pooledRef).slot.http2FrameCodecCtx();
+					Http2Pool.Slot slot = http2PooledRef(pooledRef).slot;
+					ChannelHandlerContext frameCodec = slot.http2FrameCodecCtx();
```
violetagg, line 423:
```java
						installConnectionLivenessCheckListener(connection.channel(), slot, http2Pool);
```

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2ConnectionProvider.java
violetagg, line 437:
```suggestion

		static void installConnectionLivenessCheckListener(Channel channel, Http2Pool.Slot slot, Http2Pool http2Pool) {
			// While a PING liveness probe is in flight, the connection is excluded from acquisition;
			// re-admit it and wake any pending acquire once the probe completes.
			channel.attr(Http2ConnectionLiveness.CONNECTION_LIVENESS_CHECK_LISTENER)
			       .set(inProgress -> {
			           slot.connectionLivenessCheckInProgress = inProgress;
			           if (!inProgress) {
			               http2Pool.drain();
			           }
			       });
		}

```

violetagg (changes requested):
I was thinking about something like this. What do you think?

LivingLikeKrillin:
Yes — this is cleaner than the polled-attribute version I sketched. Storing the `Consumer` in the attribute keeps the acquire path a direct field read on the `Slot` and confines the attribute lookup to probe start/end, so there's no hot-path cost — it gets B's "no public constructor" without the lookup I'd flagged.

I'll apply the suggestions as you've laid them out.

violetagg (approved):

PR #4283, opened 2026-07-07
Add RFC 10008 (QUERY Method) support

desiderantes:
This change adds support for [RFC 10008](https://www.rfc-editor.org/rfc/rfc10008.html) (The HTTP QUERY Method).

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
@@ -496,6 +496,7 @@ class HttpClientConnect extends HttpClient {
 		volatile UriEndpoint                    toURI;
 		volatile @Nullable UriEndpoint          fromURI;
 		volatile Supplier<String> @Nullable []  redirectedFrom;
+		volatile boolean                        redirectedAsGet;
 		volatile boolean                        shouldRetry;
 		volatile @Nullable HttpHeaders          previousRequestHeaders;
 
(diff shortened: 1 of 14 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/server/HttpPredicate.java
```diff
@@ -194,6 +194,24 @@ public static Predicate<HttpServerRequest> put(String uri) {
...
+	 * Creates a {@link Predicate} based on a URI template filtering.
+	 * <p>
+	 * This will listen for QUERY Method.
+	 *
+	 * @param uri The string to compile into a URI template and use for matching
+	 *
+	 * @return The new {@link Predicate}.
+	 *
+	 * @see Predicate
+	 * @since 1.3.7
```
violetagg:
```suggestion
	 * @since 1.3.7
	 * @see Predicate
```

violetagg:

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
```diff
@@ -496,6 +496,7 @@ static final class HttpClientHandler extends SocketAddress
 		volatile UriEndpoint                    toURI;
 		volatile @Nullable UriEndpoint          fromURI;
 		volatile Supplier<String> @Nullable []  redirectedFrom;
+		volatile boolean                        redirectedAsGet;
```
violetagg: Can you please explain more about this change - `redirectedAsGet`. It is backward incompatible, also I do not see anything in RFC requiring this. In our documentation it is clearly written `the body Publisher passed in will be invoked also for redirect requests`.
https://projectreactor.io/docs/netty/release/api/reactor/netty/http/client/HttpClient.RequestSender.html#send(org.reactivestreams.Publisher)

violetagg (changes requested):
Thanks for the PR!

reactor-netty-http/src/main/java/reactor/netty/http/client/HttpClientConnect.java
desiderantes: Sorry, i was trying to handle the QUERY redirect logic, but I didn't realize I messed with the other methods with this change, I'll remove this and redo it

PR #4297, opened 2026-07-17
Reduce per-request CPU in the HTTP/2 connection pool

samueldlightfoot:
## Summary

`Http2Pool`'s acquire/release drain path runs ~4× per request and is a notable CPU and allocation contributor under HTTP/2 load. This trims its fixed per-operation work on the hot path. The change is confined to `Http2Pool` with no public API change:

- **Memoize the `Http2Connection` on the event loop** so `Slot.goAwayReceived()` reads it directly instead of walking the pipeline when called off the event loop (`findConnection` runs on the acquiring thread, not the channel's loop, where the cached-context fast path is skipped).
- **Cache the `isH2cUpgrade` negative lookup and short-circuit on ALPN**, so `deliver` stops re-walking the whole pipeline for the upgrade handler that TLS and post-upgrade connections never carry.
- **Run `invalidate()`'s follow-up drain in a `finally` instead of a `doFinally`** — `destroyPoolable` is synchronous, so this drops a per-release `MonoDoFinally` and its lambda (two allocations per release).

Adds `Http2PoolTest` coverage for the `goAwayReceived` memoization (including `GO_AWAY`), and the h2c negative-lookup caching.

## Benchmarks

Baseline is `main` (`09312ca72`); patched is `main` + this change. Machine: Hetzner bare-metal, 12 cores (x86_64), idle; OpenJDK 17; JMH 1.37; DEBUG logging disabled. Built on the box from git and A/B-run back-to-back.

### Warm pool — acquire/release throughput and allocation

One connection held in steady state while 4 threads loop `acquire → deliver → release` (a few streams kept open so every op does real steady-state pool work). This isolates the fixed per-request drain cost on one hot connection. `-bm thrpt -tu s -t 4 -f 8 -wi 4 -i 4 -w 1s -r 1s -prof gc`

| Metric | Baseline | Patched | Δ |
|---|---|---|---|
| Throughput | 573,729 ± 8,985 ops/s | **632,835 ± 8,259 ops/s** | **+10.3%** | 
| Allocation | 447.9 ± 30.3 B/op | **395.7 ± 31.5 B/op** | **−11.7%** | 

An end-to-end H2C load harness (real event loops, blocking callers, warm multiplexed pool) shows ~2% request/s for the same change.

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
@@ -32,6 +32,7 @@ import java.util.function.Function;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.EventLoop;
+import io.netty.handler.codec.http2.Http2Connection;
 import io.netty.handler.codec.http2.Http2FrameCodec;
 import io.netty.handler.codec.http2.Http2MultiplexHandler;
 import io.netty.handler.ssl.ApplicationProtocolNames;
@@ -48,6 +49,7 @@ import reactor.core.publisher.Operators;
 import reactor.netty.Connection;
 import reactor.netty.FutureMono;
 import reactor.netty.NettyPipeline;
+import reactor.netty.internal.shaded.reactor.pool.AllocationStrategy;
 import reactor.netty.internal.shaded.reactor.pool.InstrumentedPool;
 import reactor.netty.internal.shaded.reactor.pool.PoolAcquirePendingLimitException;
 import reactor.netty.internal.shaded.reactor.pool.PoolAcquireTimeoutException;
@@ -163,6 +165,7 @@ class Http2Pool implements InstrumentedPool<Connection>, InstrumentedPool.PoolMe
 	final @Nullable BiPredicate<Connection, PooledRefMetadata> evictionPredicate;
 	final long maxIdleTime;
 	final int streamBatchSize;
+	final AllocationStrategy allocationStrategy;
 
 	long lastInteractionTimestamp;
 
@@ -188,6 +191,7 @@ class Http2Pool implements InstrumentedPool<Connection>, InstrumentedPool.PoolMe
 		this.poolConfig = poolConfig;
 		this.evictionPredicate = evictionPredicate;
 		this.maxIdleTime = maxIdleTime;
+		this.allocationStrategy = poolConfig.allocationStrategy();
 
 		recordInteractionTimestamp();
 		scheduleEviction();
@@ -210,7 +214,7 @@ class Http2Pool implements InstrumentedPool<Connection>, InstrumentedPool.PoolMe
 
 	@Override
 	public int allocatedSize() {
-		return poolConfig.allocationStrategy().permitGranted();
+		return allocationStrategy.permitGranted();
 	}
 
 	@Override
@@ -379,7 +383,7 @@ class Http2Pool implements InstrumentedPool<Connection>, InstrumentedPool.PoolMe
 	}
 
 	void drainLoop() {
-		recordInteractionTimestamp();
+		boolean interacted = false;
 		int maxPending = poolConfig.maxPending();
 
 		for (;;) {
@@ -394,10 +398,11 @@ class Http2Pool implements InstrumentedPool<Connection>, InstrumentedPool.PoolMe
 			int borrowersCount = pendingSize;
 
 			if (borrowersCount != 0) {
+				interacted = true;
 				// find a connection that can be used for opening a new stream
 				// when cached connections are below minimum connections, then allocate a new connection
 				boolean belowMinConnections = minConnections > 0 &&
-						poolConfig.allocationStrategy().permitGranted() < minConnections;
+						allocationStrategy.permitGranted() < minConnections;
 				boolean enableStrictReuse = strictConnectionReuse || minConnections > 0;
 				int resourcesCount = idleSize;
 				Slot slot = belowMinConnections ? null : findConnection(resources, resourcesCount);
@@ -442,11 +447,11 @@ class Http2Pool implements InstrumentedPool<Connection>, InstrumentedPool.PoolMe
 				}
 				else {
 					if (enableStrictReuse && !belowMinConnections &&
-							poolConfig.allocationStrategy().permitGranted() > resourcesCount) {
+							allocationStrategy.permitGranted() > resourcesCount) {
 						// connections allocations were triggered
 					}
 					else {
-						int permits = poolConfig.allocationStrategy().getPermits(1);
+						int permits = allocationStrategy.getPermits(1);
 						if (permits <= 0) {
 							if (maxPending >= 0) {
 								borrowersCount = pendingSize;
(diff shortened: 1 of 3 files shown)


violetagg:

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -188,6 +191,7 @@ class Http2Pool implements InstrumentedPool<Connection>, InstrumentedPool.PoolMe
 		this.poolConfig = poolConfig;
 		this.evictionPredicate = evictionPredicate;
 		this.maxIdleTime = maxIdleTime;
+		this.allocationStrategy = poolConfig.allocationStrategy();
```
violetagg: Do we really need to make this refactoring? I don't see what kind of improvement this will give us.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -379,7 +383,7 @@ void drain() {
 	}
 
 	void drainLoop() {
-		recordInteractionTimestamp();
```
violetagg, line 382: Why do we want to remove this? We need to record every single work done by the pool.

violetagg (changes requested):
Thanks for the PR!

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
samueldlightfoot: This is only cosmetic - I have reverted.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
samueldlightfoot: Thanks for the review!

The idea was to class an 'interaction' as work done within the drain loop rather than just an iteration of the drainLoop (e.g. delivering to a borrower). Reverted now.

violetagg:
@samueldlightfoot Your second commit is missing `Signed-off-by`

samueldlightfoot:
> @samueldlightfoot Your second commit is missing `Signed-off-by`

@violetagg fixed - thanks.

samueldlightfoot:
Benchmarks updated with the clock opt omtited.

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -1173,13 +1199,23 @@ boolean connectionLivenessCheckInProgress() {
...
 			ctx = connection.channel().pipeline().context(NettyPipeline.H2CUpgradeHandler);
-			h2cUpgradeHandlerCtx = ctx;
+			if (ctx == null) {
+				// The upgrade handler is added once at channel init and removes itself after the
+				// upgrade; it is never re-added, so a miss is permanent — stop re-walking.
+				h2cUpgradeHandlerAbsent = true;
+			}
+			else {
+				h2cUpgradeHandlerCtx = ctx;
+			}
```
violetagg: It is good practise to reset `h2cUpgradeHandlerCtx` once the handler is removed, thus `h2cUpgradeHandlerCtx` and `h2cUpgradeHandlerAbsent` will be consistent. What do you think?

```suggestion
			h2cUpgradeHandlerCtx = ctx;
			if (ctx == null) {
				// The upgrade handler is added once at channel init and removes itself after the
				// upgrade; it is never re-added, so a miss is permanent — stop re-walking.
				h2cUpgradeHandlerAbsent = true;
			}
```

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
violetagg, line 1175: Can we memoize the `http2Connection` here and then just use it in `goAwayReceived()` and `computeMaxConcurrentStreams()`?

```suggestion
			if (ctx != null) {
				http2Connection = ((Http2FrameCodec) ctx.handler()).connection();
			}
			return ctx;
```

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -1142,8 +1158,18 @@ int decrementConcurrencyAndGet() {
...
+				// Not yet memoized on the event loop (before the first deliver, or an h2c slot
+				// pre-upgrade / plain HTTP/1.1 slot): resolve once via the pipeline and cache.
+				ChannelHandlerContext frameCodec = http2FrameCodecCtx();
+				if (frameCodec == null) {
+					return false;
+				}
+				conn = ((Http2FrameCodec) frameCodec.handler()).connection();
+				http2Connection = conn;
+			}
+			return conn.goAwayReceived();
```
violetagg: We've memoized the `http2Connection`.

```suggestion
			Http2Connection conn = http2Connection;
			return conn != null && conn.goAwayReceived();
```

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
```diff
@@ -1089,10 +1099,16 @@ void updateMaxConcurrentStreams(long remoteMaxConcurrentStreams) {
...
+			if (frameCodec != null) {
+				// Memoize the connection on the event loop so goAwayReceived() can read it off-loop
+				// without re-walking the pipeline.
+				Http2Connection conn = ((Http2FrameCodec) frameCodec.handler()).connection();
+				this.http2Connection = conn;
+				if (http2MultiplexHandlerCtx() != null) {
+					int maxActiveStreams = conn.local().maxActiveStreams();
+					return pool.maxConcurrentStreams == -1 ? maxActiveStreams :
+							Math.min(pool.maxConcurrentStreams, maxActiveStreams);
+				}
```
violetagg: We've memoized the `http2Connection`.

```suggestion
			Http2Connection conn = http2Connection;
			if (frameCodec != null && conn != null && http2MultiplexHandlerCtx() != null) {
				int maxActiveStreams = conn.local().maxActiveStreams();
				return pool.maxConcurrentStreams == -1 ? maxActiveStreams :
						Math.min(pool.maxConcurrentStreams, maxActiveStreams);
```

violetagg (commented):
In addition to the comments below, what do you think if we also make a change in reactor.netty.http.client.Http2ConnectionProvider.DisposableAcquire#operationComplete i.e. to use the new `http2Connection`

instead of 

```
				if (!channel.isActive() || frameCodec == null ||
						((Http2FrameCodec) frameCodec.handler()).connection().goAwayReceived() ||
						!((Http2FrameCodec) frameCodec.handler()).connection().local().canOpenStream()) {
```

we can do it

```
				if (!channel.isActive() || connection == null ||
						connection.goAwayReceived() ||
						!connection.local().canOpenStream()) {
```

reactor-netty-http/src/main/java/reactor/netty/http/client/Http2Pool.java
samueldlightfoot: Moved single memoize location at http2FrameCodecCtx() which can be used by goAway and compute

violetagg (approved):

violetagg:
I back ported this to 1.3.x branch

PR #4298, opened 2026-07-18
Evaluate the address supplier once in `AddressUtils.updatePort`

samueldlightfoot:
updatePort re-invoked the upstream supplier three times, re-running the whole host/port override chain (a fresh IP-literal parse each) per request on the event loop, and could observe an inconsistent snapshot from a dynamic supplier. Reuse the value captured on the first call.


violetagg (approved):
Thanks for the PR!

violetagg:
I cherry picked the change in 1.3.x branch

PR #4300, opened 2026-07-19
Do not capture caller context into pool maintenance tasks

### LINKED ISSUE #4299: Memory leak: evictInBackground task indefinitely retains request-scoped ThreadLocal context when Hooks.enableAutomaticContextPropagation() is enabled
The connection pool's background eviction task (`ConnectionProvider.Builder#evictInBackground`) is library-internal maintenance work that has no use for user context. However, when automatic context propagation is enabled, it captures whatever request-scoped `ThreadLocal` state is present when the pool is created. It then **retains that state for the lifetime of the pool**, because the task re-captures its own restored context every time it re-schedules itself.

In our production application this retained 501MB — 52.2% of a 960MB live heap — across 44 pools, with a single pending eviction task pinning 437MB of a request that had completed more than 12 hours earlier. A self-contained reproducer (no Spring) is included below, pinned to the currently supported 1.3.x line (1.3.6). The leak also reproduces on the older 1.2.18 and 1.1.31, and the control run without the accessor does not leak.

## Expected Behavior

Request-scoped `ThreadLocal` values exposed through a Micrometer `ThreadLocalAccessor` should become collectible once the request that set them completes — or, at worst, should be retained for no longer than one eviction interval.

## Actual Behavior

With `io.micrometer:context-propagation` on the classpath, a registered `ThreadLocalAccessor` for a request-scoped object, and `Hooks.enableAutomaticContextPropagation()`, the eviction task permanently retains the context of the request that happened to create the pool. This happens in two steps:

1. **Initial capture.** `Hooks.enableAutomaticContextPropagation()` installs a `Schedulers.onScheduleHook` that wraps *every* `Runnable` scheduled through Reactor `Schedulers` with `captureThreadLocals()` + `ContextSnapshot.wrap()` ([Hooks.java#L543-L550](https://github.com/reactor/reactor-core/blob/v3.6.18/reactor-core/src/main/java/reactor/core/publisher/Hooks.java#L543-L550), [ContextPropagation.java#L127-L132](https://github.com/reactor/reactor-core/blob/v3.6.18/reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java#L127-L132)). Connection pools are created lazily on first acquire, which typically happens on a request-processing thread. As a result, the `SimpleDequePool` constructor's `scheduleEviction()` captures that request's ThreadLocals into the pending eviction task.

2. **Self-perpetuating re-capture.** When the eviction task fires, `ContextSnapshot.wrap()` restores the captured ThreadLocals around `run()`. At the end of its body, `evictInBackground()` calls `scheduleEviction()` again ([SimpleDequePool.java#L130-L138](https://github.com/reactor/reactor-pool/blob/v1.2.6/reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java#L130-L138), the version shaded into 1.3.6; unchanged on current `main`) — from *inside* the restored scope — so the next task re-captures the same, long-dead request context. This repeats every eviction interval, forever. (`scheduleEviction()` has exactly two call sites: the pool constructor and the end of `evictInBackground()`.)

The captured graph *is* released correctly when the task is cancelled or the provider is disposed. But in a healthy long-running application neither ever happens, so the retention is effectively permanent: a task that lives as long as the application keeps holding context that should live no longer than one request.

### Impact observed in production

In a Spring Boot MVC + `WebClient` application with a `ThreadLocalAccessor` exposing Spring's request-scoped `RequestAttributes` and `evictInBackground(Duration.ofSeconds(5))`: **44 pending `SchedulerTask`s — exactly one per `SimpleDequePool`, verified by OQL — retained 501MB (52.2%) of the 960.1MB live heap**. A single task pinned **437MB**: the buffered response of a request that had completed more than 12 hours earlier. The pools themselves are tiny (464-888 bytes); the weight is entirely the captured context.

MAT dominator path (reactor-netty-core 1.0.39 pod; the same mechanism reproduces locally on 1.1.31 through 1.3.6):

```
parallel-7 (NonBlockingThread, alive)                          <- GC root
 └ ScheduledThreadPoolExecutor$DelayedWorkQueue
   └ ScheduledFutureTask                                        (state = NEW, i.e. pending)
     └ reactor.core.scheduler.SchedulerTask
       ├ (SimpleDequePool.evictionTask field points here)
       └ task -> application onScheduleHook wrapper lambda      (app's own hook; leak reproduces without it — see reproducer)
           └ io.micrometer.context.ContextSnapshot$$Lambda
               ├ arg$2 -> SimpleDequePool$$Lambda               (this::evictInBackground)
               └ arg$1 -> DefaultContextSnapshot
                   └ HashMap$Node -> ServletRequestAttributes
                       └ response -> ... -> ByteArrayOutputStream
                           └ byte[437,518,336]
```

### The same pattern exists in reactor-netty proper

`PooledConnectionProvider` uses the same pattern for inactive-pool disposal: schedule from the constructor, then re-schedule from inside the task. The [constructor calls `scheduleInactivePoolsDisposal()`](https://github.com/reactor/reactor-netty/blob/v1.1.31/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java#L114), which does [`Schedulers.parallel().schedule(this::disposeInactivePoolsInBackground, ...)`](https://github.com/reactor/reactor-netty/blob/v1.1.31/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java#L394-L398), and `disposeInactivePoolsInBackground()` [re-schedules itself in its last statement](https://github.com/reactor/reactor-netty/blob/v1.1.31/reactor-netty-core/src/main/java/reactor/netty/resources/PooledConnectionProvider.java#L435) (links point at 1.1.31; the pattern is unchanged on current `main`). So applications that configure `ConnectionProvider.Builder#disposeInactivePoolsInBackground` hit the same indefinite retention through un-shaded reactor-netty code, independently of reactor-pool.

## Steps to Reproduce

Self-contained reproducer — **no Spring, no user `Schedulers.onScheduleHook`**; only reactor-netty + reactor-core + context-propagation. It sets a request-scoped value on a dedicated thread, makes one HTTP call through a pooled `HttpClient`, clears the ThreadLocal, and lets the thread die. A `WeakReference` then tells us whether the value can be collected.

`build.gradle`:

```groovy
plugins {
    id 'java'
    id 'application'
}
repositories { mavenCentral() }
dependencies {
    implementation 'io.projectreactor.netty:reactor-netty-http:1.3.6'
    implementation 'io.projectreactor:reactor-core:3.8.6'
    implementation 'io.micrometer:context-propagation:1.2.1'
}
application { mainClass = 'EvictionContextLeakDemo' }
```

`src/main/java/EvictionContextLeakDemo.java`:

```java
import io.micrometer.context.ContextRegistry;
import io.micrometer.context.ThreadLocalAccessor;
import reactor.core.publisher.Hooks;
import reactor.core.publisher.Mono;
import reactor.netty.DisposableServer;
import reactor.netty.http.client.HttpClient;
import reactor.netty.http.server.HttpServer;
import reactor.netty.resources.ConnectionProvider;

import java.lang.ref.WeakReference;
import java.time.Duration;
import java.util.concurrent.atomic.AtomicReference;

public class EvictionContextLeakDemo {

    /** Stand-in for any request-scoped ThreadLocal (Spring RequestAttributes, security context, ...). */
    static final ThreadLocal<RequestScope> REQUEST_SCOPE = new ThreadLocal<>();

    static final class RequestScope {
        final byte[] payload = new byte[32 * 1024 * 1024]; // 32MB, to make the retention visible in a heap dump
        final String id;
        RequestScope(String id) { this.id = id; }
        @Override public String toString() { return "RequestScope(" + id + ")"; }
    }

    static final class RequestScopeAccessor implements ThreadLocalAccessor<RequestScope> {
        @Override public Object key() { return "demo.requestScope"; }
        @Override public RequestScope getValue() {
            RequestScope v = REQUEST_SCOPE.get();
            if (v != null) log("ThreadLocalAccessor.getValue() -> captured " + v);
            return v;
        }
        @Override public void setValue(RequestScope value) {
            log("ThreadLocalAccessor.setValue(" + value + ") -> restored");
            REQUEST_SCOPE.set(value);
        }
        @Override public void setValue() { REQUEST_SCOPE.remove(); }
    }

    public static void main(String[] args) throws Exception {
        boolean control = args.length > 0 && args[0].equals("control");
        System.out.println("scenario            : " + (control ? "control (accessor NOT registered)" : "leak (accessor registered)"));
        System.out.println("java                : " + System.getProperty("java.vendor") + " " + System.getProperty("java.version"));
        System.out.println("reactor-netty       : " + HttpClient.class.getPackage().getImplementationVersion());
        System.out.println("reactor-core        : " + Hooks.class.getPackage().getImplementationVersion());
        System.out.println("context-propagation : " + ContextRegistry.class.getPackage().getImplementationVersion());
        System.out.println();

        if (!control) {
            ContextRegistry.getInstance().registerThreadLocalAccessor(new RequestScopeAccessor());
        }
        Hooks.enableAutomaticContextPropagation();

        DisposableServer server = HttpServer.create().port(0)
                .handle((req, res) -> res.sendString(Mono.just("ok")))
                .bindNow();

        ConnectionProvider provider = ConnectionProvider.builder("demo")
                .maxConnections(500)
                .maxIdleTime(Duration.ofSeconds(30))
                .maxLifeTime(Duration.ofSeconds(100))
                .evictInBackground(Duration.ofSeconds(1))  // required for the repro; production used 5s
                .lifo()
                .build();

        HttpClient client = HttpClient.create(provider).host("localhost").port(server.port());

        // Simulate one request-scoped unit of work on its own thread, then let the thread die,
        // so that no stack frame or ThreadLocalMap entry can keep the scope alive.
        WeakReference<RequestScope> scopeRef = simulateRequest(client);
        log("request thread terminated; the request-scoped value is unreachable from application code");
        System.out.println();

        boolean collected = awaitCollected(scopeRef, 8);
        System.out.println();
        System.out.println("=== after 8 GC + eviction cycles : request scope collected = " + collected
                + (collected ? "" : "   <-- LEAK: retained by the pending eviction task"));
        System.out.println();

        provider.disposeLater().block(Duration.ofSeconds(5));
        boolean collectedAfterDispose = awaitCollected(scopeRef, 5);
        System.out.println("=== after ConnectionProvider.dispose() : request scope collected = " + collectedAfterDispose);

        server.disposeNow();
        System.out.println();
        System.out.println(collected
                ? "RESULT: no leak - request scope was collected while the eviction task kept running"
                : "RESULT: LEAK CONFIRMED - request scope was retained by the eviction task chain and only released by dispose()");
        System.exit(0);
    }

    private static WeakReference<RequestScope> simulateRequest(HttpClient client) throws InterruptedException {
        AtomicReference<WeakReference<RequestScope>> out = new AtomicReference<>();
        Thread t = new Thread(() -> {
            RequestScope scope = new RequestScope("request-1");
            out.set(new WeakReference<>(scope));
            REQUEST_SCOPE.set(scope);
            try {
                // First use of the connection pool: the pool is created lazily HERE, while the
                // request-scoped ThreadLocal is set.
                String body = client.get().uri("/").responseContent().aggregate().asString()
                        .block(Duration.ofSeconds(10));
                log("request completed, response=" + body);
            } finally {
                REQUEST_SCOPE.remove(); // request-scoped cleanup, as a framework would do
            }
        }, "request-thread-1");
        t.start();
        t.join();
        return out.get();
    }

    private static boolean awaitCollected(WeakReference<?> ref, int seconds) throws InterruptedException {
        for (int i = 0; i < seconds; i++) {
            System.gc();
            if (ref.get() == null) return true;
            Thread.sleep(1000);
        }
        System.gc();
        return ref.get() == null;
    }

    static void log(String msg) {
        System.out.printf("[%-18s] %s%n", Thread.currentThread().getName(), msg);
    }
}
```

Run `gradle run` (leak scenario). Output — note the initial capture on the request thread, then the restore/re-capture pair on a **different `parallel-N` thread every second, after the request thread is dead**:

```
scenario            : leak (accessor registered)
java                : Amazon.com Inc. 17.0.17
reactor-netty       : 1.3.6
reactor-core        : 3.8.6
context-propagation : 1.2.1

[request-thread-1  ] ThreadLocalAccessor.getValue() -> captured RequestScope(request-1)   <- pool created lazily on the request thread
...
[request-thread-1  ] request completed, response=ok
[main              ] request thread terminated; the request-scoped value is unreachable from application code

[parallel-1        ] ThreadLocalAccessor.setValue(RequestScope(request-1)) -> restored    <- eviction run restores the dead request's context
[parallel-1        ] ThreadLocalAccessor.getValue() -> captured RequestScope(request-1)   <- ...and re-captures it into the next task
[parallel-2        ] ThreadLocalAccessor.setValue(RequestScope(request-1)) -> restored
[parallel-2        ] ThreadLocalAccessor.getValue() -> captured RequestScope(request-1)
[parallel-3        ] ThreadLocalAccessor.setValue(RequestScope(request-1)) -> restored
[parallel-3        ] ThreadLocalAccessor.getValue() -> captured RequestScope(request-1)
    ... (repeats every eviction interval)

=== after 8 GC + eviction cycles : request scope collected = false   <-- LEAK: retained by the pending eviction task

=== after ConnectionProvider.dispose() : request scope collected = true

RESULT: LEAK CONFIRMED - request scope was retained by the eviction task chain and only released by dispose()
```

Control run `gradle run --args=control` (identical, but the accessor is not registered):

```
=== after 8 GC + eviction cycles : request scope collected = true

=== after ConnectionProvider.dispose() : request scope collected = true

RESULT: no leak - request scope was collected while the eviction task kept running
```

With only the dependency versions changed, the leak also reproduces on
**reactor-netty 1.2.18** (reactor-core 3.7.19, context-propagation 1.1.4) and
**reactor-netty 1.1.31** (reactor-core 3.6.18, context-propagation 1.0.6).

Additional data points:

| condition | request scope after request ends |
|---|---|
| accessor registered + automatic propagation ON | **retained (leak)** |
| accessor NOT registered + automatic propagation ON | collected |
| provider disposed (eviction task cancelled) | collected |

## Possible Solution

A library-internal, self-rescheduling maintenance task should not participate in user-context capture:

- reactor-core could provide a capture-exempt scheduling path for library-internal periodic tasks. Today there is no way for reactor-pool/reactor-netty to opt out: every `Scheduler` built via the `Schedulers` factory methods funnels tasks through `Schedulers.onSchedule(task)`, which applies the context-capture hook.
- Alternatively, reactor-pool could schedule the initial eviction task from a clean context (the `SimpleDequePool` constructor runs on the acquiring thread) and re-schedule it from outside the restored snapshot scope, breaking the initial capture and the self-perpetuation respectively. `PooledConnectionProvider#scheduleInactivePoolsDisposal` would need the same treatment on the reactor-netty side.

I'd be happy to contribute a PR; the open question is which layer (reactor-netty, reactor-pool, or reactor-core) you'd prefer the fix in.

Precedent: in reactor/reactor-core#3468 the automatic-context-propagation queue-wrapper hook was removed because it "can be a source of `ThreadLocal` leaks, while it does not provide value in the current design" — the same reasoning seems to apply to snapshot-wrapping the pool eviction task.

Related but not sufficient: #4033 / #4038 added `evictInBackground(Duration, Scheduler)` (1.3.2). Passing a custom `Scheduler` does not avoid the capture, because any `Scheduler` created via the `Schedulers` factory methods still applies the `onSchedule` decoration; only a fully hand-written `Scheduler` implementation would bypass it.

We are aware that the context-propagation documentation places ThreadLocal cleanup responsibility on the application. However, this task is not created, scheduled, or clearable by the application: it is created inside the library, re-schedules itself, and lives as long as the application does.

### Workarounds we verified

- Do not register request-scoped `ThreadLocalAccessor`s globally when automatic propagation is enabled; propagate request context explicitly instead (`contextWrite` / snapshots at the boundaries).
- `evictInBackground(Duration.ZERO)` — no background task is scheduled; eviction then only happens on the acquire/release paths.

## Your Environment

* Production (heap dump): reactor-netty-core **1.0.39**, reactor-core 3.6.18, context-propagation 1.0.6 (micrometer-core 1.13.15), Amazon Corretto 17.0.12, Linux, Spring Boot MVC + `WebClient`
* Reproducer (no Spring): reactor-netty **1.3.6 / 1.2.18 / 1.1.31**, with reactor-core 3.8.6 / 3.7.19 / 3.6.18 and context-propagation 1.2.1 / 1.1.4 / 1.0.6, Amazon Corretto 17.0.17, macOS
* The `scheduleEviction()` code path is unchanged on current reactor-pool `main`

hocaron:
## What

`SimpleDequePool`'s background eviction task and `PooledConnectionProvider`'s inactive
pools disposal task are library-internal, self-rescheduling maintenance tasks that live as
long as the application. They were scheduled through `Schedulers.parallel()`, so every
scheduled `Runnable` goes through `Schedulers.onSchedule(...)` task decoration. When an
application enables `Hooks.enableAutomaticContextPropagation()` and registers a
`ThreadLocalAccessor` for a request-scoped object, that decoration wraps each task in a
Micrometer `ContextSnapshot` capturing the scheduling caller's ThreadLocals:

1. Connection pools are created lazily on first acquire — typically on a request
   processing thread — so the pool constructor's `scheduleEviction()` captures that
   request's context into the pending eviction task.
2. When the task runs, the snapshot is restored around `run()`, and the task re-schedules
   itself from inside that restored scope. The same long-dead request context is therefore
   re-captured into every subsequent task, indefinitely.

The result is permanent retention of the caller's object graph (437MB observed in
production for over 12 hours — see #4299 for the full analysis and a standalone
reproducer).

## Change

- Add a package-private `PoolMaintenanceScheduler`: a minimal `Scheduler` backed by a
  single shared daemon thread that submits tasks directly, deliberately bypassing
  `Schedulers.onSchedule` decoration.
- Background eviction: the default scheduler becomes `PoolMaintenanceScheduler.INSTANCE`;
  a scheduler explicitly supplied via `ConnectionProvider.Builder#evictInBackground(Duration, Scheduler)`
  (available since 1.3.2) is still honored.
- Inactive pools disposal: `Schedulers.parallel()` → `PoolMaintenanceScheduler.INSTANCE`.
- New tests in the `contextPropagationTest` source set reproducing both retention cases
  via `WeakReference` reachability; they fail before this change and pass after it.

Verified end-to-end: the standalone reproducer from #4299 leaks on 1.1.31, 1.2.18 and
1.3.6, and shows no leak against locally built `1.3.7-SNAPSHOT` and `1.2.19-SNAPSHOT`
builds containing this change.

Note: this also bypasses user-registered `onScheduleHook` decorators (e.g. MDC copying or
instrumentation) for these two internal tasks. I believe that is acceptable — these tasks
never execute user code — but it is a behavior change worth calling out.

## Alternatives considered

- A capture-exempt scheduling path in reactor-core would solve this class of problem for
  every library with internal periodic tasks (reactor-pool's eviction task affects direct
  reactor-pool users the same way). Happy to pursue that direction instead if preferred.
- Since 1.3.2, `ConnectionProvider.Builder#evictInBackground(Duration, Scheduler)` lets
  users supply their own scheduler, but any scheduler built via the `Schedulers` factory
  methods still applies the `onSchedule` decoration, so it cannot be used to opt out.

Fixes #4299.

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PoolMaintenanceScheduler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PoolMaintenanceScheduler.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright (c) 2026 VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty.resources;
+
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+import reactor.core.Disposable;
+import reactor.core.Disposables;
+import reactor.core.scheduler.Scheduler;
+
+/**
+ * A minimal {@link Scheduler} dedicated to connection pool maintenance tasks: background
+ * eviction and inactive pool disposal.
+ *
+ * <p>Tasks are submitted directly to a shared single daemon thread and deliberately do
+ * NOT go through {@link reactor.core.scheduler.Schedulers} task decoration
+ * ({@code Schedulers.onSchedule(...)}). Maintenance tasks are library-internal,
+ * self-rescheduling and live as long as the application; decorating them with hooks such
+ * as the context-capture hook installed by
+ * {@code Hooks.enableAutomaticContextPropagation()} captures the scheduling caller's
+ * {@code ThreadLocal} state (typically a request scope, because pools are created lazily
+ * on request processing threads) into a pending task and re-captures it on every
+ * reschedule, retaining the caller's object graph for the lifetime of the pool.
+ */
+final class PoolMaintenanceScheduler implements Scheduler {
+
+	static final PoolMaintenanceScheduler INSTANCE = new PoolMaintenanceScheduler();
+
+	final ScheduledThreadPoolExecutor executor;
+
+	private PoolMaintenanceScheduler() {
+		ScheduledThreadPoolExecutor e = new ScheduledThreadPoolExecutor(1, r -> {
+			Thread t = new Thread(r, "reactor-netty-pool-maintenance");
+			t.setDaemon(true);
+			return t;
+		});
+		e.setRemoveOnCancelPolicy(true);
+		this.executor = e;
+	}
+
+	@Override
+	public Disposable schedule(Runnable task) {
+		return toDisposable(executor.submit(task));
+	}
+
+	@Override
+	public Disposable schedule(Runnable task, long delay, TimeUnit unit) {
+		return toDisposable(executor.schedule(task, delay, unit));
+	}
+
+	@Override
+	public Disposable schedulePeriodically(Runnable task, long initialDelay, long period, TimeUnit unit) {
+		return toDisposable(executor.scheduleAtFixedRate(task, initialDelay, period, unit));
+	}
+
+	@Override
+	public Worker createWorker() {
+		return new PoolMaintenanceWorker(this);
+	}
+
+	@Override
+	public void dispose() {
+		// shared infrastructure scheduler backed by a daemon thread, never disposed
+	}
+
+	static Disposable toDisposable(Future<?> future) {
+		return () -> future.cancel(false);
+	}
+
+	static final class PoolMaintenanceWorker implements Worker {
+
+		final PoolMaintenanceScheduler parent;
+		final Disposable.Composite tasks = Disposables.composite();
+
+		PoolMaintenanceWorker(PoolMaintenanceScheduler parent) {
+			this.parent = parent;
+		}
+
+		@Override
+		public Disposable schedule(Runnable task) {
+			return track(parent.schedule(task));
+		}
+
+		@Override
+		public Disposable schedule(Runnable task, long delay, TimeUnit unit) {
+			return track(parent.schedule(task, delay, unit));
+		}
+
+		@Override
+		public Disposable schedulePeriodically(Runnable task, long initialDelay, long period, TimeUnit unit) {
+			return track(parent.schedulePeriodically(task, initialDelay, period, unit));
+		}
+
+		@Override
+		public void dispose() {
+			tasks.dispose();
+		}
+
+		@Override
+		public boolean isDisposed() {
+			return tasks.isDisposed();
+		}
+
+		Disposable track(Disposable disposable) {
+			tasks.add(disposable);
+			return disposable;
+		}
+	}
+}
--- a/reactor-netty-core/src/main/java/reactor/netty/resources/PoolMaintenanceScheduler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/resources/PoolMaintenanceScheduler.java
@@ -18,16 +18,18 @@ package reactor.netty.resources;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 
 import reactor.core.Disposable;
 import reactor.core.Disposables;
 import reactor.core.scheduler.Scheduler;
+import reactor.core.scheduler.Schedulers;
 
 /**
  * A minimal {@link Scheduler} dedicated to connection pool maintenance tasks: background
  * eviction and inactive pool disposal.
  *
- * <p>Tasks are submitted directly to a shared single daemon thread and deliberately do
+ * <p>Tasks are submitted directly to a shared daemon thread and deliberately do
  * NOT go through {@link reactor.core.scheduler.Schedulers} task decoration
  * ({@code Schedulers.onSchedule(...)}). Maintenance tasks are library-internal,
  * self-rescheduling and live as long as the application; decorating them with hooks such
@@ -44,8 +46,9 @@ final class PoolMaintenanceScheduler implements Scheduler {
 	final ScheduledThreadPoolExecutor executor;
 
 	private PoolMaintenanceScheduler() {
-		ScheduledThreadPoolExecutor e = new ScheduledThreadPoolExecutor(1, r -> {
-			Thread t = new Thread(r, "reactor-netty-pool-maintenance");
+		AtomicLong counter = new AtomicLong();
+		ScheduledThreadPoolExecutor e = new ScheduledThreadPoolExecutor(Schedulers.DEFAULT_POOL_SIZE, r -> {
+			Thread t = new Thread(r, "reactor-netty-pool-maintenance-" + counter.incrementAndGet());
 			t.setDaemon(true);
 			return t;
 		});
--- a/reactor-netty-core/src/contextPropagationTest/java/reactor/netty/PooledConnectionProviderContextPropagationTest.java
+++ b/reactor-netty-core/src/contextPropagationTest/java/reactor/netty/PooledConnectionProviderContextPropagationTest.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright (c) 2026 VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty;
+
+import io.micrometer.context.ContextRegistry;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Hooks;
+import reactor.netty.resources.ConnectionProvider;
+import reactor.netty.tcp.TcpClient;
+import reactor.netty.tcp.TcpServer;
+
+import java.lang.ref.WeakReference;
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Verifies that connection pool maintenance tasks (background eviction and inactive
+ * pool disposal) do not capture and retain the scheduling caller's ThreadLocal state
+ * when automatic context propagation is enabled.
+ */
+class PooledConnectionProviderContextPropagationTest {
+
+	DisposableServer disposableServer;
+	ConnectionProvider provider;
+	ContextRegistry registry;
+
+	@BeforeEach
+	void setUp() {
+		registry = ContextRegistry.getInstance();
+		registry.registerThreadLocalAccessor(new TestThreadLocalAccessor());
+		Hooks.enableAutomaticContextPropagation();
+		disposableServer =
+				TcpServer.create()
+				         .handle((in, out) -> out.send(in.receive().retain()))
+				         .bindNow();
+	}
+
+	@AfterEach
+	void tearDown() {
+		Hooks.disableAutomaticContextPropagation();
+		TestThreadLocalHolder.reset();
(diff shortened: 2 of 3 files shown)


reactor-netty-core/src/main/java/reactor/netty/resources/PoolMaintenanceScheduler.java
```diff
@@ -0,0 +1,124 @@
...
+	static final PoolMaintenanceScheduler INSTANCE = new PoolMaintenanceScheduler();
+
+	final ScheduledThreadPoolExecutor executor;
+
+	private PoolMaintenanceScheduler() {
+		ScheduledThreadPoolExecutor e = new ScheduledThreadPoolExecutor(1, r -> {
+			Thread t = new Thread(r, "reactor-netty-pool-maintenance");
+			t.setDaemon(true);
+			return t;
+		});
```
violetagg: Should we use the same pool size as `Schedulers.parallel()`?

```suggestion
		AtomicLong counter = new AtomicLong();
		ScheduledThreadPoolExecutor e = new ScheduledThreadPoolExecutor(Schedulers.DEFAULT_POOL_SIZE, r -> {
			Thread t = new Thread(r, "reactor-netty-pool-maintenance-" + counter.incrementAndGet());
			t.setDaemon(true);
			return t;
		});
```

violetagg (changes requested):
Thanks for the PR!

reactor-netty-core/src/main/java/reactor/netty/resources/PoolMaintenanceScheduler.java
hocaron: Good point — applied, and updated the class javadoc accordingly. 62ba279

violetagg (approved):

PR #4302, opened 2026-07-20
Fix `WebSocket-over-HTTP/2` server metrics gauge accounting on close

### LINKED ISSUE #4290: WebSocket over HTTP/2: server streams.active gauge leaks and connections.active goes negative after the connection closes
## Expected Behavior

After a WebSocket-over-HTTP/2 connection closes, the server-side gauges should return to the values they held before it was opened, the same way they do for WebSocket over HTTP/1.1:

- `reactor.netty.http.server.streams.active` back to `0`
- `reactor.netty.http.server.connections.active` unchanged (`0` in a single-connection test)

## Actual Behavior

After the HTTP/2 stream carrying the WebSocket closes:

- `reactor.netty.http.server.streams.active` stays at `1`. It is incremented when the upgrade request is read (`recordOpenStream`) and never decremented, so it leaks by one per WebSocket-over-HTTP/2 connection.
- `reactor.netty.http.server.connections.active` is decremented to `-1`.

The HTTP/1.1 WebSocket path is not affected; its gauges return to `0`.

Root cause (verified against `main`): when the upgrade completes, `WebsocketServerOperations#initHandshaker` replaces the HTTP metrics handler in place with the pass-through `WebsocketHttpServerMetricsHandler` added in #3229 for HTTP/1.1. Its `channelInactive` records `recorder.recordServerConnectionInactive(...)` regardless of the channel type. On an `Http2StreamChannel` the active unit is a stream — counted earlier through `recordOpenStream` — so the close should decrement `streams.active` via `recordClosedStream`. Recording a connection-inactive instead leaves `streams.active` incremented and decrements `connections.active` for a unit that was never counted there (`connections.active` is only incremented for a `SocketChannel`, never for an `Http2StreamChannel`).

The base handler already dispatches correctly — `AbstractHttpServerMetricsHandler#recordInactiveConnectionOrStream` routes an `Http2StreamChannel` to `recordClosedStream` — but the pass-through predates the HTTP/2 WebSocket path (#3691) and overrides `channelInactive` without going through it.

## Steps to Reproduce

The following test fails on `main`; the HTTP/1.1 sibling (`testServerConnectionsWebsocketMicrometer`) passes, which rules out a measurement artifact. It uses `.handle(...)` (Extended CONNECT uses the `CONNECT` method, so routing by `GET` would 404) and waits for the stream to close on the server before reading the gauges (`StreamCloseHandler` is a small `doOnConnection` helper that latches when the server-side stream channel goes inactive).

```java
@Test
void testServerConnectionsWebsocketMicrometerHttp2() throws Exception {
	disposableServer =
			createServer()
					.channelGroup(group)
					.host("127.0.0.1")
					.protocol(HttpProtocol.H2C)
					.http2Settings(spec -> spec.connectProtocolEnabled(true))
					.metrics(true, Function.identity())
					.handle((req, res) -> res.sendWebsocket((in, out) -> out.sendString(Mono.just("Hello World!"))))
					.doOnConnection(cnx -> StreamCloseHandler.INSTANCE.register(cnx.channel()))
					.bindNow();

	String address = formatSocketAddress(disposableServer.address());

	httpClient.protocol(HttpProtocol.H2C)
	          .websocket()
	          .uri("/ws")
	          .handle((in, out) -> in.receive().aggregate().asString())
	          .as(StepVerifier::create)
	          .expectNext("Hello World!")
	          .expectComplete()

	// wait until the stream is closed on the server side (pass-through channelInactive ran)
	assertThat(StreamCloseHandler.INSTANCE.awaitClientClosedOnServer()).as("awaitClientClosedOnServer timeout").isTrue();

	Gauge streamsActiveGauge = registry.find(SERVER_STREAMS_ACTIVE).tags(URI, HTTP, LOCAL_ADDRESS, address).gauge();
	Gauge connectionsActiveGauge = registry.find(SERVER_CONNECTIONS_ACTIVE).tags(URI, HTTP, LOCAL_ADDRESS, address).gauge();
	double streamsActive = streamsActiveGauge != null ? streamsActiveGauge.value() : 0d;
	double connectionsActive = connectionsActiveGauge != null ? connectionsActiveGauge.value() : 0d;

	// currently observed: streams.active = 1.0, connections.active = -1.0
	assertThat(streamsActive).isEqualTo(0d);
	assertThat(connectionsActive).isEqualTo(0d);
}
```

## Possible Solution

Have `WebsocketHttpServerMetricsHandler#channelInactive` dispatch on the channel type the way `recordInactiveConnectionOrStream` already does — for an `Http2StreamChannel`, record a stream close (`recordClosedStream`) instead of a connection-inactive. The HTTP/1.1 path keeps its current behavior. I'm happy to open a PR with the fix and this test if you'd like.

## Your Environment

* Reactor version(s) used: reactor-netty-http 1.3.6 (also reproduces on `main` / 1.3.7-SNAPSHOT)
* Other relevant libraries versions: Netty 4.2.15.Final
* JVM version (`java -version`): OpenJDK 25.0.2 (build 25.0.2+10-LTS)
* OS and version: Windows 11

LivingLikeKrillin:
This fixes the server-side metrics gauge accounting when a WebSocket-over-HTTP/2 connection closes.

On an `Http2StreamChannel` the active unit is a *stream*: reading the upgrade request increments `reactor.netty.http.server.streams.active` (via `recordOpenStream`). On close, however, the pass-through `WebsocketHttpServerMetricsHandler#channelInactive` recorded a connection-inactive (`recordServerConnectionInactive`) regardless of the channel type. As a result:

- `streams.active` was never decremented and leaked by one per WebSocket-over-HTTP/2 connection;
- `connections.active` (only ever incremented for a `SocketChannel`) was spuriously decremented and went negative (`-1`).

The fix delegates `channelInactive` to the inherited `recordInactiveConnectionOrStream`, which dispatches on the channel type — recording a stream close for an `Http2StreamChannel` and a connection-inactive for a `SocketChannel` — so the pass-through stays consistent with the base handler. The channel's `HttpServerOperations` is looked up and passed through, the way every other call site of `recordInactiveConnectionOrStream` does, so the close is recorded at the same local address source the matching open used (`ops.connectionHostAddress()`) rather than relying on the two coinciding.

HTTP/1.1 is unaffected: for the WebSocket operations types `connectionHostAddress()` resolves to `channel().localAddress()`, so a `SocketChannel` still records a connection-inactive at exactly the address the previous inline code recorded.

The new `testServerConnectionsWebsocketMicrometerHttp2` in `HttpMetricsHandlerTests` exercises the `Http2StreamChannel` (stream) branch — the same branch taken by HTTP/2-over-TLS — while the existing `testServerConnectionsWebsocketMicrometer` exercises the `SocketChannel` (connection) branch. After the stream closes, the new test asserts that `streams.active` returns to `0`, `connections.active` is not recorded, and the parent connection's `connections.total` is left intact; the HTTP/1.1 sibling continues to pass.

Fixes #4290

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/WebsocketServerOperations.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2025 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2011-2026 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -50,6 +50,7 @@ import reactor.netty.FutureMono;
 import reactor.netty.NettyOutbound;
 import reactor.netty.NettyPipeline;
 import reactor.netty.ReactorNetty;
+import reactor.netty.channel.ChannelOperations;
 import reactor.netty.http.HttpOperations;
 import reactor.netty.http.websocket.WebsocketInbound;
 import reactor.netty.http.websocket.WebsocketOutbound;
@@ -321,6 +322,13 @@ class WebsocketServerOperations extends HttpServerOperations
 			AtomicIntegerFieldUpdater.newUpdater(WebsocketServerOperations.class,
 					"closeSent");
 
+	/**
+	 * Replaces the {@code HttpMetricsHandler} slot after a WebSocket upgrade: it stops recording HTTP
+	 * request/response metrics (the exchange is over) while still carrying the connection/stream close
+	 * accounting via {@link #recordInactiveConnectionOrStream}. For the Micrometer recorder the
+	 * {@code ChannelMetricsHandler} is not present in the pipeline (see {@code HttpServerConfig}), so this
+	 * slot is the sole owner of that accounting.
+	 */
 	static final class WebsocketHttpServerMetricsHandler extends AbstractHttpServerMetricsHandler {
 
 		final HttpServerMetricsRecorder recorder;
@@ -347,11 +355,13 @@ class WebsocketServerOperations extends HttpServerOperations
 					recorder.recordServerConnectionClosed(ctx.channel().localAddress());
 				}
 
-				if (channelActivated) {
-					channelActivated = false;
-					// Always use the real connection local address without any proxy information
-					recorder.recordServerConnectionInactive(ctx.channel().localAddress());
+				// Dispatch on the channel type as the base handler does: an Http2StreamChannel close is a stream close.
+				ChannelOperations<?, ?> channelOps = ChannelOperations.get(ctx.channel());
+				HttpServerOperations ops = null;
+				if (channelOps instanceof HttpServerOperations) {
+					ops = (HttpServerOperations) channelOps;
 				}
+				recordInactiveConnectionOrStream(ctx.channel(), ops);
 			}
 			catch (RuntimeException e) {
 				// Allow request-response exchange to continue, unaffected by metrics problem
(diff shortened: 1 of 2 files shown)


violetagg (approved):
Thanks for the PR!

violetagg:
I added the fix to 1.3.x branch

PR #4303, opened 2026-07-20
Expose forwarded header selection

### LINKED ISSUE #4301: Expose an API for choosing whether to use Forwarded or X-Forwarded headers

arnabnandy7:
The default forwarded header handler always gives the standard `Forwarded` header precedence and only uses `X-Forwarded-*` headers when `Forwarded` is absent. Applications could not explicitly select one format without reimplementing the corresponding parsing logic in a custom handler.

Introduce `ForwardedHeaderMode` with the following modes:

- `AUTO` preserves the existing behavior.
- `FORWARDED` uses only the `Forwarded` header.
- `X_FORWARDED` uses only the `X-Forwarded-*` headers.

Add `HttpServer#forwarded(ForwardedHeaderMode)` while keeping `forwarded(true)` fully backward-compatible.

The reference documentation now includes a compiled example demonstrating explicit `X-Forwarded-*` selection. Parameterized tests use conflicting `Forwarded` and `X-Forwarded-For` values to verify each selection mode.

Resolves #4301

Validation:

- `ConnectionInfoTests`
- Reactor Netty examples compilation
- Spotless
- HTTP module Javadoc
- Antora resource generation

### DIFF
--- a/docs/modules/ROOT/pages/http-server.adoc
+++ b/docs/modules/ROOT/pages/http-server.adoc
@@ -292,6 +292,17 @@ include::{examples-dir}/clientaddress/Application.java[tags=snippet-application]
 `HTTP` request headers, if possible.
 <2> Returns the address of the remote (client) peer.
 
+When both header formats can be present, you can choose to use only the `Forwarded` header or only the
+`X-Forwarded-*` headers:
+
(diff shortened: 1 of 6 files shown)


violetagg:

violetagg:
@arnabnandy7 Thanks for the PR! We decided to go with the PR #4312

PR #4312, opened 2026-07-28
Require choice between `Forwarded` and `X-Forwarded` headers

### LINKED ISSUE #4301: Expose an API for choosing whether to use Forwarded or X-Forwarded headers

violetagg:
This commit introduces `HttpServer#forwarded(boolean, boolean)` to select whether to use the standard `Forwarded` header or the `X-Forwarded-*` alternative headers. The headers of the alternative that is not selected are ignored. A separate parameter enables support for `X-Forwarded-Prefix`.

`HttpServer#forwarded(boolean)` is removed, this is a breaking change.
`HttpServer#noForwarded()` is added as a replacement for `forwarded(false)`.

Fixes #4301


violetagg:
@rstoyanchev PTAL

rstoyanchev (approved):

violetagg:
After additional internal discussion, we decided to keep the old `HttpServer#forwarded(boolean)` but deprecated.

PR #4317, opened 2026-08-01
Add WebSocket server metrics

### LINKED ISSUE #4305: Support for WebSocket server metrics
## Motivation

Since #4118 the WebSocket client records WebSocket-specific metrics, but the server side has no WebSocket-level metrics. Once the upgrade completes, the HTTP metrics handler is replaced by a pass-through that stops recording HTTP request/response metrics (by design), and the channel-level metrics handler is removed for HTTP/1.1 (on HTTP/2 the parent connection keeps connection-count accounting only) — so nothing measures the handshake, the session lifetime, the WebSocket data traffic, or WebSocket errors on the server.

## Desired solution

Server-side equivalents under a `reactor.netty.websocket.server` prefix, mirroring the client implementation (#4118) and the existing server conventions — `uri` templated through the `uriTagValue` mapper, connection-level gauges dimensioned by `local.address` like `reactor.netty.http.server.connections.active`:

- **Core:** `handshake.time` (uri, status), `connection.duration`, `data.received`/`data.sent` (+ per-message timing), `errors`
- **Connection lifecycle:** a `connections.active` gauge and a `connections.closed` counter (close code + initiator)
- **In-flight health:** an outbound-backpressure gauge and ping/pong frame counters

Everything is additive-only: new recorder methods are `default` no-ops, and the public surface is limited to `WebSocketServerMetricsRecorder`, `ContextAwareWebSocketServerMetricsRecorder`, and a `Metrics` prefix constant — everything else package-private, as on the client side.

## Considered alternatives

Manual instrumentation via `doOnConnection` and custom Netty handlers — error-prone and inconsistent, the same reason the client got built-in metrics (#3820). A single large PR — to keep reviews manageable I would split this into a few small PRs, starting with the core meters.

## Additional context

The existing `WebsocketHttpServerMetricsHandler` is retained as-is — it carries the connection/stream close accounting (cf. #4290); the new WebSocket metrics handler is additive at the `WsMetricsHandler` slot. I have a working implementation. If this direction sounds good, I'll start with the core PR.

LivingLikeKrillin:
## Summary

Add WebSocket server metrics, mirroring the client WebSocket metrics (#4118). This is the first step of #4305. The meters are registered under the `reactor.netty.websocket.server` prefix, tagged with `uri` (mapped through the configured `uriTagValue` when one is set, as for the HTTP server meters):

- `handshake.time` (tags `uri`, `status`): `101` on a successful HTTP/1.1 upgrade, `200` on a successful HTTP/2 Extended CONNECT, `ERROR` when the handshake fails after the metrics handler is installed. Pre-upgrade rejections (unsupported version on HTTP/1.1, invalid Extended CONNECT on HTTP/2) are answered before installation and record no WebSocket meter
- `connection.duration`: from handler installation on upgrade to pipeline removal on close, so it includes the handshake and is also recorded when the handshake fails (same as the client)
- `data.received`, `data.sent`, `data.received.time`, `data.sent.time`: recorded once per message on the final fragment; control frames (Close, Ping, Pong) are excluded
- `errors`: pipeline exceptions and application errors raised while sending the WebSocket outbound

On the Micrometer path — as on the client — the handshake meter is produced by a `reactor.netty.websocket.server.handshake.time` observation instead of a direct recorder call.

## Changes

On upgrade, `WebsocketServerOperations` (HTTP/1.1) and `Http2WebsocketServerOperations` (HTTP/2) install a WebSocket metrics handler at the `WsMetricsHandler` slot, selected from the configured HTTP metrics handler the same way the client's `swapMetricsHandler` does. The Micrometer path always uses the built-in WebSocket recorder; on the other two, a recorder that does not implement the matching WebSocket type (`ContextAwareWebSocketServerMetricsRecorder` on the context-aware path, `WebSocketServerMetricsRecorder` on the plain one) is wrapped in an adapter that delegates the inherited HTTP methods and no-ops the two WebSocket ones. The `WebsocketHttpServerMetricsHandler` pass-through is kept unchanged, since it carries the connection/stream count accounting (#4290); the new handler is additive.

New public API: `WebSocketServerMetricsRecorder` (extends `HttpServerMetricsRecorder`, two methods) and `ContextAwareWebSocketServerMetricsRecorder` (adds the two `ContextView` overloads), plus `Metrics.WEBSOCKET_SERVER_PREFIX`. All other new types are package-private, and no meter of the built-in Micrometer recorder carries a `remote.address` or `proxy.address` tag. Native-image reflection configuration for the four new handlers and a reference documentation section are included.

## Test plan

- [x] Handshake time on HTTP/1.1 (`101`) and HTTP/2 Extended CONNECT (`200`)
- [x] Connection duration recorded on close
- [x] Data metrics in both directions, byte counts and timings
- [x] Errors counted on the application-error path
- [x] `ERROR` handshake status when the handshake fails after the handler is installed
- [x] `uriTagValue` mapping applied to the `uri` tag
- [x] Control frames excluded from the data metrics, inbound and outbound
- [x] A fragmented inbound message recorded once, on its final fragment
- [x] A burst of messages recorded per message in both directions, with exact byte attribution

These cover the Micrometer recorder path. The custom-recorder path is exercised end to end by the existing `testServerConnectionsWebsocketRecorder`, but no test asserts the WebSocket recordings made on it, and the context-aware recorder path is not covered by a test yet.

Related to #4305

### DIFF
--- a/docs/modules/ROOT/pages/http-server.adoc
+++ b/docs/modules/ROOT/pages/http-server.adoc
@@ -723,6 +723,57 @@ include::{examples-dir}/metrics/custom/Application.java[tags=snippet-application
 ----
 <1> Enables HTTP server metrics and provides {javadoc}/reactor/netty/http/server/HttpServerMetricsRecorder.html[`HttpServerMetricsRecorder`] implementation.
 
+[[websocket-metrics]]
+=== WebSocket Metrics
+When the connection is upgraded to a `WebSocket`, the HTTP server metrics handler stops recording the HTTP
+request/response metrics (it keeps recording the connection and stream metrics), and a dedicated
+`WebSocket` server metrics handler is installed alongside it. From that point on, the `WebSocket` traffic is
+exposed with a prefix of `reactor.netty.websocket.server`.
+
+The following table provides information for the `WebSocket` server metrics:
+
+[width="100%",options="header"]
+|=======
+| metric name | type | tags | description
+| reactor.netty.websocket.server.handshake.time | Timer | `uri`, `status` | Time spent for the `WebSocket` opening handshake.
+| reactor.netty.websocket.server.connection.duration | Timer | `uri` | Time spent for the lifetime of the `WebSocket` connection.
+| reactor.netty.websocket.server.data.received | DistributionSummary | `uri` | Amount of the data received, in bytes.
+| reactor.netty.websocket.server.data.sent | DistributionSummary | `uri` | Amount of the data sent, in bytes.
+| reactor.netty.websocket.server.data.received.time | Timer | `uri` | Time spent in consuming incoming data.
+| reactor.netty.websocket.server.data.sent.time | Timer | `uri` | Time spent in sending outgoing data.
+| reactor.netty.websocket.server.errors | Counter | `uri` | Number of errors that occurred, both pipeline exceptions (for example, protocol/codec errors reaching `exceptionCaught`) and application errors raised while sending the `WebSocket` outbound (for example, an error signal from the handler's outbound `Publisher`).
+|=======
+
+The four `data.*` meters are recorded once per message, when its final frame is processed. Control frames
+(Close, Ping, Pong) are excluded from all four.
+
+The `status` tag on `handshake.time` takes one of three values: `101` (a successful HTTP/1.1 upgrade),
+`200` (a successful HTTP/2 upgrade per RFC 8441 Extended CONNECT), or `ERROR` (the handshake failed
+after the `WebSocket` metrics handler was installed, see the note below). A handshake success-rate query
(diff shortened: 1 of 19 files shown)


LivingLikeKrillin:
Hi @violetagg, pushed a small fix: `@since` now says 1.3.8 rather than 1.3.7, which no longer holds now that 1.3.7 has shipped. Javadoc only.

This may be a lot to review in one pass. If smaller pieces would be easier, I am glad to split it myself: it divides cleanly along the three content commits (roughly 860, 710 and 620 lines), in whatever grouping suits you. If you would rather keep it as one PR, that works too.

violetagg:
Thanks!
I have this in my ToDo list and will make sure to merge it for 1.3.8

PR #4318, opened 2026-08-02
Reuse per-connection address and byte meters in the Micrometer client metrics path

samueldlightfoot:
### Summary

With metrics enabled, the client rebuilt per-connection-constant state on every recorded request/event: the remote address was re-formatted on each of the ~5 recorder dispatches, and the TCP-level byte meters were re-resolved (address format + key build + map lookup) on every I/O event.

- Format the remote/proxy address once per MicrometerHttpClientMetricsHandler and pass the strings to fast-path recorder methods; the public SocketAddress recorder API is kept as a thin delegating layer.
- Cache the resolved dataReceived/dataSent DistributionSummary per connection in MicrometerChannelMetricsHandler, guarded to the fixed-address case (unconnected UDP, where the address varies per datagram, keeps the per-event path).

Same meters, same tags, same recorded values; removes redundant per-request and per-event allocation and map lookups. Public API unchanged (japicmp clean).

### Benchmarks

Each rig arm runs the same load generator — client threads issuing HTTP/2 requests against a loopback server
for 180 s after a 30 s warmup — and reports steady-state throughput alongside async-profiler CPU and
allocation samples attributed to the client event-loop threads; arms differ only in the code under test, with
harness, JVM and parameters held fixed.

Rig: H2C loopback, 32 threads, JDK 17, `SimpleMeterRegistry` backing the global composite (metrics on for
both arms). JMH: 2 forks, 5×1s warmup, 5×1s measurement, JDK 17, `-prof gc` for allocation.

| Benchmark | Metric | Before | After | Δ |
|---|---|--:|--:|--:|
| Rig A/B, full client loop under real network I/O | Metrics-wrapper CPU | 4.48% of loop | 2.40% of loop | **−46.4%** |
| Rig A/B, full client loop under real network I/O | Metrics-wrapper allocation | 12,929 alloc samples | 5,645 alloc samples | **−56.3%** |
| Rig A/B, full client loop under real network I/O | Throughput | 94,367 req/s | 96,123 req/s | **+1.9%** |

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/channel/MicrometerChannelMetricsHandler.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/channel/MicrometerChannelMetricsHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2025 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2022-2026 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 package reactor.netty.channel;
 
 import io.micrometer.common.KeyValues;
+import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.Timer;
 import io.micrometer.observation.Observation;
 import io.netty.channel.ChannelHandler;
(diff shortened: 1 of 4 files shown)


violetagg:
@samueldlightfoot I added some changes, would you be able to rebase this PR? Thanks!

violetagg (approved):

violetagg:
@samueldlightfoot Thanks for the PR!

violetagg:
I back ported the commit to `1.3.x` branch

PR #4335, opened 2026-08-11
Add CheckReturnValue and CanIgnoreReturnValue annotations

filiphr:
Introduce reactor.netty.lang.CheckReturnValue and CanIgnoreReturnValue and apply CheckReturnValue at class level to Transport, ClientTransport, and HttpClient including its nested interfaces (UriConfiguration, RequestSender, ResponseReceiver, WebsocketSender, WebsocketReceiver). Error Prone and IntelliJ IDEA match these annotations by simple name, so discarding the result of the immutable fluent API is reported without requiring an extra dependency. Also fix test call sites that discarded fluent results.

The reason why `@CheckReturnValue` is applied both on method and class is due to that fact that IntelliJ does not go up the chat to look for `@CheckReturnValue`. It only does that for

* `javax.annotation.CheckReturnValue`
* `org.assertj.core.util.CheckReturnValue`
* `com.google.errorprone.annotations.CheckReturnValue`
* `org.jetbrains.annotations.CheckReturnValue`
* `org.springframework.lang.CheckReturnValue`

See https://github.com/JetBrains/intellij-community/blob/d9318167d85d96c9d2febad78c9198af853b5cb7/java/java-impl-inspections/src/com/siyeh/ig/bugs/IgnoreResultOfCallInspection.java#L141-L146

---

The reason I was inspired to do this was due to the fact that we had something like:

```java
httpClient.compress(true);

return httpClient.onConnected(...);
```

This lead to the `httpClient.compress(true)` not doing anything.

I am more than happy to apply this to other places in the codebase as well

### DIFF
--- a/reactor-netty-core/src/main/java/reactor/netty/lang/CanIgnoreReturnValue.java
+++ b/reactor-netty-core/src/main/java/reactor/netty/lang/CanIgnoreReturnValue.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2026 VMware, Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package reactor.netty.lang;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Specifies that the return value of the annotated method can be ignored.
+ *
+ * <p>This annotation overrides an enclosing {@link CheckReturnValue} annotation:
+ * when a type is annotated with {@link CheckReturnValue}, a method within that type
+ * can be annotated with this annotation to indicate that its return value is only
+ * <i>additional</i> information and callers are free to discard it. Static analysis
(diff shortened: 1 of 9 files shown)


filiphr:
@violetagg was wondering if you had the chance to look into this? Is this something that you would be interested in having in reactor netty?

violetagg:
I'll check this PR, next week. Thanks!

На ср, 19.08.2026 г. в 6:27 Filip Hrisafov ***@***.***>
написа:

filiphr:
@violetagg did you manage to have some time to look into this?

violetagg:
Give me some time, I'm reviewing the issues and PRs at the moment.
Thanks!

violetagg:
@filiphr,

Thanks for the PR and for the detailed explanation. I agree this is a real problem. However I don't think it is the right moment to add this annotation.

We already had this situation with the `nullability` annotations. We had our own and then we moved to the standard `JSpecify` ones in `1.3.0`. This took a lot of effort on our side and it was not easy for our users. There is no standard annotation for `@CheckReturnValue` today so I prefer not to add our own annotations and go through this again.

Two more reasons:

- To be consistent we need the same annotations on `HttpServer`, `TcpClient`, `TcpServer`, `UdpClient`, `UdpServer`, `QuicClient` and `QuicServer`. This is a few hundred annotations.
- Our build cannot verify this. When new methods are added the annotations can be easily forgotten.

I can reconsider when there is a standard annotation.

filiphr:
Thanks for looking into this @violetagg.

> There is no standard annotation for `@CheckReturnValue` today so I prefer not to add our own annotations and go through this again.

That's fair and makes complete sense. 

>* To be consistent we need the same annotations on HttpServer, TcpClient, TcpServer, UdpClient, UdpServer, QuicClient and QuicServer. This is a few hundred annotations.

Indeed, the PR was just a first step to understand whether or not this is something the team would like.

>* Our build cannot verify this. When new methods are added the annotations can be easily forgotten.

I don't think that this is correct, there could be ways to do it as part of the build.

>I can reconsider when there is a standard annotation.

Thanks. Currently there is https://github.com/jspecify/jspecify/issues/200 from JSpecify, so hopefully something like that would be useful for reactor-netty as well. In any case, thanks for taking the time to check the PR and providing the feedback you did. Hopefully there will be a standard way for this soon and then the project can adopt that.

PR #2883, opened 2023-08-29
Add a `http snoop server` sample

longkunbetter:
Add a 'http snoop server' sample which receives any request and tells the client the details of the request in a formatted string [#{761}](https://github.com/reactor/reactor-netty/issues/761)

This is a sample http server that corresponds to [HttpSnoopServer](https://github.com/netty/netty/blob/4.1/example/src/main/java/io/netty/example/http/snoop/HttpSnoopServer.java) in examples of project netty.

Please notice that the http snoop server is not the same as the http echo server.
It will return the details of the request, not only the request body.
So I think it is a good example for people who want to learn how to use reactor-netty to build a http server and have a full control of the request and response.

I have tested it and it works well.

I'm unfamilar with such PR operation, if there is any problem please tell me.

PR #2677, opened 2023-01-29
Add Reactor Netty version of secure chat server example reactor#761

jchenga:
Provide a corresponding Reactor Netty version of Secure chat server example as found in Netty project.

Related to #761

PR #4346, opened 2026-08-20
Add a WebSocket echo client example

anjeongkyun:
The `websocket/echo` sample only has a server. The other samples come as a client and server pair mirroring the Netty examples, so this adds the client half, corresponding to Netty's `http/websocketx/client` example.

Follows the same shape as `http/echo/EchoClient`, with the `secure`, `port`, `wiretap` and `http2` system properties.

Verified against the existing `WebsocketServer`:

```
$ ... WebsocketServer &
$ ... WebsocketClient
Hello World!
```

The send and receive are structured the way `WebsocketTest` does it, bounding the inbound with `next()` and returning `Mono<Void>`. Sending first and then subscribing to the inbound as a separate `Flux` made the server complete its side and close before the echo was read.

Related to #761

### DIFF
--- a/reactor-netty-examples/src/main/java/reactor/netty/examples/websocket/echo/WebsocketClient.java
+++ b/reactor-netty-examples/src/main/java/reactor/netty/examples/websocket/echo/WebsocketClient.java
@@ -46,7 +46,8 @@ public final class WebsocketClient {
 				Http2SslContextSpec http2SslContextSpec =
 						Http2SslContextSpec.forClient()
 						                   .configure(builder -> builder.trustManager(InsecureTrustManagerFactory.INSTANCE));
-				client = client.secure(spec -> spec.sslContext(http2SslContextSpec));
+				client = client.secure(spec -> spec.sslContext(http2SslContextSpec))
+				               .protocol(HttpProtocol.H2);
 			}
 			else {
 				Http11SslContextSpec http11SslContextSpec =
@@ -54,10 +55,6 @@ public final class WebsocketClient {
 						                    .configure(builder -> builder.trustManager(InsecureTrustManagerFactory.INSTANCE));
 				client = client.secure(spec -> spec.sslContext(http11SslContextSpec));
 			}
-		}
-
-		if (HTTP2) {
-			client = client.protocol(HttpProtocol.H2);
 		}
 
 		client.websocket()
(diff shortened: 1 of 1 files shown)


anjeongkyun:
One thing I should flag that I only noticed after opening this: there is already a WebSocket client under `documentation/http/client/websocket/Application.java`. I had excluded the `documentation` tree when I looked for gaps, which was a mistake on my part.

It looks like a different thing to me, but you are better placed to judge. That one is a snippet with `tag::` markers for the reference docs and it talks to the external `wss://echo.websocket.org`, whereas this one pairs with the `WebsocketServer` in the same package on localhost, the way `http/echo`, `tcp/echo` and the rest are laid out.

So the case for this PR is completing the client and server pair in the sample tree, not adding a first WebSocket client example. If you would rather not have both, I am happy to close this.

For the record, the merged PRs under #761 are Telnet (#2617), Quote of the Moment (#2636), secure chat (#2677), http snoop (#2883), CORS (#3554) and the WebSocket echo server (#4177). None of them added a client for the WebSocket sample.

reactor-netty-examples/src/main/java/reactor/netty/examples/websocket/echo/WebsocketClient.java
```diff
@@ -0,0 +1,73 @@
...
+			else {
+				Http11SslContextSpec http11SslContextSpec =
+						Http11SslContextSpec.forClient()
+						                    .configure(builder -> builder.trustManager(InsecureTrustManagerFactory.INSTANCE));
+				client = client.secure(spec -> spec.sslContext(http11SslContextSpec));
+			}
+		}
+
+		if (HTTP2) {
+			client = client.protocol(HttpProtocol.H2);
```
violetagg: What do you think about moving this in the 

```
if (SECURE) {
			if (HTTP2) {
```

section as H2 always needs a security configured. See the server example.

violetagg (commented):
Thanks for the PR!

anjeongkyun:
Good point, moved it into the secure branch so it sits with the TLS context, matching `WebsocketServer`.

Verified both ways against the server example:

```
$ runWsServer                          $ runWsClient
Hello World!

$ runWsServer -Dsecure -Dhttp2         $ runWsClient -Dsecure -Dhttp2
Hello World!
```

On the CodeQL comments, I left both as they are since they're the existing convention in this module rather than anything this example introduces, but happy to change them if you'd prefer:

- `Integer.parseInt(System.getProperty(...))` in the `PORT` field is what all 21 examples do.
- `spec.sslContext(...)` is used by 17 of them, including the `WebsocketServer` this pairs with.

violetagg:
Next week I'll review this PR

violetagg (approved):

PR #4347, opened 2026-08-20
Resume the read when the pipelined queue drains without a request

knutwannheden:
`HttpTrafficHandler` reads on demand, so every read has to be re-armed by an explicit `ctx.read()`. `resumeRead()` delegates to `run()` whenever the pipelined queue is non-empty, and `run()` is the only path that never re-arms it — it drains the queue, sets `overflow = false`, and returns.

That is fine as long as `run()` dispatches a request, because the resulting `HttpServerOperations` will terminate and call `resumeRead()` again. It is not fine when the queue drains without one:

1. A keep-alive request arrives before the previous request's `HttpServerOperations` has terminated, so it is buffered (`Buffering keep-alive request, previous ops not yet terminated`), along with its trailing `LastHttpContent`.
2. The previous operations terminate, `resumeRead()` sees a non-empty queue and schedules `run()`.
3. `run()` dispatches the buffered request, which is answered and terminates re-entrantly. `resumeRead()` runs again, still sees a non-empty queue — the trailing content — and schedules another `run()`.
4. The outer `run()` drains that trailing content and returns. The scheduled `run()` then finds an empty queue and returns too.

Nobody called `ctx.read()`. The idle timeout handler is not armed either, since it is installed only in the other `resumeRead()` branch, so nothing reaps the connection. The next request sits unread in the socket until the peer gives up; the server logs it only once the peer closes and the resulting read delivers it.

Captured from a server run with `reactor.netty.http.server` at `DEBUG`, on the connection that stalls:

```
41.855 [186] Buffering keep-alive request, previous ops not yet terminated
41.856 Draining next pipelined HTTP request, pending responses count: 1, queued: 2
41.857 AccessLog: "DELETE /… HTTP/1.1" 403 13 2
41.857 [187] Last HTTP packet was sent, terminating the channel
41.857 Draining next pipelined HTTP request, pending responses count: 0, queued: 1
41.858 [client] request_sent  GET /consent          <- never read
46.860 [client] Http client inbound receiver cancelled, closing channel.
46.862 [server] Increasing pending responses count: 1   <- the request, 5s late
```

## Why this is worth fixing

The window is narrow, but it does not require HTTP pipelining on the client's part — only connection reuse. A client that sends its next request as soon as it has read the previous response can land in it, and a busy event loop widens the gap between "last response frame written" and "operations terminated". We hit it as an intermittent hang in a `WebTestClient` test suite against a booted Spring Boot WebFlux application, roughly 3% of full-suite runs on a 2-vCPU CI machine, always as a request that never reaches a handler and unsticks the moment the client cancels. The same shape is reachable behind a load balancer that keeps connections alive to its targets, where it surfaces as a gateway timeout.

## The fix

Extract the read-arming into `requestRead()` and call it at the end of `run()` when the queue drained without a request being dispatched, since in that case no operations remain whose termination would do it. `IdleTimeoutHandler.addIdleTimeoutHandler` already no-ops when the handler is present, so sharing it between the two call sites is safe, and an extra `ctx.read()` is harmless.

## The test

`pipelinedRequestResumesReadWhenOnlyTrailingContentIsQueued` writes two requests in a single write so the second is buffered while the first is still in flight and is then answered off the queue, leaving only its trailing content. A third request is then sent, and the assertion is that it gets a response. It is deterministic — no retries or timing tolerance beyond the socket read timeout — and fails on an unpatched build with `java.net.SocketTimeoutException: Read timed out` on that third request.

## Verification

- `./gradlew :reactor-netty-http:test` — 2730 tests, 0 failures.
- `./gradlew check` — passes.

## Branch

Targeted at `1.2.x` as the oldest active maintenance branch, since #4189 touched this same handler and landed in 1.2.18. `resumeRead()` and `run()` are identical on `1.3.x` and `main`, so it applies cleanly to either — happy to retarget if triage prefers a different branch.

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpTrafficHandler.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpTrafficHandler.java
@@ -578,9 +578,13 @@ final class HttpTrafficHandler extends ChannelDuplexHandler implements Runnable 
 			ctx.executor().execute(this);
 		}
 		else {
-			IdleTimeoutHandler.addIdleTimeoutHandler(ctx.pipeline(), idleTimeout, HttpConnectionLiveness.CLOSE);
-			ctx.read();
+			requestRead();
 		}
+	}
+
+	void requestRead() {
+		IdleTimeoutHandler.addIdleTimeoutHandler(ctx.pipeline(), idleTimeout, HttpConnectionLiveness.CLOSE);
+		ctx.read();
 	}
 
 	@Override
@@ -662,6 +666,12 @@ final class HttpTrafficHandler extends ChannelDuplexHandler implements Runnable 
 			}
 		}
 		overflow = false;
+
+		// The queue drained without dispatching a request, so no HttpServerOperations will terminate
+		// and resume the read; without this the next request stays unread until the peer gives up.
+		if (nextRequest == null && persistentConnection) {
+			requestRead();
+		}
 	}
 
 	@Override
--- a/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
+++ b/reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
@@ -17,10 +17,13 @@ package reactor.netty.http.server;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.InetSocketAddress;
+import java.net.Socket;
 import java.net.SocketAddress;
 import java.net.URISyntaxException;
 import java.nio.charset.Charset;
@@ -105,6 +108,7 @@ import io.netty.handler.ssl.SslHandshakeTimeoutException;
 import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import io.netty.handler.ssl.util.SelfSignedCertificate;
 import io.netty.util.AttributeKey;
+import io.netty.util.CharsetUtil;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.ReferenceCounted;
 import io.netty.util.concurrent.DefaultEventExecutor;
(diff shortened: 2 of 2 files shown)


knutwannheden:
One more data point, from trying `-Dreactor.netty.http.server.lastFlushWhenNoRead=true` as a stopgap while this is unreleased. It turns out to only cover *half* of the problem, which I think is worth recording — it isolates two independent routes into the same lost read.

The pipelined queue can be populated from two places in `channelRead`:

- **`pendingResponses > 1`** — genuinely pipelined requests (`Buffering pipelined HTTP request`).
- **`existingOps instanceof HttpServerOperations`** — a keep-alive request that arrives before the previous request's operations have terminated (`Buffering keep-alive request, previous ops not yet terminated`). No pipelining by the client is required; it only needs to send its next request promptly on a reused connection.

`lastFlushWhenNoRead=true` closes the second route only. Its `terminateInternal()` call runs just before that check, and `terminate()` → `rebind(connection)` CASes the `ReactorNetty.CONNECTION` attribute away from the `HttpServerOperations`, so `ChannelOperations.get(channel)` no longer returns one and the request is dispatched inline instead of queued. It does nothing for the `pendingResponses > 1` route, which is reached before that block.

Measured both ways:

| scenario | route into the queue | `lastFlushWhenNoRead=false` | `=true` |
|---|---|---|---|
| the test added here (two requests in one write) | `pendingResponses > 1` | fails | **still fails** |
| a client that reuses a pooled connection with no pipelining | previous ops not yet terminated | fails (0/5 CI shards clean) | passes (5/5 shards clean) |

Both end in the same place — `run()` drains the queue and returns without `ctx.read()` — so the patch here covers both, and the added test deliberately exercises the route the property cannot mask.

Not suggesting the property as a fix, and I wouldn't recommend it generally given it is undocumented. Recording it because it corroborates the diagnosis from a second angle, and because anyone hitting only the keep-alive-reuse variant has a usable stopgap until this ships.

violetagg:
@knutwannheden Thanks for the PR! Unfortunately version 1.2.x is not supported anymore. Would you be able to do the changes against 1.3.x?

violetagg:
I applied the fix to the 1.3.x branch and will forward merge to the main branch.

violetagg:
a11596502249de3cc3c68123c3cbc47831bb33d0

knutwannheden:
Thank you! Somehow I'd missed the follow-up comment.

PR #3448, opened 2024-09-27
Ensure the buffer is not released twice

violetagg:
When `NettyOutbound#sendObject(java.lang.Object)` is used, it is Netty's responsibility to release the buffer on success/error

Fixes #3406

Issue #3406, opened 2024-08-23
Repeatedly releasing buf ,IllegalReferenceCountException: refCnt: 0, decrement: 1


switchYello:
## Expected Behavior

no exception, bytebuf release once.

## Actual Behavior

2024-08-23 13:47:13.682 ERROR [     demo-nio-1] [324] reactor.core.publisher.Operators Operator called default onErrorDropped
 io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
	at io.netty.util.internal.ReferenceCountUpdater.toLiveRealRefCnt(ReferenceCountUpdater.java:83)
	at io.netty.util.internal.ReferenceCountUpdater.release(ReferenceCountUpdater.java:148)
	at io.netty.buffer.AbstractReferenceCountedByteBuf.release(AbstractReferenceCountedByteBuf.java:101)
	at io.netty.util.ReferenceCountUtil.release(ReferenceCountUtil.java:90)
	at reactor.netty.channel.FluxReceive.drainReceiver(FluxReceive.java:298)
	at reactor.netty.channel.FluxReceive.onInboundNext(FluxReceive.java:403)
	at reactor.netty.channel.ChannelOperations.onInboundNext(ChannelOperations.java:425)
	at reactor.netty.channel.ChannelOperationsHandler.channelRead(ChannelOperationsHandler.java:115)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412)
	at io.netty.handler.logging.LoggingHandler.channelRead(LoggingHandler.java:280)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:442)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1407)
[...]

PR #4349, opened 2026-08-22
Guard FluxReceive.cleanQueue against already released buffers

### LINKED ISSUE #4343: FluxReceive.cleanQueue aborts the drain on an already released buffer
## Expected Behavior

When `FluxReceive` drains its inbound queue after a cancellation or a failed release, every queued
buffer is released and the subscriber gets its terminal signal — even if one of the queued items has
already been released.

## Actual Behavior

`cleanQueue` releases without guarding:

```java
final void cleanQueue(@Nullable Queue<Object> q) {
	if (q != null) {
		Object o;
		while ((o = q.poll()) != null) {
			if (log.isDebugEnabled()) {
				log.debug(format(parent.channel(), "{}: dropping frame {}"), this, parent.asDebugLogMessage(o));
			}
			ReferenceCountUtil.release(o);   // <-- throws on an already released item
		}
	}
}
```

One already-released item throws `IllegalReferenceCountException`, which escapes the `while` loop, so:

1. the remaining queued buffers are never released — a leak proportional to the queue depth, not to
   the one bad item;
2. `terminateReceiver(q, a)`, called right after `cleanQueue(q)` at three of the four call sites, is
   skipped, so the subscriber is left waiting for a terminal signal that never arrives;
3. the exception surfaces on the event loop as an unhandled task failure.

Note the asymmetry: the release inside `drainReceiver` **is** guarded (#3406, PR #3448), and its
`catch` block calls straight into the unguarded `cleanQueue`. So the hardened path routes into an
unhardened one.

**What this report is not about.** Whatever performs the first, early release is a separate question —
in our case it may well turn out to be our own code, and we have not attributed it yet. Nothing here
depends on that: the point is that `cleanQueue` is not exception-safe, so one bad item costs the whole
queue and the terminal signal. It is also not a request to tolerate an over-release silently, which is
what #1372 asked for — a caller that over-releases still gets its `IllegalReferenceCountException`,
because the guarded release in `drainReceiver` already records it as `inboundError` and hands it to the
subscriber. The only ask is that the *other* queued buffers still get released and the subscriber still
gets terminated.

How we hit it, for motivation only: an HTTP/2 (h2c) client streaming an endless multipart response,
when the downstream subscriber cancels mid-stream.

```
WARN AbstractEventExecutor - A task raised an exception. Task: reactor.netty.channel.FluxReceive$$Lambda@...
io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
    at io.netty.buffer.AbstractDerivedByteBuf.release(AbstractDerivedByteBuf.java:94)
    at io.netty.handler.codec.http.DefaultHttpContent.release(DefaultHttpContent.java:92)
    at io.netty.util.ReferenceCountUtil.release(ReferenceCountUtil.java:90)
    at reactor.netty.channel.FluxReceive.cleanQueue(FluxReceive.java:221)
    at reactor.netty.channel.FluxReceive.drainReceiver(FluxReceive.java:309)
    at reactor.netty.channel.FluxReceive.lambda$request$1(FluxReceive.java:136)
```

A heap dump of an affected pod showed 452 MB of pinned direct memory (`PoolChunk.pinnedBytes` summed
over the direct arenas) against **one** live `PooledByteBuf` in the whole heap — the buffers were
stranded with their owning objects already collected. That exhausted `-XX:MaxDirectMemorySize=1536m`.

## Steps to Reproduce

Two tests, no networking and no HTTP/2 needed — the defect is protocol-agnostic. Both fail on 1.3.x
HEAD and pass with the fix below. They cover the two `cleanQueue` call sites that matter: the
`a == null` cancellation branch of `drainReceiver`, and the `catch` around the guarded release.

```java
	@Test
	void cleanQueueReleasesEveryBufferWhenOneWasAlreadyReleased() {
		EmbeddedChannel channel = new EmbeddedChannel();
		ChannelOperations<NettyInbound, NettyOutbound> operations =
				new ChannelOperations<>(() -> channel, (connection, newState) -> {
				});
		FluxReceive receive = new FluxReceive(operations);

		// no initial request, so inbound buffers accumulate in the receiver queue instead of being delivered
		receive.subscribe(TestSubscriber.builder().initialRequest(0).build());

		ByteBuf releasedElsewhere = Unpooled.buffer().writeByte(1);
		ByteBuf queuedBehindIt = Unpooled.buffer().writeByte(2);
		receive.onInboundNext(releasedElsewhere);
		receive.onInboundNext(queuedBehindIt);
		assertThat(receive.getPending()).isEqualTo(2);

		releasedElsewhere.release();

		assertThatCode(receive::cancel).doesNotThrowAnyException();

		assertThat(queuedBehindIt.refCnt()).as("buffer queued behind an already released one").isZero();
	}

	@Test
	void cleanQueueStillTerminatesTheReceiverWhenABufferWasAlreadyReleased() {
		EmbeddedChannel channel = new EmbeddedChannel();
		ChannelOperations<NettyInbound, NettyOutbound> operations =
				new ChannelOperations<>(() -> channel, (connection, newState) -> {
				});
		FluxReceive receive = new FluxReceive(operations);

		TestSubscriber<Object> subscriber = TestSubscriber.builder().initialRequest(0).build();
		receive.subscribe(subscriber);

		ByteBuf delivered = Unpooled.buffer().writeByte(1);
		ByteBuf queuedBehindIt = Unpooled.buffer().writeByte(2);
		receive.onInboundNext(delivered);
		receive.onInboundNext(queuedBehindIt);

		// both buffers reached refCnt 0 elsewhere, the way the slices of an over-released cumulation buffer do
		delivered.release();
		queuedBehindIt.release();

		// delivers the first buffer, then fails to release it, which is the guarded release that calls cleanQueue
		assertThatCode(() -> receive.request(1)).doesNotThrowAnyException();

		assertThat(subscriber.isTerminated()).as("receiver terminated").isTrue();
		assertThat(subscriber.expectTerminalError()).isInstanceOf(IllegalReferenceCountException.class);
	}
```

On 1.3.x HEAD the first fails from the cancellation branch:

```
Expecting code not to raise a throwable but caught
  "io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
	at reactor.netty.channel.FluxReceive.cleanQueue(FluxReceive.java:224)
	at reactor.netty.channel.FluxReceive.drainReceiver(FluxReceive.java:243)
	at reactor.netty.channel.FluxReceive.doCancel(FluxReceive.java:210)
	at reactor.netty.channel.FluxReceive.cancel(FluxReceive.java:108)
```

and the second from inside the `catch` that PR #3448 added (`cleanQueue` at `drainReceiver:305`).
Swallowing that escaping exception in the test leaves `isTerminated()` false, i.e. the subscriber is
never told anything.

## Possible Solution

`ReferenceCountUtil.safeRelease(o)` in `cleanQueue`. Netty catches and logs instead of throwing, so the
drain completes, the remaining buffers are freed and `terminateReceiver` runs. One-word diff.

Happy to send a PR; the two tests above are ready. Note the project also has its own
`ReactorNetty.safeRelease` and `ChannelOperationsHandler.safeRelease`, both of which skip silently when
`refCnt() == 0`. I picked netty's because the guarded release added for #3406 *records* the failure
rather than swallowing it, and an early release is a real bug somewhere that should stay visible — but
say the word if you prefer the silent variant.

## Your Environment

* Reactor version(s) used: reproduced on `1.3.7-SNAPSHOT` (1.3.x HEAD); the production stack trace
  above is from 1.2.11. `cleanQueue` is byte-identical from 1.2.11 through `main`, so every version to
  date is affected.
* Other relevant libraries versions (eg. `netty`, ...): netty 4.1.135.Final for the tests; netty
  4.1.127.Final and spring-webflux 6.2.12 in the production run.
* JVM version (`java -version`): Temurin 25 in production; the tests run on the build's JDK 8
  toolchain.
* OS and version (eg. `uname -a`): Linux (container) in production.

arimu1:
`FluxReceive.cleanQueue` released queued frames without guarding against an already released buffer. One `IllegalReferenceCountException` aborted the drain, so remaining buffers leaked, `terminateReceiver` was skipped at the call sites that depend on `cleanQueue` completing, and the exception surfaced as an unhandled event-loop task failure.

`drainReceiver` already catches that exception when releasing the delivered frame (#3406 / #3448), but its `catch` still called the unguarded `cleanQueue`.

Catch `IllegalReferenceCountException` per queued item (same approach as `drainReceiver` / `ChannelOperationsHandler.safeRelease`) so the rest of the queue is released and the subscriber is still terminated.

Fixes #4343.

### DIFF
--- a/reactor-netty-core/src/test/java/reactor/netty/channel/FluxReceiveTest.java
+++ b/reactor-netty-core/src/test/java/reactor/netty/channel/FluxReceiveTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 VMware, Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2023-2026 VMware, Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,12 +17,18 @@ package reactor.netty.channel;
 
 import java.time.Duration;
 
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
 import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.util.IllegalReferenceCountException;
 import org.junit.jupiter.api.Test;
 import reactor.netty.NettyInbound;
 import reactor.netty.NettyOutbound;
 import reactor.test.subscriber.TestSubscriber;
 import reactor.test.util.RaceTestUtils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
 
 public class FluxReceiveTest {
 
@@ -38,5 +44,53 @@ public class FluxReceiveTest {
 
 			subscriber.block(Duration.ofSeconds(5));
 		}
+	}
+
+	@Test
+	void cleanQueueReleasesEveryBufferWhenOneWasAlreadyReleased() {
+		EmbeddedChannel channel = new EmbeddedChannel();
+		ChannelOperations<NettyInbound, NettyOutbound> operations =
+				new ChannelOperations<>(() -> channel, (connection, newState) -> {
+				});
+		FluxReceive receive = new FluxReceive(operations);
+
+		receive.subscribe(TestSubscriber.builder().initialRequest(0).build());
+
+		ByteBuf releasedElsewhere = Unpooled.buffer().writeByte(1);
+		ByteBuf queuedBehindIt = Unpooled.buffer().writeByte(2);
+		receive.onInboundNext(releasedElsewhere);
+		receive.onInboundNext(queuedBehindIt);
+		assertThat(receive.getPending()).isEqualTo(2);
+
+		releasedElsewhere.release();
+
+		assertThatCode(receive::cancel).doesNotThrowAnyException();
+
+		assertThat(queuedBehindIt.refCnt()).as("buffer queued behind an already released one").isZero();
+	}
+
+	@Test
+	void cleanQueueStillTerminatesTheReceiverWhenABufferWasAlreadyReleased() {
+		EmbeddedChannel channel = new EmbeddedChannel();
+		ChannelOperations<NettyInbound, NettyOutbound> operations =
+				new ChannelOperations<>(() -> channel, (connection, newState) -> {
+				});
+		FluxReceive receive = new FluxReceive(operations);
+
+		TestSubscriber<Object> subscriber = TestSubscriber.builder().initialRequest(0).build();
+		receive.subscribe(subscriber);
+
+		ByteBuf delivered = Unpooled.buffer().writeByte(1);
+		ByteBuf queuedBehindIt = Unpooled.buffer().writeByte(2);
+		receive.onInboundNext(delivered);
+		receive.onInboundNext(queuedBehindIt);
+
+		delivered.release();
+		queuedBehindIt.release();
+
+		assertThatCode(() -> receive.request(1)).doesNotThrowAnyException();
+
+		assertThat(subscriber.isTerminated()).as("receiver terminated").isTrue();
+		assertThat(subscriber.expectTerminalError()).isInstanceOf(IllegalReferenceCountException.class);
 	}
 }
--- a/reactor-netty-core/src/test/java/reactor/netty/channel/FluxReceiveTest.java
+++ b/reactor-netty-core/src/test/java/reactor/netty/channel/FluxReceiveTest.java
@@ -20,8 +20,8 @@ import java.time.Duration;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.embedded.EmbeddedChannel;
-import io.netty.util.IllegalReferenceCountException;
 import org.junit.jupiter.api.Test;
+import reactor.netty.ConnectionObserver;
 import reactor.netty.NettyInbound;
 import reactor.netty.NettyOutbound;
 import reactor.test.subscriber.TestSubscriber;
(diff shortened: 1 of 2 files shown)


reactor-netty-core/src/test/java/reactor/netty/channel/FluxReceiveTest.java
```diff
@@ -39,4 +45,52 @@ void disposeAndSubscribeRaceTest() {
...
+	void cleanQueueReleasesEveryBufferWhenOneWasAlreadyReleased() {
+		EmbeddedChannel channel = new EmbeddedChannel();
+		ChannelOperations<NettyInbound, NettyOutbound> operations =
+				new ChannelOperations<>(() -> channel, (connection, newState) -> {
+				});
+		FluxReceive receive = new FluxReceive(operations);
+
+		receive.subscribe(TestSubscriber.builder().initialRequest(0).build());
+
+		ByteBuf releasedElsewhere = Unpooled.buffer().writeByte(1);
```
violetagg: Please provide end-to-end test so that we can see where is this `else where`. There is a strict contract between Netty -> Reactor Netty -> application. This contract is exactly to know which component needs to release.
The contract is: 
- Once Netty delivers the buffer to Reactor Netty, Reactor Netty is responsible to release (https://netty.io/wiki/reference-counted-objects.html#inbound-messages)
- There are two scenarios for the application
  - it can consume the buffer immediately -> Reactor Netty releases
  - it wants to delay -> application calls `retain`, then Reactor Netty releases and application releases
https://projectreactor.io/docs/netty/release/api/reactor/netty/ByteBufFlux.html#retain()

arimu1: Thanks — I replaced those tests with an end-to-end Netty → Reactor Netty → application path (`ChannelOperationsHandler` inbound → `receive()` / `receive().retain()`).

The contract is now explicit:
- consume immediately: Reactor Netty releases (`refCnt == 0` after onNext)
- delay: application `retain`, Reactor Netty releases, application releases (`refCnt == 1` then `0`)

The already-released buffer is the one Reactor Netty released after consume immediately. Cancel then still releases the remaining queued buffers (`cleanQueue`) without the test releasing Reactor Netty-owned frames. The `cleanQueue` release contract is unchanged.

violetagg: @arimu1 All new tests are green even without the fix. Can you check?

PR #4362, opened 2026-08-31
Parse one-digit ports in HttpClient String URIs

### LINKED ISSUE #4357: HttpClient String URI parser treats one-digit ports as part of the hostname
## Expected Behavior

The `HttpClient` String URI overload should parse a valid one-digit port as the port:

```java
HttpClient.create()
        .get()
        .uri("http://127.0.0.1:1")
        .response()
        .block();
```

This should attempt a connection to host `127.0.0.1`, port `1`, and normally fail immediately with `Connection refused: /127.0.0.1:1`.

## Actual Behavior

The client treats `127.0.0.1:1` as the hostname and defaults the port to 80. With DNS tracing enabled, the lookup is:

```
DefaultDnsQuestion(127.0.0.1:1. IN A)
```

That can turn an intended immediate loopback refusal into a long DNS timeout. Passing the same value through the `URI` overload instead of the String overload parses host and port correctly.

## Cause

`UriEndpointFactory.URL_PATTERN` currently accepts explicit ports only when they contain 2-5 digits:

```java
"(?:(\\w+)://)?((?:\\[.+?])|(?<!\\[)(?:[^/?]+?))(?::(\\d{2,5}))?([/?].*)?"
```

Because the host group is permissive, the optional port group does not match `:1` and the colon plus digit are consumed by the host group. The same pattern is present on `main` as of 2026-08-30. Changing the port group to `\\d{1,5}` and adding one-digit-port cases to `UriEndpointFactory` tests appears sufficient.

## Environment

- Reactor Netty: 1.3.7
- Netty: 4.2.17.Final
- JVM: OpenJDK 17.0.20.1
- OS: macOS ARM64

The parser behavior is platform-independent; macOS only made the accidental DNS lookup conspicuous.

kalayciburak:
`UriEndpointFactory.URL_PATTERN` only accepted explicit ports of 2-5 digits. A String URI such as `http://127.0.0.1:1` therefore kept `:1` in the host group and defaulted the port to 80, which can turn an intended loopback refusal into a DNS lookup. The `URI` overload already parses one-digit ports correctly.

This change accepts 1-5 digit ports in the String parser and adds coverage for IPv4, IPv6, and host-only forms.

Fixes #4357

## Tests

- `./gradlew :reactor-netty-http:test --tests reactor.netty.http.client.UriEndpointFactoryTest` (GREEN)
- `./gradlew :reactor-netty-http:checkstyleMain :reactor-netty-http:checkstyleTest spotlessCheck` (GREEN)

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/UriEndpointFactory.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/UriEndpointFactory.java
@@ -31,7 +31,7 @@ final class UriEndpointFactory {
 	final BiFunction<String, Integer, InetSocketAddress> inetSocketAddressFunction;
 
 	static final Pattern URL_PATTERN = Pattern.compile(
-			"(?:(\\w+)://)?((?:\\[.+?])|(?<!\\[)(?:[^/?]+?))(?::(\\d{2,5}))?([/?].*)?");
+			"(?:(\\w+)://)?((?:\\[.+?])|(?<!\\[)(?:[^/?]+?))(?::(\\d{1,5}))?([/?].*)?");
 	static final Pattern SCHEME_PATTERN = Pattern.compile("https?|wss?");
 
 	UriEndpointFactory(Supplier<? extends SocketAddress> connectAddress, boolean defaultSecure,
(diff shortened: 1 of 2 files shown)


violetagg (approved):
Thanks for the PR!

violetagg:
I backported the fix to the `1.3.x` branch

PR #4368, opened 2026-09-04
Expose connection lifetime to custom `MeterRegistrar`

### LINKED ISSUE #4369: Expose connection lifetime to custom meters
## Motivation

Metric consumers can't get at the existing connection lifetime metrics.

## Desired solution

Expose connection lifetime to metric consumers.

I did take a stab at this in https://github.com/reactor/reactor-netty/pull/4368

caiges:
Custom `ConnectionProvider.MeterRegistrar` implementations previously never received connection lifetime data because that duration is recorded through `PoolMetricsRecorder` where `recordLifetimeDuration` was a no-op.

# Tests

```
  ./gradlew :reactor-netty-http:test --tests "reactor.netty.resources.DefaultPooledConnectionProviderTest.testCustomRegistrarReceivesConnectionLifetime" --tests "reactor.netty.http.client.Http2PooledConnectionProviderCustomMetricsTest.measureConnectionLifetime"                                                                                                                                                                                                           
```

### DIFF
--- a/reactor-netty-http/src/main/java/reactor/netty/http/client/MeterRegistrarPoolMetricsRecorder.java
+++ b/reactor-netty-http/src/main/java/reactor/netty/http/client/MeterRegistrarPoolMetricsRecorder.java
@@ -21,9 +21,10 @@ import reactor.netty.resources.ConnectionProvider.MeterRegistrar;
 import java.net.SocketAddress;
 
 /**
- * A {@link PoolMetricsRecorder} that bridges pending acquisition latencies to a user-provided
- * {@link MeterRegistrar}. This allows custom registrars to receive the {@code pending.streams.time}
- * data that is otherwise only available to the built-in Micrometer integration.
+ * A {@link PoolMetricsRecorder} that bridges pending acquisition latencies and connection lifetime
+ * durations to a user-provided {@link MeterRegistrar}. This allows custom registrars to receive the
+ * {@code pending.streams.time} data that is otherwise only available to the built-in Micrometer
+ * integration, as well as connection lifetime data that is not available even via Micrometer.
  *
  * @author ejhnsn
  * @since 1.3.7
@@ -69,7 +70,7 @@ final class MeterRegistrarPoolMetricsRecorder implements PoolMetricsRecorder {
 
 	@Override
 	public void recordLifetimeDuration(long millisecondsSinceAllocation) {
-		//noop
+		registrar.recordConnectionLifetime(poolName, id, remoteAddress, millisecondsSinceAllocation);
 	}
 
 	@Override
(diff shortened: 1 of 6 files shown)


reactor-netty-http/src/main/java/reactor/netty/http/client/MeterRegistrarPoolMetricsRecorder.java
```diff
@@ -69,7 +70,7 @@ public void recordRecycled() {
 
 	@Override
 	public void recordLifetimeDuration(long millisecondsSinceAllocation) {
-		//noop
+		registrar.recordConnectionLifetime(poolName, id, remoteAddress, millisecondsSinceAllocation);
```
violetagg, line 73: I see the use case of exposing this metric for a connection, but can you please explain what we expect with this metric for streams? This class was introduced in order to be able to expose some metrics for streams. However once a stream is obtained from the pool it is never returned to the pool because streams are not reusable. So basically a lifetime of a stream is the actual response time.

caiges: Good catch, that's very much me getting a feel for this codebase. I am now also recording the lifetime of the connection duration for HTTP/2/3 (in addition to HTTP/1) when the slot in the pool is invalidated which is the real intent here, not stream/request response time -- I added a more thorough test which was previously giving a false positive.

violetagg: What I meant is that the changes in reactor-netty-http seem not needed. Let me explain: In Reactor we have reactor.netty.resources.DefaultPooledConnectionProvider which is used for HTTP/1.1 connection pool and this uses reactor.netty.resources.MeterRegistrarPoolMetricsRecorder which I do agree needs a recording for the connection life time. On the other hand we have reactor.netty.http.client.Http2ConnectionProvider which is used for HTTP/2 connection pool and it uses reactor.netty.http.client.MeterRegistrarPoolMetricsRecorder for metrics related to streams. The relation between the 2 connection pools is that HTTP/2 pool acquires a connection from HTTP/1.1 pool, when this connection is not needed anymore it is the HTTP/1.1 pool that destroys it. IMO we do not need changes on the level of HTTP/2 pool.

caiges: Ah! That makes sense, thank you for the explanation. I guess that means metric naming for those stream-specific metrics will be under the `http2` prefixed pool.

violetagg: yes exactly

violetagg (approved):
Thanks for the PR!

violetagg:
I prepared a PR for the built-in integration with Micrometer.
#4382

PR #4373, opened 2026-09-08
Enable reproducible builds

mbhave:
This assumes that the build is reproducible for the same major version of the JDK

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -340,9 +340,14 @@ subprojects {
 		enabled = false
 	}
 
+	tasks.withType(AbstractArchiveTask).configureEach {
+        preserveFileTimestamps = false
+        reproducibleFileOrder = true
+    }
+
 	jar {
 		manifest {
-			attributes("Created-By": "${System.getProperty("java.version")} (${System.getProperty("java.specification.vendor")})",
+			attributes("Build-Jdk-Spec": "${System.getProperty("java.specification.version")}",
 					"Implementation-Title": project.name,
 					"Implementation-Version": project.version)
 		}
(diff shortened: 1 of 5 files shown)


violetagg:
@mbhave Thanks for the PR!
Please review these failures and tell us which are acceptable
GHA https://github.com/reactor/reactor-netty/actions/runs/34347135609/job/102451302222
Branch https://github.com/reactor/reactor-netty/tree/reproducibility-check

mbhave:
@violetagg the ones other than the docs-zip look like actual differences between the jdk patch versions, not much we can do there. For docs-zip, I added the reproducible flag which gets passed to asciidoctor-pdf. Can you try it on your branch and check?

violetagg:
This is the check with the second commit
https://github.com/reactor/reactor-netty/actions/runs/34572927354/job/103178799283

Please add `Signed-off-by` to the second commit so that DCO is happy. Thanks!

violetagg:
This is the check with the latest changes https://github.com/reactor/reactor-netty/actions/runs/35192220123

mbhave:
The differences are between two patches of the JDK due to a couple of reasons
- javadoc differences due to jQuery
- bytecode difference in `DomainSocketAddressUtils` which is caused by a difference in how the two patches handle `invokedDynamic` with String concatenation.

I tested this branch locally on 17.0.9 and 17.0.20 and got byte-for-byte identical binaries. I think we can merge this PR if you're happy with the changes.

violetagg:
@mbhave Thanks for the PR!

violetagg (approved):

PR #4338, opened 2026-08-12
Extend `HttpCompressionOptionsSpec` with the ability to specify desired `maxPipelineDepth`

reta:
In scope of https://github.com/netty/netty/pull/17063, the `HttpContentEncoder` got a new property `maxPipelineDepth` and the respective constructor variant. The `HttpContentCompressor` have been updated in scope of https://github.com/netty/netty/pull/17068 to allow `maxPipelineDepth` tuning as the default (`128`) value might be too low for some applications.

The pull request allows `maxPipelineDepth` to be passed through `HttpCompressionOptionsSpec`.
The change is not breaking existing APIs, adds some overloaded methods.

PR #4380, opened 2026-09-14
Extend `HttpServerConfig` with the ability to specify desired `maxPipelineDepth`

reta:
Similarly to https://github.com/reactor/reactor-netty/pull/4338, add the ability to configure `maxPipelineDepth` at the `HttpServerCodec` level, this is a new configuration option in Netty `4.2.18.Final`. 

The change is not breaking existing APIs, adds some overloaded methods.
Targeted version: 1.4.x / 1.3.x


reta:
@violetagg would appreciate if you could take a look please, this is new configuration knob in latest netty (4.2.18.Final), largely independent from the `HttpContentCompressor` although same property name.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -1467,6 +1467,24 @@ public final HttpServer wiretap(boolean enable) {
 		return super.wiretap(enable);
 	}
 
+	/**
+	 * The maximum allowed depth of the server decoding queue.
```
violetagg: Should we mention that this applies only to HTTP/1.1 pipelining?

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
```diff
@@ -1467,6 +1467,24 @@ public final HttpServer wiretap(boolean enable) {
...
+	 * @param maxPipelineDepth the maximum number of requests that may be decoded while awaiting the
+	 * corresponding responses to be written, before decoding of further requests is rejected with
+	 * an {@link IllegalStateException}.
+	 * @return a new {@link HttpServer}
+	 * @since 1.3.8
+	 */
+	public final HttpServer maxPipelineDepth(int maxPipelineDepth) {
+		if (maxPipelineDepth < 0) {
```
violetagg:
```suggestion
		if (maxPipelineDepth <= 0) {
```

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
```diff
@@ -4149,4 +4149,38 @@ void testIssue4188() throws Exception {
...
+		assertThat(server.configuration().maxPipelineDepth()).isEqualTo(128);
+
+		server = server.maxPipelineDepth(5);
+		assertThat(server.configuration().maxPipelineDepth()).isEqualTo(5);
+
+		disposableServer = server.bindNow();
+
+		HttpClient client = createClient(disposableServer.port());
+		Flux.range(0, 5)
+			.parallel(5)
```
violetagg: This opens 5 different connections to the server.
You need HTTP/1.1 pipelining.

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
```diff
@@ -309,6 +311,18 @@ public boolean redirectHttpToHttps() {
 		return uriTagValue;
 	}
 
+	/**
+	 * Returns the maximum allowed depth of the encoding pipeline queue, the default
+	 * value is set to {@link HttpContentEncoder#DEFAULT_MAX_PIPELINE_DEPTH}.
```
violetagg, line 316:
```suggestion
	 * value is set to {@link HttpServerCodec#DEFAULT_MAX_PIPELINE_DEPTH}.
```

violetagg (changes requested):
Thanks for the PR!

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServerConfig.java
reta: A bit more on that: `HttpServerCodec.DEFAULT_MAX_PIPELINE_DEPTH` is package private (so sadly we cannot refer it directly), but it is set to the same value as `HttpContentEncoder. DEFAULT_MAX_PIPELINE_DEPTH` (and this is what I use as the default), I could certainly change the docs but that would be not in-sync with the code

violetagg: ok let's keep it as is

reactor-netty-http/src/test/java/reactor/netty/http/server/HttpServerTests.java
reta: Thanks a lot for the review @violetagg , completely reworked the test test and made sure we actually hit the limit

reactor-netty-http/src/main/java/reactor/netty/http/server/HttpServer.java
reta: Also aligned the error message with `HttpCompressionOptionsSpec`

violetagg (approved):

violetagg:
I backported this to the 1.3.x branch

